PR #3, opened 2021-06-16
Add support for lirc programs

seanyoung:
Tested with rc-loopback

### DIFF
--- a/aya/src/programs/lirc_mode2.rs
+++ b/aya/src/programs/lirc_mode2.rs
@@ -1,0 +1,101 @@
+use std::os::unix::prelude::{AsRawFd, RawFd};
+
+use crate::{
+    generated::{bpf_attach_type::BPF_LIRC_MODE2, bpf_prog_type::BPF_PROG_TYPE_LIRC_MODE2},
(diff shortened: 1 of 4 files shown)


aya/src/programs/lirc_mode2.rs
```diff
@@ -0,0 +1,101 @@
...
+///
+/// # Examples
+///
+/// ```no_run
+/// use std::fs::File;
+/// use std::convert::TryInto;
+/// use aya::programs::LircMode2;
+///
+/// let file = File::open("/dev/lirc0").unwrap();
+/// let let mut bpf = Bpf::load_file("imon_rsc.o").unwrap();
```
alessandrod, line 26:
```suggestion
/// let mut bpf = Bpf::load_file("imon_rsc.o").unwrap();
```

alessandrod (approved):
Great stuff, thank you! Please fix the doc issue and it's good to go.

aya/src/programs/lirc_mode2.rs
seanyoung: done!

PR #4, opened 2021-06-17
Fix doctest and run them during CI

seanyoung:
Signed-off-by: Sean Young <sean@mess.org>


alessandrod:
Ha amazing, this was on my todo list for the weekend 🙏😄

Issue #10, opened 2021-06-21
Support iproute2 bpf_map_def layout


alessandrod:
This is how aya currently defines `bpf_map_def`:  https://github.com/alessandrod/aya/blob/bb595c4e69ff0c72c8327e7f64d43ca7a4bc16a3/aya/src/bpf.rs#L52-L60

Many older eBPF programs use an old definition of `bpf_map_def` that didn't include the `id`  and `pinning` fields. Currently aya will fail to parse those programs because of this check: https://github.com/alessandrod/aya/blob/bb595c4e69ff0c72c8327e7f64d43ca7a4bc16a3/aya/src/obj/mod.rs#L512-L517

We could easily support those older programs by relaxing the above check a little - checking that the struct contains up to the `map_flags` field, and set `id`  and `pinning` to `0`.

RafaelOrtizRC:
Where are the `id` and `pinning` fields coming from? Here's the anonymous struct in the latest 5.12:

```c
	struct { /* anonymous struct used by BPF_MAP_CREATE command */
		__u32	map_type;	/* one of enum bpf_map_type */
		__u32	key_size;	/* size of key in bytes */
		__u32	value_size;	/* size of value in bytes */
		__u32	max_entries;	/* max number of entries in a map */
		__u32	map_flags;	/* BPF_MAP_CREATE related
					 * flags defined above.
					 */
		__u32	inner_map_fd;	/* fd pointing to the inner map */
		__u32	numa_node;	/* numa node (effective only if
					 * BPF_F_NUMA_NODE is set).
					 */
		char	map_name[BPF_OBJ_NAME_LEN];
		__u32	map_ifindex;	/* ifindex of netdev to create on */
		__u32	btf_fd;		/* fd pointing to a BTF type data */
		__u32	btf_key_type_id;	/* BTF type_id of the key */
		__u32	btf_value_type_id;	/* BTF type_id of the value */
		__u32	btf_vmlinux_value_type_id;/* BTF type_id of a kernel-
						   * struct stored as the
						   * map value
						   */
	};
```

alessandrod:
> Where are the `id` and `pinning` fields coming from? Here's the anonymous struct in the latest 5.12:
> 

iproute2, see https://github.com/alessandrod/aya/pull/14#issuecomment-865325330

RafaelOrtizRC:
Closed by #14

PR #14, opened 2021-06-21
Support iproute2 bpf_map_def layout

RafaelOrtizRC:
Addresses #10 

Please note, while this passes tests I haven't had a chance to actually try it with an older/new map definitions. If someone has an old program lying around, please test it before merging.

It also breaks the `id` and `pinning` fields, replacing them with the members found [here](https://elixir.bootlin.com/linux/v5.12.12/source/include/uapi/linux/bpf.h#L491).

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -48,7 +48,7 @@ unsafe_impl_pod!(i8, u8, i16, u16, i32, u32, i64, u64);
 
 #[allow(non_camel_case_types)]
 #[repr(C)]
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
 pub(crate) struct bpf_map_def {
     // minimum features required by old BPF programs
     pub(crate) map_type: u32,
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -57,8 +57,8 @@ pub(crate) struct bpf_map_def {
     pub(crate) max_entries: u32,
     pub(crate) map_flags: u32,
     // optional features
-    pub(crate) id: u32, // id
-    pub(crate) pinning: u32,    // pinning
+    pub(crate) id: u32,
+    pub(crate) pinning: u32,
 }
 
 /// The main entry point into the library, used to work with eBPF programs and maps.
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -48,7 +48,7 @@ unsafe_impl_pod!(i8, u8, i16, u16, i32, u32, i64, u64);
 
 #[allow(non_camel_case_types)]
 #[repr(C)]
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Default)]
 pub(crate) struct bpf_map_def {
     // minimum features required by old BPF programs
     pub(crate) map_type: u32,
@@ -57,34 +57,8 @@ pub(crate) struct bpf_map_def {
     pub(crate) max_entries: u32,
     pub(crate) map_flags: u32,
     // optional features
-    pub(crate) inner_map_fd: u32, // id
-    pub(crate) numa_node: u32,    // pinning
-    pub(crate) map_name: [u8; BPF_OBJ_NAME_LEN],
-    pub(crate) map_ifindex: u32,
-    pub(crate) btf_id: u32,
-    pub(crate) btf_key_type_id: u32,
-    pub(crate) btf_value_type_id: u32,
-    pub(crate) btf_vmlinux_value_type_id: u32,
-}
-
-impl Default for bpf_map_def {
-    fn default() -> Self {
-        Self {
-            map_type: 0,
-            key_size: 0,
-            value_size: 0,
-            max_entries: 0,
-            map_flags: 0,
-            inner_map_fd: 0,
-            numa_node: 0,
-            map_name: [0u8; BPF_OBJ_NAME_LEN],
-            map_ifindex: 0,
-            btf_id: 0,
-            btf_key_type_id: 0,
-            btf_value_type_id: 0,
-            btf_vmlinux_value_type_id: 0
-        }
-    }
+    pub(crate) id: u32, // id
+    pub(crate) pinning: u32,    // pinning
 }
 
 /// The main entry point into the library, used to work with eBPF programs and maps.
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -50,13 +50,41 @@ unsafe_impl_pod!(i8, u8, i16, u16, i32, u32, i64, u64);
 #[repr(C)]
 #[derive(Copy, Clone, Debug)]
 pub(crate) struct bpf_map_def {
+    // minimum features required by old BPF programs
     pub(crate) map_type: u32,
     pub(crate) key_size: u32,
     pub(crate) value_size: u32,
     pub(crate) max_entries: u32,
     pub(crate) map_flags: u32,
-    pub(crate) id: u32,
-    pub(crate) pinning: u32,
+    // optional features
+    pub(crate) inner_map_fd: u32, // id
+    pub(crate) numa_node: u32,    // pinning
+    pub(crate) map_name: [u8; BPF_OBJ_NAME_LEN],
+    pub(crate) map_ifindex: u32,
+    pub(crate) btf_id: u32,
+    pub(crate) btf_key_type_id: u32,
+    pub(crate) btf_value_type_id: u32,
+    pub(crate) btf_vmlinux_value_type_id: u32,
+}
+
+impl Default for bpf_map_def {
+    fn default() -> Self {
+        Self {
+            map_type: 0,
+            key_size: 0,
+            value_size: 0,
+            max_entries: 0,
+            map_flags: 0,
+            inner_map_fd: 0,
+            numa_node: 0,
+            map_name: [0u8; BPF_OBJ_NAME_LEN],
+            map_ifindex: 0,
+            btf_id: 0,
+            btf_key_type_id: 0,
+            btf_value_type_id: 0,
+            btf_vmlinux_value_type_id: 0
+        }
+    }
 }
 
 /// The main entry point into the library, used to work with eBPF programs and maps.
--- a/aya/src/obj/mod.rs
+++ b/aya/src/obj/mod.rs
@@ -22,6 +22,7 @@ use crate::{
     obj::btf::{Btf, BtfError, BtfExt},
     BpfError,
 };
+use std::slice::from_raw_parts_mut;
 
 const KERNEL_VERSION_ANY: u32 = 0xFFFF_FFFE;
 /// The first five __u32 of `bpf_map_def` must be defined.
@@ -520,13 +521,9 @@ fn parse_map_def(name: &str, data: &[u8]) -> Result<bpf_map_def, ParseError> {
     if data.len() < mem::size_of::<bpf_map_def>() {
         let mut map_def = bpf_map_def::default();
         unsafe {
-            let mut data_ptr = data.as_ptr();
-            let mut map_def_ptr = &mut map_def as *mut bpf_map_def as *mut u8;
-            for _ in 0..=data.len() {
-                *map_def_ptr = *data_ptr;
-                map_def_ptr = map_def_ptr.add(1);
-                data_ptr = data_ptr.add(1);
-            }
+            let map_def_ptr =
+                from_raw_parts_mut(&mut map_def as *mut bpf_map_def as *mut u8, data.len());
+            map_def_ptr.copy_from_slice(data);
         }
         Ok(map_def)
     } else {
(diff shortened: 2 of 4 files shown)


alessandrod:
`bpf_map_def` is not the same as the args passed to `BPF_MAP_CREATE`. 

There are two versions out in the wild, one used by libbpf: https://github.com/libbpf/libbpf/blob/12eb2666d9e97d03a10347edb12112b3f638b989/src/libbpf.h#L400 and one used by iproute2: https://github.com/eworm-de/iproute2/blob/1fdea280517b9951e43bc4544210024d0346b27e/include/bpf_elf.h#L32

`id` and `pinning` are required to load programs originally written to be loaded with `tc` and friends from iproute2.

RafaelOrtizRC:
> `bpf_map_def` is not the same as the args passed to `BPF_MAP_CREATE`.

Whoops, I misunderstood how the values were being passed around. It's using the correct struct now.

aya/src/bpf.rs
```diff
@@ -48,15 +48,17 @@ unsafe_impl_pod!(i8, u8, i16, u16, i32, u32, i64, u64);
...
+    // minimum features required by old BPF programs
     pub(crate) map_type: u32,
     pub(crate) key_size: u32,
     pub(crate) value_size: u32,
     pub(crate) max_entries: u32,
     pub(crate) map_flags: u32,
-    pub(crate) id: u32,
-    pub(crate) pinning: u32,
+    // optional features
+    pub(crate) id: u32, // id
```
alessandrod, line 60:
```suggestion
    pub(crate) id: u32,
```

aya/src/bpf.rs
```diff
@@ -48,15 +48,17 @@ unsafe_impl_pod!(i8, u8, i16, u16, i32, u32, i64, u64);
...
+    pub(crate) pinning: u32,    // pinning
```
alessandrod, line 61:
```suggestion
    pub(crate) pinning: u32,
```

aya/src/obj/mod.rs
```diff
@@ -510,13 +509,24 @@ fn parse_map(section: &Section, name: &str) -> Result<Map, ParseError> {
 }
 
 fn parse_map_def(name: &str, data: &[u8]) -> Result<bpf_map_def, ParseError> {
-    if mem::size_of::<bpf_map_def>() > data.len() {
+    if data.len() > mem::size_of::<bpf_map_def>() {
```
alessandrod, line 512: We also need to check that it's at least as big as up to `map_flags`

aya/src/obj/mod.rs
```diff
@@ -510,13 +509,24 @@ fn parse_map(section: &Section, name: &str) -> Result<Map, ParseError> {
...
         return Err(ParseError::InvalidMapDefinition {
             name: name.to_owned(),
         });
     }
 
-    Ok(unsafe { ptr::read_unaligned(data.as_ptr() as *const bpf_map_def) })
+    if data.len() <  mem::size_of::<bpf_map_def>() {
+        let mut map_def = bpf_map_def { ..Default::default() };
```
alessandrod, line 519:
```suggestion
        let mut map_def = bpf_map_def::default();
```

aya/src/obj/mod.rs
```diff
@@ -510,13 +509,24 @@ fn parse_map(section: &Section, name: &str) -> Result<Map, ParseError> {
...
+        unsafe {
+            ptr::copy(data.as_ptr() as *const bpf_map_def, &mut map_def as *mut bpf_map_def, 1);
+            // id and pinning will be garbage data
```
alessandrod, line 522: This is UB, see https://doc.rust-lang.org/stable/std/ptr/fn.copy.html#safety

You need to cast both to bytes and copy bytes. After copying bytes, resetting id and pinning won't be necessary.

aya/src/obj/mod.rs
```diff
@@ -657,10 +667,19 @@ mod tests {
 
     #[test]
     fn test_parse_map_def() {
+        #![allow(unused_variables)] // map_def is used by the assertion
         assert!(matches!(
-            parse_map_def("foo", &[]),
+            parse_map_def("foo", &[0u8; std::mem::size_of::<bpf_map_def>() + 10]),
```
alessandrod, line 672: The old test is still valid, please revert this change

aya/src/obj/mod.rs
```diff
@@ -670,32 +689,45 @@ mod tests {
...
-                id: 0,
-                pinning: 0
-            })
+            Ok(map_def)
+        ));
+
+        assert!(matches!(
+            parse_map_def(
+                "foo",
+                &[0u8, 0u8, 0u8, 1u8, 0u8, 0u8, 0u8, 2u8, 0u8, 0u8, 0u8, 4u8, 0u8, 0u8, 0u8, 5u8]
```
alessandrod, line 701: use `bytes_of` here and index with a shorter range

aya/src/obj/mod.rs
```diff
@@ -670,32 +689,45 @@ mod tests {
...
+            ),
+            Ok(map_def)
         ));
+        let map = parse_map_def(
+            "foo",
+            &[0u8, 0u8, 0u8, 1u8, 0u8, 0u8, 0u8, 2u8, 0u8, 0u8, 0u8, 4u8, 0u8, 0u8, 0u8, 5u8]
```
alessandrod, line 707: same, use `bytes_of`  please

aya/src/obj/mod.rs
```diff
@@ -670,32 +689,45 @@ mod tests {
...
+        ).unwrap();
+        assert!(map.id == 0 && map.pinning == 0)
     }
 
     #[test]
     fn test_parse_map_error() {
         assert!(matches!(
-            parse_map(&fake_section("maps/foo", &[]), "foo"),
+            parse_map(&fake_section("maps/foo", &[0u8; std::mem::size_of::<bpf_map_def>() + 10]), "foo"),
             Err(ParseError::InvalidMapDefinition { .. })
```
alessandrod, line 716: the old test was correct, please revert

aya/src/obj/mod.rs
```diff
@@ -706,32 +738,34 @@ mod tests {
                         value_size: 3,
                         max_entries: 4,
                         map_flags: 5,
-                        id: 0,
```
alessandrod, line 709: I think you wanted to replace this with `def` too or why bind `def` at all?

aya/src/obj/mod.rs
```diff
@@ -706,32 +738,34 @@ mod tests {
...
-                },
+                def,
                 data
             }) if name == "foo" && data.is_empty()
         ))
     }
 
     #[test]
     fn test_parse_map_data() {
+        #![allow(unused_variables)] // def is used by the assertion
```
alessandrod, line 757: I don't think this test should be changed? Other than the Default::default() bit that is

alessandrod (changes requested):
Thanks for doing this! See my comments

alessandrod:
Oh also please make sure to `cargo fmt` your changes

aya/src/obj/mod.rs
RafaelOrtizRC: rustc doesn't like it when `..Default::default()` is called in `Ok()`. 

```
error: expected `}`, found `Default`
   --> aya/src/obj/mod.rs:709:19
    |
703 |             Ok(bpf_map_def {
    |                ----------- while parsing the fields for this pattern
...
709 |                 ..Default::default()
    |                   ^^^^^^^ expected `}`
```

Is there an alternative?

aya/src/obj/mod.rs
RafaelOrtizRC: I was getting errors due to `_map_type` and `value_size` being undefined.

aya/src/obj/mod.rs
RafaelOrtizRC: I added the check and put back the tests for a too-small definition. Now it should check for too large and too small.

aya/src/obj/mod.rs
RafaelOrtizRC: I was trying to avoid that byte copy loop but you're right - no guarantees about what will happen when copying past the end of `data`.

See 1404d12a7e29ab02225bc13fd36794f6c6206db1

aya/src/obj/mod.rs
RafaelOrtizRC: Reverted in 1404d12a7e29ab02225bc13fd36794f6c6206db1

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 1404d12a7e29ab02225bc13fd36794f6c6206db1

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 1404d12a7e29ab02225bc13fd36794f6c6206db1

aya/src/obj/mod.rs
RafaelOrtizRC: Reverted in 1404d12a7e29ab02225bc13fd36794f6c6206db1

aya/src/obj/mod.rs
```diff
@@ -510,13 +511,27 @@ fn parse_map(section: &Section, name: &str) -> Result<Map, ParseError> {
...
         });
     }
 
-    Ok(unsafe { ptr::read_unaligned(data.as_ptr() as *const bpf_map_def) })
+    if data.len() < mem::size_of::<bpf_map_def>() {
+        let mut map_def = bpf_map_def::default();
+        unsafe {
+            let mut data_ptr = data.as_ptr();
+            let mut map_def_ptr = &mut map_def as *mut bpf_map_def as *mut u8;
+            for _ in 0..=data.len() {
```
alessandrod, line 525: Instead of looping, please convert to byte slices and use `copy_from_slice`

aya/src/obj/mod.rs
```diff
@@ -657,10 +672,23 @@ mod tests {
 
     #[test]
     fn test_parse_map_def() {
+        #![allow(unused_variables)] // map_def is used by the assertion
```
alessandrod, line 675: the compiler is giving you a warning because using `matches!` map_def is not being used, it's being re-bound in the pattern.

If you want to factor out `map_def` - which looks good to me - you need to change the test to use `assert_eq!` instead of `matches!`. Eg:
```suggestion
assert_eq!(parse_map_def("foo", bytes_of(&def)).unwrap(), def)
```

For that to work, you'll have to make `bpf_map_def`  derive `PartialEq`

aya/src/obj/mod.rs
```diff
@@ -657,10 +672,23 @@ mod tests {
...
         assert!(matches!(
             parse_map_def("foo", &[]),
             Err(ParseError::InvalidMapDefinition { .. })
         ));
+        assert!(matches!(
+            parse_map_def("foo", &[0u8; std::mem::size_of::<bpf_map_def>() + 10]),
```
alessandrod, line 681:
```suggestion
            parse_map_def("foo", &[0u8; std::mem::size_of::<bpf_map_def>() + 1]),
```

aya/src/obj/mod.rs
```diff
@@ -670,32 +698,67 @@ mod tests {
...
 
     #[test]
     fn test_parse_map_error() {
         assert!(matches!(
             parse_map(&fake_section("maps/foo", &[]), "foo"),
             Err(ParseError::InvalidMapDefinition { .. })
+        ));
+        assert!(matches!(
+            parse_map(
+                &fake_section("maps/foo", &[0u8; std::mem::size_of::<bpf_map_def>() + 10]),
```
alessandrod, line 744:
```suggestion
                &fake_section("maps/foo", &[0u8; std::mem::size_of::<bpf_map_def>() + 1]),
```

aya/src/obj/mod.rs
alessandrod: yes, switch to `assert_eq` see the comment above

alessandrod (changes requested):

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 86b639b

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 86b639b

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 86b639b

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 86b639b

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 86b639b

aya/src/obj/mod.rs
RafaelOrtizRC: Fixed in 86b639b

PR #21, opened 2021-06-28
feat(bpf): add probe register utility methods

EKTehtris:
There is currently a lot of commented code in those methods because the code will panic on no_std, if possible a proper fix for assert and panic should be added, so we can write safer code when using specific arch

### DIFF
--- a/bpf/aya-bpf/src/programs/probe.rs
+++ b/bpf/aya-bpf/src/programs/probe.rs
@@ -9,7 +9,7 @@ pub struct ProbeContext {
 impl ProbeContext {
     pub fn new(ctx: *mut c_void) -> ProbeContext {
         ProbeContext {
-            regs: Regs { regs: ctx as *mut pt_regs },
+            regs: Regs::from(ctx)
         }
     }
 }
--- a/bpf/aya-bpf/src/programs/probe.rs
+++ b/bpf/aya-bpf/src/programs/probe.rs
@@ -1,6 +1,6 @@
 use core::ffi::c_void;
 
-use crate::{bindings::pt_regs, BpfContext};
+use crate::{bindings::__u64, bindings::pt_regs, BpfContext};
 
 pub struct ProbeContext {
     pub regs: Regs,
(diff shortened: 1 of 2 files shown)


bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -3,19 +3,177 @@ use core::ffi::c_void;
...
+}
+
+pub struct Regs {
+    regs: *mut pt_regs,
+}
+
+impl From<*mut c_void> for Regs {
+    fn from(ctx: *mut c_void) -> Self {
+        Regs { regs: ctx as *mut pt_regs }
+    }
```
alessandrod, line 30: This should take `*mut pt_regs` instead, and require casting on the caller

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -3,19 +3,177 @@ use core::ffi::c_void;
...
+}
+
+/// This struct allow to generate the macros we have in C as PT_REGS_*
+/// as simple method calls.
+impl Regs {
+    /// Utility to know if the context is valid
+    pub fn is_null(&self) -> bool {
+        self.regs.is_null()
+    }
```
alessandrod, line 39: When can the context be null?

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -3,19 +3,177 @@ use core::ffi::c_void;
...
+
+    /// Utility to get the raw ptr, useful to interact with the context directly.
+    pub fn as_raw_ptr(&self) -> *mut c_void {
```
alessandrod, line 42: This should be `as_mut_ptr` for consistency with https://doc.rust-lang.org/stable/std/?search=as_mut_ptr and should return `*mut pt_regs`

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -3,19 +3,177 @@ use core::ffi::c_void;
...
         self.regs as *mut c_void
     }
+
+    /// Utility to get the underlying regs, useful for advanced users to access the
+    /// underlying binding without having to recast.
+    pub fn as_regs(&self) -> *mut pt_regs {
```
alessandrod, line 48: I don't think we need this

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -3,19 +3,177 @@ use core::ffi::c_void;
...
+        self.regs
+    }
+
+    #[doc(alias = "PT_REGS_PARM1")]
+    /// Utility to get the First Parameter
+    pub fn parm1(&self) -> *const c_void {
```
alessandrod, line 54: The `parmX()` methods should return `u64`, registers can contain values not just pointers. Casting should be on the caller.

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -3,19 +3,177 @@ use core::ffi::c_void;
...
+        // assert!(self.regs.is_null(),"You should ensure that the registers are not null");
+        #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+            return unsafe { &*(self.regs) }.rdi as *const c_void;
+        // unreachable!();
+        // panic!("Aya does not support this platform yet.");
+        return 0 as *const c_void;
```
alessandrod, line 60: I think that instead of returning NULL, the methods should be defined only on archs that support a given register. Like `std::os::unix` is only available on unix.

alessandrod (changes requested):
Thanks for this! Please see my comments.

bpf/aya-bpf/src/programs/probe.rs
EKTehtris: I don't expect anyone to use this trait but you are right, I fixed it

bpf/aya-bpf/src/programs/probe.rs
EKTehtris: For now I don't think any probe can have a null context but I don't think that it's not a future possibility, in any case this is a passthrough

bpf/aya-bpf/src/programs/probe.rs
EKTehtris: this was fixed

bpf/aya-bpf/src/programs/probe.rs
EKTehtris: I made it crate only, I use it internally

bpf/aya-bpf/src/programs/probe.rs
EKTehtris: this was fixed to return option<u64>

bpf/aya-bpf/src/programs/probe.rs
EKTehtris: I fixed it by adding a build.rs similar to the bindings so I can evaluate on a function level

PR #29, opened 2021-07-05
bpf: add support for tracepoint program

tw4452852:
This patch add initial support for tracepoint program type.
Hope you enjoy.

Signed-off-by: Tw <wei.tan@intel.com>

### DIFF
--- a/bpf/aya-bpf/src/programs/tracepoint.rs
+++ b/bpf/aya-bpf/src/programs/tracepoint.rs
@@ -1,0 +1,22 @@
+use crate::{helpers::bpf_probe_read, BpfContext};
+use core::ffi::c_void;
+
+pub struct TracePointContext {
+    ctx: *mut c_void,
+}
+
+impl TracePointContext {
+    pub fn new(ctx: *mut c_void) -> TracePointContext {
+        TracePointContext { ctx }
+    }
+
+    pub unsafe fn read_at<T>(&self, offset: usize) -> Result<T, i64> {
+        bpf_probe_read(self.ctx.add(offset) as *const T)
+    }
+}
+
+impl BpfContext for TracePointContext {
+    fn as_ptr(&self) -> *mut c_void {
+        self.ctx
+    }
+}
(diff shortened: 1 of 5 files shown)


bpf/aya-bpf/src/programs/tracepoint.rs
```diff
@@ -0,0 +1,20 @@
+use crate::{helpers::bpf_probe_read, BpfContext};
+use core::ffi::c_void;
+
+pub struct TracePointContext(*mut c_void);
```
alessandrod, line 4: please make the ctx pointer a namd field for consistency with the other contexts

alessandrod (changes requested):
Great work thank you!

bpf/aya-bpf/src/programs/tracepoint.rs
```diff
@@ -0,0 +1,22 @@
...
+    ctx: *mut c_void,
+}
+
+impl TracePointContext {
+    pub fn new(ctx: *mut c_void) -> TracePointContext {
+        TracePointContext { ctx }
+    }
+
+    pub unsafe fn read_at<T>(&self, offset: usize) -> Result<T, i64> {
+        bpf_probe_read(self.ctx.add(offset) as *const T)
```
alessandrod, line 14: I think you want to cast self.ctx to usize and add usizes here. Otherwise you're adding mem::size_of::<*mut c_void>() * count.

alessandrod (changes requested):

bpf/aya-bpf/src/programs/tracepoint.rs
tw4452852: I don't think so, according to the [doc of add](https://doc.rust-lang.org/std/primitive.pointer.html#method.add), we're adding `count * mem::size_of::<c_void>` here which is expected.

alessandrod: oops yes of course you're right

PR #31, opened 2021-07-09
fix: pass BTF object by reference in order to allow multiple eBPF programs to share it and save memory (closes #30).

### LINKED ISSUE #30: allow multiple BPF programs to share the same BTF
I've noticed that multiple BPF programs loaded by the same userspace process can't share the same host BTF because [this method](https://github.com/alessandrod/aya/blob/d996a88de47d60053504695036a14157ce6b3aa6/aya/src/bpf.rs#L114) consumes the `target_btf` object instead of accepting it by reference. However, [later in the code](https://github.com/alessandrod/aya/blob/d996a88de47d60053504695036a14157ce6b3aa6/aya/src/obj/btf/relocation.rs#L181), in the only part where `target_btf` is used (immutably), its reference is used. If i'm not mistaken this means that the very first method could accept a reference instead of a value to begin with.

On my Ubuntu each BTF object is around 13MB (in debug mode) and I'm loading several different eBPF programs, so it'd be nice to be able to share the relocation info among them and save a lot of memory.

evilsocket:
As per title, this PR changtes `Bpf::load` to use `Option<&Btf>` instead of `Option<Btf>` so that a single instance of Btf information can be shared among multiple programs, thus decreasing overall memory usage.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -90,7 +90,7 @@ impl Bpf {
                 path: path.to_owned(),
                 error,
             })?,
-            Btf::from_sys_fs().ok(),
+            Btf::from_sys_fs().ok().as_ref(),
         )
     }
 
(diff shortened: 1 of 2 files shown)


alessandrod (approved):

alessandrod:
Thank you!

Issue #2, opened 2021-06-16
aya is missing query features


seanyoung:
libbpf has the `bpf_prog_query()` function to get a list of the all attached programs; it is also possible to deattach them and get their names.

I am currently re-writing `ir-keytable` in rust https://github.com/seanyoung/ir (early stages). For aya to be useful for this work, it would require this functionality.

alessandrod:
The query I'm thinking should probably be a helper function in the `aya::programs`, and be something like 

```rust
fn query<T: AsRawFd>(target_fd: T, attach_type: XXX, flags: u32) -> Result<(Vec<u32>, u32), io::Error>
```

Where the result is a `Vec` of prog ids and the attach flags. As for the `XXX` type , we could re-export `crate::generated::bpf_attach_type`  as `aya::programs::AttachType` for now, or we could create a new enum and keep it in sync, `bpf_attach_type` doesn't change that often anyway.

To get the name then you'd use a separate syscall right?

As for detaching, I'm not sure yet what we could do. I know I want to add a `forget()` method to the `Link` trait (related to your other PR), but haven't thought about what the API to detach an arbitrary fd/type would look like.

seanyoung:
I do wonder if it is possible to create a higher level of abstraction than `(Vec<u32>, u32)`. I think it would be nicer if we returned a list of programs, which had a `detach()` and `info()` method of some sort. `struct bpf_prog_info` contains a lot of information, not just the name.

alessandrod:
Yes you're right, we should definitely create a higher level API. I guess my thinking was that coming up with a higher level interface would probably require some work (see below), so we could start by wrapping the syscall. Now that I've thought about it though, I think I was overthinking it.

Agreed on `info()`, we should have it return a type that contains everything useful from `struct bpf_prog_info`. And about `detach()`, I was thinking that it would be tricky ensuring that it doesn't break things if you have other `Link`s laying around, but that's not the case. All the existing `Link::detach()` impls already ignore errors on detach. And regardless of what interface we come up with, we need to handle other programs (eg bpftool) potentially detaching our links anyway.

So yeah, this is easier than I first thought :)

alessandrod:
@seanyoung are you planning to do or have you done any work on this? Otherwise I might get to it next week

seanyoung:
@alessandrod  I had started on this, see my query branch https://github.com/seanyoung/aya/commits/query. The query of programs and their names works, but not detaching or any other field than name.

[...]

PR #32, opened 2021-07-25
Implement query for lirc programs

seanyoung:
Signed-off-by: Sean Young <sean@mess.org>

### DIFF
--- a/aya/src/programs/lirc_mode2.rs
+++ b/aya/src/programs/lirc_mode2.rs
@@ -2,8 +2,8 @@ use std::os::unix::prelude::{AsRawFd, RawFd};
 
 use crate::{
     generated::{bpf_attach_type::BPF_LIRC_MODE2, bpf_prog_type::BPF_PROG_TYPE_LIRC_MODE2},
-    programs::{load_program, Link, LinkRef, ProgramData, ProgramError},
-    sys::{bpf_prog_attach, bpf_prog_detach},
+    programs::{load_program, query, Link, LinkRef, ProgramData, ProgramError, ProgramInfo},
+    sys::{bpf_obj_get_info_by_fd, bpf_prog_attach, bpf_prog_detach, bpf_prog_get_fd_by_id},
 };
 
 use libc::{close, dup};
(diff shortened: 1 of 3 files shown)


seanyoung:
Fixes #2

aya/src/programs/lirc_mode2.rs
```diff
@@ -78,10 +83,63 @@ impl LircMode2 {
 
         Ok(self.data.link(LircLink::new(prog_fd, lircdev_fd)))
     }
+
+    /// Query lirc device for attached programs
+    pub fn query<T: AsRawFd>(target_fd: T) -> Result<Vec<LircLink>, ProgramError> {
+        let mut prog_ids = vec![0u32; 64];
```
alessandrod, line 89: Could you move the part that calls `bpf_prog_query` and populates the return ids to a shared  `pub(crate)` util in `aya::programs::query()`?

aya/src/programs/lirc_mode2.rs
```diff
@@ -93,6 +151,17 @@ impl LircLink {
             target_fd: Some(unsafe { dup(target_fd) }),
         }
     }
+
+    pub fn info(&self) -> Result<bpf_prog_info, ProgramError> {
+        if let Some(fd) = self.prog_fd {
```
alessandrod, line 156: I'm not sure how I feel about exposing `bpf_prog_info` as part of the API. What's your use case? I'm thinking we should maybe return some kind of wrapper type.

alessandrod (changes requested):
Thanks for this! Overall I like this approach and I think it will work for other programs as well.

aya/src/programs/lirc_mode2.rs
seanyoung: done

aya/src/programs/lirc_mode2.rs
seanyoung: The bpf_prog_info struct has a lot of fields and a lot of them are simple `u64` fields, so the wrapper will be pretty thin.

So we could have a `struct ProgramInfo(bpf_prog_info)` with a load of accessor functions. Would that work?

alessandrod: That sounds good yes, and for now I would implement only the accessors that you need. If people need more accessors they'll open an issue or send a PR, we'll learn about more use cases, and see if there's a better API we can provide.

seanyoung: Great, thanks. I've done that, let me know what you think

aya/src/programs/mod.rs
```diff
@@ -533,3 +579,24 @@ impl_try_from_program!(
     CgroupSkb,
     LircMode2
 );
+
+pub struct ProgramInfo(bpf_prog_info);
+
+impl ProgramInfo {
+    pub fn name(&self) -> Option<String> {
```
alessandrod, line 586: Shouldn't this return Option<&[u8]>? It's valid for the name not to be utf8, I think callers should (optionally) do the conversion here

aya/src/programs/mod.rs
```diff
@@ -533,3 +579,24 @@ impl_try_from_program!(
     CgroupSkb,
     LircMode2
 );
+
+pub struct ProgramInfo(bpf_prog_info);
```
alessandrod, line 583: Can you add some docs for the struct and the accessors please?

aya/src/programs/lirc_mode2.rs
alessandrod: Looks good thanks!

aya/src/programs/mod.rs
alessandrod: > Shouldn't this return &[u8]? It's valid for the name not to be utf8, I think callers should (optionally) do the conversion here

I just realised that we use String for program names everywhere else so I'm leaving this up to you, I'm ok with either

seanyoung: So we have both of them now as an option: fn name() which gives you `&[u8]` and fn name_as_str() which returns `Option<&str>`.

aya/src/programs/mod.rs
seanyoung: I've added some rustdoc (a bit rudimentary)

alessandrod (approved):

alessandrod:
thank you!

PR #33, opened 2021-07-25
examples: Add an XDP example

dave-tucker:
This PR adds an example that shows:

1. How to create a simple XDP program that simply increments a counter, stored in a map.
1. How to write a userspace program that loads the program and reads the data in the map.

The BPF program must be compiled using:
`cargo +nightly build --examples -Z build-std=core --target=bpfel-unknown-none`

The userspace example can be run using:
`cargo run --example xdp -- <your-interface-name>`

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,8 +21,14 @@ jobs:
       - uses: actions/checkout@v2
       - uses: Swatinem/rust-cache@v1
 
+      - name: Install bpf-linker
+        run: cargo install bpf-linker
+
       - name: Build
         run: cargo build --verbose
+
+      - name: Build examples
+        run: cargo xtask examples
 
       - name: Run tests
         run: RUST_BACKTRACE=full cargo test --verbose
--- a/README.md
+++ b/README.md
@@ -74,6 +74,19 @@ let cgroup = File::open("/sys/fs/cgroup/unified")?;
 ingress.attach(cgroup, CgroupSkbAttachType::Ingress)?;
 ```
 
+### Examples
+
+There are some examples of how to use Aya to write BPF programs and userspace program in the `examples` directory. These can be compiled by running `cargo xtask examples`.
+The BPF examples require you to have a rust nightly toolchain installed.
+
+`cargo xtask examples`
+
+# Run
+
+## XDP
+
+`sudo ./examples/target/debug/xdp_userspace ./examples/target/bpfel-unknown-none/release/xdp wlp2s0`
+
 ## Community
 
 Join [the conversation on Discord][chat-url] to discuss anything related to aya.
--- a/bpf/aya-bpf/src/maps/array.rs
+++ b/bpf/aya-bpf/src/maps/array.rs
@@ -1,10 +1,10 @@
 use core::{marker::PhantomData, mem};
 
-use aya_bpf_cty::c_void;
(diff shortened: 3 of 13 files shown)


dave-tucker:
Running the program currently fails with `thread 'main' panicked at 'attempt to subtract with overflow', aya/src/obj/relocation.rs:288:17`

alessandrod:
I think examples should go under a new `aya-examples` package. I actually had one at some point. As for the relocation issue, I'll take a look!

dave-tucker:
@alessandrod thanks for the feedback. I've re-orgainzed this now.
I also noted something curious in the `// FIXME` comment.

```rust
// FIXME: We should be able to share structs between BPF and Userspace.
// However, the Array<MapRef, Value> requires that Value implements the aya::Pod trait
// This can't be done from `bpf/src/xdp/mod.rs` as the aya paclage is not no_std
// It can't be done from here because:
// error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
#[derive(Copy, Clone)]
#[repr(C)]
pub struct XdpData {
    pub packet_count : u32,
}
unsafe impl aya::Pod for XdpData {}
```

Wondering if I've approached this wrong, or maybe we can drop the requirement of the Pod trait and/or relocate it somewhere `no_std` so structs can be shared between eBPF and userspace code.

alessandrod:
> @alessandrod thanks for the feedback. I've re-orgainzed this now.
> I also noted something curious in the `// FIXME` comment.
> 
> ```rust
> // FIXME: We should be able to share structs between BPF and Userspace.
> // However, the Array<MapRef, Value> requires that Value implements the aya::Pod trait
> // This can't be done from `bpf/src/xdp/mod.rs` as the aya paclage is not no_std
> // It can't be done from here because:
> // error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
> #[derive(Copy, Clone)]
> #[repr(C)]
> pub struct XdpData {
>     pub packet_count : u32,
> }
> unsafe impl aya::Pod for XdpData {}
> ```
> 
> Wondering if I've approached this wrong, or maybe we can drop the requirement of the Pod trait and/or relocate it somewhere `no_std` so structs can be shared between eBPF and userspace code.

The way I've done this is having two features, "user" and "kernel" in the bpf crate. Then have userspace activate the user feature with:

```
bpf = { path = "./bpf", default-features = false, features = ["user"] }
```

and in the bpf crate implement `aya::Pod` with

```
[cfg(feature = "user")]
pub mod user {
    use super::*;

    unsafe impl aya::Pod for SharedStruct {}
}
```

dave-tucker:
@alessandrod getting closer now :+1: 

- Examples are in their own directory
- Adjusted the module layout as suggested above to add a `user` feature to the `bpf` crate for shared data structures.
- I've added `cargo xtask examples` which will compile the bpf examples and userspace programs
- I've added a section to the readme to show how to use the examples

The example still isn't working correctly though.
Now there isn't a relocation error, but instead `Bpf::load_file` appears to time out, and I don't get an error on the console.
I'll keep debugging and see if I can figure out what's going on

dave-tucker:
Ok, this feels mostly complete now and ready for review.
I've put a workaround for #34 in place.
The initial relocation problem was actually quite funny actually... as it turns out it was the overflow checks in the Cargo dev profile :joy:

dave-tucker:
Going to close this one for now since the book example is better.

Issue #34, opened 2021-07-27
can't load programs: btf relocations consume all available memory


dave-tucker:
Tyring to load a simple `XDP` program that was built with Aya:

```
./examples/target/bpfel-unknown-none/debug/xdp: file format elf64-bpf


Disassembly of section xdp/xdp_stats:

0000000000000000 <xdp_stats>:
       0:       b7 00 00 00 02 00 00 00 r0 = 2
       1:       95 00 00 00 00 00 00 00 exit
```

I've verified that this program can be loaded with `sudo bpftool prog load ./examples/target/bpfel-unknown-none/debug/xdp /sys/fs/bpf/xdp_stats`

However, loading this from Aya is not possible due to something weird with the BTF relocations.

The snippet of my main.rs that loads the code looks like this:
```rust
let mut bpf = Bpf::load_file(path)?;
```

When run, the program gets stopped by the oom-killer.
Running with strace reveals the following, after reading from `/sys/kernel/btf/vmlinux`

```
mmap(NULL, 1863680, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fa0e61aa000
brk(0x5619e237b000)                     = 0x5619e237b000
brk(0x5619e239c000)                     = 0x5619e239c000
mmap(NULL, 331776, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fa0e6159000
mremap(0x7fa0e6159000, 331776, 659456, MREMAP_MAYMOVE) = 0x7fa0e60b8000
mremap(0x7fa0e60b8000, 659456, 1314816, MREMAP_MAYMOVE) = 0x7fa0e5f77000
mremap(0x7fa0e5f77000, 1314816, 2625536, MREMAP_MAYMOVE) = 0x7fa0e5cf6000
mremap(0x7fa0e5cf6000, 2625536, 5246976, MREMAP_MAYMOVE) = 0x7fa0e57f5000
mremap(0x7fa0e57f5000, 5246976, 10489856, MREMAP_MAYMOVE) = 0x7fa0e4df4000
mremap(0x7fa0e4df4000, 10489856, 20975616, MREMAP_MAYMOVE) = 0x7fa0e39f3000
mremap(0x7fa0e39f3000, 20975616, 41947136, MREMAP_MAYMOVE) = 0x7fa0e11f2000
[...]

PR #35, opened 2021-07-28
btf: Add bindings for BTF_KIND_FLOAT

dave-tucker:
Fixes: #34

### DIFF
--- a/aya/src/obj/btf/btf.rs
+++ b/aya/src/obj/btf/btf.rs
@@ -110,7 +110,7 @@ impl Btf {
         }
 
         // safety: btf_header is POD so read_unaligned is safe
-        let header = unsafe { ptr::read_unaligned(data.as_ptr() as *const btf_header) };
+        let header = unsafe { read_btf_header(data) };
 
         let str_off = header.hdr_len as usize + header.str_off as usize;
         let str_len = header.str_len as usize;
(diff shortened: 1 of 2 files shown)


alessandrod:
Thanks for this Dave! Planning to compile a newer and test later today.

dave-tucker:
@alessandrod I've expanded this PR a little.
1. Added some test cases for parsing the various BTF types using data from `/sys/kernel/btf/vmlinux` on my system.
2. Made `Btf::parse` issue an error if we run in to an Unknown btf type, as that was what caused the infinite loop.

That last commit got quite big since the `BtfType::Unknown` variant was then never constructed... and therefore a whole lot of tunctions that were returning `Option` didn't need to any more.

aya/src/obj/btf/btf.rs
```diff
@@ -143,7 +143,7 @@ impl Btf {
         }
 
         let mut data = &data[base..base + type_len];
-        let mut types = vec![BtfType::Unknown];
+        let mut types = vec![];
         while !data.is_empty() {
```
alessandrod, line 147: The first type must always be `Unknown` or it will break `type_by_id`

aya/src/obj/btf/types.rs
```diff
@@ -150,10 +154,9 @@ impl BtfType {
 
         use BtfType::*;
         match self {
-            Unknown => 0,
-            Fwd(_) | Const(_) | Volatile(_) | Restrict(_) | Ptr(_) | Typedef(_) | Func(_) => {
-                ty_size
-            }
+            Unknown => ty_size,
```
alessandrod, line 157: As discussed on discord, to fix the infinite loop on `Unknown` bug, this should be `Unknown => ty_size`

alessandrod:
Good stuff! See the comments then it's good to go

alessandrod (changes requested):

aya/src/obj/btf/btf.rs
dave-tucker: fixed!

aya/src/obj/btf/types.rs
dave-tucker: fixed in https://github.com/alessandrod/aya/pull/35/commits/4465131199e36f288c552a9cbc609fa6720a3d2f

alessandrod (approved):

PR #42, opened 2021-08-05
docs: Initial Commit of Aya Book

dave-tucker:
This commit adds the start of an Aya book with a tutorial for a simple XDP-based firewall.
I plan to add a few more chapters to extend the tutorial shortly.
It also includes a Github Actions workflow to push the compiled book to the `gh_pages` branch.
If pages is enabled on this repository and set to point to that branch, then it will look something like https://dave-tucker.github.io/aya/


alessandrod:
This is truly amazing, thank you so much Dave 🙏

PR #48, opened 2021-08-12
Fix rustup command in the docs.

mstange:
`rustup install nightly --component rust-src`

gives:

```
error: Found argument '--component' which wasn't expected, or isn't valid in this context

USAGE:
    rustup install [FLAGS] [OPTIONS] <toolchain>...

For more information try --help
```

The correct command seems to be `rustup toolchain install nightly --component rust-src`.


dave-tucker (approved):
Nice catch! Thanks for the contribution.

PR #52, opened 2021-08-13
Add support for PerfEvent programs

mstange:
This seems to work, in my very basic tests. I would appreciate if somebody else could push it over the finish line, because I'm not sure how much time I'll be able to spend on this.

Things that need to be tweaked:

 - Public API: I'm not sure if the "PerfEvent" name is appropriate in all the cases. It clashes a little bit with the existing name `PerfEventBuffer`.
 - Public API: The arguments `perf_type` and `config` are currently exposed as u32 and u64. At least for `perf_type` it would be preferable to use the `perf_type_id` enum, but it wasn't straightforward to me how to make that type public.
 - Comments: I don't have a deep understanding of how all of this works, and who determines what kinds of perf events exist (kernel version? CPU model? user privileges?) so I left this mostly blank.

### DIFF
--- a/aya/src/programs/perf_event.rs
+++ b/aya/src/programs/perf_event.rs
@@ -1,0 +1,115 @@
+use crate::{generated::bpf_prog_type::BPF_PROG_TYPE_PERF_EVENT, sys::perf_event_open};
+
+pub use crate::generated::perf_type_id;
+
+use super::{load_program, perf_attach, LinkRef, ProgramData, ProgramError};
+
+#[derive(Debug, Clone)]
+pub enum SamplePolicy {
+    Period(u64),
+    Frequency(u64),
+}
+
+#[derive(Debug, Clone)]
+#[allow(clippy::enum_variant_names)]
+pub enum PerfEventScope {
+    CallingProcessAnyCpu,
+    CallingProcessOneCpu { cpu: u32 },
+    OneProcessAnyCpu { pid: u32 },
+    OneProcessOneCpu { cpu: u32, pid: u32 },
+    AllProcessesOneCpu { cpu: u32 },
+}
+
+/// A program that can be attached at a perf event.
+///
+/// TODO: Explain the different types of perf events and how to get a list.
+/// Maybe just link to the man page of `perf list`.
+/// (But it's not clear to me how to translate those strings into numbers.)
+///
+/// # Minimum kernel version
+///
+/// TODO: minimum kernel version?
+///
+/// # Examples
+///
+/// ```no_run
+/// # #[derive(Debug, thiserror::Error)]
+/// # enum Error {
+/// #     #[error(transparent)]
+/// #     IO(#[from] std::io::Error),
+/// #     #[error(transparent)]
+/// #     Map(#[from] aya::maps::MapError),
+/// #     #[error(transparent)]
+/// #     Program(#[from] aya::programs::ProgramError),
+/// #     #[error(transparent)]
+/// #     Bpf(#[from] aya::BpfError)
+/// # }
+/// # let mut bpf = aya::Bpf::load(&[], None)?;
+/// use std::convert::TryInto;
+/// use aya::programs::{PerfEvent, PerfEventScope, SamplePolicy };
+/// use aya::util::online_cpus;
+///
+/// let prog: &mut PerfEvent = bpf.program_mut("observe_cpu_clock")?.try_into()?;
+/// prog.load()?;
+///
+/// for cpu in online_cpus()? {
+///     prog.attach(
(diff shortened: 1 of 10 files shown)


aya/src/programs/perf_event.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// A program that can be attached at a perf event.
+///
+/// TODO: Explain the different types of perf events and how to get a list.
+/// Maybe just link to the man page of `perf list`.
+/// (But it's not clear to me how to translate those strings into numbers.)
+///
+/// # Minimum kernel version
+///
+/// TODO: minimum kernel version?
```
alessandrod, line 31: 4.9

aya/src/programs/perf_event.rs
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    /// See also [`Program::load`](crate::programs::Program::load).
+    pub fn load(&mut self) -> Result<(), ProgramError> {
+        load_program(BPF_PROG_TYPE_PERF_EVENT, &mut self.data)
+    }
+
+    /// Attaches to a given perf event.
+    pub fn attach(
+        &mut self,
+        perf_type: u32, // perf_type_id
```
alessandrod, line 82: I think we should probably define a `PerfTypeId` enum here instead of re-exporting an auto generated binding of `perf_type_id`

aya/src/programs/perf_event.rs
```diff
@@ -0,0 +1,115 @@
...
+/// use aya::programs::{PerfEvent, PerfEventScope, SamplePolicy };
+/// use aya::util::online_cpus;
+///
+/// let prog: &mut PerfEvent = bpf.program_mut("observe_cpu_clock")?.try_into()?;
+/// prog.load()?;
+///
+/// for cpu in online_cpus()? {
+///     prog.attach(
+///         1, /* PERF_TYPE_SOFTWARE */
+///         0, /* PERF_COUNT_SW_CPU_CLOCK */
```
alessandrod, line 58: We should generate bindings for ids and flags and whatever else can be passed as config

aya/src/programs/perf_event.rs
```diff
@@ -0,0 +1,115 @@
...
+    OneProcessOneCpu { cpu: u32, pid: u32 },
+    AllProcessesOneCpu { cpu: u32 },
+}
+
+/// A program that can be attached at a perf event.
+///
+/// TODO: Explain the different types of perf events and how to get a list.
+/// Maybe just link to the man page of `perf list`.
+/// (But it's not clear to me how to translate those strings into numbers.)
+///
```
alessandrod, line 28: We should generate bindings for all these https://github.com/torvalds/linux/blob/a0e31f3a38e77612ed8967aaad28db6d3ee674b5/include/uapi/linux/perf_event.h#L60

alessandrod (changes requested):
Thanks a lot for this! It's 99% done, let me know if you're going to do the changes or if you want me to pick it up :)

alessandrod:
> This seems to work, in my very basic tests. I would appreciate if somebody else could push it over the finish line, because I'm not sure how much time I'll be able to spend on this.
> 
> Things that need to be tweaked:
> 
>     * Public API: I'm not sure if the "PerfEvent" name is appropriate in all the cases. It clashes a little bit with the existing name `PerfEventBuffer`.

I think the name is fine. It's unfortunate that the perf API is used for a bunch of barely related things, but it's how it is.


> 
>     * Public API: The arguments `perf_type` and `config` are currently exposed as u32 and u64. At least for `perf_type` it would be preferable to use the `perf_type_id` enum, but it wasn't straightforward to me how to make that type public.

I think we should generate the bindingsd for `perf_type_id`, but then export a hand written `PerfTypeId` enum with snake case etc. `perf_event_id` hasn't been changed in 9 years so it's not going to be much of a maintenance burden 😅


>     * Comments: I don't have a deep understanding of how all of this works, and who determines what kinds of perf events exist (kernel version? CPU model? user privileges?) so I left this mostly blank.

I think it's a combination of all 3. As a start I'd be fine with generating bindings for the common ids defined in `perf.h` and pointing to those in the doc.

mstange:
Thanks for the review! I'm planning to address the comments during the week. If, by the end of next weekend, I still haven't gotten to it, I'll hand it off to you.

alessandrod:
I have tweaked this a bit as per review and just merged it. Thanks again for working on it!

I wasn't able to improve the docs much, since well, the perf API is a black box. I have created an issue about improving the docs https://github.com/aya-rs/aya/issues/60

mstange:
Perfect, thanks!

PR #63, opened 2021-09-24
Implement Pinning For Maps and Programs

### LINKED ISSUE #45: Add support for map and program pinning
Program pinning looks simple enough:
It should be possible to add `pub fn pin(&mut self, path: Path )` to the `Link` trait.
Where `Path` would be a relative path on `/sys/fs/bpf` or an absolute path to `/sys/fs/bpf`.
The implementation would call the `bpf` syscall and `BPF_OBJ_PIN` (see: `bpf_obj_pin` in libbpf for reference).

For Map pinning, the API depends on where the maps are defined.
Aya seems to rely on the eBPF program to define the map, which is the created when the program is loaded:
As such I propose we use the following API:
```rust
pub fn load_file<P: AsRef<Path>>(path: P, map_pin_directory: Option<P>) -> Result<Bpf, BpfError>
```
The `map_pin_directory` is then passed down through function calls so it's available when the maps are created.
The usage of Path in this API is the same as the one on the `Link` trait.

dave-tucker:
This commit adds 2 new methods to aya::sys
- bpf_pin_object
- bpf_get_object

Which allow the pinning and retrieval of programs/maps to bpffs.

For map pinning, the user must ensure the `pinning u32` in the
`bpf_map_def` is set to 1, maps will be pinned using a new builder API.

```rust
BpfLoader::new().map_pin_path("/sys/fs/bpf/myapp").load_file("myapp.o")
```

This will pin all maps whose definition requests pinning to path + name.

Closes #45

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -2,7 +2,7 @@ use std::{
     collections::HashMap,
     error::Error,
     fs, io,
-    os::raw::c_int,
+    os::{raw::c_int, unix::io::RawFd},
     path::{Path, PathBuf},
 };
 
@@ -58,64 +58,92 @@ pub(crate) struct bpf_map_def {
     pub(crate) map_flags: u32,
     // optional features
     pub(crate) id: u32,
-    pub(crate) pinning: u32,
+    pub(crate) pinning: PinningType,
 }
 
-/// The main entry point into the library, used to work with eBPF programs and maps.
-#[derive(Debug)]
-pub struct Bpf {
-    maps: HashMap<String, MapLock>,
-    programs: HashMap<String, Program>,
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub(crate) enum PinningType {
+    None = 0,
+    #[allow(dead_code)] // ByName is constructed from the BPF side
+    ByName = 1,
 }
 
-impl Bpf {
+impl Default for PinningType {
+    fn default() -> Self {
+        PinningType::None
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct BpfLoader {
+    btf: Option<Btf>,
+    map_pin_path: Option<PathBuf>,
+}
+
+impl BpfLoader {
+    pub fn new() -> BpfLoader {
+        BpfLoader {
+            btf: None,
+            map_pin_path: None,
+        }
+    }
+
+    // Set the target BTF
+    pub fn btf(&mut self, btf: Btf) -> &mut BpfLoader {
+        self.btf = Some(btf);
+        self
+    }
+
+    // Set the map pin path
+    pub fn map_pin_path<P: AsRef<Path>>(&mut self, path: P) -> &mut BpfLoader {
+        self.map_pin_path = Some(path.as_ref().to_owned());
+        self
+    }
+
     /// Loads eBPF bytecode from a file.
     ///
-    /// Parses the given object code file and initializes the [maps](crate::maps) defined in it. If
-    /// the kernel supports [BTF](Btf) debug info, it is automatically loaded from
-    /// `/sys/kernel/btf/vmlinux`.
+    /// Parses the given object code file and initializes the [maps](crate::maps) defined in it.
     ///
     /// # Examples
     ///
     /// ```no_run
-    /// use aya::Bpf;
+    /// use aya::BpfLoader;
     ///
-    /// let bpf = Bpf::load_file("file.o")?;
+    /// let bpf = BpfLoader::new().load_file("file.o")?;
     /// # Ok::<(), aya::BpfError>(())
     /// ```
-    pub fn load_file<P: AsRef<Path>>(path: P) -> Result<Bpf, BpfError> {
+    pub fn load_file<P: AsRef<Path>>(&mut self, path: P) -> Result<Bpf, BpfError> {
         let path = path.as_ref();
-        Bpf::load(
-            &fs::read(path).map_err(|error| BpfError::FileError {
-                path: path.to_owned(),
-                error,
-            })?,
-            Btf::from_sys_fs().ok().as_ref(),
-        )
+        self.load(&fs::read(path).map_err(|error| BpfError::FileError {
+            path: path.to_owned(),
+            error,
+        })?)
     }
 
     /// Load eBPF bytecode.
     ///
     /// Parses the object code contained in `data` and initializes the [maps](crate::maps) defined
-    /// in it. If `target_btf` is not `None` and `data` includes BTF debug info, [BTF](Btf) relocations
-    /// are applied as well. In order to allow sharing of a single [BTF](Btf) object among multiple
-    /// eBPF programs, `target_btf` is passed by reference.
+    /// in it. If `BpfLoader.btf` is not `None` and `data` includes BTF debug info, [BTF](Btf) relocations
+    /// are applied as well. Any maps that require pinning will be pinned to `BpfLoader.map_pin_path`
     ///
     /// # Examples
     ///
     /// ```no_run
-    /// use aya::{Bpf, Btf};
+    /// use aya::{BpfLoader, Btf};
     /// use std::fs;
     ///
     /// let data = fs::read("file.o").unwrap();
     /// // load the BTF data from /sys/kernel/btf/vmlinux
-    /// let bpf = Bpf::load(&data, Btf::from_sys_fs().ok().as_ref());
+
+    /// let target_btf = Btf::from_sys_fs().unwrap();
+    /// let bpf = BpfLoader::new().btf(target_btf).load(&data);
     /// # Ok::<(), aya::BpfError>(())
     /// ```
-    pub fn load(data: &[u8], target_btf: Option<&Btf>) -> Result<Bpf, BpfError> {
+    pub fn load(&mut self, data: &[u8]) -> Result<Bpf, BpfError> {
         let mut obj = Object::parse(data)?;
 
-        if let Some(btf) = target_btf {
+        if let Some(btf) = &self.btf {
             obj.relocate_btf(btf)?;
         }
 
@@ -130,8 +158,32 @@ impl Bpf {
                     })?
                     .len() as u32;
             }
-            let mut map = Map { obj, fd: None };
-            let fd = map.create()?;
+            let mut map = Map {
+                obj,
+                fd: None,
+                pinned: false,
+            };
+            let fd = match map.obj.def.pinning {
+                PinningType::ByName => {
+                    let path = match &self.map_pin_path {
+                        Some(p) => p,
+                        None => return Err(BpfError::NoPinPath),
+                    };
+                    // try to open map in case it's already pinned
+                    match map.from_pinned(path) {
+                        Ok(fd) => {
+                            map.pinned = true;
+                            fd as RawFd
+                        }
+                        Err(_) => {
+                            let fd = map.create()?;
+                            map.pin(path)?;
+                            fd
+                        }
+                    }
+                }
+                PinningType::None => map.create()?,
+            };
             if !map.obj.data.is_empty() && map.obj.name != ".bss" {
                 bpf_map_update_elem_ptr(fd, &0 as *const _, map.obj.data.as_mut_ptr(), 0).map_err(
                     |(code, io_error)| MapError::SyscallError {
@@ -208,7 +260,6 @@ impl Bpf {
                 (name, program)
             })
             .collect();
-
         Ok(Bpf {
             maps: maps
                 .drain(..)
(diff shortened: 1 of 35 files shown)


dave-tucker:
This needs some tests, and some discussion on how the API for this will work.

alessandrod:
Thanks for getting this started! 

The code so far makes sense. I think before we decide on an API, we should think about how we would implement loading pinned maps and programs at `load()` time. Not necessarily implement it, but at least come up with a plan to make sure that the whole pinning story feels consistent.

dave-tucker:
@alessandrod I had a go at implementing this for maps

I think this should work as follows:

1. You set the `pinning u32` on your map definition to 1, which libbpf interprets as `pin this map by name`
2. On load, you provide a map pin path, which i'm currently assuming is an absolute path to a location on bpffs (e.g `/sys/fs/bpf/myapp`)
3. For each map, if the `pinning u32` == `pin by name`, we first look on bpffs to see if the map exists... if it does, we use bpf_obj_get to get a new fd and use that to perform relocations, otherwise we create the map as usual, and call `pin` to pin it to the location on bpffs.

For programs, this should be easier, as I think exposing `program.pin("/sys/fs/bpffs/myapp/prog_myapp")` is sufficient.

a couple of things of note in the implementation so far:

1. I need to implement bpf_obj_get and not use fs::Open
2. Using `Option<P>` where `P: AsRef<Path>` is horrible as the `None` type needs to be annotated :vomiting_face:

alessandrod:
> @alessandrod I had a go at implementing this for maps

Amazing! 🙏 This is actually less invasive than I thought! See some generic comments below, i'll do a proper review once you think the code is ready.

> I think this should work as follows:
> 
>     1. You set the `pinning u32` on your map definition to 1, which libbpf interprets as `pin this map by name`
> 
>     2. On load, you provide a map pin path, which i'm currently assuming is an absolute path to a location on bpffs (e.g `/sys/fs/bpf/myapp`)
> 
>     3. For each map, if the `pinning u32` == `pin by name`, we first look on bpffs to see if the map exists... if it does, we use bpf_obj_get to get a new fd and use that to perform relocations, otherwise we create the map as usual, and call `pin` to pin it to the location on bpffs.

Yup this all makes sense.

  
>     2. Using `Option<P>` where `P: AsRef<Path>` is horrible as the `None` type needs to be annotated 🤮

I think it's time we move to a builder-like API for loading, which would avoid the problem above entirely.  I think we should:

- have `Bpf::load(slice)` for loading from memory, using default BTF info if available
- have `Bpf::load_file(path)` for loading from file, using default BTF info if available
- create a new `BpfLoader` type that would allow fine grained control over all the loading options, eg:
```rust
let bpf = BpfLoader::new()
    .btf(path_to_btf)
    .map_pin_path(path_to_bpffs_dir)
    .load(buf)?; // or load_file(path)
```

I think this would at the same time simplify the API for the simple case (load() and load_file()), and give us a powerful and future proof API for more complex cases.

dave-tucker:
@alessandrod this is ready for review/testing now - i'll try it out myself tomorrow, but figured i'd tag you as you may get to it before me. I added the `BpfLoader` to `aya` as you suggested... and as a bonus I added a new `MapBuilder` to `aya-bpf` to simplify the creation of maps and passing options (like max_entries, pinning etc...) from Bpf code, eg.

```rust
let arr : Array<u32> = MapBuilder::new().max_entries(1024).pinning(BpfPinningType::PinByName).build();
```

aya/src/maps/mod.rs
```diff
@@ -166,6 +170,24 @@ impl Map {
             name: self.obj.name.clone(),
         })
     }
+
+    pub fn pin<P: AsRef<Path>>(&self, path: P) -> Result<(), MapError> {
+        let fd = self.fd_or_err()?;
+        let path_string = match CString::new(path.as_ref().to_str().unwrap()) {
```
alessandrod, line 176: I think it would be better to use `to_string_lossy` here than unwrap. Also use `map_err` please instead of matching ok and err.

aya/src/maps/mod.rs
```diff
@@ -128,6 +131,7 @@ pub enum MapError {
 pub struct Map {
     pub(crate) obj: obj::Map,
     pub(crate) fd: Option<RawFd>,
+    pub pinned: bool,
```
alessandrod, line 134: We don't seem to be reading this from anywhere, and we only set it to true when `bpf_object_get`-ting a map, but not after pinning it with `pin()`. Unless I'm missing something, I'd say let's remove this and we can add it back when and if we need it?

aya/src/programs/mod.rs
```diff
@@ -136,6 +137,9 @@ pub enum ProgramError {
     #[error("unexpected program type")]
     UnexpectedProgramType,
 
+    #[error("invalid pin path `{error}`")]
+    InvalidPin { error: String },
```
alessandrod, line 141: This is called `InvalidPin` and the map one is called `InvalidPath`. I think we should use the same name, perhaps `InvalidPinPath`?

aya/src/bpf.rs
```diff
@@ -58,7 +60,243 @@ pub(crate) struct bpf_map_def {
...
+}
+
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub(crate) enum BpfPinningType {
+    PinNone = 0,
+    PinByName = 1,
+}
+
+impl TryFrom<u32> for BpfPinningType {
```
alessandrod, line 73: I don't think this is ever used? If so I would remove it and remove `UnexpectedPinningType` too

aya/src/bpf.rs
```diff
@@ -58,7 +60,243 @@ pub(crate) struct bpf_map_def {
...
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(BpfPinningType::PinNone),
+            1 => Ok(BpfPinningType::PinByName),
+            n => Err(BpfError::UnexpectedPinningType { name: n }),
+        }
+    }
+}
+
+impl Default for BpfPinningType {
```
alessandrod, line 85: Samme for this I don't think it's used

aya/src/bpf.rs
```diff
@@ -58,7 +60,243 @@ pub(crate) struct bpf_map_def {
...
+    }
+
+    // Set the target BTF
+    pub fn btf(&mut self, btf: Btf) -> &mut BpfLoader {
+        self.btf = Some(btf);
+        self
+    }
+
+    // Set the target BTF to /sys/kernel/btf/vmlinux
+    pub fn default_btf(&mut self) -> &mut BpfLoader {
```
alessandrod, line 112: I don't think we need this, I think we can let the caller use `Btf::from_sys_fs()` and let them figure out whether they want to error out if BTF is not available or not.

alessandrod (changes requested):

alessandrod:
Also please don't forget to update `Bpf::load` and `Bpf::load_file` to use the loader. I think `load` now can be:

```
 pub fn load(data: &[u8]) -> Result<Bpf, BpfError>
```

And it'll try to use BTF from sysfs if available and otherwise load without BTF. If people want to fail if BTF is not present, or want to use custom BTF data, they can use the loader directly.

aya/src/bpf.rs
dave-tucker: This is needed because we have `#[derive(Default)]` on `bpf_map_def`. As `BpfPinningType` is a member, we need an `impl Default` as we can't derive one.

tw4452852:
Sorry for chime in. Just a minor question, are we able to create a map from a pinned map with this patch? I think this is very useful for sharing maps between different eBPF programs.

aya/src/bpf.rs
alessandrod: Oh yep you're right

aya/src/maps/mod.rs
dave-tucker: good catch, i actually meant to include it in the `map.pin` logic as a guard against calling `pin` if we've already called it once before. I'd think this could happen if you wanted to share the same map between two programs that shared the same loader.
i've updated the pin logic to make use of it...

dave-tucker:
@tw4452852 can you explain your use case a bit more?

`MapBuilder` is a convenience for map construction from within the BPF program. If your goal was to share map definitions between programs I think you could share them in a crate.

If you wanted to include a pinned map that, wasn't defined in the program you are loading, but you wanted to interact with it from userspace, then I think we could probably add a `BpfLoader::new().with_pinned_map("/path/to/map")` function in a follow up PR.

tw4452852:
@dave-tucker 

I think what I want is demonstrated in this [example](https://github.com/foniod/redbpf/commit/6d156fbde60166dca11c6acae52b55f2690e2e91) with [redbpf::Map::from_pin_file](https://github.com/foniod/redbpf/blob/9b3afe57158e6214b1a2049c5cce1cfca79d13cf/redbpf/src/lib.rs#L1468).

dave-tucker:
@tw4452852 ack. then I think the the answer is "yes" this PR will support that use case assuming that you use the same loader E.g:

```rust
let loader = BpfLoader::new().map_pin_path("/sys/fs/bpf/myapp");
// assuming our map is called foo, this will create /sys/fs/bpf/myapp/foo
let prog1 = loader.load_file("prog1.o")?;
// assuming this program also wants to use the map foo, it will use the pinned map
let prog1 = loader.load_file("prog2.o")?;
```

dave-tucker:
@alessandrod the BPF part of this PR is a no-go.
The builder pattern won't work as maps are static, and I can't make the trait functions const :cry: 

I'm going to go with "add another parameter to `with_max_entries`" for now, which isn't very nice, but that will at least prove this works. Then we can figure out a cleaner API for creating maps with options - possibly via macro.

alessandrod:
> @tw4452852 ack. then I think the the answer is "yes" this PR will support that use case assuming that you use the same loader E.g:
> 
> ```rust
> let loader = BpfLoader::new().map_pin_path("/sys/fs/bpf/myapp");
> // assuming our map is called foo, this will create /sys/fs/bpf/myapp/foo
> let prog1 = loader.load_file("prog1.o")?;
> // assuming this program also wants to use the map foo, it will use the pinned map
> let prog1 = loader.load_file("prog2.o")?;
> ```

I think there's a case for not going through the loader at all, eg if you're implementing a headless UI for some BPF stuff running in background, where you load the UI and start pulling out data without loading any programs at all.

I think we'll need a separate method for that, but if anyone wants to have a go at it, let's definitely do it in a separate PR please.

tw4452852:
> assuming

Thanks @dave-tucker 

What if there are two different loaders, is it possible to share a pinned map?

dave-tucker:
ok @alessandrod this should now be ready to go for maps.
I've tested it with the xdp packet logging example from the book, and I see the map get pinned to `/sys/fs/bpf/xdpfw/EVENTS`, the maps persists after the program has exited, and i'm able to load it again re-using the existing map.

program pinning works too, but the API is a bit ugly:
```rust
let probe: &mut Xdp = bpf.program_mut("xdp")?.try_into()?;
probe.load()?;
probe.attach(&iface, XdpFlags::default())?;
let program: &mut Program = bpf.program_mut("xdp")?;
program.pin("/sys/fs/bpf/xdpfw/myapp")?;
```

aya/src/maps/mod.rs
```diff
@@ -153,6 +167,31 @@ impl Map {
         Ok(fd)
     }
 
+    pub fn from_pinned<P: AsRef<Path>>(&mut self, path: P) -> Result<RawFd, MapError> {
```
alessandrod, line 170: I think this shouldn't be pub for now.

At some point I think we'll want to expose something like `let map = Map::from_pinned(path)` to load a pinned map before going through the loader (and without having a `bpf_map_def` for the map).

aya/src/maps/mod.rs
```diff
@@ -166,6 +205,29 @@ impl Map {
...
     }
+
+    pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), MapError> {
+        if self.pinned {
+            return Err(MapError::AlreadyPinned {
+                name: self.name().to_string(),
+            });
+        }
+        let map_path = path.as_ref().join(self.name());
+        println!("in map.pin, trying to get the fd");
```
alessandrod, line 216: remove the print please

aya/src/maps/mod.rs
```diff
@@ -166,6 +205,29 @@ impl Map {
             name: self.obj.name.clone(),
         })
     }
+
+    pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), MapError> {
```
alessandrod, line 209: This needs docs and an example. (I know the other map methods don't have docs, but this is the first method you'd actually use from the generic `Map` type.)

aya/src/bpf.rs
```diff
@@ -58,64 +58,92 @@ pub(crate) struct bpf_map_def {
...
 
-/// The main entry point into the library, used to work with eBPF programs and maps.
-#[derive(Debug)]
-pub struct Bpf {
-    maps: HashMap<String, MapLock>,
-    programs: HashMap<String, Program>,
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub(crate) enum BpfPinningType {
+    PinNone = 0,
```
alessandrod, line 67: Should the variants be `None` and `ByName`? I'm not sure it makes sense to repeat the prefix for each variant.

aya/src/bpf.rs
```diff
@@ -58,64 +58,92 @@ pub(crate) struct bpf_map_def {
...
 }
 
-/// The main entry point into the library, used to work with eBPF programs and maps.
-#[derive(Debug)]
-pub struct Bpf {
-    maps: HashMap<String, MapLock>,
-    programs: HashMap<String, Program>,
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub(crate) enum BpfPinningType {
```
alessandrod, line 66: I think this should be `PinningType`. It makes sense to have the prefix for `BpfLoader` since you `use aya::BpfLoader` and it's possible that you have another `Loader` symbol in the module.

bpf/aya-bpf/src/maps/mod.rs
```diff
@@ -1,3 +1,6 @@
+pub(crate) const PIN_NONE: u32 = 0;
```
alessandrod, line 1: I think this could be a `#[repr(u32)] enum`?

alessandrod (changes requested):
Thank you so much for this work Dave! I left a few minor comments then it's ready to go.

aya/src/maps/mod.rs
dave-tucker: done

aya/src/maps/mod.rs
dave-tucker: ack, made it `pub(crate)` instead

aya/src/maps/mod.rs
dave-tucker: I've made this `pub(crate)` for now too since it's only ever called from within a `load` or `load_file`.
As such, I haven't added any docs as it's not a public API, but I can do if you'd like me to.

aya/src/bpf.rs
dave-tucker: done

aya/src/bpf.rs
dave-tucker: done

bpf/aya-bpf/src/maps/mod.rs
dave-tucker: done

alessandrod (approved):

PR #70, opened 2021-10-18
Implement HashOfMaps and ArrayOfMaps

dave-tucker:
This PR adds user and kernel space implementations of HashOfMaps and ArrayOfMaps.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/70)

### DIFF
--- a/test/integration-test/bpf/progarray.bpf.c
+++ b/test/integration-test/bpf/progarray.bpf.c
@@ -1,0 +1,26 @@
+// clang-format off
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+// clang-format on
+
+struct {
+  __uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+  __uint(key_size, sizeof(uint32_t));
+  __uint(value_size, sizeof(uint32_t));
+  __uint(max_entries, 2);
+  __array(values, int());
+} jump_table SEC(".maps") = {
+    .values =
+        {
+            [1] = &xdp_pass,
+        },
+};
+
+SEC("xdp")
+int prog_array_test(struct xdp_md *ctx) {
+  bpf_tail_call(ctx, &jump_table, 1);
+  return XDP_ABORTED;
+}
+
+SEC("xdp")
+int xdp_pass(struct xdp_md *ctx) { return XDP_PASS; }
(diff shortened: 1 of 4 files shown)


alessandrod:
Thanks for taking this up!

I only skimmed through the changes. The user space side looks ok, the ebpf side I think should probably be a bit different. Right now maps of maps return an integer from `get()` (which is presumably a map id). I think they should return typed maps, eg:

```rust
#[map]
static mut OUTER: ArrayOfMaps<HashMap<X, Y>> = ...

fn some_probe(...) -> Result {
    let inner_map: HashMap<X, Y> = unsafe { OUTER.get(0).ok_or(())? };
    inner_map.insert(x_value, y_value);
    ...
}
```

This will also require adding some kind of `from_id(u32)` method to all maps, and we'll have to refactor maps to remove the assumption that we always have a `bpf_map_def` in `self.def`.

jeromegn:
This is of interest to us!

dave-tucker:
@jeromegn awesome! i'll get back to this after #127... and probably after #151 since may as well handle internal changes to fetching internal maps/programs by id together... 

@alessandrod I haven't looked into the kernel side yet, but are all maps in the array/hashmap constrained to being of the same type? if not, it won't be possible to return a strongly typed wrapper....

kakkoyun:
@dave-tucker Any interest in continuing on this? 
We actually need it and are happy to contribute. 

> I haven't looked into the kernel side yet, but are all maps in the array/hashmap constrained to being of the same type? if not, it won't be possible to return a strongly typed wrapper....

If I'm not mistaken, all the values are the same for the map of maps. They're refs u32.

astoycos:
I'd like to take a stab at getting this up to day/ finishing if need be! WDYT @dave-tucker?

kckeiks:
Our project could use this feature. Any updates? Glad to lend a hand in finishing this up if needed.

alessandrod:
> Our project could use this feature. Any updates? Glad to lend a hand in finishing this up if needed.

I've needed this too recently so I'd like to get it landed. We've changed the implementation of maps so the code in this PR needs to be updated. Getting it to compile/work would be a good first step I guess?

anfedotoff:
HI, @dave-tucker, thank you for maintaining such a cool project!
Can you  tell me, please, if you still have plans to implement HashOfMaps and ArrayOfMaps? These are very useful map types, I think.  I would be happy to use them in my project:).

vadorovsky:
Superseded by #1446

PR #78, opened 2021-10-27
aya: fix include_bytes_aligned! macro to work in some corner cases

willfindlay:
I found a corner case in my own development workflow that caused the existing macro to not work properly. The following changes appear to fix things.

Ideally, we could add some test cases to CI to prevent regressions. This would require creating a dedicated directory to
hold test BPF programs so that we can "include" them at compile time.

Kind of off-topic but still relevant: A nice side effect of having such a directory of test programs is that it would allow us to potentially unit test aya's program loading/attaching functionality in the future, independently of the underlying aya-bpf implementation. We are doing something similar with libbpf-rs. @alessandrod WDYT?

### DIFF
--- a/aya/src/util.rs
+++ b/aya/src/util.rs
@@ -125,9 +125,12 @@ pub(crate) fn tc_handler_make(major: u32, minor: u32) -> u32 {
 #[macro_export]
 macro_rules! include_bytes_aligned {
     ($path:literal) => {{
+        #[repr(align(32))]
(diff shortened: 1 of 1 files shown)


KirillLykov:
who remembers why 32b alignment has been chosen? My maybe naive understanding is that object when reads file checks the elf type in header of elf file (`Class:                             ELF64` typically) and assumes that the alignment matches (8b in this case).

PR #83, opened 2021-10-28
aya-bpf: introduce argument coercion

willfindlay:
This PR introduces argument coercion for pt_regs and BTF program types and implements it for kprobes, uprobes, LSM programs, and BTF tracepoints.

You can see a [more concrete example of using this API here][1].

Features:
  - Uses a FromBtfArgs trait with blanket implementations to enable coercion to both primitive types _and_ raw pointers
  - No need for the user to perform any kind of manual casting on the BPF side
  - Similar logic for pt_regs programs, abstracting over the PT_REGS access

TODO:
  - Need to support architectures other than x86 for k/uprobe arguments (bindings need to be fixed)
  
While I was at it, I modified the Cargo.toml file of aya-bpf-macros to fix broken doctests. Perhaps we should add doctests to the CI workflow?

Closes #21

[1]: https://github.com/willfindlay/aya-lsm-args-example

### DIFF
--- a/bpf/aya-bpf/src/args.rs
+++ b/bpf/aya-bpf/src/args.rs
@@ -1,0 +1,186 @@
+use crate::cty::c_void;
+
+// aarch64 uses user_pt_regs instead of pt_regs
+#[cfg(not(bpf_target_arch = "aarch64"))]
+use crate::bindings::pt_regs;
+#[cfg(bpf_target_arch = "aarch64")]
+use crate::bindings::user_pt_regs as pt_regs;
+
+/// A trait that indicates a valid type for an argument which can be coerced from a BTF
+/// context.
+///
+/// Users should not implement this trait.
+///
+/// SAFETY: This trait is _only_ safe to implement on primitive types that can fit into
+/// a `u64`. For example, integers and raw pointers may be coerced from a BTF context.
+pub unsafe trait FromBtfArgument: Sized {
+    /// Coerces a `T` from the `n`th argument from a BTF context where `n` starts
+    /// at 0 and increases by 1 for each successive argument.
+    ///
+    /// SAFETY: This function is deeply unsafe, as we are reading raw pointers into kernel
+    /// memory. In particular, the value of `n` must not exceed the number of function
+    /// arguments. Moreover, `ctx` must be a valid pointer to a BTF context, and `T` must
+    /// be the right type for the given argument.
+    unsafe fn from_argument(ctx: *const c_void, n: usize) -> Self;
+}
+
+unsafe impl<T> FromBtfArgument for *const T {
+    unsafe fn from_argument(ctx: *const c_void, n: usize) -> *const T {
+        // BTF arguments are exposed as an array of `usize` where `usize` can
+        // either be treated as a pointer or a primitive type
+        *(ctx as *const usize).add(n) as _
+    }
+}
+
+/// Helper macro to implement [`FromBtfArgument`] for a primitive type.
+macro_rules! unsafe_impl_from_btf_argument {
+    ($type:ident) => {
+        unsafe impl FromBtfArgument for $type {
+            unsafe fn from_argument(ctx: *const c_void, n: usize) -> Self {
+                // BTF arguments are exposed as an array of `usize` where `usize` can
+                // either be treated as a pointer or a primitive type
+                *(ctx as *const usize).add(n) as _
+            }
+        }
+    };
+}
+
+unsafe_impl_from_btf_argument!(u8);
+unsafe_impl_from_btf_argument!(u16);
+unsafe_impl_from_btf_argument!(u32);
+unsafe_impl_from_btf_argument!(u64);
+unsafe_impl_from_btf_argument!(i8);
+unsafe_impl_from_btf_argument!(i16);
+unsafe_impl_from_btf_argument!(i32);
+unsafe_impl_from_btf_argument!(i64);
+unsafe_impl_from_btf_argument!(usize);
+unsafe_impl_from_btf_argument!(isize);
+
+/// A trait that indicates a valid type for an argument which can be coerced from
+/// a pt_regs context.
+///
+/// Any implementation of this trait is strictly architecture-specific and depends on the
+/// layout of the underlying pt_regs struct and the target processor's calling
+/// conventions. Users should not implement this trait.
+pub trait FromPtRegs: Sized {
+    /// Coerces a `T` from the `n`th argument of a pt_regs context where `n` starts
+    /// at 0 and increases by 1 for each successive argument.
+    fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self>;
+
+    /// Coerces a `T` from the return value of a pt_regs context.
+    fn from_retval(ctx: &pt_regs) -> Option<Self>;
+}
+
+#[cfg(bpf_target_arch = "x86_64")]
+impl<T> FromPtRegs for *const T {
+    fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self> {
+        match n {
+            0 => ctx.rdi().map(|v| v as _),
+            1 => ctx.rsi().map(|v| v as _),
+            2 => ctx.rdx().map(|v| v as _),
+            3 => ctx.rcx().map(|v| v as _),
+            4 => ctx.r8().map(|v| v as _),
+            5 => ctx.r9().map(|v| v as _),
+            _ => None,
+        }
+    }
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.rax().map(|v| v as _)
+    }
+}
+
+#[cfg(bpf_target_arch = "armv7")]
+impl<T> FromPtRegs for *const T {
+    fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self> {
+        if n <= 6 {
+            ctx.uregs().map(|regs| regs[n] as _)
+        } else {
+            None
+        }
+    }
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.uregs().map(|regs| regs[0] as _)
+    }
+}
+
+#[cfg(bpf_target_arch = "aarch64")]
+impl<T> FromPtRegs for *const T {
+    fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self> {
+        if n <= 7 {
+            ctx.regs().map(|regs| regs[n] as _)
+        } else {
+            None
+        }
+    }
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.regs().map(|regs| regs[0] as _)
+    }
+}
+
+/// Helper macro to implement [`FromPtRegs`] for a primitive type.
+macro_rules! impl_from_pt_regs {
+    ($type:ident) => {
+        #[cfg(bpf_target_arch = "x86_64")]
+        impl FromPtRegs for $type {
+            fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self> {
+                match n {
+                    0 => ctx.rdi().map(|v| v as _),
+                    1 => ctx.rsi().map(|v| v as _),
+                    2 => ctx.rdx().map(|v| v as _),
+                    3 => ctx.rcx().map(|v| v as _),
+                    4 => ctx.r8().map(|v| v as _),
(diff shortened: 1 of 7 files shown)


bpf/aya-bpf/src/btf_args.rs
```diff
@@ -0,0 +1,46 @@
...
+    ///
+    /// SAFETY: This function is deeply unsafe, as we are reading raw pointers into kernel memory.
+    /// In particular, the value of `n` must not exceed the number of function arguments.
+    /// Luckily, the BPF verifier will catch this for us.
+    unsafe fn from_argument(ctx: *const c_void, position: usize) -> Self;
+}
+
+unsafe impl<T> FromBtfArgument for *const T {
+    unsafe fn from_argument(ctx: *const c_void, position: usize) -> *const T {
+        *(ctx as *const *const c_void).add(position * 8) as _
```
alessandrod, line 22: I'm not sure I understand, shouldn't it be `ctx.add((1 + position) * mem::size_of::<*const c_void>()) as _`?

willfindlay: Good catch, I'll fix it

willfindlay: @alessandrod I've updated this to be more semantically meaningful. While I was at it I just switched to use ptr.offset instead of ptr.add. Much simpler and more correct.

willfindlay:
@alessandrod Any thoughts on how this looks from an API perspective? Once I have the green light I will add supports for `pt_regs`-style contexts. Tracepoints are a bit trickier, since their contexts are actually just interpreted as a pointer to a tracepoint-specific C struct, so the API might diverge a bit there.

alessandrod:
It generally looks great! I think you should go ahead and try it with `pt_regs` as that will inform the design more. For example I think you'll need something like `FromBtfArgument` there too, so that might have to be renamed. 

Then I'd probably call it `ctx.arg(n)` instead of argument, for consistency with `std::env::args()`, `.arg()` from clap etc.

willfindlay:
@alessandrod I've added support for pt_regs probes. I have some example code over at https://github.com/willfindlay/aya-kprobe-arg-example demonstrating the API on a try_to_wake_up kprobe.

I'll definitely squash these commits before we land this, but I'm leaving the history mostly as-is for now to make things easier.

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,186 @@
...
+/// a `u64`. For example, integers and raw pointers may be coerced from a BTF context.
+pub unsafe trait FromBtfArgument: Sized {
+    /// Coerces a `T` from the `n`th argument from a BTF context where `n` starts
+    /// at 0 and increases by 1 for each successive argument.
+    ///
+    /// SAFETY: This function is deeply unsafe, as we are reading raw pointers into kernel
+    /// memory. In particular, the value of `n` must not exceed the number of function
+    /// arguments. Moreover, `ctx` must be a valid pointer to a BTF context, and `T` must
+    /// be the right type for the given argument.
+    unsafe fn from_argument(ctx: *const c_void, n: isize) -> Self;
```
alessandrod, line 24: Why is `n: isize` and not `usize`? (here and in other places)

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.uregs().map(|regs| regs[14] as _)
+    }
+}
+
+#[cfg(bpf_target_arch = "aarch64")]
+impl<T> FromPtRegs for *const T {
+    fn from_argument(ctx: &pt_regs, n: isize) -> Option<Self> {
+        if n <= 5 {
```
alessandrod, line 111: Shouldn't this be `if n <= 6`? (r0-r7)

bpf/aya-bpf/src/programs/lsm.rs
```diff
@@ -10,6 +10,50 @@ impl LsmContext {
     pub fn new(ctx: *mut c_void) -> LsmContext {
         LsmContext { ctx }
     }
+
+    /// Coerces a `T` from the `n`th argument to the LSM hook where `n` starts at 0 and
```
alessandrod, line 14: I think this is a bit too technical, I'd rephrase it as:

```
/// Returns the `n`th argument passed to the LSM hook, starting from 0.
```

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -12,6 +18,52 @@ impl ProbeContext {
             regs: ctx as *mut pt_regs,
         }
     }
+
+    /// Coerces a `T` from the `n`th argument to the probed function where `n` starts at
```
alessandrod, line 22: And this as:

```
Returns the `n`th argument passed to the probed function, starting from 0.
```

bpf/aya-bpf/src/programs/tp_btf.rs
```diff
@@ -10,6 +10,43 @@ impl BtfTracePointContext {
     pub fn new(ctx: *mut c_void) -> BtfTracePointContext {
         BtfTracePointContext { ctx }
     }
+
+    /// Coerces a `T` from the `n`th argument to the LSM hook where `n` starts at 0 and
```
alessandrod, line 14:
```
Returns the `n`th argument passed to the tracepoint, starting from 0.
```

alessandrod (changes requested):

bpf/aya-bpf/src/args.rs
willfindlay: isize is needed for `ptr::offset`

willfindlay: other places is just for api consistency

bpf/aya-bpf/src/programs/lsm.rs
willfindlay: Sure

bpf/aya-bpf/src/args.rs
willfindlay: I suppose we could also accept a usize and just cast to an isize here though, up to you

alessandrod: That's an implementation detail leaking out tho, we should take `usize` since you can't get negative arguments, and cast as `isize` in the impl. We don't allow large usize values anyway.

bpf/aya-bpf/src/args.rs
willfindlay: libbpf only defines PT_REGSPARM0 up to PT_REGS_PARM5, so I did the same here. See https://github.com/libbpf/libbpf/blob/17d7f04e7cb8cca664dfccd27c3848d105c56b5e/src/bpf_tracing.h#L200-L222. Are these extra registers documented somewhere?

bpf/aya-bpf/src/programs/probe.rs
willfindlay: Sure

bpf/aya-bpf/src/args.rs
alessandrod: if we go the libbpf route, then it should be `n <= 4`. See https://en.wikipedia.org/wiki/Calling_convention

willfindlay:
@alessandrod I've applied your documentation nits, changed all signatures to accept a `usize` rather than an `isize`, and also changed the maximum allowed `n` for aarch64 and armv7. I mostly followed https://man7.org/linux/man-pages/man2/syscall.2.html (see the section labelled Architecture calling convention).  Hopefully, I haven't gotten anything wrong, since we're kind of going off the reservation w.r.t. libbpf.

P.S. I'm noticing a pattern where PRs I work on after 2am all have off-by-one errors, so maybe it's time to rethink my sleeping habits :)

willfindlay:
Ah clippy doesn't like casting for offsets, I'll refactor.

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,196 @@
...
+        // "Architecture calling conventions"
+        if n <= 6 {
+            ctx.uregs().map(|regs| regs[n] as _)
+        } else {
+            None
+        }
+    }
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.uregs().map(|regs| regs[14] as _)
```
alessandrod, line 108: I think this should be `regs[0]`?

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.rax().map(|v| v as _)
+    }
+}
+
+#[cfg(bpf_target_arch = "armv7")]
+impl<T> FromPtRegs for *const T {
+    fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self> {
+        // See https://man7.org/linux/man-pages/man2/syscall.2.html under
```
alessandrod, line 98: I don't think this link is too relevant - kprobes can execute in non-syscall contexts, and uprobes use `ProbeContext` too

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,196 @@
...
+        // "Architecture calling conventions"
+        if n <= 7 {
+            ctx.regs().map(|regs| regs[n] as _)
+        } else {
+            None
+        }
+    }
+
+    fn from_retval(ctx: &pt_regs) -> Option<Self> {
+        ctx.regs().map(|regs| regs[30] as _)
```
alessandrod, line 125: I think this should be `regs[0]` too

bpf/aya-bpf/src/programs/probe.rs
```diff
@@ -12,6 +18,50 @@ impl ProbeContext {
...
+    ///     // Do something with pid or something else with tp
+    ///
+    ///     Ok(0)
+    /// }
+    /// ```
+    pub fn arg<T: FromPtRegs>(&self, n: usize) -> Option<T> {
+        T::from_argument(unsafe { &*self.regs }, n)
+    }
+
+    /// Coerces a `T` from the return value of the probed function.
```
alessandrod, line 47: I'd phrase this as:

```
Returns the return value of the probed function.
```

alessandrod (changes requested):
I'm starting to think that we should probably not try to implement `ProbeContext::ret()` at all. The registers used for return values depend on the return type (eg int vs float), and can actually span multiple registers.

bpf/aya-bpf/src/args.rs
willfindlay: Ah whoops good catch

bpf/aya-bpf/src/args.rs
willfindlay: Right but this at least appears to be the calling convention they are referencing in libbpf so I figured this would be the best place to consult for pt_regs-related accesses beyond the scope of what libbpf does (just the first five arguments). Alternatively, we could just truncate everything to 5 arguments like libbpf does and call it a day.

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,196 @@
...
+                if n <= 7 {
+                    ctx.regs().map(|regs| regs[n] as _)
+                } else {
+                    None
+                }
+            }
+
+            fn from_retval(ctx: &pt_regs) -> Option<Self> {
+                ctx.regs().map(|regs| regs[30] as _)
+            }
```
alessandrod, line 182: this should be `regs[0]`

bpf/aya-bpf/src/args.rs
```diff
@@ -0,0 +1,196 @@
...
+                // "Architecture calling conventions"
+                if n <= 6 {
+                    ctx.uregs().map(|regs| regs[n] as _)
+                } else {
+                    None
+                }
+            }
+
+            fn from_retval(ctx: &pt_regs) -> Option<Self> {
+                ctx.uregs().map(|regs| regs[14] as _)
```
alessandrod, line 164: This should be `regs[0]`

alessandrod (changes requested):
Oops we missed a couple. This should be the last one 🤞

willfindlay:
> Oops we missed a couple. This should be the last one 🤞

Ah whoops. I'll have some time to fix this later tonight!

willfindlay:
@alessandrod should be fixed now

alessandrod (approved):

alessandrod:
Great work! 🙏

PR #86, opened 2021-10-29
aya-bpf/helpers: add documentation and implement more bpf_probe_read_* wrappers

willfindlay:
This patch adds some documentation to aya-bpf/helpers and adds documentation
for the module itself and for all of the wrappers currently defined in the module.

It also implements the rest of the bpf_probe_read_* wrappers that were missing from this
file. In the future, it probably also makes sense to add some bpf_probe_read_* wrappers
that can read directly into a map pointer, avoiding the BPF stack altogether. I'm going to
call this out of scope for this PR, but plan to submit a subsequent one that addresses
this use case.

### DIFF
--- a/bpf/aya-bpf/src/helpers.rs
+++ b/bpf/aya-bpf/src/helpers.rs
@@ -1,3 +1,12 @@
+//! This module contains kernel helper functions that may be exposed to specific BPF
+//! program types. These helpers can be used to perform common tasks, query and operate on
+//! data exposed by the kernel, and perform some operations that would normally be denied
+//! by the BPF verifier.
+//!
+//! Here, we provide some higher-level wrappers around the underlying kernel helpers, but
+//! also expose bindings to the underlying helpers as a fall-back in case of a missing
+//! implementation.
+
 use core::mem::{self, MaybeUninit};
 
 pub use aya_bpf_bindings::helpers as gen;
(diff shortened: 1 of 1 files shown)


bpf/aya-bpf/src/helpers.rs
```diff
@@ -20,6 +40,89 @@ pub unsafe fn bpf_probe_read<T>(src: *const T) -> Result<T, c_long> {
     Ok(v.assume_init())
 }
 
+/// Read bytes stored at the _userspace_ pointer `src` and store them as a `T`.
+///
+/// The size of `T` should be the same as the size of whatever is behind `src`. Typically,
+/// `T` would be a binding such as a struct or primitive that directly maps to the pointee
+/// of `src`.
+///
```
alessandrod, line 48: The description is valid, but I think it's a bit cryptic. I'd say something like:

```
/// Reads the memory pointed by the _user space_ pointer `src`.
///
/// Returns a bitwise copy of `mem::size_of::<T>()` bytes stored at the user space address `src`. See 
/// `bpf_probe_read_kernel` for  reading kernel space memory.
///
/// # Example
///
/// ...a one liner here...
///
/// # Errors
/// 
/// On failure, returns a negative error code. 
```

alessandrod (commented):
Thanks a lot for this!

See my comment - I'd rephrase the doc a bit, add examples and put error values in an #Error section for consistency with aya user space.

alessandrod:
>  it probably also makes sense to add some bpf_probe_read_* wrappers that can read directly into a map pointer, avoiding the BPF stack altogether. I'm going to call this out of scope for this PR, but plan to submit a subsequent one that addresses this use case.

Yes we should add wrappers that read into a user provided slice, like `bpf_probe_read_user_str` does.

alessandrod:
> `bpf_get_current_pid_tgid`

Maybe have `BpfContext::current_pid_tgid()` return a tuple?

willfindlay:
@alessandrod I've addressed your comments. Let me know what you think.

Also, I was thinking it might make more sense to just export `gen` instead of `gen::*`. This would help clearly separate our wrappers from the generated helpers and also makes the documentation look _much_ cleaner. WDYT?

bpf/aya-bpf/src/helpers.rs
```diff
@@ -20,6 +54,149 @@ pub unsafe fn bpf_probe_read<T>(src: *const T) -> Result<T, c_long> {
...
+
+    Ok(v.assume_init())
+}
+
+/// Read a null-terminated string stored at `src` into `dest`.
+///
+/// Generally speaking, the more specific [`bpf_probe_read_user_str`] and
+/// [`bpf_probe_read_kernel_str`] should be preferred over this function.
+///
+/// The length of `dest` should be the same or smaller than the length or `src`.
```
alessandrod, line 136: This is not true, the destination slice can be larger

bpf/aya-bpf/src/helpers.rs
```diff
@@ -20,6 +54,149 @@ pub unsafe fn bpf_probe_read<T>(src: *const T) -> Result<T, c_long> {
...
+    let mut len = len as usize;
+    if len > dest.len() {
+        // this can never happen, it's needed to tell the verifier that len is
+        // bounded
+        len = dest.len();
+    }
+    Ok(len as usize)
+}
+
+/// Read a null-terminated from _userspace_ stored at `src` into `dest`.
```
alessandrod, line 177: missing "string".

Here and in other places, I'd use user space, and kernel space, with a... space :P

alessandrod (approved):

alessandrod:
🙏🙏🙏

PR #91, opened 2021-10-31
Add xdp example from book to aya repo

nitishm:
Adding the source for the xdp example from the bookfor easier reference.

Signed-off-by: Nitish Malhotra <nitish.malhotra@gmail.com>


nitishm:
@willfindlay @dave-tucker @alessandrod Adding the book example source to examples directory for easier reference for a new user, instead of having to try and copy the example from the book site.

PR #94, opened 2021-11-02
aya: netlink: use NETLINK_EXT_ACK from libc crate

tklauser:
NETLINK_EXT_ACK is available since libc crate version 0.2.105, see
https://github.com/rust-lang/libc/releases/tag/0.2.105


vadorovsky (approved):

alessandrod:
Thank you! I think we need to bump the libc dep in Cargo.toml too?

tklauser:
@alessandrod thanks, I've updated the libc version in Cargo.toml as well.

PR #96, opened 2021-11-05
Fix bindings for archs other than x86

alessandrod:
This fixes binding generation for archs other than x86, and makes it possible to specify sysroot paths when running codegen on distros other than ubuntu.

### DIFF
--- a/xtask/src/codegen/mod.rs
+++ b/xtask/src/codegen/mod.rs
@@ -53,6 +53,17 @@ pub struct Options {
     #[structopt(long)]
     libbpf_dir: PathBuf,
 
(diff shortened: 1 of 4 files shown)


xtask/src/codegen/mod.rs
```diff
@@ -53,6 +53,17 @@ pub struct Options {
     #[structopt(long)]
     libbpf_dir: PathBuf,
 
+    // sysroot options. Default to ubuntu headers installed by the
+    // libc6-dev-{arm64,armel}-cross packages.
```
willfindlay, line 57: If you make this a doc comment it will show up when you run aya-gen --help. Could be useful to add docs to other flags too.

Issue #100, opened 2021-11-05
Access packet data as &[u8] from XdpContext


fcantournet:
Hi,
 
I'm trying to use `pdu` to do some packet parsing from the `XdpContext`.
This seemed to work at some point in `redbpf` examples at https://github.com/rebpf/rebpf/blob/50e235721228c1ece2c685f9357a954bd4a322d3/examples/packet_parser/src/kern.rs#L44

So I ported this function to XdpContext in aya-bpf https://sourcegraph.com/github.com/rebpf/rebpf@50e235721228c1ece2c685f9357a954bd4a322d3/-/blob/rebpf/src/libbpf.rs?L547:12#tab=def

I'm trying to access it like so :

```
#[inline(always)]
fn try_dns_snoop(ctx: XdpContext) -> Result<u32, Error> {

    let buf: &[u8] = if let Some(buf) = ctx.data_buffer() {
        buf
    } else {
        return Ok(xdp_action::XDP_PASS);
    };

    //let ether = EthernetPdu::new(buf)?;
    let first_byte = buf[0];

    let log_entry = PacketLog {
        first_byte: first_byte,
    };
    unsafe {
        EVENTS.output(&ctx, &log_entry, 0);
    }
    Ok(xdp_action::XDP_PASS)
}
```

But the verifier complains about out of packet access :
```Error: the BPF_PROG_LOAD syscall failed. Verifier output: func#0 @0
0: R1=ctx(id=0,off=0,imm=0) R10=fp0
0: (61) r2 = *(u32 *)(r1 +0)
1: R1=ctx(id=0,off=0,imm=0) R2_w=pkt(id=0,off=0,r=0,imm=0) R10=fp0
1: (15) if r2 == 0x0 goto pc+17
 R1=ctx(id=0,off=0,imm=0) R2_w=pkt(id=0,off=0,r=0,imm=0) R10=fp0
2: R1=ctx(id=0,off=0,imm=0) R2_w=pkt(id=0,off=0,r=0,imm=0) R10=fp0
2: (61) r3 = *(u32 *)(r1 +4)
3: R1=ctx(id=0,off=0,imm=0) R2_w=pkt(id=0,off=0,r=0,imm=0) R3_w=pkt_end(id=0,off=0,imm=0) R10=fp0
3: (3d) if r2 >= r3 goto pc+15
[...]

PR #101, opened 2021-11-05
xdp: Add XdpContext method to access data as &[u8]

fcantournet:
This is a port of data_buffer and data_pointer from redbpf
https://sourcegraph.com/github.com/rebpf/rebpf@50e235721228c1ece2c685f9357a954bd4a322d3/-/blob/rebpf/src/libbpf.rs?L547:12#tab=def

This allows code like this (with a fork of pdu to make it bpf compatible) : 
```
#[inline(always)]
fn try_dns_snoop(ctx: XdpContext) -> Result<u32, Error> {
    let buf: &[u8] = if let Some(buf) = ctx.data_buffer() {
        buf
    } else {
        return Ok(xdp_action::XDP_PASS);
    };
    let (_, end) = ctx.data_pointer();
    let ether = EthernetPdu::new(buf, end)?;

    match ether.ethertype() {
        EtherType::IPV4 => {
          // do stuff
        },
        _ => return Ok(xdp_action::XDP_PASS),
    }
    Ok(xdp_action::XDP_PASS)
}
```

The API extension is not strictly necessary, you can get by with `XdpContext:{data,data_end}` but this seems to me like somewhat common usecase.
see #100 for reference

### DIFF
--- a/bpf/aya-bpf/src/programs/xdp.rs
+++ b/bpf/aya-bpf/src/programs/xdp.rs
@@ -7,16 +7,42 @@ pub struct XdpContext {
 }
 
 impl XdpContext {
+
+    #[inline(always)]
     pub fn new(ctx: *mut xdp_md) -> XdpContext {
         XdpContext { ctx }
     }
 
+    #[inline(always)]
     pub fn data(&self) -> usize {
         unsafe { (*self.ctx).data as usize }
     }
 
+    #[inline(always)]
     pub fn data_end(&self) -> usize {
         unsafe { (*self.ctx).data_end as usize }
+    }
+
+    #[inline(always)]
+    pub fn data_buffer(&self) -> Option<&[u8]> {
+        unsafe {
+            let data_buffer: *const u8 = (*self.ctx).data as usize as *const u8;
+            if (*self.ctx).data_end <= (*self.ctx).data {
+                return None;
+            }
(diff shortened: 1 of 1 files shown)


dave-tucker:
Thanks for the contribution @fcantournet!
I'm not a big fan of this API, because  `data_buffer` is going to allocate a slice on the stack, but our program stack space is limited to 512 bytes (256 bytes if we're using tail calls). This means that you need to be careful on what size packets you use this on. Awesome to see that you've got pdu working though, we could definitely use something like that.

fcantournet:
I think it does allocate does it ? `core::slice::from_raw_parts` doesn't allocate if I understand correctly.
I'm logging
```
IP packet 140.82.121.3:0 => 192.168.0.34:0 BUFFSIZE: 1458
```
Where BUFFSIZE is `buf.len()` in the code above.

shaneutt:
Hi @fcantournet :wave:

In the community meeting today we were discussing some of these older PRs and wanted to check in. Did you end up implementing what you wanted in your own project, or as a separate crate somewhere? Are you still hoping to add this functionality here?

fcantournet:
Hey @shaneutt,
I'm not using aya immediately, so you can close this PR, I think the issue opened is still a valid use case but I'm not sure this is enough to achieve the goals of the issue and it might be extending the API for nothing.

PR #102, opened 2021-11-10
bpf: Add macros for sk_skb programs

dave-tucker:
This commit adds the stream_parser and stream_verdict macros for use in
bpf programs

### DIFF
--- a/bpf/aya-bpf-macros/src/expand.rs
+++ b/bpf/aya-bpf-macros/src/expand.rs
@@ -441,3 +441,53 @@ impl BtfTracePoint {
         })
     }
 }
+
+#[allow(clippy::enum_variant_names)]
+#[derive(Debug, Copy, Clone)]
+pub enum SkSkbKind {
+    StreamVerdict,
+    StreamParser,
+}
+
+impl std::fmt::Display for SkSkbKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use SkSkbKind::*;
+        match self {
+            StreamVerdict => write!(f, "stream_verdict"),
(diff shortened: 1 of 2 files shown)


bpf/aya-bpf-macros/src/expand.rs
```diff
@@ -441,3 +441,53 @@ impl BtfTracePoint {
...
+        } else {
+            format!("sk_skb/{}", kind)
+        };
+        let fn_name = &self.item.sig.ident;
+        let item = &self.item;
+        Ok(quote! {
+            #[no_mangle]
+            #[link_section = #section_name]
+            fn #fn_name(ctx: *mut ::aya_bpf::bindings::__sk_buff) -> u32 {
+                return #fn_name(::aya_bpf::programs::SkSkbContext::new(ctx));
```
willfindlay, line 487: I know StreamVerdict programs are expected to return non-zero values but is this true for StreamParsers as well? If not, maybe we should match on `self.kind` here and ignore the return value for StreamParsers.

dave-tucker: I believe that StreamVerdict programs are supposed to return the number of bytes read.
Here's an (old) example from the kernel source https://elixir.bootlin.com/linux/v4.14.46/source/samples/sockmap/sockmap_kern.c#L49

willfindlay: Cool, makes sense.

dave-tucker: yes i do lol

alessandrod (approved):

willfindlay (approved):

PR #104, opened 2021-11-10
obj: fix name parsing for sk_skb sections

dave-tucker:
This commit fixes name parsing of sk_skb sections such that both named
and unnamed variants will work correctly.

Before this patch, a section named `sk_skb/stream_parser` would result in an error

### DIFF
--- a/aya/src/obj/mod.rs
+++ b/aya/src/obj/mod.rs
@@ -146,6 +146,15 @@ impl FromStr for ProgramSection {
             }
             "socket_filter" => SocketFilter { name },
             "sk_msg" => SkMsg { name },
+            "sk_skb" => match &*name {
+                "stream_parser" => SkSkbStreamParser { name },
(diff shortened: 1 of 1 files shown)


aya/src/obj/mod.rs
```diff
@@ -146,6 +146,13 @@ impl FromStr for ProgramSection {
             }
             "socket_filter" => SocketFilter { name },
             "sk_msg" => SkMsg { name },
+            "sk_skb" => match parts[0] {
```
alessandrod, line 149:
```suggestion
            "sk_skb" => match &*name {
```

alessandrod (approved):
Thank you! See minor nit

aya/src/obj/mod.rs
dave-tucker: fixed. thanks

PR #107, opened 2021-11-16
fix: make maps compatible with kernel <= 4.14

deverton:
In kernel 4.15 and additional parameter was added to allow maps to have names but using this breaks on older kernels. This change makes it so the name is only added on kernels 4.15 and newer.

Symptoms of the problem would show up like

```
error: map error: failed to create map `IPV4_EVENTS` with code -1: Invalid argument (os error 22)`
```

and in `strace` output would look like

```
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_PERF_EVENT_ARRAY, key_size=4, value_size=4, max_entries=8, map_flags=0, inner_map_fd=0, map_name="IPV4_EVENTS", map_ifindex=0}, 128) = -1 EINVAL (Invalid argument)
```

After this patch the `strace` output looks like 

```
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_PERF_EVENT_ARRAY, key_size=4, value_size=4, max_entries=2, map_flags=0, inner_map_fd=0, map_name="", map_ifindex=0}, 128) = 9
```

### DIFF
--- a/aya/src/sys/bpf.rs
+++ b/aya/src/sys/bpf.rs
@@ -32,7 +32,7 @@ pub(crate) fn bpf_create_map(name: &CStr, def: &bpf_map_def) -> SysResult {
 
     // https://github.com/torvalds/linux/commit/ad5b177bd73f5107d97c36f56395c4281fb6f089
     // The map name was added as a parameter in kernel 4.15+ so we skip adding it on
-    // older kernels for compatibility 
+    // older kernels for compatibility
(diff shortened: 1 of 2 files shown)


deverton:
Ended up just doing the easy stub to make this work since the `kernel_version` function isn't used extensively in areas that have tests.

alessandrod:
Thank you! 🙏

PR #108, opened 2021-11-17
Support k/uprobes on older kernels.

deverton:
Prior to kernel 4.17 probes were attached via debugfs and this patch
attempts to make that work.

Tested on kernel 4.14.

### DIFF
--- a/aya/src/programs/probe.rs
+++ b/aya/src/programs/probe.rs
@@ -1,11 +1,16 @@
 use libc::pid_t;
-use std::{fs, io};
+use std::{
+    fs,
+    io::{self, Write},
+    process,
+};
 
 use crate::{
     programs::{
-        kprobe::KProbeError, perf_attach, uprobe::UProbeError, LinkRef, ProgramData, ProgramError,
+        kprobe::KProbeError, perf_attach, trace_point::read_sys_fs_trace_point_id,
+        uprobe::UProbeError, LinkRef, ProgramData, ProgramError,
     },
-    sys::perf_event_open_probe,
+    sys::{kernel_version, perf_event_open_probe, perf_event_open_trace_point},
 };
 
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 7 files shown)


aya/src/programs/probe.rs
```diff
@@ -48,14 +120,139 @@ pub(crate) fn attach(
...
+                format!(
+                    "{}:{}s/{} {}",
+                    probe_type_prefix, event_type, event_alias, fn_name
+                )
+            }
+        }
+    };
+
+    events_file
+        .write_all(p.as_bytes())
```
alessandrod, line 219: do we need to write a line here? because we aren't but then in `find_in_sys_kernel_debug_tracing_events` we iterate lines

aya/src/programs/probe.rs
```diff
@@ -23,10 +29,76 @@ pub enum ProbeKind {
...
+     * it is safe to skip the cleaning up process (write -:... to the file).
+     */
+
+    let event_type = match kind {
+        KProbe | KRetProbe => "kprobe",
+        UProbe | URetProbe => "uprobe",
+    };
+
+    let events_file_name = format!("/sys/kernel/debug/tracing/{}_events", event_type);
+
```
alessandrod, line 68: I would open the file here, then make `find_in_sys_kernel_debug_tracing_events` and `delete_in_sys_kernel_debug_tracing_events` closures

aya/src/programs/trace_point.rs
```diff
@@ -80,7 +80,7 @@ impl TracePoint {
     }
 }
 
-fn read_sys_fs_trace_point_id(category: &str, name: &str) -> Result<u32, TracePointError> {
+pub fn read_sys_fs_trace_point_id(category: &str, name: &str) -> Result<u32, TracePointError> {
```
alessandrod, line 83: this should be `pub(crate)`

alessandrod (changes requested):

aya/src/programs/probe.rs
deverton: Writing to the file is how the trace point is deleted. It seems to work (verified with strace) so I'm not sure what you mean by it not writing a line?

aya/src/programs/probe.rs
deverton: The behaviour of libbpf is to read the file with `O_RDONLY` and then close and reopen the file with `O_APPEND`. I'm not sure if that's required by the kernel interface, but that's what I was trying to emulate.

aya/src/programs/probe.rs
alessandrod: Sorry I meant we're not adding a final "\n" when we write

aya/src/programs/probe.rs
alessandrod: I would be very surprised if that was the case!

aya/src/programs/probe.rs
deverton: Oh right, newline vs line :) So the docs here https://www.kernel.org/doc/Documentation/trace/kprobetrace.txt don't seem to indicate either way, but use `echo` in  the examples which would add a new line. So I'm guessing the kernel doesn't rely on it, but it would be more consistent to add a `\n` on write since it would match the expectations from reading line-by-line.

alessandrod (approved):
Great work, thank you!

PR #109, opened 2021-11-18
Use current kernel version as default if not specified

deverton:
When a BPF program doesn't specify the target kernel version, the
most compatible option is to set the program kernel version to match
the currently running kernel.

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -400,9 +400,9 @@ fn load_program(prog_type: bpf_prog_type, data: &mut ProgramData) -> Result<(), 
     let target_kernel_version = match *kernel_version {
         KernelVersion::Any => {
             let (major, minor, patch) = crate::sys::kernel_version().unwrap();
-            KernelVersion::Version((major << 16) + (minor << 8) + patch)
+            (major << 16) + (minor << 8) + patch
         }
-        _ => *kernel_version,
+        _ => (*kernel_version).into(),
     };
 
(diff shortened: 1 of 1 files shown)


aya/src/programs/mod.rs
```diff
@@ -397,6 +397,14 @@ fn load_program(prog_type: bpf_prog_type, data: &mut ProgramData) -> Result<(),
         ..
     } = obj;
 
+    let target_kernel_version = match *kernel_version {
+        KernelVersion::Any => {
+            let (major, minor, patch) = crate::sys::kernel_version().unwrap();
+            KernelVersion::Version((major << 16) + (minor << 8) + patch)
+        }
+        _ => *kernel_version,
+    };
```
alessandrod, line 406: I don't think we need to wrap here. We can return `(major << 16) + (minor << 8) + patch` in the `Any` branch, and `v.into()` in the `Version(v)` branch

aya/src/programs/mod.rs
```diff
@@ -405,7 +413,7 @@ fn load_program(prog_type: bpf_prog_type, data: &mut ProgramData) -> Result<(),
             ty: prog_type,
             insns: instructions,
             license,
-            kernel_version: (*kernel_version).into(),
+            kernel_version: target_kernel_version.into(),
             expected_attach_type: data.expected_attach_type,
```
alessandrod, line 417: then we don't need this `into()` anymore

PR #112, opened 2021-11-20
bpf: Fix cgroup_skb macro

dave-tucker:
This ensures the macro expands to the link section named `cgroup/skb` when no attach direction is provided.
Otherwise it uses `cgroup_skb/ingress` or `cgroup_skb/egress`.
I also added some tests, hopefully these can be added to check the output of other macros.


alessandrod (approved):
Thank you! This looks great and can go in! User space doesn't seem to be parsing "cgroup/skb" so that needs to be done too but can be done separately.

PR #115, opened 2021-11-25
aya: remove unnecessary usage of &dyn trait in favor of impl trait.

eero-thia:
This should improve performance in most situations by eliminating
unnecessary fat pointer indirection.


vadorovsky (approved):

alessandrod (approved):
Thanks!

PR #117, opened 2021-11-25
aya: set .rodata maps as BPF_F_RDONLY.

eero-thia:


willfindlay:
I remember trying this at some point and it ended up breaking things since the .rodata map ends up doing some pre-flight that conflict with the BPF_F_READONLY flag. Have you had a chance to verify that this works as expected?

eero-thia:
I have not. Do you have any suggestions on how to get the compiler to reliably emit a .rodata section?

willfindlay:
> I have not. Do you have any suggestions on how to get the compiler to reliably emit a .rodata section?

Yes, you can use a static global variable in your aya-bpf code and annotate it with #[used].

eero-thia:
Thanks for teaching me about #[used]!

It looks like it's still broken, sadly. Thanks for the feedback.

alessandrod:
> Thanks for teaching me about #[used]!
> 
> It looks like it's still broken, sadly. Thanks for the feedback.

What is broken exactly?

eero-thia:
The map fails to load altogether, but I didn't inspect the exact map error. I could probably take a more detailed look at it later tonight or tomorrow.

alessandrod:
> The map fails to load altogether, but I didn't inspect the exact map error. I could probably take a more detailed look at it later tonight or tomorrow.
> 
> Constructing the map with zeroed flags doesn't exhibit this issue, so BPF_F_RDONLY is definitely the culprit.

I see. I think this is probably worth looking into - I've had that TODO about making rodata readonly for ages.

eero-thia:
I did some sleuthing today. Amusingly, in a way, it is actually working as intended.

When loading a Bpf, all non-bss map values are updated to their ELF section values via bpf_update_elem https://github.com/aya-rs/aya/blob/main/aya/src/bpf.rs#L232. If the map is created with BPF_F_RDONLY, this call *should* fail.

...I don't actually know how one is supposed to initialize a RDONLY map with a non-zero value.

PR #118, opened 2021-11-25
aya: remove RwLocks and Ref wrappers on bpf maps.

eero-thia:
This improves performance by removing the runtime checking of borrowing
rules. In order to retain a similar level of flexibility as the previous
implementation permitted, `take_map` is added as a method to Bpf,
allowing callers to take ownership of a Map.


eero-thia:
In our private branch, this has a dependency on https://github.com/aya-rs/aya/pull/120. It could be rebased to not be dependent on it, but then merge conflicts would arise in one or the other, depending on which one is merged first.

I'm leaving this as a draft until / if https://github.com/aya-rs/aya/pull/120 is approved.

alessandrod:
This generally looks good. I went back and forth a couple of times on this when I first implemented it, and I'm now convinced that for the general case having this `MapRef`/`MapRefMut` API is overkill. 

I'm planning to do a proper review after I port some heavily multi threaded code I have to this and get a better feel for it.

eero-thia:
It looks like something went horribly wrong as I tried rebasing this one, let me give it another rebase from our fork and force-push...

eero-thia:
> This generally looks good. I went back and forth a couple of times on this when I first implemented it, and I'm now convinced that for the general case having this `MapRef`/`MapRefMut` API is overkill.
> 
> I'm planning to do a proper review after I port some heavily multi threaded code I have to this and get a better feel for it.

Seeing as Bpf doesn't implement Send or Sync, I imagine that usage should be pretty similar to the single-threaded case, since your Bpf has to be behind a Mutex or RwLock, right?

shaneutt:
Hi @eero-thia :wave: 

In the community meeting today we discussed some of these older PRs. Looks like this change may still be relevant. We're curious to know what your thoughts are on the current state of this: Are you using Aya without this still in your projects and are you still in need of it?

alessandrod:
This was actually implemented as part of https://github.com/aya-rs/aya/pull/397

PR #121, opened 2021-11-25
aya: programs_mut iterator to complement programs.

eero-thia:


aya/src/bpf.rs
```diff
@@ -534,6 +534,24 @@ impl Bpf {
...
         self.programs.values()
     }
+
+    /// An iterator mutably referencing all the programs.
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # let mut bpf = aya::Bpf::load(&[])?;
+    /// for program in bpf.programs_mut() {
+    ///     println!(
```
alessandrod, line 544: I think the example should mutate the programs, so that it makes sense to iterate over mutable refs

alessandrod (approved):

Issue #51, opened 2021-08-13
lifecyle: Add a `forget()` API for `Link`


dave-tucker:
Default behaviour is that the `Drop` trait calls `detach()` when the program is terminated (or panics).
For some types of long running programs you may wish to intentionally `forget()` the fd wrapped by `Link`, so you can keep the program running after exit. Query and detach can happen out-of-band for some program types (e.g via netlink, or `bpf_prog_query`)

seanyoung:
I'm not sure it should be called `forget()`, but then again I am struggling to come up with a better name. Maybe something like `keep_program_attached()`, `disown()`, or `keep_attached()`. Maybe I'll come up with a better name.

PR #123, opened 2021-11-25
aya: remove Rc<RefCell<T>> usage in BPF program links.

eero-thia:
This is a breaking change which transfers all ownership of a link to an
eBPF program to the caller return value of the "attach" functions. This
eliminates the overhead of Rc<RefCell<T>>, as well as prevents the
links section of ProgramData from growing without bound as programs are
attached and detached.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -251,7 +251,6 @@ impl<'a> BpfLoader<'a> {
                 let data = ProgramData {
                     obj,
                     fd: None,
-                    links: Vec::new(),
                     expected_attach_type: None,
                     attach_btf_obj_fd: None,
                     attach_btf_id: None,
--- a/aya/src/maps/sock/mod.rs
+++ b/aya/src/maps/sock/mod.rs
@@ -11,3 +11,9 @@ pub use sock_map::SockMap;
 pub trait SocketMap {
     fn fd_or_err(&self) -> Result<RawFd, MapError>;
 }
+
+impl<'a, S: SocketMap> SocketMap for &'a S {
+    fn fd_or_err(&self) -> Result<RawFd, MapError> {
+        (*self).fd_or_err()
+    }
+}
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -5,11 +5,9 @@ use crate::{
         bpf_attach_type::{BPF_CGROUP_INET_EGRESS, BPF_CGROUP_INET_INGRESS},
         bpf_prog_type::BPF_PROG_TYPE_CGROUP_SKB,
     },
-    programs::{load_program, LinkRef, ProgAttachLink, ProgramData, ProgramError},
+    programs::{load_program, FdLink, OwnedLink, ProgAttachLink, ProgramData, ProgramError},
     sys::{bpf_link_create, bpf_prog_attach, kernel_version},
 };
-
-use super::FdLink;
 
 /// A program used to inspect or filter network activity for a given cgroup.
 ///
@@ -78,7 +76,7 @@ impl CgroupSkb {
         &mut self,
         cgroup: T,
         attach_type: CgroupSkbAttachType,
-    ) -> Result<LinkRef, ProgramError> {
+    ) -> Result<OwnedLink, ProgramError> {
         let prog_fd = self.data.fd_or_err()?;
         let cgroup_fd = cgroup.as_raw_fd();
 
@@ -95,7 +93,7 @@ impl CgroupSkb {
                         io_error,
                     }
                 })? as RawFd;
-            Ok(self.data.link(FdLink { fd: Some(link_fd) }))
+            Ok(FdLink { fd: Some(link_fd) }.into())
         } else {
             bpf_prog_attach(prog_fd, cgroup_fd, attach_type).map_err(|(_, io_error)| {
                 ProgramError::SyscallError {
@@ -104,9 +102,7 @@ impl CgroupSkb {
                 }
             })?;
 
-            Ok(self
-                .data
-                .link(ProgAttachLink::new(prog_fd, cgroup_fd, attach_type)))
+            Ok(ProgAttachLink::new(prog_fd, cgroup_fd, attach_type).into())
         }
     }
 }
--- a/aya/src/programs/kprobe.rs
+++ b/aya/src/programs/kprobe.rs
@@ -7,7 +7,7 @@ use crate::{
     programs::{
         load_program,
         probe::{attach, ProbeKind},
-        LinkRef, ProgramData, ProgramError,
+        OwnedLink, ProgramData, ProgramError,
     },
 };
 
@@ -65,7 +65,7 @@ impl KProbe {
     /// If the program is a `kprobe`, it is attached to the *start* address of the target function.
     /// Conversely if the program is a `kretprobe`, it is attached to the return address of the
     /// target function.
-    pub fn attach(&mut self, fn_name: &str, offset: u64) -> Result<LinkRef, ProgramError> {
+    pub fn attach(&mut self, fn_name: &str, offset: u64) -> Result<OwnedLink, ProgramError> {
         attach(&mut self.data, self.kind, fn_name, offset, None)
     }
 }
--- a/aya/src/programs/lirc_mode2.rs
+++ b/aya/src/programs/lirc_mode2.rs
@@ -2,7 +2,7 @@ use std::os::unix::prelude::{AsRawFd, RawFd};
 
 use crate::{
     generated::{bpf_attach_type::BPF_LIRC_MODE2, bpf_prog_type::BPF_PROG_TYPE_LIRC_MODE2},
-    programs::{load_program, query, Link, LinkRef, ProgramData, ProgramError, ProgramInfo},
+    programs::{load_program, query, Link, OwnedLink, ProgramData, ProgramError, ProgramInfo},
     sys::{bpf_obj_get_info_by_fd, bpf_prog_attach, bpf_prog_detach, bpf_prog_get_fd_by_id},
 };
 
@@ -60,7 +60,7 @@ impl LircMode2 {
     }
 
     /// Attaches the program to the given lirc device.
-    pub fn attach<T: AsRawFd>(&mut self, lircdev: T) -> Result<LinkRef, ProgramError> {
+    pub fn attach<T: AsRawFd>(&mut self, lircdev: T) -> Result<OwnedLink, ProgramError> {
         let prog_fd = self.data.fd_or_err()?;
         let lircdev_fd = lircdev.as_raw_fd();
 
@@ -71,7 +71,7 @@ impl LircMode2 {
             }
         })?;
 
-        Ok(self.data.link(LircLink::new(prog_fd, lircdev_fd)))
+        Ok(LircLink::new(prog_fd, lircdev_fd).into())
     }
 
     /// Queries the lirc device for attached programs.
--- a/aya/src/programs/lsm.rs
+++ b/aya/src/programs/lsm.rs
@@ -6,7 +6,7 @@ use thiserror::Error;
 use crate::{
     generated::{bpf_attach_type::BPF_LSM_MAC, bpf_prog_type::BPF_PROG_TYPE_LSM},
     obj::btf::{Btf, BtfError, BtfKind},
-    programs::{load_program, FdLink, LinkRef, ProgramData, ProgramError},
+    programs::{load_program, FdLink, OwnedLink, ProgramData, ProgramError},
     sys::bpf_raw_tracepoint_open,
 };
 
@@ -84,13 +84,13 @@ impl Lsm {
     }
 
     /// Attaches the program.
-    pub fn attach(&mut self) -> Result<LinkRef, ProgramError> {
+    pub fn attach(&mut self) -> Result<OwnedLink, ProgramError> {
         attach_btf_id(&mut self.data)
     }
 }
 
 /// Common logic for all BPF program types that attach to a BTF id.
-pub(crate) fn attach_btf_id(program_data: &mut ProgramData) -> Result<LinkRef, ProgramError> {
+pub(crate) fn attach_btf_id(program_data: &mut ProgramData) -> Result<OwnedLink, ProgramError> {
     let prog_fd = program_data.fd_or_err()?;
 
     // Attaching LSM programs doesn't require providing attach name. LSM
@@ -102,5 +102,5 @@ pub(crate) fn attach_btf_id(program_data: &mut ProgramData) -> Result<LinkRef, P
         }
     })? as RawFd;
 
-    Ok(program_data.link(FdLink { fd: Some(pfd) }))
+    Ok(FdLink { fd: Some(pfd) }.into())
 }
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -56,19 +56,18 @@ mod xdp;
 
 use libc::{close, dup, ENOSPC};
 use std::{
-    cell::RefCell,
     cmp,
     convert::TryFrom,
     ffi::{CStr, CString},
     io,
     os::unix::io::{AsRawFd, RawFd},
     path::Path,
-    rc::Rc,
 };
 use thiserror::Error;
 
 pub use cgroup_skb::{CgroupSkb, CgroupSkbAttachType};
 pub use kprobe::{KProbe, KProbeError};
+use lirc_mode2::LircLink;
 pub use lirc_mode2::LircMode2;
 pub use lsm::{Lsm, LsmLoadError};
 use perf_attach::*;
@@ -78,11 +77,14 @@ pub use raw_trace_point::RawTracePoint;
 pub use sk_msg::SkMsg;
 pub use sk_skb::{SkSkb, SkSkbKind};
 pub use sock_ops::SockOps;
+use socket_filter::SocketFilterLink;
 pub use socket_filter::{SocketFilter, SocketFilterError};
+use tc::TcLink;
 pub use tc::{SchedClassifier, TcAttachType, TcError};
 pub use tp_btf::{BtfTracePoint, BtfTracePointError};
 pub use trace_point::{TracePoint, TracePointError};
 pub use uprobe::{UProbe, UProbeError};
+use xdp::NlLink;
 pub use xdp::{Xdp, XdpError, XdpFlags};
 
 use crate::{
@@ -286,7 +288,6 @@ impl Program {
 pub(crate) struct ProgramData {
     pub(crate) obj: obj::Program,
     pub(crate) fd: Option<RawFd>,
-    pub(crate) links: Vec<Rc<RefCell<dyn Link>>>,
     pub(crate) expected_attach_type: Option<bpf_attach_type>,
     pub(crate) attach_btf_obj_fd: Option<u32>,
     pub(crate) attach_btf_id: Option<u32>,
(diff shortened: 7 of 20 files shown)


eero-thia:
In retrospect, this could actually be a drop-in change if OwnedLink is named LinkRef. My concern is that it's a little misleading of a name. On the other hand, I don't think that users will typically import and directly refer to LinkRef / OwnedLink in the first place, so perhaps this is a moot point.

alessandrod:
> This is a breaking change which transfers all ownership of a link to an eBPF program to the caller return value of the "attach" functions. This eliminates the overhead of Rc<RefCell>, 

I'm wondering is this hypothetical overhead or is it something you were able to measure? Based on this and other PRs I'm assuming you're using aya in some kind of embedded system. 

I'm not against this change, but it's not something that I'd do lightly since it does make the API harder to use, and it makes it easier to misuse it (programs can get detached silently because you failed to store or propagate a link up the stack).

Also if we do go this route, I think we should make the returned link type opaque. The way programs get attached - and so the type of link used - can change between kernel versions, and will change more in the future as more programs adopt `BPF_LINK_CREATE`. At some point in the next 6 months I think we should release 1.0, at which point we won't be able to break API easily.

> as well as prevents the links section of ProgramData from growing without bound as programs are attached and detached.

This is something that should be fixed.

eero-thia:
> I'm wondering is this hypothetical overhead or is it something you were able to measure? Based on this and other PRs I'm assuming you're using aya in some kind of embedded system.

I did not measure; the overhead introduced by the Rc<RefCell<dyn Link>> should be very minor in an "attach-once-and-forget" use case. We are indeed using aya in an embedded system, and we regularly have to attach or detach XDP programs from network interfaces, so every memory allocation - and to a lesser extent, every pointer indirection - counts.

> I'm not against this change, but it's not something that I'd do lightly since it does make the API harder to use, and it makes it easier to misuse it (programs can get detached silently because you failed to store or propagate a link up the stack).

I can empathize that this is a more contentious change. In my eyes, the fact that dropping the returned LinkRef still leaves a resource occupied (ie: the program is still attached) is not idiomatic Rust behavior, especially given that there is no way to detach it from that point on until the Bpf is dropped.

> Also if we do go this route, I think we should make the returned link type opaque. The way programs get attached - and so the type of link used - can change between kernel versions, and will change more in the future as more programs adopt BPF_LINK_CREATE. At some point in the next 6 months I think we should release 1.0, at which point we won't be able to break API easily.

Totally understandable and I agree; it should be trivial to add that on to this PR with a wrapper struct around the OwnedLink enum.

> as well as prevents the links section of ProgramData from growing without bound as programs are attached and detached.

I think that this would have to become a HashMap keyed on some unique specifier so that links with the same inputs are always slotted to the same bucket, but perhaps this is an XDP-centric perspective. Or of course get rid of it entirely as this PR does. :)

willfindlay:
> > I'm wondering is this hypothetical overhead or is it something you were able to measure? Based on this and other PRs I'm assuming you're using aya in some kind of embedded system.
> 
> I did not measure; the overhead introduced by the Rc<RefCell> should be very minor in an "attach-once-and-forget" use case. We are indeed using aya in an embedded system, and we regularly have to attach or detach XDP programs from network interfaces, so every memory allocation - and to a lesser extent, every pointer indirection - counts.
> 
> > I'm not against this change, but it's not something that I'd do lightly since it does make the API harder to use, and it makes it easier to misuse it (programs can get detached silently because you failed to store or propagate a link up the stack).
> 
> I can empathize that this is a more contentious change. In my eyes, the fact that dropping the returned LinkRef still leaves a resource occupied (ie: the program is still attached) is not idiomatic Rust behavior, especially given that there is no way to detach it from that point on until the Bpf is dropped. It _resembles_ a leak.

I would tend to agree with this as well. But I think we should very carefully document the potential repercussions of dropping a link unexpectedly if we end up going this route.

> 
> > Also if we do go this route, I think we should make the returned link type opaque. The way programs get attached - and so the type of link used - can change between kernel versions, and will change more in the future as more programs adopt BPF_LINK_CREATE. At some point in the next 6 months I think we should release 1.0, at which point we won't be able to break API easily.
> 
> Totally understandable and I agree; it should be trivial to add that on to this PR with a wrapper struct around the OwnedLink enum.
> 
> > as well as prevents the links section of ProgramData from growing without bound as programs are attached and detached.
> 
> I think that this would have to become a HashMap keyed on some unique specifier so that links with the same inputs are always slotted to the same bucket, but perhaps this is an XDP-centric perspective. Or of course get rid of it entirely as this PR does. :)

This is actually how it's done in libbpf-rs IIRC (they are actually keyed by the program name in the libbpf-rs skeleton, but of course that assumes that the mapping of programs to links is one-to-one, which is definitely not necessarily the case. We may need to put some thought into what exactly the keys of this map would contain.). I think this would be a totally valid approach.

eero-thia:
I've updated the PR with implementation hiding and a longer documentation note.

alessandrod:
> > > I'm not against this change, but it's not something that I'd do lightly since it does make the API harder to use, and it makes it easier to misuse it (programs can get detached silently because you failed to store or propagate a link up the stack).
> > 
> > 
> > I can empathize that this is a more contentious change. In my eyes, the fact that dropping the returned LinkRef still leaves a resource occupied (ie: the program is still attached) is not idiomatic Rust behavior, especially given that there is no way to detach it from that point on until the Bpf is dropped. It _resembles_ a leak.
> 
> I would tend to agree with this as well. But I think we should very carefully document the potential repercussions of dropping a link unexpectedly if we end up going this route.

I disagree that it's not idiomatic rust. I think it's conceptually not very different form say [`tokio::spawn`](https://docs.rs/tokio/1.14.0/tokio/fn.spawn.html) (or `std::thread::spawn`): you can use the returned handle if you want, but if you don't, the task is still managed by the runtime.

I could have used other examples but the reason I chose `tokio::spawn` is that I think it makes the same trade off that `attach()` currently makes: if one had to keep alive all the task handles returned by `spawn()` the API would objectively become a lot less ergonomic (and incidentally, futures have this problem in rust and it's one of the things that make them a bit painful to use).

The way I see it, the whole point of having the `Bpf` object is to manage resources. In fact a first iteration of aya didn't have the `Bpf` object at all; I wanted the API to be as flexible as possible, so you'd use the `obj::*` API to parse ELF, then manually create maps and programs and call relocation functions. I started using that API, realised it was extremely tedious and decided that convenience was more important than flexibility. 

For example consider the following code with the current API:

```
fn init_some_feature(bpf: &mut Bpf) -> Result<(), anyhow::Error> {
    let program: &mut TracePoint = bpf.program_mut("sys_enter_connect")?.try_into()?;
    program.load()?;
    program.attach("syscalls", "sys_enter_connect")?;

    let program: &mut TracePoint = bpf.program_mut("sys_enter_accept")?.try_into()?;
    program.load()?;
    program.attach("syscalls", "sys_enter_accept")?;
    program.attach("syscalls", "sys_enter_accept4")?;

    // [snip - I'm attaching to a bunch more places here] 

   Ok(())
}

fn init_some_other_feature(bpf: &mut Bpf) -> Result<(), anyhow::Error> {
    for sym in ["sym1", "sym2", "sym3"] {
        let program: &mut UProbe = bpf.program_mut(&sym)?.try_into()?;
        program.load()?;
        program.attach(Some(sym), 0, "libXXX", None)?;

        if let Ok(program) = bpf.program_mut(&format!("{}_ret", sym)) {
            let program: &mut UProbe = program.try_into()?;
            program.load()?;
            program.attach(Some(sym), 0, "libXXX", None)?;
        }
    }

    Ok(())
}
```

After having called `Bpf::load()` the calling code can treat `init_some_feature()` and `init_some_other_feature()` pretty much as black boxes. Those two functions can do whatever BPF operations they need to do and be essentially self contained.

```
// ... call Bpf::load()
init_some_feature(&mut bpf)?;
init_some_other_feature(&mut bpf)?;
```

The code above is taken from a real program that includes about ~10 such "features" that can be turned on or off depending on configuration settings. 

Now if we merge this PR, the code becomes:

```
fn init_some_feature(bpf: &mut Bpf) -> Result<Vec<OwnedLink>, anyhow::Error> {
    let mut links = Vec::new();

    let program: &mut TracePoint = bpf.program_mut("sys_enter_connect")?.try_into()?;
    program.load()?;
    links.push(program.attach("syscalls", "sys_enter_connect")?);

    let program: &mut TracePoint = bpf.program_mut("sys_enter_accept")?.try_into()?;
    program.load()?;
    links.push(program.attach("syscalls", "sys_enter_accept")?);
    links.push(program.attach("syscalls", "sys_enter_accept4")?);

    // [snip - many more times] 

   Ok(links)
}

fn init_some_other_feature(bpf: &mut Bpf) -> Result<Vec<OwnedLink>, anyhow::Error> {
    let mut links = Vec::new();

    for sym in ["sym1", "sym2", "sym3"] {
        let program: &mut UProbe = bpf.program_mut(&sym)?.try_into()?;
        program.load()?;
        links.push(program.attach(Some(sym), 0, "libXXX", None)?);

        if let Ok(program) = bpf.program_mut(&format!("{}_ret", sym)) {
            let program: &mut UProbe = program.try_into()?;
            program.load()?;
            links.push(program.attach(Some(sym), 0, "libXXX", None)?);
        }
    }

    Ok(links)
}
```

And the caller:

```
// ... call Bpf::load()
let mut links = init_some_feature(&mut bpf)?;
links.append(init_some_other_feature(&mut bpf)?));

// now the caller needs to keep both the `bpf` instance and  `links` around
```

Now the `init_*` functions _leak_ this weird link stuff I'm supposed to always keep in a vec (like - I literally never want them to be dropped). This seems a strictly worse UX to me. We're pushing some tedious bookkeeping  work we're now doing internally back to the user, and I'm not sure what the benefit is?

eero-thia:
> I disagree that it's not idiomatic rust. I think it's conceptually not very different form say tokio::spawn (or std::thread::spawn): you can use the returned handle if you want, but if you don't, the task is still managed by the runtime.

Understandable. I have never been fond of these APIs for this exact reason, but it is a measured trade-off in terms of UX.

> Now the init_* functions leak this weird link stuff I'm supposed to always keep in a vec (like - I literally never want them to be dropped). This seems a strictly worse UX to me. We're pushing some tedious bookkeeping work we're now doing internally back to the user, and I'm not sure what the benefit is?

For the use case of "attach-once-and-leave it for the entire program's / Bpf's lifetime", there is indeed minimal benefit. When one has to frequently attach/detach, or has a fixed-size use case (we can get away with an array of MaybeUninit<OwnedLink> or Option<OwnedLink>), then we don't have to deal with as much dynamic memory allocation or pointer indirection under-the-hood.

A concrete example, if a little raw:
`
enum Event
{
    AttachLink(usize),
    DetachLink(usize),
}

fn event_stream() -> impl Stream<Item = Event + Unpin>
{
    unimplemented!()
}


async fn event_loop(mut bpf: Bpf)
{
    let links = [None, None];
    let programs = [ "xdp0", "xdp1" ];
    let interfaces = [ "eth0", "eth1" ];

    while let Some(ev) = stream.next().await
    {
        match ev
        {
            Event::AttachLink(index) =>
            {
                let xdp: &mut Xdp = unsafe {
                    bpf.program_mut(programs[index])
                        .unwrap_unchecked()
                        .try_into()
                        .unwrap_unchecked()
                };
                let link = match xdp.attach(interfaces[index], XdpFlags::SkbMode)
                {
                    Ok(l) => l,
                    Err(_) => continue,
                };
                links[index].replace(link);
            },
            Event::DetachLink(index) =>
            {
                links[index].take();
            },
        }
    }
}
`

This change puts the onus on the caller to decide what trade-off to make, but at least allows the caller to make that trade-off rather than deciding for them. "The Rust way" has always come across to me as _opting in_ to any sort of additional runtime costs for convenience.

Having said all of that, you're the maintainer here, and we can continue to maintain our own fork of aya for this specific change.

There is still a real problem with the links vector growing without bound in the current implementation, but I don't have a definitive answer for how to uniquely key on a link in such a way that it covers all potential use cases.

alessandrod:
> > Now the init_* functions leak this weird link stuff I'm supposed to always keep in a vec (like - I literally never want them to be dropped). This seems a strictly worse UX to me. We're pushing some tedious bookkeeping work we're now doing internally back to the user, and I'm not sure what the benefit is?
> 
> For the use case of "attach-once-and-leave it for the entire program's / Bpf's lifetime", there is indeed minimal benefit. When one has to frequently attach/detach, or has a fixed-size use case (we can get away with an array of MaybeUninit or Option), then we don't have to deal with as much dynamic memory allocation or pointer indirection under-the-hood.

Even for cases where one has to frequently attach and detach, I'd argue that not having to keep links around is preferable. Attach and detach do syscalls, so those will trump whatever performance cost the tiny heap allocation might have. Unless you have very limited heap space - in which case fair enough this change might make a difference - I doubt you'd see those link `Rc`s in any perf profile. 

> 
> This change puts the onus on the caller to decide what trade-off to make, but at least allows the caller to make that trade-off rather than deciding for them. "The Rust way" has always come across to me as _opting in_ to any sort of additional runtime costs for convenience.

If this was in a performance sensitive area I would agree with you, and it's for example why the `PerfEventArray` API looks the way it does. It could be made a lot easier to use, but only at the expense of performance. Since it is an API that is often going to be in a hot path, it makes sense to prioritise performance over ease of use. 

> Having said all of that, you're the maintainer here, and we can continue to maintain our own fork of aya for this specific change.

I think that might be the best course of action. That said if you do measure and see that the allocation does have a significant performance cost on your system, I'd certainly be open to changing my mind.

> 
> There is still a real problem with the links vector growing without bound in the current implementation, but I don't have a definitive answer for how to uniquely key on a link in such a way that it covers all potential use cases.

Yep this is definitely something we need to fix. Not sure exactly how either, will require some thinking.

alessandrod:
This morning I had a shower thought: to properly clean up `ProgramData::links` on `link.detach()`, links will probably have to include some kind of cookie that uniquely identifies them. If we change the API from `link.detach()` to `program.detach(link)` then we can remove the `Rc`s, make the links `Copy` and do nothing on `Drop`

eero-thia:
I like this; we'll probably stick to our approach on our end, but this would probably be good on the upstream side. Something like:

```
pub type LinkId = usize;

pub(crate) struct ProgramData {
    pub(crate) obj: obj::Program,
    pub(crate) fd: Option<RawFd>,
    pub(crate) links: HashMap<LinkId, OwnedLink>,
    link_ind: LinkId,
    pub(crate) expected_attach_type: Option<bpf_attach_type>,
    pub(crate) attach_btf_obj_fd: Option<u32>,
    pub(crate) attach_btf_id: Option<u32>,
}

impl ProgramData {
    pub fn attach<T: Into<OwnedLink>>(&mut self, link: T) -> LinkId {
        let id = self.link_id;
        self.links.insert(id, link.into());
        self.link_id += 1;
        id
    }

    pub fn detach(&mut self, id: LinkId) {
        self.links.remove(id);
    }
```

alessandrod:
> I like this; we'll probably stick to our approach on our end, but this would probably be good on the upstream side. Something like:

Yep something like that

eero-thia:
It does feel silly to me for each Program / ProgramData to track its own links, since we're ultimately trying to tie them to the Bpf's lifetime, right? Perhaps we can migrate this map there...

alessandrod:
> It does feel silly to me for each Program / ProgramData to track its own links, since we're ultimately trying to tie them to the Bpf's lifetime, right? Perhaps we can migrate this map there...

It seems logical to me? Programs are owned by `Bpf`, so transitively the links are too. And it also provides a kind of namespacing which will simplify things once we add support for unloading programs, as we'll be able to iterate `ProgramData::links` and detach all of a program's links before unloading.

Also keep in mind that while everything happens inside of `Bpf` at the moment, it probably won't always be like that. For example see https://github.com/aya-rs/aya/issues/2. There are eBPF APIs that we don't yet (fully) support that allow working with maps and programs loaded by other user space programs. It's not clear yet whether those APIs will go through `Bpf` or not.

eero-thia:
I've pushed a new commit which includes this change if you want to play around with it. I think that my initial enthusiasm is substantially dulled.

The caller has to now keep track of which program a given LinkRef belongs to, which makes manual detaching a lot more painful with many programs to manage. But we do retain the advantage of keeping the API for the simplest, likely most common, use case trivial.

alessandrod:
> I've pushed a new commit which includes this change if you want to play around with it. 

Thank you! I'll play a bit with it today

> 
> The caller has to now keep track of which program a given LinkRef belongs to, which makes manual detaching a lot more painful with many programs to manage. But we do retain the advantage of keeping the API for the simplest, likely most common, use case trivial.

I see, in which case doing `bpf.detach(link)` would make sense. I guess it wouldn't be hard to make `LinkRef` work for both `bpf.detach(link)` and `program.detach(link)` tho? (LinkRef would have to store the program name too)

eero-thia:
If we go that route, I think that it ends up being more expensive than the original `Rc<RefCell<dyn Link>>` implementation due to having to clone Strings for each LinkRef.

seanyoung:
This change breaks lirc. With lirc, you want to be able to load a BPF program, and then the program must be able to exit **without** detaching the program.

This is the whole point of BPF IR decoding: you load the BPF program once during boot, and then the BPF program keeps on decoding. There is no user-space program running.

This is the problem:
```
impl Drop for OwnedLink {
    fn drop(&mut self) {
        let _ = self.detach();
    }
}
```

Please remove it.

alessandrod:
> This change breaks lirc. With lirc, you want to be able to load a BPF program, and then the program must be able to exit **without** detaching the program.

Many other programs work that way too, say you want an userspace program to launch, setup some eBPF programs then exit leaving the programs attached. It's a very common scenario.

I think that needs to be explicitly handled tho, see https://github.com/aya-rs/aya/issues/51

seanyoung:
I am not following. For lirc, this currently work in aya. I think this PR breaks that. You're saying that is ok?

alessandrod:
> I am not following. For lirc, this currently work in aya. I think this PR breaks that. You're saying that is ok?

I'm saying if this ever goes in it's probably fine as it is, but before it can go in in order not to break lirc (and enable other people to exit userspace leaving eBPF programs attached), we should implement #51.

dave-tucker:
> I'm saying if this ever goes in it's probably fine as it is, but before it can go in in order not to break lirc (and enable other people to exit userspace leaving eBPF programs attached), we should implement #51.

And I'd also require that we add some documentation to the book that explains eBPF program lifecycle in Aya and what your options are for managing it :wink:

eero-thia:
@seanyoung It is a breaking change, but I think that detach on drop is the better default behavior and what most users would expect. In order to intentionally leak a value, one can always `forget(link)`. You're right that this needs to be documented in more detail.

Looking at lirc and similar programs specifically, I think that their Drop implementation should be changed to detach, and that we should implement https://github.com/aya-rs/aya/issues/51.

alessandrod:
> If we go that route, I think that it ends up being more expensive than the original `Rc<RefCell<dyn Link>>` implementation due to having to clone Strings for each LinkRef.

We could probably store programs by their name hash, then include the hash in the link ref?

seanyoung:
Makes sense.

This is probably superfluous, but if a link is created via `bpf_prog_query()` then the default should not be to detach on drop.

eero-thia:
@seanyoung I don't think that aya currently has a data type for links from queries, but I would definitely qualify it as a separate data type from the `OwnedLink`.

eero-thia:
I've drafted an example forget implementation building off the concept of OwnedLink at https://github.com/aya-rs/aya/pull/124. It should be possible to apply this to the LinkRef approach, but I don't think that I want to invest the energy in figuring that out; the API seems worse to me than OwnedLink.

In light of this, I think that there is currently a footgun: certain classes of eBPF programs (anything but LIRC2) are *detached* when dropping or terminating the program, while others are merely forgotten unless explicitly detached (LIRC2). I think that the behavior should be to consistently detach on drop unless explicitly forgotten (my preference), or vice-versa.

seanyoung:
> @seanyoung I don't think that aya currently has a data type for links from queries, but I would definitely qualify it as a separate data type from the `OwnedLink`.

 `LircLink` can be used exactly for this, see: https://github.com/aya-rs/aya/blob/main/aya/src/programs/lirc_mode2.rs#L83-L104

dave-tucker:
Closing as replaced by #249

PR #124, opened 2021-11-30
aya: extend Link trait with forget. Fix LircLink drop handling.

eero-thia:
This builds atop the change to OwnedLinks in order to handle link forgetting.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -251,7 +251,6 @@ impl<'a> BpfLoader<'a> {
                 let data = ProgramData {
                     obj,
                     fd: None,
-                    links: Vec::new(),
                     expected_attach_type: None,
                     attach_btf_obj_fd: None,
                     attach_btf_id: None,
--- a/aya/src/maps/sock/mod.rs
+++ b/aya/src/maps/sock/mod.rs
@@ -11,3 +11,9 @@ pub use sock_map::SockMap;
 pub trait SocketMap {
     fn fd_or_err(&self) -> Result<RawFd, MapError>;
 }
+
+impl<'a, S: SocketMap> SocketMap for &'a S {
+    fn fd_or_err(&self) -> Result<RawFd, MapError> {
+        (*self).fd_or_err()
+    }
+}
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -93,7 +93,7 @@ impl CgroupSkb {
                         io_error,
                     }
                 })? as RawFd;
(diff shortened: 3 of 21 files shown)


eero-thia:
I'm not really fond of the double-forget, but I can't currently think of another way to indicate a partial clean-up.

seanyoung:
> I'm not really fond of the double-forget, but I can't currently think of another way to indicate a partial clean-up.

Both detach and forget could take ownership, like `fn forget(self)` rather than taking a reference. I don't know why it needs to return anything.

eero-thia:
D'oh! You're absolutely correct.

eero-thia:
Well, that was a hefty change to make, but I dare say that it has made things *a lot* cleaner; I was able to rip out a lot of interior Options.

eero-thia:
Looks like the current implementation does double-drop though; I'll need to revise this further.

eero-thia:
This should be fixed, with no unsafe code now. Instead, the exterior Link trait takes links by value, while the inner implementation takes them by mutable reference.

Since we don't expose types implementing InnerLink but not Link publicly, we can do away with interior Options without fear that we'll call their detach/forget methods multiple times (although I don't think any of them are actually unsafe to call multiple times).

The one open question I have is whether we want to propagate any errors from closing file descriptors; the existing implementation does not.

seanyoung:
> This should be fixed, with no unsafe code now. Instead, the exterior Link trait takes links by value, while the inner implementation takes them by mutable reference.

I think this looks much nicer. Thanks!

> Since we don't expose types implementing InnerLink but not Link publicly, we can do away with interior Options without fear that we'll call their detach/forget methods multiple times (although I don't think any of them are actually unsafe to call multiple times).
> 
> The one open question I have is whether we want to propagate any errors from closing file descriptors; the existing implementation does not.

I think, if `close()` returns an error, that means there is a bug in aya. I think we should either ignore it, or `panic!()` or `assert!(). The caller cannot do anything with this information, it's the "this cannot happen" category. 

Eventually the process may run out of file descriptors if we're not closing the right file descriptor, for example.

alessandrod:
This generally looks ok but as discussed in the other PR, I don't think we should go for OwnedLinks. I do understand if you're not interested in making forget work with LinkRefs, in which case I'm happy to do that myself.

eero-thia:
It is quite a bit of effort with what I would argue is a worse API, and we won't be using it in our fork, so I don't think that I will tackle it.

Hopefully this PR serves as a useful starting point at the very least.

dave-tucker:
Closing as implemented on top of #249 in #253

PR #126, opened 2021-12-02
obj: Add Btf::to_bytes

dave-tucker:
This allows for parsed BTF to be re-encoded such that it could be loaded
in to the kernel


aya/src/obj/btf/btf.rs
```diff
@@ -291,6 +293,26 @@ impl Btf {
             type_id: root_type_id,
         })
     }
+
+    #[allow(dead_code)]
```
alessandrod, line 297: 🤔

alessandrod (changes requested):
I think you should use this instead of all those from_raw_parts: https://github.com/aya-rs/aya/blob/352e54b72405b5e9f21a947ff0146f3ba162b78a/aya/src/obj/mod.rs#L638

Maybe put it as `pub(crate)` in utils.rs or something?

aya/src/obj/btf/btf.rs
dave-tucker: oops, fixed it by making it pub, which was what I intended to do :laughing:

dave-tucker:
ugh. going to close this as it doesn't make sense on it's own as the code that uses it isn't here..

PR #127, opened 2021-12-02
Support for BPF_PROG_TYPE_EXT

dave-tucker:
This PR add support for `BPF_PROG_TYPE_EXT`

These program types can be used to replace a function (freplace) of an already loaded BPF program.
In order to to this we must ensure that when the main program is loaded, its BTF is also loaded into the kernel.
This requires that we also provide `func_info` and `line_info` 

In order to load BTF to the kernel safely, this PR also implements feature probing.
This allows us to detect BTF feature support and conditionally enable/disable parts of the loading process and/or fix the BTF provided by LLVM to work with the current running kernel.

1. We properly adjust the `func_info` and `line_info` from the `.BTF.ext` ELF section, including in cases where we link functions from other ELF sections
1. If a program contains a valid `.BTF` and `.BTF.ext` and section, we will load these in to the kernel
1. We also set the program name in the `BPF_PROG_LOAD` syscall for ease of finding programs later

As a result, when loading a BPF program (written in C, with BTF) we get line information included in the output from `bpftool prog dump xlated`, as well as the ability to inspect a loaded programs BTF with `bpftool btf dump`.

The API for `BPF_PROG_TYPE_EXT`. The API is as follows:

1. Get the program from the loaded BPF as usual `let ext : &mut Extension = bpf.programs_mut("xdp_pass").unwrap().try_into()`
1. In order to load an `Extension` program, we need the BPF FD of the program we are targeting. This can be obtained from a `LinkRef` if we have one in scope, using `LinkRef::fd()`. Since it's likely that the root program may be pinned, I also implemented `LinkRef::from_pinned_path()`.

The API for load and attach is as follows:
```rust
let prog_ref = LinkRef::from_pinned_path(path)?;
let link_fd = prog_ref.fd()?;
let pass: &mut Extension = bpf.program_mut("pass").unwrap().try_into()?;
pass.load(link_fd, "prog0".to_string())?;
pass.attach(link_fd)?;
```


dave-tucker:
Here's a proposed API for `BPF_PROG_TYPE_EXT`

1. Use `BpfLoader::new().prog_freplace()` which will treat all programs you try and load as `BPF_PROG_TYPE_EXT`. This is useful because you may write them as a normal XDP or TC program, but we convert them on load.

2. We need a way to get `RawFd` from `LinkRef`. I added an `fd()` function for now, but I fully suspect this will need to change

3. The API for load and attach will look something like this. We need a program `fd` on load, so we can lookup program BTF and validate the function name we're attaching to - this could actually be a the BTF Fd instead, if we had some API to get that for a given program that's been loaded. `attach` takes the FD of the program instance you want to attach to.

```rust
let pass: &mut Extension = bpf2.program_mut("pass").unwrap().try_into()?;
pass.load(link_fd, "prog0".to_string())?;
pass.attach(link_fd)?;
```

It's still not working as yet, as I've got to add `func_info` and `line_info` support to the relocations.

This is the full example:

```rust
use aya::programs::{Xdp, XdpFlags, Extension, LinkRef};
use aya::{include_bytes_aligned, Bpf, BpfLoader};

use std::{
    convert::{TryInto},
    sync::atomic::{AtomicBool, Ordering},
    sync::Arc,
    thread,
    time::Duration,
};
use structopt::StructOpt;

fn main() {
    env_logger::init();
    if let Err(e) = try_main() {
        eprintln!("error: {:#}", e);
    }
}

#[derive(Debug, StructOpt)]
struct Opt {
    #[structopt(short, long, default_value = "eth0")]
    iface: String,
}

fn try_main() -> Result<(), anyhow::Error> {
    let opt = Opt::from_args();
    let mut bpf = Bpf::load(include_bytes_aligned!("../../xdp_dispatcher.bpf.o"))?;
    let program: &mut Xdp = bpf.program_mut("dispatcher").unwrap().try_into()?;
    program.load()?;
    let path = "/sys/fs/bpf/xdp_dispatcher";
    program.attach(&opt.iface, XdpFlags::default())?;
    bpf.program_mut("dispatcher").unwrap().pin(path)?;
    
    let prog_ref = LinkRef::from_pinned_path(path)?;
    
    let link_fd = prog_ref.fd().unwrap();

    let mut bpf2 = BpfLoader::new().prog_freplace().load(include_bytes_aligned!("../../xdp_pass.bpf.o"))?;
    let pass: &mut Extension = bpf2.program_mut("pass").unwrap().try_into()?;
    pass.load(link_fd, "prog0".to_string())?;
    pass.attach(link_fd)?;
    
    let running = Arc::new(AtomicBool::new(true));
    let r = running.clone();

    ctrlc::set_handler(move || {
        r.store(false, Ordering::SeqCst);
    })
    .expect("Error setting Ctrl-C handler");

    println!("Waiting for Ctrl-C...");
    while running.load(Ordering::SeqCst) {
        thread::sleep(Duration::from_millis(500))
    }
    println!("Exiting...");

    Ok(())
}
```

dave-tucker:
Updated PR with a better description since this is 90% done now. Will push one more commit to do more fixups before loading to the kernel. We'll have to probe for kernel support since I guess it's possible that some distro vendors might backport some of these features at somepoint.

dave-tucker:
@alessandrod 

1. Rebased on #145 to pick up codegen changes
1. Removed `BpfLoader::new().programs_as_extensions()`
1. Implemented `TryFrom<&'a Program>` and `TryFrom<&'a mut Program>` for `Extension`
1. Implemented `AsRef<Extension>` and `AsMut<Extension>` to perform reference to reference conversion, since this is required for the `TryFrom<&'a Program>` implementation

dave-tucker:
Rebased!

aya/src/obj/btf/btf.rs
```diff
@@ -71,6 +84,28 @@ pub enum BtfError {
...
+        io_error: io::Error,
+        /// The error log produced by the kernel verifier.
+        verifier_log: String,
+    },
+
+    #[error("Offset not found for symbol `{symbol_name}`")]
+    SymbolOffsetNotFound { symbol_name: String },
+
+    #[error("BTF type that is not VAR was found in DATASEC")]
+    NonVarTypeInDatasec,
```
alessandrod, line 102: I think this should be InvalidDatasecType

aya/src/obj/btf/btf.rs
```diff
@@ -90,6 +125,40 @@ pub struct Btf {
...
+
+    pub fn add_string(&mut self, name: String) -> u32 {
+        let str = CString::new(name).unwrap();
+        let name_off = self.strings.len();
+        self.strings.extend(str.as_c_str().to_bytes_with_nul());
+        self.header.str_len = self.strings.len() as u32;
+        name_off as u32
+    }
+
+    pub fn add_type(&mut self, type_: BtfType) -> Result<u32, BtfError> {
```
alessandrod, line 154: Can this actually fail?

aya/src/obj/btf/btf.rs
```diff
@@ -291,6 +360,212 @@ impl Btf {
...
+                    // While I figure out if this needs fixing in the Kernel or LLVM, we'll
+                    // do a fixup here
+                    ty.name_off = 0;
+                    types.push(BtfType::Ptr(ty));
+                }
+                BtfType::DataSec(mut ty, data) => {
+                    // Start DataSec Fixups
+                    let sec_name = self.type_name(t)?.ok_or(BtfError::InvalidTypeInfo)?;
+                    let name = sec_name.to_string();
+                    // There are cases when the compiler does indeed populate the
```
alessandrod, line 400: Do you have a test case that repros when LLVM sets the size vs when it doesn't? I'm happy to fix this in LLVM if it's a bug

aya/src/obj/btf/btf.rs
```diff
@@ -472,4 +815,237 @@ mod tests {
         assert_eq!(header.str_off, 0x2a5464);
         assert_eq!(header.str_len, 0x1c6410);
     }
+
+    #[test]
+    fn test_parse_btf() {
+        let data: &[u8] = &[
```
alessandrod, line 821: a comment on how to generate this would help when this test inevitably breaks :P

alessandrod (changes requested):
This is incredible, you're a hero

aya/src/obj/btf/btf.rs
dave-tucker: It's expected behaviour. See https://reviews.llvm.org/D59441

> At the time of debug info emission, the data section
size is unknown, so the btf_type.size = 0 for
BTF_KIND_DATASEC.

I don't remember if I saw a comment in libbpf or LLVM that triggered me to add this conditional

aya/src/obj/btf/btf.rs
dave-tucker: Nope :laughing: . Fixed

aya/src/programs/mod.rs
```diff
@@ -291,18 +316,22 @@ impl Program {
...
             Program::FExit(p) => &mut p.data,
+            Program::Extension(p) => &mut p.data,
         }
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct ProgramData {
+    pub(crate) name: Option<String>,
```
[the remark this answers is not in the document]
dave-tucker: I was using this so that `name` only gets set on `ProgramData` if we're going to use it for loading.
That test happens via the feature detection code when you construct a `BpfLoader`.
This is consistent with other optional load-time features - e.g `btf_fd`.

The alternative was for `load_program` to always load a test program first to figure out if loading a program name is supported or not... or to find a way to pass the detected features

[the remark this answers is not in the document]
alessandrod: Ah, yeah ok that makes sense. I'm ok with landing this as is, we can refactor later

aya/src/obj/btf/types.rs
```diff
@@ -87,6 +88,36 @@ impl TryFrom<u32> for BtfKind {
...
 
+impl Display for BtfKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            BtfKind::Unknown => write!(f, "[UNKNOWN]"),
+            BtfKind::Int => write!(f, "[INT]"),
+            BtfKind::Float => write!(f, "[FLOAT]"),
+            BtfKind::Ptr => write!(f, "[PTR]"),
+            BtfKind::Array => write!(f, "[ARRAY]"),
+            BtfKind::Struct => write!(f, "[STRUCT"),
```
alessandrod, line 99: missing closing bracket.

aya/src/obj/btf/types.rs
```diff
@@ -149,6 +180,80 @@ impl BtfType {
...
+            BtfType::Fwd(btf_type)
+            | BtfType::Const(btf_type)
+            | BtfType::Volatile(btf_type)
+            | BtfType::Restrict(btf_type)
+            | BtfType::Ptr(btf_type)
+            | BtfType::Typedef(btf_type)
+            | BtfType::Func(btf_type)
+            | BtfType::Float(btf_type) => bytes_of::<btf_type>(btf_type).to_vec(),
+            BtfType::Int(btf_type, len) => {
+                // Safety: btf_type is POD
```
alessandrod, line 198: Pls remove all these comments

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn load<T: AsRawFd>(&mut self, program: T, func_name: String) -> Result<(), ProgramError> {
+        let target_prog_fd = program.as_raw_fd();
+
+        let info = sys::bpf_obj_get_info_by_fd(target_prog_fd).map_err(|io_error| {
+            ProgramError::SyscallError {
+                call: "bpf_obj_get_info_by_fd".to_owned(),
+                io_error,
+            }
+        })?;
+        if info.btf_id == 0 {
```
alessandrod, line 61: if this would indicate a kernel bug, make it an assert!()

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,178 @@
...
+#[doc(alias = "BPF_PROG_TYPE_EXT")]
+pub struct Extension {
+    pub(crate) data: ProgramData,
+}
+
+impl Extension {
+    /// Loads the program inside the kernel.
+    ///
+    /// See also [`Program::load`](crate::programs::Program::load).
+    pub fn load<T: AsRawFd>(&mut self, program: T, func_name: String) -> Result<(), ProgramError> {
```
alessandrod, line 52: func_name should be `&str`

aya/src/programs/extension.rs
dave-tucker: i've added a typed error to catch this. it's not a kernel bug as it's possible for a user to try and replace f unction in a loaded program which hasn't loaded BTF into the kernel.

aya/src/programs/extension.rs
alessandrod: Instead of AsRawFd you should use ProgramFd

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,187 @@
...
+                    let btf_info =
+                        sys::btf_obj_get_info_by_fd(btf_fd, &mut buf).map_err(|io_error| {
+                            ProgramError::SyscallError {
+                                call: "bpf_obj_get_info_by_fd".to_owned(),
+                                io_error,
+                            }
+                        })?;
+                    Ok(btf_info)
+                } else {
+                    Ok(info)
```
alessandrod, line 95: this is unnecessary, you're slicing in the next line

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,187 @@
...
+        )
+        .map_err(ProgramError::Btf)?;
+
+        let btf_id = btf
+            .id_by_type_name_kind(func_name, BtfKind::Func)
+            .map_err(ProgramError::Btf)?;
+
+        self.data.attach_btf_obj_fd = Some(btf_fd as u32);
+        self.data.attach_prog_fd = Some(target_prog_fd);
+        self.data.attach_btf_id = Some(btf_id);
```
alessandrod, line 117: surely the attach type shouldn't be hardcoded here

aya/src/programs/mod.rs
```diff
@@ -509,9 +518,38 @@ impl LinkRef {
...
+        let fd =
+            bpf_get_object(&path_cstring).map_err(|(code, io_error)| ProgramError::LoadFromPin {
+                path: path_str.to_owned(),
+                code,
+                io_error,
+            })? as RawFd;
+        let fdlink = FdLink { fd: Some(fd) };
+        let link: Rc<RefCell<dyn Link>> = Rc::new(RefCell::new(fdlink));
+        Ok(LinkRef::new(link))
+    }
```
alessandrod, line 545: What is this for?

alessandrod (changes requested):
A few small things left.

And not sure why we need Link::fd(), it doesn't seem to be used?

aya/src/programs/mod.rs
```diff
@@ -491,8 +499,9 @@ pub(crate) fn query<T: AsRawFd>(
     }
 }
 
-/// Detach an attached program.
+/// Detach an attached program or retrive an Fd to it
 pub trait Link: std::fmt::Debug {
+    fn fd(&mut self) -> Option<RawFd>;
```
alessandrod, line 504: what is this for?

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,187 @@
...
+        load_program(BPF_PROG_TYPE_EXT, &mut self.data)
+    }
+
+    /// Attaches the program to the given program
+    pub fn attach(&mut self) -> Result<LinkRef, ProgramError> {
+        let prog_fd = self.data.fd_or_err()?;
+        let target_fd = self.data.attach_prog_fd.ok_or(ProgramError::NotLoaded)?;
+        let btf_id = self.data.attach_btf_id.ok_or(ProgramError::NotLoaded)?;
+        let link_fd = bpf_link_create(prog_fd, target_fd, BPF_CGROUP_INET_INGRESS, Some(btf_id), 0)
```
alessandrod, line 126: surely the attach type shouldn't be hardcoded here

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,187 @@
...
+                    Ok(info)
+                }
+            }
+            Err(io_error) => Err(ProgramError::SyscallError {
+                call: "bpf_obj_get_info_by_fd".to_owned(),
+                io_error,
+            }),
+        }?;
+
+        let btf_data = buf.as_slice();
```
alessandrod, line 104: this is unnecessary, you're slicing in the next line

aya/src/programs/extension.rs
dave-tucker: actually it does need to be set to `0` here, i'll add a comment as the BPF_GROUP_INET_INGRESS is misleading

aya/src/programs/mod.rs
dave-tucker: originally I thought it was better to get the fd of a program from the `LinkRef`...
but your comments are making me think again...

removing this and making the `load` function of Extension accept `ProgramFd` would result in a better API I think...

1. I loaded the program that I wish to extend.

```rust
let mut bpf = BpfLoader::new().load_file("extension.o")?;
let prog: &mut Xdp = bpf.program_mut("main").unwrap().try_into()?;
let ext: &mut Extension = bpf.program_mut("extension").unwrap().try_into()?;
// <snip> load the main program and attach it....
ext.load(prog, "prog0")?;
ext.attach()?;
```

However, I'm not sure how to solve for the other use case:

2. Someone else (i.e another Aya program) loaded the program that I wish to extend

```rust
let mut bpf = BpfLoader::new().load_file("extension.o")?;
let prog: &mut Xdp = program_from_pinned_path("/sys/fs/bpf/prog").try_into()?; // <- What should this API be?
let ext: &mut Extension = bpf.program_mut("extension").unwrap().try_into()?;
// <snip> load the main program and attach it....
ext.load(prog, "prog0")?;
ext.attach()?;
```

dave-tucker: resolved based on discussion in discord. we remove this for now, and will figure out a way to support use case 2 later

aya/src/sys/bpf.rs
```diff
@@ -371,6 +441,233 @@ pub(crate) fn bpf_raw_tracepoint_open(name: Option<&CStr>, prog_fd: RawFd) -> Sy
...
+        slice::from_raw_parts(name_bytes.as_ptr() as *const c_char, len)
+    });
+    u.prog_name = name;
+
+    let prog: &[u8] = &[
+        0xb7, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // mov64 r0, 0
+        0x95, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // exit
+    ];
+
+    let gpl = CString::new("GPL").unwrap();
```
alessandrod, line 488: This should be:

```
let gpl = b"GPL\0";
u.license = gpl.as_ptr() as u64;
```

aya/src/obj/btf/types.rs
```diff
@@ -1,23 +1,24 @@
...
+        BTF_KIND_DATASEC, BTF_KIND_ENUM, BTF_KIND_FLOAT, BTF_KIND_FUNC, BTF_KIND_FUNC_PROTO,
+        BTF_KIND_FWD, BTF_KIND_INT, BTF_KIND_PTR, BTF_KIND_RESTRICT, BTF_KIND_STRUCT,
+        BTF_KIND_TYPEDEF, BTF_KIND_UNION, BTF_KIND_UNKN, BTF_KIND_VAR, BTF_KIND_VOLATILE,
     },
     obj::btf::{Btf, BtfError, MAX_RESOLVE_DEPTH},
 };
 
 #[derive(Clone, Debug)]
-pub(crate) enum BtfType {
+pub enum BtfType {
```
alessandrod, line 21: why does this need to be public? Making this public is problematic since we don't control the variants, and it uses bindgen generated types which we also don't control

aya/src/obj/mod.rs
```diff
@@ -236,10 +254,30 @@ impl Object {
...
             }
         }
 
+        // .BTF and .BTF.ext sections must be parsed first
+        // as they're required to prepare function and line information
+        // when parsing program sections
+        if let Some(s) = obj.section_by_name(".BTF") {
+            bpf_obj.parse_section(Section::try_from(&s)?)?;
+        }
+        if let Some(s) = obj.section_by_name(".BTF.ext") {
```
alessandrod, line 270: should we parse .BTF.ext only if we parse .BTF first? If so, move inside if above. If not, the code that parses BTF.ext uwnrap()s if self.btf is None, so that needs to be fixed.

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,186 @@
...
+#[doc(alias = "BPF_PROG_TYPE_EXT")]
+pub struct Extension {
+    pub(crate) data: ProgramData,
+}
+
+impl Extension {
+    /// Loads the program inside the kernel.
+    ///
+    /// See also [`Program::load`](crate::programs::Program::load).
+    pub fn load<T: AsRawFd>(&mut self, program: T, func_name: &str) -> Result<(), ProgramError> {
```
alessandrod, line 62: The docs need to explain what program and func_name do

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,186 @@
...
+            .map_err(ProgramError::Btf)?;
+
+        self.data.attach_btf_obj_fd = Some(btf_fd as u32);
+        self.data.attach_prog_fd = Some(target_prog_fd);
+        self.data.attach_btf_id = Some(btf_id);
+        load_program(BPF_PROG_TYPE_EXT, &mut self.data)
+    }
+
+    /// Attaches the program to the given program
+    pub fn attach(&mut self) -> Result<LinkRef, ProgramError> {
```
alessandrod, line 120: To what? 😜

alessandrod: And what happens if you detach the returned link? Does the extension stay attached? Is the old program reinstated? (unlikely). The doc should explain this.

aya/src/obj/btf/info.rs
```diff
@@ -0,0 +1,189 @@
...
+ *     but may not be identical
+ *   struct btf_sec_func_info for section #2
+ *   a list of bpf_func_info records for section #2
+ *   ......
+ */
+#[derive(Debug, Clone, Default)]
+pub struct FuncSecInfo {
+    pub sec_name_offset: u32,
+    pub num_info: u32,
+    pub func_info: Vec<bpf_func_info>,
```
alessandrod, line 26: Surely this needs to be an inline array and not a Vec? The kernel won't be able to work with Vec

aya/src/obj/btf/info.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub(crate) fn get(&self, name: &str) -> FuncSecInfo {
+        match self.data.get(name) {
+            Some(d) => d.clone(),
+            None => FuncSecInfo::default(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct LineSecInfo {
```
alessandrod, line 102: does this need to be public?

aya/src/obj/btf/info.rs
```diff
@@ -0,0 +1,189 @@
...
+        buf
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.func_info.len()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct FuncInfo {
```
alessandrod, line 82: does this need to be public?

aya/src/obj/btf/info.rs
```diff
@@ -0,0 +1,189 @@
...
+ *   struct btf_sec_func_info for section #1
+ *   a list of bpf_func_info records for section #1
+ *     where struct bpf_func_info mimics one in include/uapi/linux/bpf.h
+ *     but may not be identical
+ *   struct btf_sec_func_info for section #2
+ *   a list of bpf_func_info records for section #2
+ *   ......
+ */
+#[derive(Debug, Clone, Default)]
+pub struct FuncSecInfo {
```
alessandrod, line 23: does this need to be public?

aya/src/obj/btf/info.rs
```diff
@@ -0,0 +1,189 @@
...
+        buf
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.line_info.len()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct LineInfo {
```
alessandrod, line 172: public?

aya/src/programs/xdp.rs
```diff
@@ -72,6 +74,7 @@ impl Xdp {
     ///
     /// See also [`Program::load`](crate::programs::Program::load).
     pub fn load(&mut self) -> Result<(), ProgramError> {
+        self.data.expected_attach_type = Some(bpf_attach_type::BPF_XDP);
```
alessandrod, line 77: do we actually need to set this?

aya/src/sys/bpf.rs
```diff
@@ -371,6 +441,233 @@ pub(crate) fn bpf_raw_tracepoint_open(name: Option<&CStr>, prog_fd: RawFd) -> Sy
...
+    let cstring = CString::new("test").unwrap();
+    let name_bytes = cstring.to_bytes();
+    let len = min(name.len(), name_bytes.len());
+    name[..len].copy_from_slice(unsafe {
+        slice::from_raw_parts(name_bytes.as_ptr() as *const c_char, len)
+    });
+    u.prog_name = name;
+
+    let prog: &[u8] = &[
+        0xb7, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // mov64 r0, 0
```
alessandrod, line 484: The comment should say `r0 = 0` which is BPF asm format

alessandrod (changes requested):
A few more comments.

How are you testing this? I'm 99% sure that setting funcinfo/lineinfo isn't working (because of that Vec stuff).

Do we need to parse funcinfo/lineinfo at all? Can't we just proxy it verbatim to the kernel? (legit question, it's not clear to me at the moment)

aya/src/programs/xdp.rs
dave-tucker: it does work fine without it, but setting it doesn't hurt. especially as libbpf does.
I noticed it comparing the strace of aya vs libbpf.

aya/src/sys/bpf.rs
dave-tucker: > A few more comments.
> 
> How are you testing this? I'm 99% sure that setting funcinfo/lineinfo isn't working (because of that Vec stuff).

I have a test program - i'll push it tomorrow so you can have a play.

> Do we need to parse funcinfo/lineinfo at all? 

Yes. You can't load BTF to the kernel without it (I tried and failed).

> Can't we just proxy it verbatim to the kernel? (legit question, it's not clear to me at the moment)

No. Because the ELF and kernel API are subtly different in that offsets in ELF format are in bytes, whereas in kernel format they are in instructions. So we have to parse it (unfortunately).

dave-tucker:
> A few more comments.
> 
> How are you testing this? I'm 99% sure that setting funcinfo/lineinfo isn't working (because of that Vec stuff).

I have a test program - i'll push it tomorrow so you can have a play.

> Do we need to parse funcinfo/lineinfo at all? 

Yes. You can't load BTF to the kernel without it (I tried and failed).

> Can't we just proxy it verbatim to the kernel? (legit question, it's not clear to me at the moment)

No. Because the ELF and kernel API are subtly different in that offsets in ELF format are in bytes, whereas in kernel format they are in instructions. So we have to parse it (unfortunately).

aya/src/obj/btf/info.rs
dave-tucker: I used a vec here because we don't know the size ahead of time.
This struct isn't used verbatim by the kernel - it only cares about the `Vec<bpf_func_info>` and there is a `func_info_bytes()` function that does the conversion via `bytes_of`... I guess (not certain) that you could probably use `func_info_bytes.as_ptr()` - which would be nice, less code to maintain :tada: but I'd have to test it out.

alessandrod:
> > A few more comments.
> > How are you testing this? I'm 99% sure that setting funcinfo/lineinfo isn't working (because of that Vec stuff).
> 
> I have a test program - i'll push it tomorrow so you can have a play.
> 
> > Do we need to parse funcinfo/lineinfo at all?
> 
> Yes. You can't load BTF to the kernel without it (I tried and failed).
> 
> > Can't we just proxy it verbatim to the kernel? (legit question, it's not clear to me at the moment)
> 
> No. Because the ELF and kernel API are subtly different in that offsets in ELF format are in bytes, whereas in kernel format they are in instructions. So we have to parse it (unfortunately).

My bad! I got confused and thought `BpfLoadProgramAttrs` was a bindgen type so https://github.com/dave-tucker/aya/blob/f025b0f8a9fd88c033dc5e6d20e47e153d21d16d/aya/src/programs/mod.rs#L426 looked pretty confusing. Please ignore the vec comments

aya/src/obj/btf/info.rs
alessandrod: Never mind! I got confused see https://github.com/aya-rs/aya/pull/127#issuecomment-1004416147

dave-tucker:
@alessandrod comments addressed, docs improved and here's a working example https://github.com/dave-tucker/aya-ext-test

aya/src/programs/extension.rs
alessandrod: I would say something like:

```
Loads the program extension.

Prepares the code included in the extension to replace the code of the function `func_name` within the eBPF program represented by the `program` file descriptor. The extension code will be loaded but inactive until it's attached.

# Example 
...

```

alessandrod: It should also explicitly say whether you can replace a top level program or not, eg:

```
#[xdp]
fn outer(ctx: XdpContext) -> XdpAction {
	inner(ctx)
}

fn inner(ctx: XdpContext) -> XdpAction {
	...
}
```

```
let outer = ...;
let ext = ...;

ext.load(outer, "outer")? // <- does this work?
```

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,191 @@
...
+            .id_by_type_name_kind(func_name, BtfKind::Func)
+            .map_err(ProgramError::Btf)?;
+
+        self.data.attach_btf_obj_fd = Some(btf_fd as u32);
+        self.data.attach_prog_fd = Some(target_prog_fd);
+        self.data.attach_btf_id = Some(btf_id);
+        load_program(BPF_PROG_TYPE_EXT, &mut self.data)
+    }
+
+    /// Attaches the program
```
alessandrod, line 124:
```
Attaches the extension.

Attaches the extension effectively replacing the original target function. Detaching the returned link restores the original function.
```

aya/src/programs/mod.rs
```diff
@@ -670,6 +682,46 @@ impl_try_from_program!(
     FExit,
 );
 
+macro_rules! impl_as_extension {
+    ($($ty:ident),+ $(,)?) => {
+        $(
+            impl AsRef<Extension> for $ty {
+                fn as_ref(&self) -> &Extension {
+                    // Safety: These types both encapsulate the same underlying data
+                    unsafe { & *(self as * const $ty as * const Extension) }
```
alessandrod, line 691: 😱

I had missed this! I don't think we need it. Unless I'm missing something `Extension` can store `program: &'a Program` and use `program.data()` to access `ProgramData`

alessandrod (changes requested):
See the doc comments (including https://github.com/aya-rs/aya/pull/127#discussion_r780650063 which vscode put in the old review omg) and the AsRef stuff

aya/src/programs/mod.rs
dave-tucker: Removed it and adjusted the API per discord discussion

dave-tucker:
@alessandrod addressed docs comments and removed the AsRef stuff. Now you need to use `BpfLoader::new().extension("myfunc")` to mark programs as being an extension...

aya/src/obj/btf/btf.rs
```diff
@@ -90,6 +125,40 @@ pub struct Btf {
 }
 
 impl Btf {
+    pub fn new() -> Btf {
```
alessandrod, line 128: I think this should be pub(crate)?

aya/src/obj/btf/info.rs
```diff
@@ -0,0 +1,189 @@
...
+ */
+#[derive(Debug, Clone, Default)]
+pub(crate) struct FuncSecInfo {
+    pub _sec_name_offset: u32,
+    pub num_info: u32,
+    pub func_info: Vec<bpf_func_info>,
+}
+
+impl FuncSecInfo {
+    pub fn parse(
```
alessandrod, line 30: pub(crate) here and the other methods

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,138 @@
...
+#[derive(Debug)]
+#[doc(alias = "BPF_PROG_TYPE_EXT")]
+pub struct Extension {
+    pub(crate) data: ProgramData,
+}
+
+impl Extension {
+    /// Loads the extension inside the kernel.
+    ///
+    /// See also [`Program::load`](crate::programs::Program::load).
```
alessandrod, line 56: Move this to the end

aya/src/programs/extension.rs
```diff
@@ -0,0 +1,138 @@
...
+    ///
+    /// Prepares the code included in the extension to replace the code of the function
+    /// `func_name` within the eBPF program represented by the `program` file descriptor.
+    /// This requires that both the [`Extension`] and `program` have had their BTF
+    /// loaded into the kernel as the verifier must check that the function signatures
+    /// match. The extension code will be loaded but inactive until it's attached.
```
alessandrod, line 62: Close the paragraph at "match." and start a new paragraph with "The extension code..."

aya/src/programs/mod.rs
```diff
@@ -175,9 +177,30 @@ pub enum ProgramError {
...
+    /// An error occurred while working with an [`Extension`] program.
+    #[error(transparent)]
+    ExtensionError(#[from] ExtensionError),
+
     /// An error occurred while working with BTF.
     #[error(transparent)]
     Btf(#[from] BtfError),
+
+    // An error ocurred loading a pinned program
+    #[error("failed to retrieve pinned program from path `{path}` with code {code}")]
```
alessandrod, line 189: Remove this as it isn't used

aya/src/sys/bpf.rs
```diff
@@ -82,6 +108,23 @@ pub(crate) fn bpf_load_program(aya_attr: BpfLoadProgramAttrs) -> SysResult {
     u.insn_cnt = aya_attr.insns.len() as u32;
     u.license = aya_attr.license.as_ptr() as u64;
     u.kern_version = aya_attr.kernel_version;
+
+    let line_info_buf = aya_attr.line_info.line_info_bytes();
```
alessandrod, line 112: move inside if

aya/src/sys/bpf.rs
```diff
@@ -371,6 +441,233 @@ pub(crate) fn bpf_raw_tracepoint_open(name: Option<&CStr>, prog_fd: RawFd) -> Sy
...
+        Ok(v) => Ok(v as RawFd),
+        Err((_, err)) => Err(err),
+    }
+}
+
+pub(crate) fn is_prog_name_supported() -> bool {
+    let mut attr = unsafe { mem::zeroed::<bpf_attr>() };
+    let u = unsafe { &mut attr.__bindgen_anon_3 };
+    let mut name: [c_char; 16] = [0; 16];
+    let cstring = CString::new("test").unwrap();
```
alessandrod, line 475: future me looking at strace kindly requests this to be named "aya_name_check"

aya/src/obj/relocation.rs
```diff
@@ -296,6 +301,39 @@ impl<'a> FunctionLinker<'a> {
...
+        let off_adj = start - (fun.section_offset as usize / INS_SIZE);
+        let func_info = &fun.func_info.func_info;
+        let func_info = func_info
+            .iter()
+            .map(|f| {
+                let mut new = *f;
+                new.insn_off = f.insn_off + off_adj as u32;
+                new
+            })
+            .collect::<Vec<bpf_func_info>>();
```
alessandrod, line 320: I don't think you need to collect here?

aya/src/obj/relocation.rs
```diff
@@ -296,6 +301,39 @@ impl<'a> FunctionLinker<'a> {
...
+
+        let line_info = &fun.line_info.line_info;
+        let line_info = line_info
+            .iter()
+            .map(|l| {
+                let mut new = *l;
+                new.insn_off = l.insn_off + off_adj as u32;
+                new
+            })
+            .collect::<Vec<bpf_line_info>>();
```
alessandrod, line 332: and here

alessandrod (approved):
Some minor changes, and we're finally done!

alessandrod:
Thank you so much this PR is 🔥

PR #130, opened 2021-12-03
aya: use correct program name when relocating

wg:
Use `ProgramSection` to extract the correct program name from the ELF section name when performing BTF relocations. This fixes a relocation failure when relocating tracepoints in a section named `tracepoint/<category>/name`.


alessandrod:
thank you!

PR #129, opened 2021-12-02
Fix #128: Add support for BPF_MAP_TYPE_LPM_TRIE map

nimrodshn:
## What this does?
Adds support for `BPF_MAP_TYPE_LPM_TRIE` map. 

Issue: #128 .

Some resources:
- https://www.ferrisellis.com/content/ebpf_syscall_and_maps/#bpf_map_type_lpm_trie
- https://blog.apnic.net/2020/10/23/journeying-into-xdp-augmenting-the-dns/
- https://www.lewuathe.com/longest-prefix-match-with-trie-tree.html

### TODO:

- [ ] Unit Tests
- [x] Support for keys with _specific_ struct as described [here](https://blog.apnic.net/2020/10/23/journeying-into-xdp-augmenting-the-dns/).

cc: @dave-tucker @alessandrod

PR #137, opened 2021-12-09
Generate bindings for bpf_lpm_trie_key

nimrodshn:
cc: @dave-tucker @alessandrod 

This is required in order to support the `LPMTrie` map (see #129 ).


nimrodshn:
Following discussion with @alessandrod and @dave-tucker  I'm going to create a seperate commit for these changes in #129 and drop the commit once were confident enough it works to be added in a seperate PR for CI to generate.

PR #140, opened 2021-12-10
Support Parsing of BTF Maps

dave-tucker:
This PR allows for BTF maps in the .maps ELF section to be parsed.
It reads the necessary information from the BTF section of the ELF file.
It provides the BTF IDs to the `bpf_map_create` call.

This allows for `bpftool` to pretty print maps! For example:

```bash
$ sudo bpftool map dump id 224 
[{
        "key": 0,
        "value": {
            "rx_packets": 0
        }
    },{
        "key": 1,
        "value": {
            "rx_packets": 0
        }
    },{
        "key": 2,
        "value": {
            "rx_packets": 148
        }
    },{
        "key": 3,
        "value": {
            "rx_packets": 0
        }
    },{
        "key": 4,
        "value": {
            "rx_packets": 0
        }
    }
]
```

### DIFF
--- a/aya/src/obj/mod.rs
+++ b/aya/src/obj/mod.rs
@@ -1,6 +1,7 @@
 pub(crate) mod btf;
 mod relocation;
 
+use log::debug;
 use object::{
     read::{Object as ElfObject, ObjectSection, Section as ObjSection},
     Endianness, ObjectSymbol, ObjectSymbolTable, RelocationTarget, SectionIndex, SectionKind,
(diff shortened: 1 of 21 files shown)


dave-tucker:
Parsing works fine, but relocations do not. marking as draft until this is resolved.

aya/src/obj/btf/btf.rs
```diff
@@ -156,6 +156,15 @@ pub enum BtfError {
     /// unable to get symbol name
     #[error("Unable to get symbol name")]
     InvalidSymbolName,
+
+    /// BTF Map parse error. Unexpected Type Encountered
+    #[error("map parse error. expected type {expected} but found type {found}")]
+    MapParseUnexpectedType {
```
alessandrod, line 162: I think instead of this we should probably use the existing `UnexpectedBtfType` variant. I don't think it's an error that we're actually expecting to happen (except for corrupt files), so it's not worth adding an extra variant (API).

aya/src/obj/mod.rs
```diff
@@ -977,6 +1137,29 @@ fn parse_version(data: &[u8], endianness: object::Endianness) -> Result<KernelVe
     })
 }
 
+fn get_map_field(btf: &Btf, type_id: u32) -> Result<u32, BtfError> {
```
alessandrod, line 1140: What does this do exactly? It returns `arr.nelems` so it's not obviously clear. Needs a comment and maybe a better name? :P

aya/src/obj/mod.rs
```diff
@@ -1939,6 +2209,85 @@ mod tests {
...
         let map = obj.maps.get(".rodata").unwrap();
-        assert_eq!(test_data, map.data);
+        assert_eq!(test_data, map.data());
+    }
+
+    #[test]
+    fn test_parse_btf_map_section() {
+        let mut obj = fake_obj();
+        fake_sym(&mut obj, 0, 0, "my_map", 0);
+        let data: &[u8] = &[
```
alessandrod, line 2219: Add comment on how you generated this pls? So when it inevitably starts failing in 2 years we can generate it again :P

alessandrod (commented):
This looks really great - thanks so much for working on it!

Almost there, just a couple of nits.

PR #146, opened 2021-12-15
Mark .rodata maps as readonly and freeze on load

dave-tucker:
This commit marks .rodata maps as BPF_F_RDONLY_PROG when loaded to
prevent a BPF program mutating them.

Initial map data is populated by the loader using the new `BpfLoader::set_global()` API.
The loader will mark a map as frozen using bpf_map_freeze to prevent map data
being changed from userspace.

Related to #117

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -18,15 +18,15 @@ use crate::{
     maps::{Map, MapError, MapLock, MapRef, MapRefMut},
     obj::{
         btf::{Btf, BtfError},
-        Object, ParseError, ProgramSection,
+        MapKind, Object, ParseError, ProgramSection,
     },
     programs::{
         BtfTracePoint, CgroupSkb, CgroupSkbAttachType, FEntry, FExit, KProbe, LircMode2, Lsm,
         PerfEvent, ProbeKind, Program, ProgramData, ProgramError, RawTracePoint, SchedClassifier,
         SkMsg, SkSkb, SkSkbKind, SockOps, SocketFilter, TracePoint, UProbe, Xdp,
     },
-    sys::bpf_map_update_elem_ptr,
-    util::{possible_cpus, POSSIBLE_CPUS},
+    sys::{bpf_map_freeze, bpf_map_update_elem_ptr},
+    util::{bytes_of, possible_cpus, POSSIBLE_CPUS},
 };
 
 pub(crate) const BPF_OBJ_NAME_LEN: usize = 16;
@@ -102,6 +102,7 @@ impl Default for PinningType {
 pub struct BpfLoader<'a> {
     btf: Option<Cow<'a, Btf>>,
     map_pin_path: Option<PathBuf>,
+    globals: HashMap<&'a str, &'a [u8]>,
 }
 
 impl<'a> BpfLoader<'a> {
@@ -110,6 +111,7 @@ impl<'a> BpfLoader<'a> {
         BpfLoader {
             btf: Btf::from_sys_fs().ok().map(Cow::Owned),
             map_pin_path: None,
+            globals: HashMap::new(),
         }
     }
 
@@ -155,6 +157,36 @@ impl<'a> BpfLoader<'a> {
         self
     }
 
+    /// Sets the value of a global variable
+    ///
+    /// From Rust eBPF, a global variable would be constructed as follows:
+    /// ```no run
+    /// #[no_mangle]
+    /// const VERSION = 0;
+    /// ```
+    /// If using a struct, ensure that it is `#[repr(C)]` to ensure the size will
+    /// match that of the corresponding ELF symbol.
+    ///
+    /// From C eBPF, you would annotate a variable as `volatile const`
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use aya::BpfLoader;
+    ///
+    /// let bpf = BpfLoader::new()
+    ///     .set_global("VERSION", &2)
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::BpfError>(())
+    /// ```
+    ///
+    pub fn set_global<V: Pod>(&mut self, name: &'a str, value: &'a V) -> &mut BpfLoader<'a> {
+        // Safety: value is POD
+        let data = unsafe { bytes_of(value) };
+        self.globals.insert(name, data);
+        self
+    }
+
     /// Loads eBPF bytecode from a file.
     ///
     /// # Examples
@@ -187,6 +219,7 @@ impl<'a> BpfLoader<'a> {
     /// ```
     pub fn load(&mut self, data: &[u8]) -> Result<Bpf, BpfError> {
         let mut obj = Object::parse(data)?;
+        obj.patch_map_data(self.globals.clone())?;
 
         if let Some(btf) = &self.btf {
             obj.relocate_btf(btf)?;
@@ -229,7 +262,7 @@ impl<'a> BpfLoader<'a> {
                 }
                 PinningType::None => map.create(&name)?,
             };
-            if !map.obj.data.is_empty() && name != ".bss" {
+            if !map.obj.data.is_empty() && map.obj.kind != MapKind::Bss {
                 bpf_map_update_elem_ptr(fd, &0 as *const _, map.obj.data.as_mut_ptr(), 0).map_err(
                     |(code, io_error)| MapError::SyscallError {
                         call: "bpf_map_update_elem".to_owned(),
@@ -237,6 +270,13 @@ impl<'a> BpfLoader<'a> {
                         io_error,
                     },
                 )?;
+            }
+            if map.obj.kind == MapKind::Rodata {
+                bpf_map_freeze(fd).map_err(|(code, io_error)| MapError::SyscallError {
+                    call: "bpf_map_freeze".to_owned(),
+                    code,
+                    io_error,
+                })?;
             }
             maps.insert(name, map);
         }
--- a/aya/src/obj/mod.rs
+++ b/aya/src/obj/mod.rs
@@ -19,7 +19,7 @@ use relocation::*;
 
 use crate::{
     bpf_map_def,
-    generated::{bpf_insn, bpf_map_type::BPF_MAP_TYPE_ARRAY},
+    generated::{bpf_insn, bpf_map_type::BPF_MAP_TYPE_ARRAY, BPF_F_RDONLY_PROG},
     obj::btf::{Btf, BtfError, BtfExt},
     BpfError,
 };
@@ -43,11 +43,34 @@ pub struct Object {
     pub(crate) symbols_by_index: HashMap<usize, Symbol>,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum MapKind {
+    Bss,
+    Data,
+    Rodata,
+    Other,
+}
+
+impl From<&str> for MapKind {
+    fn from(s: &str) -> Self {
+        if s == ".bss" {
+            MapKind::Bss
+        } else if s.starts_with(".data") {
+            MapKind::Data
+        } else if s.starts_with(".rodata") {
+            MapKind::Rodata
+        } else {
+            MapKind::Other
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct Map {
     pub(crate) def: bpf_map_def,
     pub(crate) section_index: usize,
     pub(crate) data: Vec<u8>,
+    pub(crate) kind: MapKind,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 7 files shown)


aya/src/util.rs
```diff
@@ -143,6 +144,25 @@ macro_rules! include_bytes_aligned {
...
+pub(crate) fn page_size() -> usize {
+    // Safety: libc
+    (unsafe { sysconf(_SC_PAGESIZE) }) as usize
+}
+
+// while not currently used, it feels like this may be useful
+// for some of the alignment FIXMEs so leaving in place for now
+#[allow(dead_code)]
+// round_up is used to round a value X up to the nearest Y
+pub(crate) fn round_up(x: usize, y: usize) -> usize {
```
alessandrod, line 156: I don't think we need this. The alignment comments are about something else - pointers needing to be aligned before rust references can be created from them, which is not something that this can help with. Also, I'm pretty sure there's already two other versions of this, one in the BTF code and one in the netlink code

aya/src/bpf.rs
```diff
@@ -155,6 +156,36 @@ impl<'a> BpfLoader<'a> {
...
+    /// ```no_run
+    /// use aya::BpfLoader;
+    ///
+    /// let bpf = BpfLoader::new()
+    ///     .set_global("VERSION", &2)
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::BpfError>(())
+    /// ```
+    ///
+    pub fn set_global<V: Pod>(&mut self, name: &'a str, value: &'a V) -> &mut BpfLoader<'a> {
```
alessandrod, line 182: I don't love that we need to pass V by reference here. One way to avoid it would be by adding a `as_bytes()` default trait method to the `Pod` trait, which is something that I've been thinking about for a while. Then we could store V in the hash map, and call `v.as_bytes()` when needed.

Anyway I can do this in a followup PR.

aya/src/bpf.rs
```diff
@@ -238,6 +269,16 @@ impl<'a> BpfLoader<'a> {
                     },
                 )?;
             }
+            match map.obj.kind {
+                MapKind::Bss | MapKind::Data | MapKind::Rodata => {
```
alessandrod, line 273: Are you sure we need to freeze bss and data too? Those are mutable sections

aya/src/bpf.rs
```diff
@@ -578,4 +619,11 @@ pub enum BpfError {
 
     #[error("program error")]
     ProgramError(#[from] ProgramError),
+
+    #[error("error mmaping data to map")]
+    Mmap {
```
alessandrod, line 624: This is unused I think?

aya/src/obj/mod.rs
```diff
@@ -43,11 +43,40 @@ pub struct Object {
     pub(crate) symbols_by_index: HashMap<usize, Symbol>,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum MapKind {
+    Default,
```
alessandrod, line 48: I would rename this variant `Other`

aya/src/obj/mod.rs
```diff
@@ -43,11 +43,40 @@ pub struct Object {
...
+    Bss,
+    Data,
+    Rodata,
+}
+
+impl Default for MapKind {
```
alessandrod, line 54: Is anything using this? If not I'd remove it

aya/src/obj/mod.rs
```diff
@@ -43,11 +43,40 @@ pub struct Object {
     pub(crate) symbols_by_index: HashMap<usize, Symbol>,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum MapKind {
```
alessandrod, line 47: I don't think this needs to be public?

aya/src/obj/mod.rs
```diff
@@ -342,16 +375,18 @@ impl Object {
 
         match section.kind {
             BpfSectionKind::Data => {
-                self.maps
-                    .insert(section.name.to_string(), parse_map(&section, section.name)?);
+                self.maps.insert(
```
alessandrod, line 378: I think you should revert parse_map as it was before, and instead do something like:

```
let map = parse_map(&section, name)?;
patch_data_map(map.data, globals, self.symbols_by_index)?;
self.maps.insert(name.to_string(), map);
```

aya/src/obj/mod.rs
```diff
@@ -374,6 +409,75 @@ impl Object {
...
+                    .filter(|(_, s)| {
+                        if let Some(idx) = s.section_index {
+                            idx == section.index
+                        } else {
+                            false
+                        }
+                    })
+                    .map(|(_, s)| s)
+                    .collect();
+                for symbol in symbols {
```
alessandrod, line 434: You should do the opposite, iterate on globals as the number of globals is always going to be stricly smaller than the number of symbols

alessandrod: And also if a global can't be found in the symbol table, you should error out

aya/src/obj/mod.rs
```diff
@@ -374,6 +409,75 @@ impl Object {
...
+                    if let Some(name) = &symbol.name {
+                        if let Some(symbol_data) = globals.get(name.as_str()) {
+                            if symbol_data.len() as u64 != symbol.size {
+                                return Err(ParseError::InvalidGlobalData {
+                                    name: name.to_string(),
+                                    sym_size: symbol.size,
+                                    data_size: symbol_data.len(),
+                                });
+                            }
+                            data.splice(
```
alessandrod, line 444: You should check the range here as splice will panic otherwise

aya/src/obj/mod.rs
```diff
@@ -374,6 +409,75 @@ impl Object {
...
+                                symbol.address as usize..symbol.size as usize,
+                                symbol_data.to_vec().into_iter(),
```
alessandrod, line 446: This should be

```
                                symbol_data.into_iter().cloned(),
```

aya/src/obj/mod.rs
```diff
@@ -662,14 +747,14 @@ mod tests {
     }
 
     fn bytes_of<T>(val: &T) -> &[u8] {
-        let size = mem::size_of::<T>();
-        unsafe { slice::from_raw_parts(slice::from_ref(val).as_ptr().cast(), size) }
+        // Safety: This is for testing only
+        unsafe { crate::util::bytes_of(val) }
```
alessandrod, line 751: You can remove this 2nd definition and just import the original?

aya/src/obj/mod.rs
```diff
@@ -181,7 +210,7 @@ impl FromStr for ProgramSection {
 }
 
 impl Object {
-    pub(crate) fn parse(data: &[u8]) -> Result<Object, BpfError> {
+    pub(crate) fn parse(data: &[u8], globals: HashMap<&str, &[u8]>) -> Result<Object, BpfError> {
```
alessandrod, line 213: Ok after realising that this is a "big diff" because of this change, I'm even more convinced that it should be done in a separate step.

Please revert parse() and parse_map() as they were before. Then add a Object::patch_data_maps(globals) method that does:

```
for (_, map) in self.maps.iter_mut() {
	patch_data(map.data, globals, self.symbols_by_index)?;
}
```

alessandrod (changes requested):
This looks great! As mentioned on discord tho, please separate patching from parsing. I just had a deeper look and don't think you'll run into any borrow check issues.

dave-tucker:
thanks! will address these comments over the weekend

aya/src/util.rs
dave-tucker: ack, it was handy to page align offsets but i've removed it for now.

aya/src/bpf.rs
dave-tucker: I did look at that, but it seemed like a breaking change. Anyone that's implemented the `Pod` trait for a Map K/V will have to implement `as_bytes` also. It may make sense to make the `unsafe_impl_pod` macro public to cover the default case, or perhaps use a proc_macro so we can use can `#[derive(Pod)]` instead.

Even better if bindgen could also add that derive to C structs :tada:

aya/src/obj/mod.rs
dave-tucker: The only problem with that is it requires adding `unsafe` in 30 something places all over the tests, so I figured a "safe" wrapper here was cheaper

aya/src/obj/mod.rs
dave-tucker: removed

aya/src/obj/mod.rs
dave-tucker: fixed

aya/src/obj/mod.rs
dave-tucker: done

aya/src/obj/mod.rs
dave-tucker: fixed

aya/src/obj/mod.rs
dave-tucker: sort of done

dave-tucker:
@alessandrod comments addressed. ptal when you have some time...

> You should do the opposite, iterate on globals as the number of globals is always going to be stricly smaller than the number of symbols
> And also if a global can't be found in the symbol table, you should error out

Whichever way we cut it it's going to be ~somewhat bad.
I've implemented this approach with a couple of caveats:

1. I have to make one full pass over `symbols_by_index` to turn them in to what is effectively `symbols_by_name`. This means that a lookup for whether a global is in the symbols table is fast.
2. I do a `self.maps.iter_mut().find()` to get the Map where the section index is the same as the symbol I'm working with. A hashmap is no good here since some sections (i.e ".maps" may have more than one map in them) but typically those we're patching with data do not.

aya/src/bpf.rs
alessandrod: Traits can have default method implementations, and implementors don't have to implement those. So we could implement it on trait Pod and that's it

aya/src/util.rs
```diff
@@ -143,6 +144,17 @@ macro_rules! include_bytes_aligned {
     }};
 }
 
+pub(crate) fn page_size() -> usize {
+    // Safety: libc
+    (unsafe { sysconf(_SC_PAGESIZE) }) as usize
+}
+
+// bytes_of converts a <T> to a byte slice
+pub(crate) unsafe fn bytes_of<T>(val: &T) -> &[u8] {
```
alessandrod, line 153: The type T should have a Pod bound, and then it could be safe since Pod guarantees that T can be a slice of bytes

aya/src/util.rs
```diff
@@ -143,6 +144,17 @@ macro_rules! include_bytes_aligned {
     }};
 }
 
+pub(crate) fn page_size() -> usize {
+    // Safety: libc
```
alessandrod, line 148: do not want 🖐

aya/src/bpf.rs
```diff
@@ -229,7 +262,7 @@ impl<'a> BpfLoader<'a> {
                 }
                 PinningType::None => map.create(&name)?,
             };
-            if !map.obj.data.is_empty() && name != ".bss" {
+            if !map.obj.data.is_empty() && map.obj.kind != MapKind::Bss {
```
alessandrod, line 265: I think the bss check needs to stay? - bss shouldn't be initialized (it's the whole point)

aya/src/obj/mod.rs
```diff
@@ -238,6 +261,54 @@ impl Object {
...
+        for (name, data) in globals {
+            if let Some(symbol) = symbols.get(name) {
+                if let Some((_, map)) = self
+                    .maps
+                    .iter_mut()
+                    // assumption: there is only one map created per section where we're trying to
+                    // patch data. this assumption holds true for the .rodata section at least
+                    .find(|(_, m)| symbol.section_index == Some(SectionIndex(m.section_index)))
+                {
+                    if data.len() as u64 != symbol.size {
```
alessandrod, line 281: looks like this if should be moved one level up?

aya/src/obj/mod.rs
```diff
@@ -238,6 +261,54 @@ impl Object {
...
+                        return Err(ParseError::InvalidGlobalData {
+                            name: name.to_string(),
+                            sym_size: symbol.size,
+                            data_size: data.len(),
+                        });
+                    }
+                    let start = symbol.address as usize;
+                    let end = symbol.size as usize;
```
alessandrod, line 289: This should be start + symbol.size?

aya/src/obj/mod.rs
```diff
@@ -238,6 +261,54 @@ impl Object {
...
+        let symbols: HashMap<String, &Symbol> = self
+            .symbols_by_index
+            .iter()
+            .filter(|(_, s)| s.name.is_some())
+            .map(|(_, s)| (s.name.as_ref().unwrap().clone(), s))
+            .collect();
+
+        for (name, data) in globals {
+            if let Some(symbol) = symbols.get(name) {
+                if let Some((_, map)) = self
```
alessandrod, line 274: This should be ok_or_else instead of an if let

aya/src/obj/mod.rs
```diff
@@ -570,27 +654,36 @@ impl From<KernelVersion> for u32 {
...
-        let def = bpf_map_def {
-            map_type: BPF_MAP_TYPE_ARRAY as u32,
-            key_size: mem::size_of::<u32>() as u32,
-            // We need to use section.size here since
-            // .bss will always have data.len() == 0
-            value_size: section.size as u32,
-            max_entries: 1,
-            map_flags: 0, /* FIXME: set rodata readonly */
-            ..Default::default()
+    let (def, data, kind) =
```
alessandrod, line 657: i think you should have `let kind = MapKind::from(name)` here then match on it instead of using strings in the if condition

aya/src/obj/mod.rs
```diff
@@ -570,27 +654,36 @@ impl From<KernelVersion> for u32 {
...
+        if name == ".bss" || name.starts_with(".data") || name.starts_with(".rodata") {
+            let def = bpf_map_def {
+                map_type: BPF_MAP_TYPE_ARRAY as u32,
+                key_size: mem::size_of::<u32>() as u32,
+                // We need to use section.size here since
+                // .bss will always have data.len() == 0
+                value_size: section.size as u32,
+                max_entries: 1,
+                map_flags: if name.starts_with(".rodata") {
```
alessandrod, line 666: this should use kind too

aya/src/obj/mod.rs
alessandrod: I see ok

alessandrod:
Almost there! See comments

aya/src/bpf.rs
dave-tucker: >  Then we could store V in the hash map, and call v.as_bytes() when needed.

The problem is that `Pod` is not object-safe since it's `Copy + `static` :cry:

aya/src/util.rs
dave-tucker: I can't do that unless I also `unsafe impl Pod for bpf_map_def`, `unsafe imple Pod for bpf_insn` etc...
Happy to do it if that works for you... or we can defer it to a cleanup of the `Pod` trait with a default `as_bytes` impl

aya/src/bpf.rs
dave-tucker: The check is still there. We're just using an enum variant now instead.

aya/src/obj/mod.rs
dave-tucker: Yeah I think you're right :blush:

aya/src/obj/mod.rs
dave-tucker: much better! thanks!

aya/src/util.rs
alessandrod: Oh yeah sorry I didn't think about that. Yeah no need to do that now

aya/src/bpf.rs
alessandrod: I'm sorry I think I reviewed some stale files, I don't know how i managed to get vscode in that state

alessandrod (approved):

PR #149, opened 2021-12-21
bpf/maps: implement ProgramArray

willfindlay:
This PR implements the ProgramArray map type in aya-bpf. Includes a convenient tail_call
method that wraps the bpf_tail_call helper.


bpf/aya-bpf/src/maps/program_array.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    /// Perform a tail call into a program indexed by this map.
+    ///
+    /// # Return Value
+    ///
+    /// On success, this function **does not return** into the original program.
+    /// On failure, a negative error is returned, wrapped in `Err()`.
+    pub fn tail_call<C: BpfContext>(&mut self, ctx: &C, index: u32) -> Result<(), c_long> {
```
alessandrod, line 72: This should return `Result<!, c_long>`

alessandrod: And it should also be `unsafe`. This is unsafe as it gets - flow control just jumps somewhere else, drop methods don't get called etc

willfindlay: > This should return `Result<!, c_long>`

I thought about doing this, but `never_type` isn't a stable feature yet so I wanted to wait for confirmation first. I'll go ahead and add this.

willfindlay:
@alessandrod I've made the requested changes. Thanks!

alessandrod (approved):
Thank you!

PR #152, opened 2022-01-07
ci: Add autogenerated release notes

dave-tucker:

### DIFF
--- a/.github/changelog-base.json
+++ b/.github/changelog-base.json
@@ -1,0 +1,31 @@
+{
+    "categories": [
+      {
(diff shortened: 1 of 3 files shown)


dave-tucker:
Ok for this to work properly we must:

1. Label PRs with the crate they are for (multiple labels is fine)
2. Label PRs with one of "feature", "fix", or "test"

Then when `cargo release` creates a new tag, this action will build release notes based on PRs merged between current and previous tag. I might see if i can get an action to handle 1.

alessandrod:
This is amazing 🙏

Issue #99, opened 2021-11-05
Documentation TODOs


willfindlay:
This is a general-purpose tracking issue for several areas of improvement in Aya's documentation. I'm planning on spending some cycles on some of these over the next couple of weeks. If anyone has any comments/suggestions on something to add or change here, do let me know.

Short-term goals (to help new contributors):
- [ ] Add an [ARCHITECTURE.md][1] file detailing the architecture of the project
- [ ] Add a description of steps needed to 1) implement a new program type; and 2) implement a new map type to CONTRIBUTING.md

Long-term goals (to help library users):
- [ ] Document every public-facing part of the userspace API
- [ ] Document every public-facing part of aya-bpf
- [ ] It probably also would make sense to `#![deny(missing_docs)]` at that point

[1]: https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html

PR #157, opened 2022-01-12
aya: document the public api

dave-tucker:
Related to #99


alessandrod (approved):
Thanks! I'm merging this because it's an improvement over what we have. But before 1.0 we need to:

1. Make all the errors `#[non_exhaustive]`
2. Try to remove some of the error variants - i feel like they're too fine grained right now
3. Improve the error strings so that they're consistent

PR #160, opened 2022-01-13
test: Add regression tests

dave-tucker:
This uses a mix of rust-script, bash, qemu and a test runner called RTF
to add a regression test suite.


alessandrod (approved):
Top 3 PRs I ever reviewed 🙌

PR #164, opened 2022-01-14
aya: Fix BTF verifier output

dave-tucker:
Currently errors can occur if the verifier output is > buffer as we get
ENOSPC. We should only provide a log_buf if initial load failed, then
retry up to 10 times to get full verifier output.


dave-tucker:
@alessandrod i've pushed a commit so we only have one place where we do the whole `loop {}` and grow the log_buf etc...
PTAL and let me know if this is what you had in mind. I'll squash and rebase once you give the OK, then this will be ready for a proper review.

alessandrod:
Yup this is exactly what I had in mind 🙌

PR #169, opened 2022-01-15
aya: Fix BTF type resolution for Arrays and Ints

dave-tucker:
The union of `size` and `type` is unused in BTF_KIND_ARRAY.
Type information of elements is in the btf_array struct that follows in
the type_ field while the index type is in the index_type field.

For BTF_KIND_INT, only the offset should be compared and size and
signedness should be ignored.


dave-tucker:
@alessandrod per discord discussion, added a test case

alessandrod (approved):
Thank you!

PR #172, opened 2022-01-16
aya: Fix name truncation

dave-tucker:


dave-tucker:
I'm not 100% sure this is the right fix. I guess we should convert to an CStr that doesn't have the nul byte at the end...

alessandrod:
The name must be null terminated, see https://github.com/torvalds/linux/blob/a52171ae7b803f4587b8172d1768313b4d093d0a/kernel/bpf/syscall.c#L739

PR #173, opened 2022-01-17
Add fixup for FuncProto

alessandrod:
If an argument has a type, it must also have a name, see `btf_func_check` in the kernel.

Given:

```
SEC("lsm/syslog")
int BPF_PROG(syslog_audit, int type, int ret_prev)
{
		return 0;
}
```

Fixes:

```
error: BTF error: the BPF_BTF_LOAD syscall failed. Verifier output: magic: 0xeb9f
version: 1
flags: 0x0
hdr_len: 24
type_off: 0
type_len: 76
str_off: 76
str_len: 128
btf_total_size: 228
[1] FUNC_PROTO (anon) return=2 args=(3 (anon))
[2] INT int size=4 bits_offset=0 nr_bits=32 encoding=SIGNED
[3] PTR (anon) type_id=4
[4] INT long long unsigned int size=8 bits_offset=0 nr_bits=64 encoding=(none)
[5] FUNC syslog_audit type_id=1
[5] FUNC syslog_audit type_id=1 Invalid arg#1
: Invalid argument (os error 22)
```


aya/src/obj/btf/btf.rs
```diff
@@ -437,11 +435,8 @@ impl Btf {
         section_sizes: &HashMap<String, u64>,
         symbol_offsets: &HashMap<String, u64>,
     ) -> Result<(), BtfError> {
-        // FIXME: there is probably a more elegant way of doing operation in place
```
dave-tucker, line 440: Thank you :pray:

dave-tucker (approved):
LGTM

PR #174, opened 2022-01-17
btf: fix sanitization if BTF_FUNC_GLOBAL is not supported

alessandrod:
The lower 16 bits were not actually being cleared.


dave-tucker (approved):
LGTM! Thanks @alessandrod

PR #175, opened 2022-01-17
aya: Merge Fixup and Sanitzation to single step

dave-tucker:
Aya will now perform sanitzation and fixups in a single phase, requiring
only one pass over the BTF. This modifies the parsed BTF in place.

### DIFF
--- a/aya/src/obj/btf/btf.rs
+++ b/aya/src/obj/btf/btf.rs
@@ -416,173 +416,159 @@ impl Btf {
     ) -> Result<(), BtfError> {
         let mut types = mem::take(&mut self.types);
         for i in 0..types.types.len() {
-            let kind = types.types.get(i).unwrap().kind()?.unwrap_or_default();
-            match kind {
+            let t = &types.types[i];
+            let kind = t.kind()?.unwrap_or_default();
+            match t {
                 // Fixup PTR for Rust
                 // LLVM emits names for Rust pointer types, which the kernel doesn't like
                 // While I figure out if this needs fixing in the Kernel or LLVM, we'll
                 // do a fixup here
-                BtfKind::Ptr => {
-                    if let Some(BtfType::Ptr(ty)) = types.types.get_mut(i) {
-                        ty.name_off = 0;
-                    }
+                BtfType::Ptr(ty) => {
+                    let mut fixed_ty = *ty;
+                    fixed_ty.name_off = 0;
+                    types.types[i] = BtfType::Ptr(fixed_ty)
                 }
                 // Sanitize VAR if they are not supported
-                BtfKind::Var if !features.btf_datasec => {
-                    if let Some(BtfType::Var(ty, _)) = types.types.get(i) {
-                        types.types[i] = BtfType::new_int(ty.name_off, 1, 0, 0);
-                    }
+                BtfType::Var(ty, _) if !features.btf_datasec => {
+                    types.types[i] = BtfType::new_int(ty.name_off, 1, 0, 0);
                 }
                 // Sanitize DATASEC if they are not supported
-                BtfKind::DataSec if !features.btf_datasec => {
-                    if let Some(BtfType::DataSec(ty, data)) = types.types.get(i) {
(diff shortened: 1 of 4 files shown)


aya/src/obj/btf/btf.rs
```diff
@@ -422,143 +404,129 @@ impl Btf {
...
-                    // Rust emits names for pointer types, which the kernel doesn't like
-                    // While I figure out if this needs fixing in the Kernel or LLVM, we'll
-                    // do a fixup here
-                    ty.name_off = 0;
+            match kind {
+                // Fixup PTR for Rust
+                // LLVM emits names for Rust pointer types, which the kernel doesn't like
+                // While I figure out if this needs fixing in the Kernel or LLVM, we'll
+                // do a fixup here
+                BtfKind::Ptr => {
```
alessandrod, line 428: why match on the kind, then right after match on the actual type? (here and in all the other match arms)

aya/src/obj/btf/btf.rs
```diff
@@ -422,143 +404,129 @@ impl Btf {
...
+                BtfKind::Var if !features.btf_datasec => {
+                    if let Some(BtfType::Var(ty, _)) = types.types.get(i) {
+                        types.types[i] = BtfType::new_int(ty.name_off, 1, 0, 0);
                     }
+                }
+                // Sanitize DATASEC if they are not supported
+                BtfKind::DataSec if !features.btf_datasec => {
+                    if let Some(BtfType::DataSec(ty, data)) = types.types.get(i) {
+                        debug!("{}: not supported. replacing with STRUCT", kind);
+                        let mut members = vec![];
```
alessandrod, line 443: iiuc you're just changing the offset here, so you could clone data and then just mutate the offset in place like we do for the BtfType::DataSec arm

aya/src/obj/btf/btf.rs
```diff
@@ -422,143 +404,129 @@ impl Btf {
...
-                        let struct_type = BtfType::new_struct(ty.name_off, members, 0);
-                        btf.add_type(struct_type);
-                    } else {
-                        btf.add_type(BtfType::DataSec(*ty, data.to_vec()));
-                    }
-                }
-                BtfType::FuncProto(ty, vars) => {
-                    if !features.btf_func {
+                // Sanitize FUNC_PROTO
+                BtfKind::FuncProto if features.btf_func => {
```
alessandrod, line 528: you need to negate the if condition here

aya/src/obj/btf/btf.rs
```diff
@@ -857,6 +821,75 @@ impl<'a> Iterator for SecInfoIter<'a> {
...
+
+impl Default for BtfTypes {
+    fn default() -> Self {
+        Self {
+            types: vec![BtfType::Unknown],
+        }
+    }
+}
+
+impl Deref for BtfTypes {
```
alessandrod, line 839: Do we really need this? Deref being for smart pointers etc

aya/src/obj/btf/btf.rs
```diff
@@ -1045,14 +1078,6 @@ mod tests {
         let func = BtfType::new_func(add, func_proto_type_id, btf_func_linkage::BTF_FUNC_GLOBAL);
         btf.add_type(func);
 
-        let name_offset = btf.add_string("int".to_string());
```
alessandrod, line 1048: did you remove these on purpose?

alessandrod:
Thanks for this, almost there!

aya/src/obj/btf/btf.rs
dave-tucker: we're actually changing between types here. we could use mem::transmute since the layout is the same, but this is safer

aya/src/obj/btf/btf.rs
dave-tucker: yeah so this is annoying, i'll open a follow up issue to address it

dave-tucker: tracked in #178

dave-tucker: nevermind, fixed it

PR #176, opened 2022-01-17
aya: Merge BTF Sanitize and Fixup

dave-tucker:
This combines both in to one iteration.
It effectively copies the ELF BTF and applies sanitizations and fixups
and the original is retained. The original may still be used for
relocations.

It's an alternative to #175, which I couldn't get to pass the borrow checker.
Even so, keeping the original, unaltered BTF as read from ELF around isn't such a bad idea....


alessandrod:
Superseded by https://github.com/aya-rs/aya/pull/175

PR #179, opened 2022-01-18
btf: Replace / in DATASEC before load to kernel

dave-tucker:
This replaces the / character with a . which is allowed in the kernel
names. Not allowing a forward slash is perhaps a kernel bug, but lets
fix it up here as it's commonly used for Aya

### DIFF
--- a/aya/src/obj/btf/btf.rs
+++ b/aya/src/obj/btf/btf.rs
@@ -453,57 +453,64 @@ impl Btf {
                     // Start DataSec Fixups
                     let sec_name = self.string_at(ty.name_off)?;
                     let name = sec_name.to_string();
-                    // There are some cases when the compiler does indeed populate the
-                    // size
-                    if unsafe { ty.__bindgen_anon_1.size > 0 } {
-                        debug!("{} {}: fixup not required", kind, name);
-                        continue;
-                    }
 
                     let mut fixed_ty = *ty;
                     let mut fixed_data = data.clone();
-                    // We need to get the size of the section from the ELF file
-                    // Fortunately, we cached these when parsing it initially
-                    // and we can this up by name in section_sizes
-                    let size =
-                        section_sizes
-                            .get(&name)
-                            .ok_or_else(|| BtfError::UnknownSectionSize {
+
+                    // Handle any "/" characters in section names
+                    // Example: "maps/hashmap"
+                    let fixed_name = name.replace('/', ".");
+                    if fixed_name != name {
+                        fixed_ty.name_off = self.add_string(fixed_name);
+                    }
+
+                    // There are some cases when the compiler does indeed populate the
+                    // size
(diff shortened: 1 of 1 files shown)


aya/src/obj/btf/btf.rs
```diff
@@ -453,6 +477,13 @@ impl Btf {
                     // Start DataSec Fixups
                     let sec_name = self.string_at(ty.name_off)?;
                     let name = sec_name.to_string();
+
+                    // Handle any "/" characters in section names
+                    // Example: "maps/hashmap"
+                    let fixed_name = str::replace(&name, "/", ".");
```
alessandrod, line 483: This should use name.replace

aya/src/obj/btf/btf.rs
```diff
@@ -453,6 +477,13 @@ impl Btf {
                     // Start DataSec Fixups
                     let sec_name = self.string_at(ty.name_off)?;
                     let name = sec_name.to_string();
```
alessandrod, line 479: to_string() isn't needed here

aya/src/obj/btf/btf.rs
```diff
@@ -308,6 +308,30 @@ impl Btf {
...
+            mut str_off,
+            str_len,
+            ..
+        } = self.header;
+        str_off += hdr_len;
+
+        let c_str = CString::new(value).unwrap();
+        let new = c_str.as_bytes_with_nul();
+
+        if offset >= str_off + str_len || offset + new.len() as u32 > str_off + str_len {
```
alessandrod, line 323: This needs some tests. The first if condition is a subset of the second so it's redundant.

The API is also a bit odd: it allows replacing a string with a replacement that is the same length or shorter, but not longer. Seems pretty arbitrary.

aya/src/obj/btf/btf.rs
```diff
@@ -308,6 +308,30 @@ impl Btf {
...
+            return Err(BtfError::InvalidStringOffset {
+                offset: offset as usize,
+            });
+        }
+
+        let offset = offset as usize;
+        self.strings
+            .splice(offset..offset + new.len(), new.iter().cloned());
```
alessandrod, line 331: What happens if new.len() is < old.len()? We leave random garbage bytes in there? I'm wondering if we shouldn't just add a new sanitized string and leave the old one as is, instead of having this replace_string_at at all.

aya/src/obj/btf/btf.rs
dave-tucker: It is as we need it to be `String` in `section_sizes.get` later on...
otherwise we have to use `section_sizes.get(&name.to_string())` and another `name.to_string()` in the error message that follows :thinking:

dave-tucker:
@alessandrod per discord discussion, just used `add_string` instead.

alessandrod (approved):

PR #181, opened 2022-01-19
aya: Support multiple maps in map sections

dave-tucker:
This commit uses the symbol table to discover all maps inside an ELF
section. We then divide the data in to equal sized chunks and parse them
using parse_map_def

### DIFF
--- a/aya/src/obj/mod.rs
+++ b/aya/src/obj/mod.rs
@@ -519,6 +519,36 @@ impl Object {
         Ok(())
     }
 
+    fn parse_map_section(
+        &mut self,
+        section: &Section,
+        symbols: Vec<Symbol>,
+    ) -> Result<(), ParseError> {
+        if symbols.is_empty() {
+            return Err(ParseError::NoSymbolsInMapSection {});
+        }
+        for (i, sym) in symbols.iter().enumerate() {
+            let start = sym.address as usize;
+            let end = start + sym.size as usize;
+            let data = &section.data[start..end];
+            let name = sym
+                .name
+                .as_ref()
+                .ok_or(ParseError::MapSymbolNameNotFound { i })?;
+            let def = parse_map_def(name, data)?;
+            self.maps.insert(
+                name.to_string(),
+                Map {
+                    section_index: section.index.0,
+                    def,
+                    data: Vec::new(),
+                    kind: MapKind::Other,
+                },
+            );
+        }
+        Ok(())
+    }
+
     fn parse_section(&mut self, mut section: Section) -> Result<(), BpfError> {
         let mut parts = section.name.rsplitn(2, '/').collect::<Vec<_>>();
         parts.reverse();
@@ -542,9 +572,19 @@ impl Object {
             BpfSectionKind::Btf => self.parse_btf(&section)?,
             BpfSectionKind::BtfExt => self.parse_btf_ext(&section)?,
             BpfSectionKind::Maps => {
-                let name = section.name.splitn(2, '/').last().unwrap();
-                self.maps
-                    .insert(name.to_string(), parse_map(&section, name)?);
+                let symbols: Vec<Symbol> = self
+                    .symbols_by_index
+                    .values()
+                    .filter(|s| {
+                        if let Some(idx) = s.section_index {
+                            idx == section.index.0
+                        } else {
+                            false
+                        }
+                    })
+                    .cloned()
+                    .collect();
+                self.parse_map_section(&section, symbols)?
             }
             BpfSectionKind::Program => {
                 let program = self.parse_program(&section)?;
@@ -625,6 +665,12 @@ pub enum ParseError {
 
     #[error("map for section with index {index} not found")]
     MapNotFound { index: usize },
+
+    #[error("the map number {i} in the `maps` section doesn't have a symbol name")]
+    MapSymbolNameNotFound { i: usize },
+
+    #[error("no symbols found for the maps included in the maps section")]
+    NoSymbolsInMapSection {},
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


alessandrod:
I'll review this soon but first - have you tried it with stripped binaries?

dave-tucker:
I haven't, but I can do

aya/src/obj/mod.rs
```diff
@@ -600,6 +642,15 @@ pub enum ParseError {
 
     #[error("map for section with index {index} not found")]
     MapNotFound { index: usize },
+
+    #[error("map section does not have {num_syms} definitions")]
+    InvalidMapSection { num_syms: usize },
```
alessandrod, line 647: This should probably be InvalidMapSectionSize

alessandrod, line 647: This should probably be `InvalidMapSectionSize { size: usize }`

aya/src/obj/mod.rs
```diff
@@ -494,6 +494,38 @@ impl Object {
...
 
+    fn parse_map_section(
+        &mut self,
+        section: &Section,
+        symbols: Vec<Symbol>,
+    ) -> Result<(), ParseError> {
+        if symbols.is_empty() {
+            return Err(ParseError::NoSymbolsInMapSection {});
+        }
+        let chunk_size = if section.size % symbols.len() as u64 == 0 {
```
alessandrod, line 505: Can the `.maps` section contain variable length `bpf_map_def`s? (we need to check libbpf). See `MINIMUM_MAP_SIZE` which we support to support `iproute2` style maps.

aya/src/obj/mod.rs
```diff
@@ -600,6 +642,15 @@ pub enum ParseError {
 
     #[error("map for section with index {index} not found")]
     MapNotFound { index: usize },
+
+    #[error("map section does not have {num_syms} definitions")]
+    InvalidMapSection { num_syms: usize },
+
+    #[error("no map name provided")]
+    NoMapName {},
```
alessandrod, line 650:
```
#[error("the map number {i} in the `maps` section doesn't have a symbol name")]
MapSymbolNotFound { i: usize }
```

aya/src/obj/mod.rs
```diff
@@ -600,6 +642,15 @@ pub enum ParseError {
...
+
+    #[error("no symbols in map section")]
```
alessandrod, line 652: "no symbols found for the maps included in the `maps` section"

aya/src/obj/mod.rs
```diff
@@ -517,9 +549,19 @@ impl Object {
...
+                    .values()
+                    .cloned()
+                    .filter(|s| {
+                        if let Some(idx) = s.section_index {
+                            idx == section.index
+                        } else {
+                            false
+                        }
+                    })
+                    .collect();
```
alessandrod, line 563: I think this needs to be sorted by `s.address`?

aya/src/obj/mod.rs
```diff
@@ -517,9 +549,19 @@ impl Object {
             BpfSectionKind::Btf => self.parse_btf(&section)?,
             BpfSectionKind::BtfExt => self.parse_btf_ext(&section)?,
             BpfSectionKind::Maps => {
-                let name = section.name.splitn(2, '/').last().unwrap();
-                self.maps
-                    .insert(name.to_string(), parse_map(&section, name)?);
+                let symbols = self
+                    .symbols_by_index
+                    .values()
+                    .cloned()
```
alessandrod, line 555: Not that it makes much of a difference but cloned should probably be after filter

aya/src/obj/mod.rs
```diff
@@ -1107,6 +1174,36 @@ mod tests {
...
+            map_type: 1,
+            key_size: 2,
+            value_size: 3,
+            max_entries: 4,
+            map_flags: 5,
+            ..Default::default()
+        })
+        .to_vec();
+        let mut buf = vec![];
+        buf.extend(map_data.iter());
```
alessandrod, line 1194: &map_data instead of iter should work and be more readable

alessandrod (changes requested):

aya/src/obj/mod.rs
dave-tucker: `maps` can contain multiple map defs.... but they all MUST be the same size.
Here's the map counting/iteration logic and you can see the assumption in the comments.
https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L1968

aya/src/obj/mod.rs
dave-tucker: Hmmm, why would we need to do that?

dave-tucker:
@alessandrod ready for another look. This time around I decided to read into the section data from the address/size in the symbols table, so we can support definitions that are different lengths. The only drawback is that we don't verify all of the section data has been consumed...

aya/src/obj/mod.rs
```diff
@@ -625,6 +665,15 @@ pub enum ParseError {
 
     #[error("map for section with index {index} not found")]
     MapNotFound { index: usize },
+
+    #[error("map section does not have {num_syms} definitions")]
+    InvalidMapSectionSize { num_syms: usize },
```
alessandrod, line 670: This is unused now

aya/src/obj/mod.rs
```diff
@@ -625,6 +665,15 @@ pub enum ParseError {
...
+
+    #[error("the map number {i} in the `maps` section doesn't have a symbol name")]
+    MapSymbolNotFound { i: usize },
```
alessandrod, line 673: This should actually be `MapSymbolNameNotFound` - we have the symbol, just not a name for it.

aya/src/obj/mod.rs
```diff
@@ -873,6 +922,22 @@ mod tests {
...
 
+    fn fake_sym(obj: &mut Object, section_index: usize, name: &str, size: u64) {
+        let idx = obj.symbols_by_index.len();
+        obj.symbols_by_index.insert(
+            idx + 1,
+            Symbol {
+                index: idx + 1,
+                section_index: Some(section_index),
+                name: Some(name.to_string()),
+                address: 0,
```
alessandrod, line 933: The tests are parsing only the first map - you need to set the `address` field now

alessandrod (changes requested):
couple of nits and fix tests pls

dave-tucker:
@alessandrod nits addressed... and while I was editing the test I threw in some padding between the last `bpf_map_def`s just to check that the parsing logic would work as expected.

alessandrod (approved):

PR #183, opened 2022-01-22
ci: Fix image build workflow

dave-tucker:


alessandrod:
We need to disable push on PR. We can do that in a follow up PR so I can merge this and fix main builds.

PR #224, opened 2022-02-08
aya: implement Pod for arrays of Pod types

Tuetuopay:
If a type is POD (meaning it can be converted to a byte array), then an array of such type is POD.


Tuetuopay:
Huh that's weird, tests run fine locally

nak3:
I think https://github.com/aya-rs/aya/pull/230 is fixing the issue (aya-bpf).

Tuetuopay:
@nak3 ah thanks! I just rebased on main

dave-tucker (approved):
LGTM

Issue #225, opened 2022-02-09
`BPF_PROG_TYPE_CGROUP_SKB` program fails to run when no `attach` was specified


nak3:
#### 1. Create a project.

```
cargo generate https://github.com/aya-rs/aya-template \
         --name cgroup-test \
	 -d program_type=cgroup_skb \
	 -d direction=Ingress
```

#### 2. Build and run

```
cargo xtask build-ebpf
cargo xtask run
```

#### 3. Fail to run due to invalid program section

```
$ cargo xtask run
 ...(snip)...
Error: error parsing BPF object

Caused by:
    invalid program section `cgroup/skb/cgroup_test`
```

### Root cause

The root cause is eBPF program misses `attach` attribute.

```
$ vim cgroup-test-ebpf/src/main.rs

#[cgroup_skb(name="cgroup_test")]
pub fn cgroup_test(ctx: SkBuffContext) -> i32 {
    match unsafe { try_cgroup_test(ctx) } {
        Ok(ret) => ret,
        Err(ret) => ret,
    }
}
```

It is an issue in aya-template but the code in aya is also not
correct. The missing `attach` attribute should not be allowed as
cgroup SKB must specify Ingress or Egress.

PR #226, opened 2022-02-09
Do not allow missing `attach` attribute for cgroup SKB

nak3:
Even though cgroup SKB (`BPF_PROG_TYPE_CGROUP_SKB`) has to set Ingress
or Egress, aya allows missing `attach` attribute.
It causes an error and difficult to debug.

This patch adds the validation.

Fix https://github.com/aya-rs/aya/issues/225

### DIFF
--- a/bpf/aya-bpf-macros/src/expand.rs
+++ b/bpf/aya-bpf-macros/src/expand.rs
@@ -238,10 +238,8 @@ impl CgroupSkb {
             } else {
                 format!("cgroup_skb/{}", attach)
             }
-        } else if let Some(name) = &self.name {
-            format!("cgroup/skb/{}", name)
         } else {
-            ("cgroup/skb").to_owned()
+            return Err(Error::new_spanned(&self.expected_attach_type, "missing attach type"));
         };
         let fn_name = &self.item.sig.ident;
         let item = &self.item;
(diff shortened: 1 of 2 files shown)


bpf/aya-bpf-macros/src/expand.rs
```diff
@@ -238,10 +238,8 @@ impl CgroupSkb {
             } else {
                 format!("cgroup_skb/{}", attach)
             }
-        } else if let Some(name) = &self.name {
-            format!("cgroup/skb/{}", name)
         } else {
-            ("cgroup/skb").to_owned()
+            return Err(Error::new_spanned(&self.expected_attach_type, "missing attach type"));
```
dave-tucker, line 242: This is not correct. It's actually fine for the section to not explicitly declare the attach type.
Please back out these changes.

dave-tucker (changes requested):
Thanks for the PR @nak3.
The fix is only partially correct since we **do** want to allow cgroup skb programs with no attach type in the section name, While I was leaving comments, I figured it would be just as quick to open #229 which should fix this.

nak3:
Thank you so much @dave-tucker for https://github.com/aya-rs/aya/pull/229 and leaving your comments. I understand that now.
Let's fix the issue by your PR https://github.com/aya-rs/aya/pull/229
I'm closing here.

Issue #227, opened 2022-02-09
`BPF_PROG_TYPE_SOCKET_FILTER` program always fails to run


nak3:
#### 2. Run `cargo xtask run`

```
$ cargo xtask run
  ...
Error: error parsing BPF object

Caused by:
    invalid program section `socket/my_filter`
```

It causes `invalid program section `socket/socket_filter``.

The root cause is that the sectionname (kind) expand with `socket` not socket_filter so it fails to `socket/socket_filter`.

https://github.com/aya-rs/aya/blob/923cd9b7679ee03d0ebe884588eb0a8f984b65f0/bpf/aya-bpf-macros/src/expand.rs#L514

PR #228, opened 2022-02-09
Fix socket_filter section match

nak3:
`BPF_PROG_TYPE_SOCKET_FILTER` program expands the sectionname's kind with `socket` not `socket_filter`.
So current eBPF program with socket filter always fails. (Please refer to  https://github.com/aya-rs/aya/issues/227.)

This patch fixes it.

Fix https://github.com/aya-rs/aya/issues/227

### DIFF
--- a/aya/src/obj/mod.rs
+++ b/aya/src/obj/mod.rs
@@ -1236,7 +1236,7 @@ mod tests {
         assert_matches!(
             obj.parse_section(fake_section(
                 BpfSectionKind::Program,
-                "socket_filter/foo",
+                "socket/foo",
                 bytes_of(&fake_ins())
             )),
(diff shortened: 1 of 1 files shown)


dave-tucker:
@nak3 this LGTM, but you'll need to also update the test case here: https://github.com/aya-rs/aya/blob/251bbd01d12708875fe5ec574d01819c82517423/aya/src/obj/mod.rs#L1239

nak3:
Thank you! Fixed unit test.

dave-tucker (approved):
LGTM

nak3:
(just rebased to pass tests.)

alessandrod (approved):
Thank you!

PR #229, opened 2022-02-09
aya: Fix Loading from cgroup/skb sections

dave-tucker:
fa037a88e2f0820d2a64bbaae12464bf5dce083d allowed for cgroup skb programs
that did not specify an attach direction to use the cgroup/skb section
name per the convention established in libbpf. It did not add the
necessary code to load programs from those sections which is added in
this commit

Fixes: #225 

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -450,6 +450,10 @@ impl<'a> BpfLoader<'a> {
                                 },
                             })
                         }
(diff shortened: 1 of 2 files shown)


nak3:
LGTM

alessandrod:
Thank you for the fix @dave-tucker and @nak3 for the review! 🙏

PR #230, opened 2022-02-09
ci: Fix aya-bpf workflow

dave-tucker:
This was testing the aya crate with the nightly toolchain which wasn't
what was intented


nak3:
LGTM

PR #239, opened 2022-03-25
aya-bpf: expose bpf_sock_ops of SockOpsContext

nak3:
This patch expose bpf_sock_ops of `SockOpsContext`.


nak3:
Without this, I cannot access to the `bpf_sock_ops` and not call this [`update()` func]( https://github.com/aya-rs/aya/blob/f01497e021aa6d2ab98141c53713a4457e6b2d68/bpf/aya-bpf/src/maps/sock_hash.rs#L50-L55 ) with `bpf_sock_ops` arg.
(Sorry if there is any possible way to achieve this :sweat_drops: )

PR #248, opened 2022-04-27
bpf: Maps live in maps section

dave-tucker:
This forces all maps to the maps section so we remain compatible with
libbpf. This requires #181 to avoid breaking userspace.


dave-tucker:
Note to self: add userspace component to the test case that uses the maps...

bpf/aya-bpf-macros/src/expand.rs
```diff
@@ -49,11 +49,13 @@ impl Map {
     }
 
     pub fn expand(&self) -> Result<TokenStream> {
-        let section_name = format!("maps/{}", self.name);
+        let section_name = "maps".to_string();
```
chenhengqi, line 52: libbpf expects maps live in `.maps` section.

kakkoyun: It only expects them in `.maps` BTF maps.

But that being said, I think we need a way to specify it by using this macro.

I'm not suggesting it's the right PR to do this, it might be add as part of BTF map support.

chenhengqi: The legacy one is deprecated.

dave-tucker: The legacy, `bpf_map_def` style of map definition is the only style that we support in `aya-bpf` right now.

> But that being said, I think we need a way to specify it by using this macro.
Not really. If we supported BTF defs, all maps would use BTF map defs.
I'm not sure if libbpf's deprecation notice means support will be removed from the kernel (I doubt it) or whether it's to give users a push to use BTF.

Anyhow, userspace support for BTF maps is proposed in https://github.com/aya-rs/aya/pull/140 but still needs work.
Support in `aya-bpf` for BTF maps is sadly not as simple as the BTF generated for Rust programs by LLVM isn't correct.

alessandrod (approved):

PR #253, opened 2022-05-06
aya: Implement forget_link

### LINKED ISSUE #51: lifecyle: Add a `forget()` API for `Link`

dave-tucker:
Fixes #51

### DIFF
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -9,7 +9,8 @@ use crate::{
         bpf_prog_type::BPF_PROG_TYPE_CGROUP_SKB,
     },
     programs::{
-        define_link_wrapper, load_program, FdLink, Link, ProgAttachLink, ProgramData, ProgramError,
+        define_link_wrapper, load_program, FdLink, Link, OwnedLink, ProgAttachLink, ProgramData,
+        ProgramError,
     },
     sys::{bpf_link_create, bpf_prog_attach, kernel_version},
 };
@@ -116,6 +117,17 @@ impl CgroupSkb {
         }
     }
 
+    /// Takes ownership of the link referenced by the provided link_id.
+    ///
+    /// The link will be detached on `Drop` and the caller is now responsible
+    /// for managing its lifetime.
+    pub fn forget_link(
+        &mut self,
+        link_id: CgroupSkbLinkId,
+    ) -> Result<OwnedLink<CgroupSkbLink>, ProgramError> {
+        Ok(OwnedLink::new(self.data.forget_link(link_id)?))
+    }
+
     /// Detaches the program.
     ///
     /// See [CgroupSkb::attach].
@@ -155,6 +167,7 @@ impl Link for CgroupSkbLinkInner {
 }
 
 define_link_wrapper!(
+    /// The link used by [CgroupSkb] programs.
     CgroupSkbLink,
     /// The type returned by [CgroupSkb::attach]. Can be passed to [CgroupSkb::detach].
     CgroupSkbLinkId,
--- a/aya/src/programs/extension.rs
+++ b/aya/src/programs/extension.rs
@@ -6,7 +6,9 @@ use object::Endianness;
 use crate::{
     generated::{bpf_attach_type::BPF_CGROUP_INET_INGRESS, bpf_prog_type::BPF_PROG_TYPE_EXT},
     obj::btf::BtfKind,
-    programs::{define_link_wrapper, load_program, FdLink, FdLinkId, ProgramData, ProgramError},
+    programs::{
+        define_link_wrapper, load_program, FdLink, FdLinkId, OwnedLink, ProgramData, ProgramError,
+    },
     sys::{self, bpf_link_create},
     Btf,
 };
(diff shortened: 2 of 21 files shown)


dave-tucker:
@alessandrod The impl is a bit rough, perhaps a trait would reduce the copy/paste... but this SHOULD work (and I've verified that nothing calls `detach()` on the link).
however, the integration test case fails because the XDP program isn't attached after the program has stopped.
Could do with some fresh set of eyes on this to see if I've missed something :eyes:

dave-tucker:
@alessandrod per discussion on Discord an `OwnedLink` type was added to deal ensure that we detach links on drop. I've verified with `strace` that `program::forget_link` will detach on drop and that `mem::forget(link)` does not detach on drop.... 

however, the link does not stay alive after the userspace program exits.
I've tried `mem::forget()`ing everything (Program, Bpf, OwnedLink) and can't get the program to persist at all - even without these changes - so I've got to conclude that either something broke in the links refactor... i'm doing it wrong,... or the kernel is doing something helpful

alessandrod:
> @alessandrod per discussion on Discord an `OwnedLink` type was added to deal ensure that we detach links on drop. I've verified with `strace` that `program::forget_link` will detach on drop and that `mem::forget(link)` does not detach on drop...

Amazing! 😊 I think we should have `OwnedLink::forget` so that you don't have to `mem::forget` explicitly which seems ugly.

> 
> however, the link does not stay alive after the userspace program exits. I've tried `mem::forget()`ing everything (Program, Bpf, OwnedLink) and can't get the program to persist at all - even without these changes - so I've got to conclude that either something broke in the links refactor... i'm doing it wrong,... or the kernel is doing something helpful

Hrm ok, I'm going to take a look at this

dave-tucker:
@alessandrod per discord discuission,

1, Removed the failing test as it relied on `mem::forget`, which won't work as we cant intentionally leak fds to make this work
2. Implemented `Deref` for `OwnedLink`

Should be ready for review now.

aya/src/programs/cgroup_skb.rs
```diff
@@ -155,6 +167,7 @@ impl Link for CgroupSkbLinkInner {
 }
 
 define_link_wrapper!(
+    /// An owned link, obtained using [CgroupSkb::forget_link]
```
alessandrod, line 170: The description seems wrong? These are not owned, and can be returned by more than forget_link (eg see lirc). (Also missing period at the end of the sentence here and in other link docs)

I'd say:

```
The link used by [CgroupSkb] programs.
```

...or something

aya/src/programs/extension.rs
```diff
@@ -146,9 +148,21 @@ impl Extension {
     pub fn detach(&mut self, link_id: ExtensionLinkId) -> Result<(), ProgramError> {
         self.data.links.remove(link_id)
     }
+
+    /// Takes ownership of the link referenced by the provided link_id.
+    ///
+    /// The link will be detached on Drop and the caller is now responsible
```
alessandrod, line 154: `Drop` or link to https://doc.rust-lang.org/stable/std/ops/trait.Drop.html whatever you prefer

aya/src/programs/links.rs
```diff
@@ -1,19 +1,54 @@
...
 
 use crate::{generated::bpf_attach_type, programs::ProgramError, sys::bpf_prog_detach};
 
-pub(crate) trait Link: std::fmt::Debug + 'static {
+/// A Link
+pub trait Link: std::fmt::Debug + 'static {
+    /// Unique Id
     type Id: std::fmt::Debug + std::hash::Hash + Eq + PartialEq;
 
+    /// Returns the Link UUID
```
alessandrod, line 17: I'd change this to say 

```
Returns the link id.
```

I'm actually thinking that we should split this in two traits, where `detach` would be public and `id` wouldn't. `id` is an internal method really I don't think it makes sense to expose it. But happy to do that in a followup PR if you prefer (I expect it's probably going to be a bit of a chore).

aya/src/programs/links.rs
```diff
@@ -1,19 +1,54 @@
...
     fn id(&self) -> Self::Id;
 
+    /// Detaches the Link
     fn detach(self) -> Result<(), ProgramError>;
 }
 
+/// OwnedLink is a wrapper that is used to return a Link to the caller
```
alessandrod, line 24:
```
An owned link that automatically detaches the inner link when dropped.
```

aya/src/programs/links.rs
```diff
@@ -257,4 +297,32 @@ mod tests {
         assert!(*l1_detached.borrow() == 1);
         assert!(*l2_detached.borrow() == 1);
     }
+
+    #[test]
+    fn test_forget() {
```
alessandrod, line 302: this should be `test_owned_detach` and we need a  new`test_owned_drop` that doesn't call `detach()` but checks that the link gets detached on drop

alessandrod (changes requested):
Smol mostly doc nits, almost there!

aya/src/programs/links.rs
```diff
@@ -257,4 +298,60 @@ mod tests {
...
+            let mut links = LinkMap::new();
+            let id1 = links.insert(l1).unwrap();
+            links.insert(l2).unwrap();
+
+            // manually forget one link and wrap in OwnedLink
+            let owned_l1 = OwnedLink {
+                inner: Some(links.forget(id1).unwrap()),
+            };
+            // check that deref works by calling TestLink::foo()
+            owned_l1.foo();
```
alessandrod, line 347: This should be:

```
let _: &TestLink = &*owned_link;
```

and no foos :P

alessandrod (changes requested):

PR #255, opened 2022-05-11
aya-bpf: Set skb_buff visbility to pub

NoneTirex:
It is possible to access the skb pointer with as_ptr method of BpfContext, but direct access to skb is better, because as_ptr requires a cast.
Accessing skb is necessary to write missing functions or to access skb_buff properties such as ifindex, data, data_end, hash, etc

You can implement missing methods directly into SkBuffContext implementation, but in the future a new method may suddenly appear.
So I think that solution is good.


alessandrod (approved):
Thanks!

Issue #262, opened 2022-05-19
BPF_PROG_TYPE_CGROUP_SKB always calls BPF_PROG_LOAD with Ingress attach type


nak3:
As per title, `BPF_PROG_TYPE_CGROUP_SKB` always calls `BPF_PROG_LOAD` even setting `attach="egress"`.
`strace` for the cgroup_skb code with egress (setting `attach="egress"`):

```rust
#[cgroup_skb(name="cgroup_skb_tmp", attach="egress")]
pub fn cgroup_skb_tmp(ctx: SkBuffContext) -> i32 {
 ...
```

`BPF_PROG_LOAD` always has `BPF_CGROUP_INET_INGRESS`.

```
bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_CGROUP_SKB, .. (snip) .. expected_attach_type=BPF_CGROUP_INET_INGRESS,
```

`BPF_CGROUP_INET_EGRESS` ("e"gress) is expected for the attach_type.

PR #263, opened 2022-05-19
Set attach type during load for BPF_PROG_TYPE_CGROUP_SKB

nak3:
As per title, this patch sets `expected_attach_type` during load.

Fix https://github.com/aya-rs/aya/issues/262

### DIFF
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -62,11 +62,12 @@ pub struct CgroupSkb {
 impl CgroupSkb {
     /// Loads the program inside the kernel.
     pub fn load(&mut self) -> Result<(), ProgramError> {
-        self.data.expected_attach_type = match self.expected_attach_type {
-            Some(CgroupSkbAttachType::Ingress) => Some(BPF_CGROUP_INET_INGRESS),
-            Some(CgroupSkbAttachType::Egress) => Some(BPF_CGROUP_INET_EGRESS),
-            None => None,
-        };
+        self.data.expected_attach_type =
+            self.expected_attach_type
+                .map(|attach_type| match attach_type {
+                    CgroupSkbAttachType::Ingress => BPF_CGROUP_INET_INGRESS,
+                    CgroupSkbAttachType::Egress => BPF_CGROUP_INET_EGRESS,
+                });
         load_program(BPF_PROG_TYPE_CGROUP_SKB, &mut self.data)
     }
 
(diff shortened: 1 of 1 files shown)


vadorovsky (approved):

dave-tucker:
@nak3 Thanks for the PR!

Does the current behaviour cause an issue (other than misleading strace output)?
Since `CgroupSkb` programs can be attached in either direction, and [you have to provide a direction on `attach` anyway,]( 
https://github.com/aya-rs/aya/blob/2fca4aee4e98b78c928be0c7aeb4c2166d93548a/aya/src/programs/cgroup_skb.rs#L89), I guess this change is just cosmetic?

I'm not against the change at all... I just want to understand the motivation.

aya/src/programs/cgroup_skb.rs
```diff
@@ -62,6 +62,11 @@ pub struct CgroupSkb {
 impl CgroupSkb {
     /// Loads the program inside the kernel.
     pub fn load(&mut self) -> Result<(), ProgramError> {
+        self.data.expected_attach_type = match self.expected_attach_type {
```
alessandrod, line 65: this should use map()

nak3:
Hi @dave-tucker

Yes, one very small issue is caused. (Sorry I should explain it in the linked issue.)
As [Kernel code](https://github.com/torvalds/linux/commit/5cf1e91456301f8c4f6bbc63ff76cff12f92f31b) allows only Egress to use the return values (2 or 3), currently attach type "Egress" causes the following error:

```rust
#[cgroup_skb(name="cgroup_skb_tmp", attach="egress")]
pub fn cgroup_skb_tmp(ctx: SkBuffContext) -> i32 {
    3
}
```

this (return `3`) causes the following error:

```
Error: the BPF_PROG_LOAD syscall failed. Verifier output: func#0 @0
0: R1=ctx(id=0,off=0,imm=0) R10=fp0
0: (b7) r0 = 3                        ; R0_w=inv3
1: (95) exit
At program exit the register R0 has value (0x3; 0x0) should have been in (0x0; 0x1)
verification time 10 usec
stack depth 0
processed 2 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0


Caused by:
```

After applied this patch, the return value `2` or `3` for Egress is allowed.

aya/src/programs/cgroup_skb.rs
nak3: Thank you! updated.

alessandrod (approved):

PR #264, opened 2022-05-19
Program unload API

davibe:
Add API to unload programs

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -386,6 +386,18 @@ impl<T: Link> ProgramData<T> {
     }
 }
 
+fn unload_program<T: Link>(data: &mut ProgramData<T>, detach: bool) -> Result<(), ProgramError> {
+    if detach {
+        data.links.remove_all()?;
+    }
+    let fd = data.fd_or_err()?;
+    unsafe {
+        libc::close(fd);
+    }
+    data.fd = None;
+    Ok(())
+}
+
 fn load_program<T: Link>(
     prog_type: bpf_prog_type,
     data: &mut ProgramData<T>,
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -593,6 +593,42 @@ impl_program_unload!(
     SockOps
 );
 
+#[cfg(test)]
+mod tests {
+    use super::Program;
+
+    #[allow(dead_code)]
+    // When a new program is added, this fn will break as a reminder: consider adding unload()
+    // See [impl_program_unload!()]
+    fn program_implements_unload(a: Program) {
+        let detach = false;
+        let _ = match a {
+            Program::KProbe(mut p) => p.unload(detach),
+            Program::UProbe(mut p) => p.unload(detach),
+            Program::TracePoint(mut p) => p.unload(detach),
+            Program::SocketFilter(mut p) => p.unload(detach),
+            Program::Xdp(mut p) => p.unload(detach),
+            Program::SkMsg(mut p) => p.unload(detach),
+            Program::SkSkb(mut p) => p.unload(detach),
+            Program::SockOps(mut p) => p.unload(detach),
+            Program::SchedClassifier(mut p) => p.unload(detach),
+            Program::CgroupSkb(mut p) => p.unload(detach),
+            Program::CgroupSysctl(mut p) => p.unload(detach),
+            Program::CgroupSockopt(mut p) => p.unload(detach),
+            Program::LircMode2(mut p) => p.unload(detach),
+            Program::PerfEvent(mut p) => p.unload(detach),
+            Program::RawTracePoint(mut p) => p.unload(detach),
+            Program::Lsm(mut p) => p.unload(detach),
+            Program::BtfTracePoint(mut p) => p.unload(detach),
+            Program::FEntry(mut p) => p.unload(detach),
+            Program::FExit(mut p) => p.unload(detach),
+            Program::Extension(mut p) => p.unload(detach),
+            Program::CgroupSockAddr(mut p) => p.unload(detach),
+            Program::SkLookup(mut p) => p.unload(detach),
+        };
+    }
+}
+
 macro_rules! impl_program_fd {
     ($($struct_name:ident),+ $(,)?) => {
         $(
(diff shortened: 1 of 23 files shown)


aya/src/programs/mod.rs
```diff
@@ -386,6 +386,18 @@ impl<T: Link> ProgramData<T> {
     }
 }
 
+fn unload_program<T: Link>(data: &mut ProgramData<T>, detach: bool) -> Result<(), ProgramError> {
+    if detach {
+        data.links.remove_all()?;
+    }
+    let fd = data.fd_or_err()?;
```
dave-tucker, line 393: This would be better:
```rust
if let Some(fd) = data.fd.take() {
   unsafe {
        libc::close(fd);
    }
} else {
   return Err(ProgramError::NotLoaded)
}
```

aya/src/programs/mod.rs
```diff
@@ -540,6 +552,47 @@ impl<'a, P: ProgramFd> ProgramFd for &'a P {
...
 }
 
+macro_rules! impl_program_unload {
+    ($($struct_name:ident),+ $(,)?) => {
+        $(
+            impl $struct_name {
+                /// Unloads the program from the kernel.
+                ///
+                /// If `detach` is true, links will be detached before unloading the program.
+                /// Note that OwnedLinks you obtained using [KProbe::forget_link] will not be detached.
```
dave-tucker, line 562: `s/KProbe/$struct_name/g`

aya/src/programs/mod.rs
```diff
@@ -386,6 +386,18 @@ impl<T: Link> ProgramData<T> {
     }
 }
 
+fn unload_program<T: Link>(data: &mut ProgramData<T>, detach: bool) -> Result<(), ProgramError> {
```
dave-tucker, line 389: I'm not sure that `detach` should not be an option here.

There are a few reasons for this:
1. I don't think `unload(data, false)` will actually unload the program if it's attached... as any links created will still reference the program fd... so the kernel won't remove it. Will have to test it out to be sure.
1. If you `unload(data, false)` and `load()` again, any links that were not detached are still pointing at the old program fd which is a bit weird

dave-tucker (changes requested):
Thanks for this @davibe. I've left a few comments but I think this is almost there.

aya/src/programs/mod.rs
davibe: ok

aya/src/programs/mod.rs
alessandrod: `unload(false)` means "unload the program once the links are detached" (presumably manually). Or in other words, it's "decrement_program_refcount(program)". 

Even without the boolean arg, it's still not guaranteed that `unload()` will unload a program, since you might have owned links.  I think `unload(bool)` is fine.

aya/src/programs/mod.rs
alessandrod: I don't think you can expand struct_name in the comment.

I'd go for "Note that owned links obtained using `forget_link()` will not be detached."

test/cases/010_load/030_unload/test.ebpf.rs
```diff
@@ -0,0 +1,30 @@
...
+#![no_std]
+#![no_main]
+
+use aya_bpf::{
+    bindings::xdp_action,
+    macros::xdp,
+    programs::XdpContext,
+};
+
+#[xdp(name="ihaveaverylongname")]
```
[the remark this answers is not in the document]
alessandrod: Yes please, maybe `test_unload` so that below...

test/cases/010_load/030_unload/test.rs
```diff
@@ -0,0 +1,59 @@
...
+    programs::{Xdp, XdpFlags},
+    Bpf,
+};
+use std::convert::TryInto;
+use std::process::Command;
+
+fn is_loaded() -> bool {
+    let output = Command::new("bpftool").args(&["prog"]).output().unwrap();
+    let stdout = String::from_utf8(output.stdout).unwrap();
+    stdout.contains("xdp  name ihaveaverylongn  tag")
```
alessandrod, line 16: This should probably be just `stdout.contains("test_unload")`

alessandrod (changes requested):
Thanks! See comments

aya/src/programs/mod.rs
dave-tucker: That still leaves `2.` in my list, which is what happens if you call `unload(false)`, followed by `load()`

Because `Program` owns `Links` and the fd has changed (thus the links are not really for this program, but the old version), what do we do here?

Keep them? Feels weird. Especially as `attach()` might fail because the old link will keep the old program alive... and that link may (in the case of XDP for example) prevent attach.

Drop them all? In which case, we could just make that default and there's no need for the bool

Something else?...

test/cases/010_load/030_unload/test.rs
```diff
@@ -0,0 +1,59 @@
...
+    let state = is_loaded();
+    if state == loaded {
+        return;
+    }
+    panic!("Expected loaded: {} but was loaded: {}", loaded, state);
+}
+
+fn main() {
+    println!("Loading XDP program");
+    let mut bpf = Bpf::load_file("test.o").unwrap();
```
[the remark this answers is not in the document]
dave-tucker: There is no `impl Drop for Bpf` here... but I can add it as a follow up as it should clean up maps and programs too.

aya/src/programs/mod.rs
alessandrod: > That still leaves `2.` in my list, which is what happens if you call `unload(false)`, followed by `load()`
> 
> Because `Program` owns `Links` and the fd has changed (thus the links are not really for this program, but the old version), what do we do here?

Agreed this is weird.

> Keep them? Feels weird. Especially as `attach()` might fail because the old link will keep the old program alive... and that link may (in the case of XDP for example) prevent attach.

This is just how attach works: it can fail and in many cases you have to handle possible failures unless you completely control the whole system. Take the xdp case - another process might have attached something, or someone via iproute, or the driver might not support your attach mode, etc. I don't think this is a problem.

>  Drop them all? In which case, we could just make that default and there's no need for the bool

Yes now I think we should do this, and document that if you have owned links, those will stay attached and will keep the old instance of the program loaded.

test/cases/010_load/030_unload/test.rs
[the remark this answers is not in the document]
davibe: @dave-tucker I would leave this to you, or to a separate PR anyway

alessandrod (approved):
Minor nit left

alessandrod:
@dave-tucker are you happy with this now? If so imma squash and merge

dave-tucker:
Yep LGTM

Issue #219, opened 2022-01-31
Add support for BPF_PROG_TYPE_CGROUP_SOCKOPT


dave-tucker:

nak3:
I will send the PR for this soon.

PR #266, opened 2022-05-20
xtask: Add `bpf_.*` instead of `bpf_map_.*` to allowed type

nak3:
This patch replaces `bpf_map_.*` with `bpf_.*`.

Currently some types that are not used in helper functions are not generated for bindings - e.g. `bpf_sk_lookup`, `bpf_sockopt` and etc.

Part of https://github.com/aya-rs/aya/issues/219

Note, this PR does not include bindings files as it would be better to be created by auto script.
The bindings can be updated by `cargo xtask codegen --libbpf-dir /<PATH_TO>/libbp`.

### DIFF
--- a/xtask/src/codegen/aya_bpf_bindings.rs
+++ b/xtask/src/codegen/aya_bpf_bindings.rs
@@ -51,7 +51,6 @@ pub fn codegen(opts: &Options) -> Result<(), anyhow::Error> {
             "pt_regs",
             "user_pt_regs",
             "xdp_action",
-            "bpf_adj_room_mode",
         ];
         let vars = ["BPF_.*", "bpf_.*", "TC_ACT_.*", "SOL_SOCKET", "SO_.*"];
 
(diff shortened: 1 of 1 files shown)


xtask/src/codegen/aya_bpf_bindings.rs
```diff
@@ -46,7 +46,7 @@ pub fn codegen(opts: &Options) -> Result<(), anyhow::Error> {
             .constified_enum("BPF_FLOW_.*");
 
         let types = [
-            "bpf_map_.*",
+            "bpf_.*",
```
alessandrod, line 49: Further down we whitelist `bpf_adj_room_mode` which I think can be removed now?

alessandrod (approved):
Thanks! See minor issue

xtask/src/codegen/aya_bpf_bindings.rs
nak3: Yes, you are correct! Thank you. Updated.

alessandrod (approved):

PR #271, opened 2022-05-20
Don't use #[no_mangle] (clashes with `#[export_name]`)

### LINKED ISSUE #270: `#[map(name = "foobar")]` macro doesn't behave as expected
It's defined as:

```rust
    pub fn expand(&self) -> Result<TokenStream> {
        let section_name = "maps".to_string();
        let name = &self.name;
        let item = &self.item;
        Ok(quote! {
            #[no_mangle]
            #[link_section = #section_name]
            #[export_name = #name]
            #item
        })
    }
```

But something like:

```rust
#[map(name = "anycast_v4_subnets")]
static mut ANYCAST_V4_SUBNETS: LpmTrie<SubnetIpv4Key, u8> = LpmTrie::pinned(512, BPF_F_NO_PREALLOC);
```

Actually gives a program file with the uppercase name:

image

After some discord discussion, we started suspecting that `#[no_mangle]` was the guilty party here, and indeed a PR follows that shows that it this is the case.

fasterthanlime:
This closes #270.

With this, the symbol name is correct in the intermediate LLVM bitcode
object file (`.rcgu.o`) and in the final BPF program.

---

`llvm-nm` on the intermediate object file:

image

`readelf -a` on the final BPF program:

image

### DIFF
--- a/bpf/aya-bpf-macros/src/expand.rs
+++ b/bpf/aya-bpf-macros/src/expand.rs
@@ -97,7 +97,6 @@ impl Map {
         let name = &self.name;
         let item = &self.item;
         Ok(quote! {
-            #[no_mangle]
             #[link_section = #section_name]
             #[export_name = #name]
             #item


dave-tucker (approved):
LGTM

alessandrod (approved):

Issue #207, opened 2022-01-31
Add support for BPF_PROG_TYPE_PERF_EVENT,


dave-tucker:

nak3:
It seems `BPF_PROG_TYPE_PERF_EVENT` is already supported 
https://github.com/aya-rs/aya/blob/main/aya/src/programs/perf_event.rs :tada:

ishitatsuyuki:
It looks like we still lack the context type support for perf_event, which looks like this:
```
struct bpf_perf_event_data {
        bpf_user_pt_regs_t regs;
        __u64 sample_period;
        __u64 addr;
};
```

It would be a sensible addition to PerfEventContext, but I wonder if we should keep exposing regs as direct methods on the context like `ProbeContext::{arg, ret}`, or factor them into a different struct since the perf context also contains other data?

alessandrod:
To summarize our conversation on discord: yes it makes sense to have a wrapper, and we'll have to document that "PtRegs::arg(n)" means "get the value of the register used to pass arg n" instead of "get the value of the nth argument", where the latter is only true in some cases (if the register hasn't been modified)

dave-tucker:
Closing this one with the remaining TODO tracked in #291

PR #272, opened 2022-05-20
bpf: Add PtRegs wrapper

dave-tucker:
This adds a portable wrapper around pt_regs and user_pt_regs.
It makes writing Raw Tracepoint or KProbe programs easier when the
arguments are one of these types while also ensuring code is portable
across architectures.

Updates: #207

### DIFF
--- a/bpf/aya-bpf/src/args.rs
+++ b/bpf/aya-bpf/src/args.rs
@@ -56,6 +56,32 @@ unsafe_impl_from_btf_argument!(i64);
 unsafe_impl_from_btf_argument!(usize);
 unsafe_impl_from_btf_argument!(isize);
 
+pub struct PtRegs {
+    regs: *mut pt_regs,
+}
+
+/// A portable wrapper around pt_regs and user_pt_regs.
+impl PtRegs {
+    pub fn new(regs: *mut pt_regs) -> Self {
+        PtRegs { regs }
+    }
+
+    /// Returns the value of the register used to pass arg `n`.
+    pub fn arg<T: FromPtRegs>(&self, n: usize) -> Option<T> {
+        T::from_argument(unsafe { &*self.regs }, n)
+    }
+
+    /// Returns the value of the register used to pass the return value.
+    pub fn ret<T: FromPtRegs>(&self) -> Option<T> {
+        T::from_retval(unsafe { &*self.regs })
+    }
+
+    /// Returns a pointer to the wrapped value.
+    pub fn as_ptr(&self) -> *mut pt_regs {
+        self.regs
+    }
+}
+
 /// A trait that indicates a valid type for an argument which can be coerced from
 /// a pt_regs context.
 ///
--- a/bpf/aya-bpf/src/lib.rs
+++ b/bpf/aya-bpf/src/lib.rs
@@ -5,6 +5,7 @@
 pub use aya_bpf_bindings::bindings;
 
 mod args;
+pub use args::PtRegs;
 pub mod helpers;
 pub mod maps;
 pub mod programs;


bpf/aya-bpf/src/args.rs
```diff
@@ -56,6 +56,30 @@ unsafe_impl_from_btf_argument!(i64);
...
 
+pub struct PtRegs {
+    regs: *mut pt_regs,
+}
+
+/// PtRegs is a portable wrapper around
+/// pt_regs and user_pt_regs
+impl PtRegs {
+    pub fn new(ctx: *mut c_void) -> Self {
+        PtRegs {
```
ishitatsuyuki, line 67: Not sure if pt_regs will always the first element of a context, better leave the cast to the caller.

bpf/aya-bpf/src/args.rs
```diff
@@ -56,6 +56,30 @@ unsafe_impl_from_btf_argument!(i64);
 unsafe_impl_from_btf_argument!(usize);
 unsafe_impl_from_btf_argument!(isize);
 
+pub struct PtRegs {
+    regs: *mut pt_regs,
```
ishitatsuyuki, line 60: Users can’t take the pointer out of the struct in the current state. Need something like as_ptr.

ishitatsuyuki (commented):
Thanks for the PR. Here are some review comments.

Also, #207 should be left open until fields are added to PerfEventContext.

bpf/aya-bpf/src/args.rs
dave-tucker: That's correct, but `new` isn't only callable using a `ctx` - any `*mut c_void` will do - have renamed the arg to make that clearer.

example code from a kprobe on  `__x64_sys_execve`, where `ctx.arg(0)` is a `pt_regs`
```rust
let pt_regs = PtRegs::new(ctx.arg(0).unwrap());
```

ishitatsuyuki (approved):

bpf/aya-bpf/src/args.rs
```diff
@@ -56,6 +56,35 @@ unsafe_impl_from_btf_argument!(i64);
...
+}
+
+/// PtRegs is a portable wrapper around
+/// pt_regs and user_pt_regs
+impl PtRegs {
+    pub fn new(regs: *mut c_void) -> Self {
```
alessandrod, line 66: this should take `*mut pt_regs` and it should be on the caller to cast

bpf/aya-bpf/src/args.rs
```diff
@@ -56,6 +56,35 @@ unsafe_impl_from_btf_argument!(i64);
 unsafe_impl_from_btf_argument!(usize);
 unsafe_impl_from_btf_argument!(isize);
 
+pub struct PtRegs {
+    regs: *mut pt_regs,
+}
+
+/// PtRegs is a portable wrapper around
+/// pt_regs and user_pt_regs
```
alessandrod, line 64: This was written on a 20 columns terminal 📺

bpf/aya-bpf/src/args.rs
```diff
@@ -56,6 +56,35 @@ unsafe_impl_from_btf_argument!(i64);
...
+        T::from_argument(unsafe { &*self.regs }, n)
+    }
+
+    /// Returns the value of the register used to pass the return value.
+    pub fn ret<T: FromPtRegs>(&self) -> Option<T> {
+        T::from_retval(unsafe { &*self.regs })
+    }
+
+    /// Returns a pointer to the wrapped value
+    pub fn as_ptr(&self) -> *mut c_void {
```
alessandrod, line 83: this should be `*mut pt_regs`

alessandrod (changes requested):
Looks good! See comments

PR #274, opened 2022-05-23
Riscv64 support

krsh:
Add support for riscv64-unknown-linux-gnu


dave-tucker:
Hey @krsh, thanks for the PR!

Let's do this in stages:

1. Open a PR with just the changes to `src/xtask/codegen` and please also include a change to our CI workflow that generates the bindings - https://github.com/aya-rs/aya/blob/3262f85925459bdc636afccadc3ae3fc926dc9ac/.github/workflows/gen.yml#L26

2. Once merged, I'll trigger the workflow and CI will generate the bindings

3. Then you can open a final PR to make sure that the generated bindings are included in the right `mod.rs/lib.rs`.

Lastly, please make sure to run `cargo +nightly fmt` and `cargo +nightly clippy` against your patches to make sure the linter is happy!

PR #278, opened 2022-05-23
riscv scaffolding for codegen

dave-tucker:


alessandrod:
And we also need some boilerplate in src/lib.rs to import the riscv64 module

PR #281, opened 2022-05-24
aya: Export program modules

dave-tucker:
This allows access to XdpLink, XdpLinkId etc... which is currently
unavailable since these modules are private

### DIFF
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -1,3 +1,4 @@
+//! Cgroup skb programs.
 use std::{
     hash::Hash,
     os::unix::prelude::{AsRawFd, RawFd},
--- a/aya/src/programs/lirc_mode2.rs
+++ b/aya/src/programs/lirc_mode2.rs
@@ -1,3 +1,4 @@
+//! Lirc programs.
 use std::os::unix::prelude::{AsRawFd, RawFd};
 
 use crate::{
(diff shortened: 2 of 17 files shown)


aya/src/programs/lirc_mode2.rs
```diff
@@ -1,3 +1,4 @@
+//! Lirc programs
```
alessandrod, line 1: Missing period

aya/src/programs/cgroup_skb.rs
```diff
@@ -1,3 +1,4 @@
+//! CGroup SKB programs.
```
alessandrod, line 1: I'd call these "Cgroup skb programs." "Cgroup sockaddr programs". "Cgroup sockopt programs." etc. That is, capital C just because it's the start of a sentence, then the well known thing like "skb" "sockaddr" etc with its usual spelling. And periods, periods everywhere!

aya/src/programs/perf_attach.rs
```diff
@@ -7,9 +8,11 @@ use crate::{
     PERF_EVENT_IOC_DISABLE, PERF_EVENT_IOC_ENABLE, PERF_EVENT_IOC_SET_BPF,
 };
 
+/// PerfLinkId is the identifer of a PerfLink
```
alessandrod, line 11: The identifier of a `PerfLink`.

(period)

alessandrod (changes requested):
.

PR #282, opened 2022-05-24
Add atomic replacement of XDP progs and multihoming of Extension progs

dave-tucker:
This PR adds:
- `bpf_link_update`
- `Xdp::attach_to_link` - which allows for atomic replacement of XDP programs
- `Extension::attach_to_program` - which allows for an Extension program to be attached to a program other than the one provided at `load()` - which is required as the verifier needs to know where it will be attached. The program/function should be BTF compatible or the call will fail.

These changes are collected together to enable the following use case in `bpfd`

1. Attach an `Xdp` [dispatcher program ](https://github.com/xdp-project/xdp-tools/blob/master/lib/libxdp/protocol.org) to an interface with an `Xdp` program loaded as an `Extension` to `func1`.
1. A user requests that a new Xdp extension be loaded at `func2`
1. Create a new `Xdp` dispatcher program.
1. Attach the existing extension to `func1` - it is now attached to the old and new dispatcher simultaneously.
1. Attach the new extension to `func2`
1. Use `Xdp::attach_to_link` to perform atomic replacement of the dispatcher programs

### DIFF
--- a/aya/src/maps/array/program_array.rs
+++ b/aya/src/maps/array/program_array.rs
@@ -4,7 +4,7 @@ use std::{
     convert::TryFrom,
     mem,
     ops::{Deref, DerefMut},
-    os::unix::prelude::RawFd,
+    os::unix::prelude::{AsRawFd, RawFd},
 };
 
 use crate::{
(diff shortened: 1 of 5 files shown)


aya/src/maps/array/program_array.rs
```diff
@@ -26,26 +26,28 @@ use crate::{
...
+/// use aya::programs::CgroupSkb;
 /// use std::convert::{TryFrom, TryInto};
 ///
 /// let mut prog_array = ProgramArray::try_from(bpf.map_mut("JUMP_TABLE")?)?;
-/// let prog_0: &CgroupSkb = bpf.program("example_prog_0").unwrap().try_into()?;
-/// let prog_1: &CgroupSkb = bpf.program("example_prog_1").unwrap().try_into()?;
-/// let prog_2: &CgroupSkb = bpf.program("example_prog_2").unwrap().try_into()?;
-///
+/// let prog_0: &mut CgroupSkb = bpf.program_mut("example_prog_0").unwrap().try_into()?;
+/// let prog_0_fd =  prog_0.fd().unwrap();
```
alessandrod, line 36: Why did you make it program_mut? Shouldn't this now be `let prog_0 = bpf.program("example_prog_0").unwrap().fd().unwrap();`

aya/src/programs/mod.rs
```diff
@@ -214,10 +214,13 @@ pub enum ProgramError {
     },
 }
 
-/// Allows the Fd of a loaded [`Program`] to be retrieved
-pub trait ProgramFd {
-    /// Returns the [`RawFd`] of the program if it has been loaded, or `None`
-    fn fd(&self) -> Option<RawFd>;
+/// A [`Program`] file descriptor.
+pub struct ProgramFd(RawFd);
```
alessandrod, line 218: This needs to derive `Copy`

aya/src/programs/xdp.rs
```diff
@@ -141,6 +141,46 @@ impl Xdp {
...
+                mem::forget(link);
+                self.data
+                    .links
+                    .insert(XdpLink(XdpLinkInner::FdLink(FdLink::new(link_fd))))
+            }
+            XdpLinkInner::NlLink(nl_link) => {
+                let if_index = nl_link.if_index;
+                let old_prog_fd = nl_link.prog_fd;
+                let flags = nl_link.flags;
+                let replace_flags = flags & XdpFlags::REPLACE;
```
alessandrod, line 169: I think you meant `flags | XdpFlags::REPLACE` here?

aya/src/sys/bpf.rs
```diff
@@ -333,6 +335,27 @@ pub(crate) fn bpf_link_create(
...
+    old_prog_fd: Option<RawFd>,
+    flags: u32,
+) -> SysResult {
+    let mut attr = unsafe { mem::zeroed::<bpf_attr>() };
+
+    attr.link_update.link_fd = link_fd as u32;
+    attr.link_update.new_prog_fd = new_prog_fd as u32;
+    if let Some(fd) = old_prog_fd {
+        attr.link_update.old_prog_fd = fd as u32;
+        attr.link_update.flags = BPF_F_REPLACE;
```
alessandrod, line 351: same here, think you want to `flags | BPF_F_REPLACE`

alessandrod (changes requested):
This is looking great now! See comments

aya/src/maps/array/program_array.rs
dave-tucker: Because `fd` required `&mut self`. Fixed it to be `&self` so `program()` works as expected

dave-tucker:
@alessandrod comments addressed :+1:

alessandrod (approved):

PR #285, opened 2022-05-27
aya: Improved BTF Type API

dave-tucker:
This commit removes reliance on generated BtfType structs, as
well as adding a dedicated struct for each BTF type. As such,
we can now add nice accessors like `bits()` and `encoding()`
for Int vs. inlined shift/mask operations.


dave-tucker:
Per a discord discussion looong ago... this was tested against reading /sys/kernel/btf/vmlinux and parsing works great.
Merging to unblock more BTF work.

Issue #288, opened 2022-05-28
`pop()` for Queue map does not return error event it calls against empty queue


nak3:
When `pop()` for Queue map is called against empty queue, the error should be returned.
However, it always return `0` value.

Please refer to the reproduce code in https://github.com/nak3/aya-showcase/tree/main/queue.
It calls `pop()` against empty queue,

```rust
    match Queue::<u32>::pop(&mut QUEUE) {
        Some(val) => info!(&ctx, "value {} found", val), // This should be empty.
        None => info!(&ctx, "not found"),
    }
```

but it always prints `value 0 found`.

```
06:52:42 [INFO] queue: [queue/src/main.rs:56] Waiting for Ctrl-C...
06:52:43 [INFO] queue: [src/main.rs:36] value 0 found
06:52:43 [INFO] queue: [src/main.rs:40] value 0 found
06:52:43 [INFO] queue: [src/main.rs:36] value 0 found
```

This is because `bpf_map_pop_elem()` returns negative value when it fails, but the negative value is not handled by Rust side (I am thinking it is not possible).

PR #289, opened 2022-05-28
Return non-error only when helper function returns `0` in queue map

nak3:
As described in https://github.com/aya-rs/aya/issues/288, queue does not return `error` even calling pop() from the empty queue.

This is because the helper functions (`bpf_map_pop_elem()`)
does not return negative value for rust/aya correctly.

The `bpf_map_pop_elem()` and `bpf_map_push_elem()`
returns `0` on success - https://man7.org/linux/man-pages/man7/bpf-helpers.7.html, so this patch
changes the condition as only `0` on success.

Fix https://github.com/aya-rs/aya/issues/288

### DIFF
--- a/bpf/aya-bpf/src/maps/queue.rs
+++ b/bpf/aya-bpf/src/maps/queue.rs
@@ -51,7 +51,7 @@ impl<T> Queue<T> {
                 flags,
             )
         };
-        (ret >= 0).then(|| ()).ok_or(ret)
+        (ret == 0).then(|| ()).ok_or(ret)
     }
 
     pub fn pop(&mut self) -> Option<T> {
@@ -61,7 +61,7 @@ impl<T> Queue<T> {
                 &mut self.def as *mut _ as *mut _,
                 value.as_mut_ptr() as *mut _,
             );
-            (ret >= 0).then(|| value.assume_init())
+            (ret == 0).then(|| value.assume_init())
         }
     }
 }


nak3:
This same issue exists in other map codes but I send this fix only for queue for now because I am not 100% confidence for this approach. Is it possible to fix `bpf_map_pop_elem()` to return negative value when it fails ? :thinking:

alessandrod:
> This same issue exists in other map codes but I send this fix only for queue for now because I am not 100% confidence for this approach. Is it possible to fix `bpf_map_pop_elem()` to return negative value when it fails ? 🤔

What value is `bpf_map_pop_elem()` returning on the empty queue? It should return `-2` https://github.com/torvalds/linux/blob/a37fb7ef24a475012547fa28f0148d2e538ad5d4/kernel/bpf/queue_stack_maps.c#L110

nak3:
It returns `4294967294` on my env. It seems to be treated as a 32-bit signed integer.

nak3:
No, I'm sure I am running on a 64 bit system :sweat_smile: (I'm neither using VM nor any special environment but just a 64-bit physical Linux machine.)
```
$ uname -a
Linux fedora 5.17.6-200.fc35.x86_64 #1 SMP PREEMPT Mon May 9 14:22:05 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

dave-tucker:
> It returns 4294967294 on my env. It seems to be treated as a 32-bit signed integer.

That's -2 as Alessandro suggested, but it's been converted to a u32 somewhere.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=160f328ce33d04fdd4a4798c9a63a325

nak3:
The original cast issue could be reproduced without aya - https://github.com/nak3/ffi-cast
Rust fails to cast from the function returning `int` to `c_long`.

I will bring up this cast issue for Rust.

PR #290, opened 2022-05-29
bpf: Replace map types to use &self, remove HashMap::get_mut

ishitatsuyuki:
The bpf_map_defs are now wrapped with UnsafeCell, which also happens to
provide a cozy way to get a *mut pointer. An UnsafeCell isn't strictly
required as the struct fields are practically opaque to us, but using an
UnsafeCell follows the general best practices against miscompilation and
also prevents some obvious errors.

HashMap::get_mut was removed because the idea is completely unsound.
Previous users should wrap their data with UnsafeCell instead or use
atomics.

Closes: #233

### DIFF
--- a/bpf/aya-bpf/src/maps/array.rs
+++ b/bpf/aya-bpf/src/maps/array.rs
@@ -1,4 +1,4 @@
-use core::{marker::PhantomData, mem, ptr::NonNull};
+use core::{cell::UnsafeCell, marker::PhantomData, mem, ptr::NonNull};
 
 use aya_bpf_cty::c_void;
 
@@ -10,14 +10,16 @@ use crate::{
 
 #[repr(transparent)]
 pub struct Array<T> {
-    def: bpf_map_def,
+    def: UnsafeCell<bpf_map_def>,
     _t: PhantomData<T>,
 }
+
+unsafe impl<T: Sync> Sync for Array<T> {}
 
 impl<T> Array<T> {
     pub const fn with_max_entries(max_entries: u32, flags: u32) -> Array<T> {
         Array {
-            def: bpf_map_def {
+            def: UnsafeCell::new(bpf_map_def {
                 type_: BPF_MAP_TYPE_ARRAY,
                 key_size: mem::size_of::<u32>() as u32,
                 value_size: mem::size_of::<T>() as u32,
@@ -25,14 +27,14 @@ impl<T> Array<T> {
                 map_flags: flags,
                 id: 0,
                 pinning: PinningType::None as u32,
-            },
+            }),
             _t: PhantomData,
         }
     }
 
     pub const fn pinned(max_entries: u32, flags: u32) -> Array<T> {
         Array {
-            def: bpf_map_def {
+            def: UnsafeCell::new(bpf_map_def {
                 type_: BPF_MAP_TYPE_ARRAY,
                 key_size: mem::size_of::<u32>() as u32,
                 value_size: mem::size_of::<T>() as u32,
@@ -40,15 +42,15 @@ impl<T> Array<T> {
                 map_flags: flags,
                 id: 0,
                 pinning: PinningType::ByName as u32,
-            },
+            }),
             _t: PhantomData,
         }
     }
 
-    pub fn get(&mut self, index: u32) -> Option<&T> {
+    pub fn get(&self, index: u32) -> Option<&T> {
         unsafe {
             let value = bpf_map_lookup_elem(
-                &mut self.def as *mut _ as *mut _,
+                self.def.get() as *mut _,
                 &index as *const _ as *const c_void,
             );
             // FIXME: alignment
--- a/bpf/aya-bpf/src/maps/hash_map.rs
+++ b/bpf/aya-bpf/src/maps/hash_map.rs
@@ -1,4 +1,4 @@
-use core::{marker::PhantomData, mem, ptr::NonNull};
+use core::{cell::UnsafeCell, marker::PhantomData, mem, ptr::NonNull};
 
 use aya_bpf_bindings::bindings::bpf_map_type::{
     BPF_MAP_TYPE_LRU_HASH, BPF_MAP_TYPE_LRU_PERCPU_HASH, BPF_MAP_TYPE_PERCPU_HASH,
@@ -13,15 +13,22 @@ use crate::{
 
 #[repr(transparent)]
 pub struct HashMap<K, V> {
-    def: bpf_map_def,
+    def: UnsafeCell<bpf_map_def>,
     _k: PhantomData<K>,
     _v: PhantomData<V>,
 }
 
+unsafe impl<K: Sync, V: Sync> Sync for HashMap<K, V> {}
+
 impl<K, V> HashMap<K, V> {
     pub const fn with_max_entries(max_entries: u32, flags: u32) -> HashMap<K, V> {
         HashMap {
-            def: build_def::<K, V>(BPF_MAP_TYPE_HASH, max_entries, flags, PinningType::None),
+            def: UnsafeCell::new(build_def::<K, V>(
+                BPF_MAP_TYPE_HASH,
+                max_entries,
+                flags,
+                PinningType::None,
+            )),
             _k: PhantomData,
             _v: PhantomData,
         }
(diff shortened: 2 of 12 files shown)


ishitatsuyuki:
This is a breaking change as it removes `HashMap::get_mut`. I don't know whether you do special changelog entries for them, so let me know if you need a more curated description for changelog.

I'll follow up with a change to aya-book examples as well.

ishitatsuyuki:
Reformatted.

bpf/aya-bpf/src/maps/array.rs
```diff
@@ -10,45 +10,47 @@ use crate::{
 
 #[repr(transparent)]
 pub struct Array<T> {
-    def: bpf_map_def,
+    def: UnsafeCell<bpf_map_def>,
     _t: PhantomData<T>,
 }
 
+unsafe impl<T: Sync> Sync for Array<T> {}
```
dave-tucker, line 17: Why do we need to implement `Sync` we don't have multiple threads here

dave-tucker (commented):
Adding some comments, but this is a very welcome change! Thanks

bpf/aya-bpf/src/maps/array.rs
ishitatsuyuki: `static` vars are assumed to be accessible from multiple threads. eBPF programs in general can be called from multiple threads as well.

bpf/aya-bpf/src/maps/array.rs
```diff
@@ -10,45 +10,47 @@ use crate::{
 
 #[repr(transparent)]
 pub struct Array<T> {
-    def: bpf_map_def,
+    def: UnsafeCell<bpf_map_def>,
```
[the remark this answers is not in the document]
ishitatsuyuki: As above, `Cell` is not thread safe.

bpf/aya-bpf/src/maps/array.rs
ishitatsuyuki: To be clear: it doesn't matter if the Map itself is thread safe or not because it's actually immutable anyway. But in terms of API design, this static object has to be thread safe, especially T, or you will have races when you have something like `Map<RefCell<T>>` and multiple threads tries to access that.

bpf/aya-bpf/src/maps/hash_map.rs
```diff
@@ -229,28 +232,19 @@ const fn build_def<K, V>(ty: u32, max_entries: u32, flags: u32, pin: PinningType
 }
 
 #[inline]
-fn get<'a, K, V>(def: &mut bpf_map_def, key: &K) -> Option<&'a V> {
+fn get<'a, K, V>(def: *mut bpf_map_def, key: &K) -> Option<&'a V> {
```
alessandrod, line 235: This is still unsound btw, since some other program might mutate the key while there's an outstanding `&V`. 

I think we should probably have a `get() -> Option<V>` and then `unsafe fn get_ptr() -> Option<*const V>` and `unsafe fn get_ptr_mut() -> Option<*mut V>` for cases where soundness is guaranteed externally (eg a non-preemptable program working on a per-cpu map)

ishitatsuyuki: Is it? For `BPF_MAP_TYPE_HASH`, "map_update_elem() replaces existing elements atomically" (I think it uses a RCU inside).

ishitatsuyuki: I think the "tearing" unsoundness is unavoidable at this point (not sure why the kernel people decided that reading garbage value is a good idea under any circumstance), but in that case copying values doesn't help either (you can get "teared" while reading that value too). So I propose the following API instead:
```rust
unsafe fn get(&self) -> Option<&V>; // Only safe if the map is "tear-free". Let's put a large warning here
fn get_ptr() -> Option<*const V>; // Unsafe is only required while dereferencing, where invariants could be violated (but there is no use-after-free)
```
As for get_ptr_mut, I still think an UnsafeCell should be used instead ([raw_get](https://doc.rust-lang.org/stable/std/cell/struct.UnsafeCell.html#method.raw_get) will get you decent pointer arithmetic). There are synchronization and reentrancy issues here, both should be taken care of separately.

dave-tucker: Maybe we should just have `get` that returns a `Cell` (which could be our own Cell impl if it needs to be Sync).
Then API-wise, users would have `get()` for a copy of `T` , `set()` and `as_ptr()` :thinking:

ishitatsuyuki: How does that help with the issue at all? No, if you have a “tearing” implementation you can’t even expect to read a proper value. You always have to deal with the possibility of reading garbage, hence only types without invariant are possible.

dave-tucker: > How does that help with the issue at all?

Maybe it doesn't? I'm learning as I go here and I'm right at the limits of my understanding.

I'm just trying to think through how I would use this API...
By convention (enforced by aya-template) all map access occurs in an `unsafe fn try_my_app(ctx: TracePointContext) -> Result<i32,i32>` because almost everything is unsafe. As a result, rustc isn't going to nag at me about unsafety and I probably won't read the docs as to why something is unsafe.

Having lots of options (i.e `get`, `get_ptr`, `get_ptr_mut`) doesn't really feel intuitive to me as it's not clear from the method name which I should use - I'd immediately discount `_ptr_*` because no-one really wants to work with pointers unless they have to... and I fell we'd end up with users asking "where's `get_mut`"?

`get` is most logical to use for immutable access, BUT per your comment it's only really safe to use with `BPF_F_NO_PREALLOC`... which most people probalby aren't using.

So the thought that triggered my comment was, "lets just have `get()` that returns a type that more accurately represents the mutable memory location that I get from the helper function... which ideally has nice setters/getters that explain all the edge cases around reading values etc..."

If that idea isn't helpful, then `get` and `get_mut` returning raw pointers is probably the best option IMHO... and would be similar to working with the helpers from C.

That said, it's really down to you and @alessandrod to agree as you both are far more knowledgeable in this area than myself.

alessandrod: > I think the "tearing" unsoundness is unavoidable at this point (not sure why the kernel people decided that reading garbage value is a good idea under any circumstance),

In a closed system, where you control all access to maps and know that there's no concurrent readers/writers, the kernel API is sound. For cases where you need synchronization, you can use bpf_locks.

> but in that case copying values doesn't help either (you can get "teared" while reading that value too). So I propose the following API instead:
> 
> ```rust
> unsafe fn get(&self) -> Option<&V>; // Only safe if the map is "tear-free". Let's put a large warning here

The problem isn't only tearing. The problem is that the value behind a `&V` can be mutated even in absence of races. The value can be mutated to another perfectly valid value (no tearing, no corruption), but that's a problem in itself since it violates rust aliasing rules (if you have a `&T`, the value can't be mutated as long as the `&T` exists).

> fn get_ptr(&self) -> Option<*const V>; // Unsafe is only required while dereferencing, where invariants could be violated (but there is no use-after-free)

ack

> ```
> As for `get_ptr_mut`, I still think an UnsafeCell should be used instead ([raw_get](https://doc.rust-lang.org/stable/std/cell/struct.UnsafeCell.html#method.raw_get) will get you decent pointer arithmetic). There are synchronization and reentrancy issues here, both should be taken care of separately. Finally, nothing stops you from looking up the same element twice.

I'm not entirely sure I understand what you mean, can you provide an example of the API you have in mind and its use to write to an existing map element?

alessandrod: > > How does that help with the issue at all?
> 
> Maybe it doesn't? I'm learning as I go here and I'm right at the limits of my understanding.
> 
> I'm just trying to think through how I would use this API... By convention (enforced by aya-template) all map access occurs in an `unsafe fn try_my_app(ctx: TracePointContext) -> Result<i32,i32>` because almost everything is unsafe. As a result, rustc isn't going to nag at me about unsafety and I probably won't read the docs as to why something is unsafe.
> 

Yeah that needs to be changed, the generated programs shouldn't be unsafe, people should opt in to unsafe.

> Having lots of options (i.e `get`, `get_ptr`, `get_ptr_mut`) doesn't really feel intuitive to me as it's not clear from the method name which I should use - I'd immediately discount `_ptr_*` because no-one really wants to work with pointers unless they have to... and I fell we'd end up with users asking "where's `get_mut`"?

This is exactly the problem, to come up with a generic API, they *must* deal with pointers because the eBPF platform doesn't provide facilities to return anything else in rust. If we yolo and provide APIs that return `&V` and `&mut V` like we do now, people will discount `ptr_*` methods and think their code is sound until it starts failing in obscure ways.

> `get` is most logical to use for immutable access, BUT per your comment it's only really safe to use with `BPF_F_NO_PREALLOC`... which most people probalby aren't using.
> 
> So the thought that triggered my comment was, "lets just have `get()` that returns a type that more accurately represents the mutable memory location that I get from the helper function... which ideally has nice setters/getters that explain all the edge cases around reading/writing values etc..."


> If that idea isn't helpful, then `get` and `get_mut` returning raw pointers is probably the best option IMHO... and would be similar to working with the helpers from C.

We are going to provide APIs that are always safe, and those will work with pointers. The question is, in *addition*, should we also provide a `get(...) -> &V` and document what the caveats are so that in 90% of the cases people can still safely use it?

alessandrod: > > The problem isn't only tearing. The problem is that the value behind a &V can be mutated even in absence of races.
> 
> It's really the same problem. The `update` calls are not coded to overwrite the existing element — even in the case of prealloc, they write to a per-cpu temporary slot. However they only have a single slot in this case — so if a second write comes in without waiting for the grace period, we're now seeing either an overwrite or a tear.
> 
> As such tear-free maps are also overwrite-free.
> 
> > I'm not entirely sure I understand what you mean, can you provide an example of the API you have in mind and its use to write to an existing map element?
> 
> ```rust
> UnsafeCell::raw_get(map.get_ptr(k))
> ```
> 
> returns a `*mut` pointer. The user should `mem::write` through that pointer.

raw_get takes UnsafeCell, so that wouldn't work

ishitatsuyuki: It *should* be an UnsafeCell. UnsafeCell makes it explicit that the application will not alias or use the *mut pointer in a racy way.

ishitatsuyuki: The user should specify their V as UnsafeCell<T>.

alessandrod: That seems unsound too? There's no way to have an `UnsafeCell` held inside a bpf map that respects the invariants required by `UnsafeCell`

alessandrod: I'm not sure I understand how handing out `*mut V` is worse than using UnsafeCell. Pointers have no aliasing rules, so once you get a pointer, all bets are off, the API kinda sucks, you must know what you're doing, but there's no unsoundness *in the API*.

If we tell people to use UnsafeCell, we're creating unsoundness from the start, since UnsafeCell isn't compatibile with the BPF threading model.

ishitatsuyuki: > I'm not sure I understand how handing out *mut V is worse than using UnsafeCell. Pointers have no aliasing rules, so once you get a pointer, all bets are off, the API kinda sucks, you must know what you're doing, but there's no unsoundness in the API.

Sure, both involves unsafe and both leave the responsibility to the user. But the BPF map handing out *mut pointers make no sense, stuffs like lazy_static or thread_local doesn't hand out mut pointers for obvious reason.

> If we tell people to use UnsafeCell, we're creating unsoundness from the start, since UnsafeCell isn't compatibile with the BPF threading model.

UnsafeCell isn't compatible with *any* threading model, the point is that the user is intentionally specifying that they would do it unchecked anyway.

alessandrod: I should say that I do like the idea of _not_ having `get_ptr_mut` so people are less tempted to use it. But I don't like that the alternative is to tell people to do something that isn't sound.

ishitatsuyuki: Is `get_ptr_mut` any more sound than using an `UnsafeCell`?

alessandrod: Yes? `get_ptr_mut` gives you a pointer, the pointer is always valid. Doesn't violate aliasing, doesn't violate the memory model. It is still a huge foot gun, but it's not unsound. 

If we tell you to put UnsafeCell in a map and the cell gets written my multiple programs without using bpf_lock, that's unsound. Ok we could say in the docs this is only sound if you do this and that, but then, is it worth potentially creating unsoundness to make it slightly harder to get a `*mut V`? This is not a rhetorical question, this is part of what we need to decide here.

alessandrod: > UnsafeCell isn't compatible with any threading model, the point is that the user is intentionally specifying that they would do it unchecked anyway.

From the UnsafeCell docs: "At all times, you must avoid data races. If multiple threads have access to the same UnsafeCell, then any writes must have a proper happens-before relation to all other accesses (or use atomics)."

Which is precisely not the case in our case.

ishitatsuyuki: If the docs say so, then it must be right. I'll put a get_ptr_mut function in then.

ishitatsuyuki:
`get()` is now unsafe and additional methods are now added to hash map variants. Additionally I dropped `Sync` requirements on per-cpu variants.

Issue #193, opened 2022-01-31
Add support for BPF_MAP_TYPE_DEVMAP


dave-tucker:

pchaseh:
I currently require this for a project I'm doing with aya-rs, what's the status on having official support for this? (I see that there's a PR which includes it, but has been sitting for quite some time)

Issue #195, opened 2022-01-31
Add support for BPF_MAP_TYPE_CPUMAP


dave-tucker:

tamird:
Done in #527.

Issue #194, opened 2022-01-31
Add support for BPF_MAP_TYPE_XSKMAP


dave-tucker:

tamird:
Done in #527.

PR #292, opened 2022-05-30
Implement XDP Map Types

dave-tucker:
This commit adds implementations for:
- xskmap
- devmap
- devmap_hash
- cpumap

Which can all be used to redirect XDP packets to various different
locations

TODO: 
- [x] Implementation in `aya-bpf`
- [ ] Test program to verify that these work as expected

Closes: #193 #194 #195 #199

### DIFF
--- a/bpf/aya-bpf/src/maps/xdp/dev_map.rs
+++ b/bpf/aya-bpf/src/maps/xdp/dev_map.rs
@@ -1,0 +1,55 @@
+use core::{mem, ptr::NonNull};
+
+use aya_bpf_cty::c_void;
+
+use crate::{
+    bindings::{bpf_map_def, bpf_map_type::BPF_MAP_TYPE_DEVMAP},
+    helpers::bpf_map_lookup_elem,
+    maps::PinningType,
+};
+
+#[repr(transparent)]
+pub struct DevMap {
+    def: bpf_map_def,
+}
+
+impl DevMap {
+    pub const fn with_max_entries(max_entries: u32, flags: u32) -> DevMap {
+        DevMap {
+            def: bpf_map_def {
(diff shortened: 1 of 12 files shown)


bpf/aya-bpf/src/maps/xdp/dev_map.rs
```diff
@@ -0,0 +1,55 @@
...
+        unsafe {
+            let value = bpf_map_lookup_elem(
+                &mut self.def as *mut _ as *mut _,
+                &index as *const _ as *const c_void,
+            );
+            // FIXME: alignment
+            NonNull::new(value as *mut u32).map(|p| p.as_ref())
+        }
+    }
+}
```
pchaseh, line 55: I think that a method such as the following is also necessary, as there's otherwise no way to access the raw map (that I've found) so that `bpf_redirect_map` can be called by a user of this type

```suggestion
    pub unsafe fn redirect(&mut self, key: u64, flags: u64) -> i64 {
        bpf_redirect_map(
            &mut self.def as *mut _ as *mut _,
            key,
            flags,
        )
    }
}
```

brevilo: Correct, I just wondered about the "missing" `bpf_redirect_map()` wrapper myself.

brevilo:
Please consider enabling these new maps for use in multi-buffer contexts (via their [specific section](https://github.com/aya-rs/aya/blob/cfa693bc3b4442a8c97cfcd24551ea6439f25e50/aya-obj/src/obj.rs#L161) names), similar to #519.

Thanks for your efforts!

dave-tucker:
Closing in favour of #527

tamird:
> Closing in favour of #527

Closing for real.

PR #293, opened 2022-05-30
Getters be gone

dave-tucker:
Removes the getters from `aya-gen` and from the generated code checked into the repo.

The changes to `arg.rs` could use some scrutiny as I *think* they are OK, but a few more pairs of eyes (and testing on aarch64 or armv7) would be useful.


alessandrod (approved):
✨✨✨

Issue #12, opened 2021-06-21
Add support for BPF ring buffer


alessandrod:
We should add support for the new ring buffer API: https://www.kernel.org/doc/html/latest/bpf/ringbuf.html

willfindlay:
Perhaps we can take some inspiration from my ringbuf implementation in [libbpf-rs](https://github.com/libbpf/libbpf-rs/blob/master/libbpf-rs/src/ringbuf.rs). We ended up getting a nice API that supports capturing closures as callbacks.

ishitatsuyuki:
#201 would be a duplicate of this, which one should we keep open?

PR #294, opened 2022-06-01
aya: Implement RingBuf

ishitatsuyuki:
Based on @willfindlay's branch, fixed the bugs I found and it appears to be working fine for my setup.

Branch based on #290.

Closes: #12
Closes: #201


ishitatsuyuki:
Doing async with tokio is a little bit more verbose, but I'm doing it like this:

```
    loop {
        let mut guard = ring.readable_mut().await?;
        guard.get_inner_mut().process_ring(&mut |e| {
            // Do things
            Ok(())
        })?;
        guard.clear_ready();
    }
```

ishitatsuyuki:
Looks like this can get stuck with epoll (miss the notification somehow). I'll investigate.

aya/src/maps/ringbuf.rs
```diff
@@ -4,16 +4,333 @@
...
+            page_size,
+            mask,
+        })
+    }
+
+    /// Retrieve an event from the ring, pass it to the callback, mark it as consumed, then repeat.
+    /// Returns when either the callback returns an Err or there's no more events.
+    pub fn process_ring(
+        &mut self,
+        callback: &mut dyn FnMut(&[u8]) -> Result<(), CallbackError>,
```
dave-tucker, line 222: I would make `callback` just a `FnMut(&[u8])`. I don't think we ought to stop ring processing on error...
The only way to stop would be to panic in `callback`, which given this isn't async is probably the best we could hope for

dave-tucker (changes requested):
Looking good so far.
I think that in async version of this code we should - instead of using callback - use the the tx side of an mpsc channel.
i.e: 
- https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html
- https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html

So `Sender<T>`

This way: 
- Ring processing speed is not bounded by the efficiency of `callback`
- Ring processing is not stopped on error
- We don't need to wrap user error types to return back to the caller - so no `anyhow` dependency

Depending on the implementation we may want to use `try_send()` to give errors on channel closed or buffer full

willfindlay:
Thanks for taking this over!

ishitatsuyuki:
Re: the "stuck" polling issue, it's an actual upstream flaw. The ringbuf *claims* to be designed around acquire-release ordering but it's actually relying on sequential consistency implicitly in the kernel space, while the userspace counterpart is simply busted. Consider the following simplified example, where the "reserve" and "commit" operation is fused and simply increments the producer counter. A notification is sent if and only if the incremented producer counter == consumer counter + 1.

```
kernel: write p 2 -> read c X -> write p 3 -> read c 1 (X doesn't matter)
user  : write c 2 -> read p 2
```

With sequential consistency it's guaranteed that once p is updated from 2 to 3, c == 1 cannot be observed since the write of c == 2 precedes the read of p == 2 which must happen before the write of p = 3. However, with acq-rel there's no meaningful causal relation here at all: the relationship is only established on reads, and the read of c == 1 nor p == 2 only builds causal relationship for operations happened before the first operation in the above diagram.

And what this means is that we need a SeqCst fence after both sides of the write. For the userspace, we use a SeqCst store; but the kernel space might seem doomed. Luckily, it actually does a `xchg` operation to update the BUSY flag which implicitly serves as a full SeqCst barrier. Therefore, the kernel side doesn't actually need to be changed; but rather unintentionally, since the comments only suggest acq-rel relationship which is insuffcient.

ishitatsuyuki:
A bunch of updates — documentation for `aya-bpf`, a reserve/commit API that is actually safe, and addressing review comments about process_ring error handling, plus the fix (+ a small explanation) for the awful coherence issue that I tried figured out the right solution for days. I'll open a separate PR for the codegen soon.

ishitatsuyuki:
Updated `RingBufEntry` to consume `self` on `discard/submit`.

ishitatsuyuki:
Suppressed a clippy lint, see code for details.

dave-tucker:
@ishitatsuyuki codegen changes are in `main` so feel free to rebase to pick those up when you're ready.
will take a look at the updates and see if I've got any review comments to add, but I think we're waiting on @alessandrod to take a look before this is ok to merge.

dave-tucker:
> Using mpsc necessitates a copy — ring processing speed isn't really the concern here, as the application has the freedom to size the ring buffer. (Also mpsc means double ring buffering — not sure that's on point.)
> 
> The only other zero-copy option would be returning a guard object that updates consumer pointer on Drop — but that's not much better I'd say.

Having thought on this more, I agree - you're right. Different applications are going to have different requirements, some which can be met by sizing the ringbuf, data notification control etc.. Either way, the API as it stands seems flexible enough that I could do whatever I want.

ishitatsuyuki:
Rebased now that the &self PR is merged.

aya/src/maps/ringbuf.rs
```diff
@@ -0,0 +1,355 @@
...
+            consumer_pos_ptr: consumer_page as *mut _,
+            producer_pos_ptr: producer_pages as *mut _,
+            page_size,
+            mask,
+        })
+    }
+
+    /// Retrieve an event from the ring, pass it to the callback, mark it as consumed, then repeat.
+    /// Returns when there's no more events.
+    pub fn process_ring(&mut self, callback: &mut impl FnMut(&[u8])) {
```
alessandrod, line 213: Have we considered returning an iterator that consumes events instead of having a callback?

aya/src/maps/ringbuf.rs
```diff
@@ -0,0 +1,355 @@
...
+            map_fd,
+            data_ptr: unsafe { (producer_pages as *mut u8).add(page_size) },
+            consumer_pos_ptr: consumer_page as *mut _,
+            producer_pos_ptr: producer_pages as *mut _,
+            page_size,
+            mask,
+        })
+    }
+
+    /// Retrieve an event from the ring, pass it to the callback, mark it as consumed, then repeat.
```
alessandrod, line 211: Meta comment about docs: for consistency the format we use is the same as rust's libstd. The format is roughly:

```
First line describes briefly and in 3rd person what the function does.

Second paragraph expands on details, mentioning in discursive form the arguments and the return value.

More paragraphs might be needed to explain caveats and edge cases.

# Example 
...
```

Sentences always end with periods. Paragraphs are separated by an empty line.

alessandrod (changes requested):
This looks amazing, see comments

alessandrod:
Also forgot to say in the review: I think we should implement the async bits of this, like we do for perf buffers

bpf/aya-bpf/src/maps/ringbuf.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+unsafe impl Sync for RingBuf {}
+
+/// A ring buffer entry, returned from [`RingBuf::reserve`].
+/// You must [`submit`] or [`discard`] this entry before this gets dropped.
+///
+/// [`submit`]: RingBufEntry::submit
+/// [`discard`]: RingBufEntry::discard
+#[must_use = "BPF verifier requires ring buffer entries to be either submitted or discarded"]
```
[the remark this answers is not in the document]
ishitatsuyuki: The verifier will complain.

For ergonomics, I think we want to commit not discard on Drop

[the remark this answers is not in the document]
alessandrod: What is the best argument for committing by default? Since the verifier will complain, now I'm thinking that we shouldn't do anything. I think we want to teach people to commit manually: the sooner you commit, the faster the ringbuf advances; if you're doing multi-event commits, you need to know that ordering matters etc.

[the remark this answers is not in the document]
ishitatsuyuki: Commit order doesn’t matter (much), the ring order is based on reservation order.

Committing by default might be ergonomic because the reserve/commit API is used a lot for efficiency purpose (zero copy), and discard is something rarely wanted in that case.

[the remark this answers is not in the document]
alessandrod: It does matter: if you reserve s0 s1 and s2, if you commit s1 and s2 the buffer won't notify, it will only do so once s0 is committed.

aya/src/maps/ringbuf.rs
```diff
@@ -0,0 +1,355 @@
...
+///         // Do something with the data bytes
+///     });
+/// }
+/// # Ok::<(), Error>(())
+/// ```
+///
+/// [`PerfEventArray`]: crate::maps::PerfEventArray
+#[doc(alias = "BPF_MAP_TYPE_RINGBUF")]
+pub struct RingBuf<T: DerefMut<Target = Map>> {
+    _map: Arc<T>,
```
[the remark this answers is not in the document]
ishitatsuyuki: Removed.

bpf/aya-bpf/src/maps/ringbuf.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// requests. If you use this with a `T` that isn't properly aligned, this function will
+    /// be compiled to a panic and silently make your eBPF program fail to load.
+    pub fn reserve<T: 'static>(&self, flags: u64) -> Option<RingBufEntry<T>> {
+        // The reserved pointer may be null, which we handle with an Option.
+        // We also need to ensure that the returned pointer is of a proper sized allocation and
+        // satisfies T's alignment requirements.
+        // Finally, cast it to an MaybeUninit as creating a reference to uninitialized memory is UB.
+
+        // ringbuf allocations are aligned to 8 bytes (hardcoded in kernel code).
+        assert!(8 % mem::align_of::<T>() == 0);
```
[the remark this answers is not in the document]
ishitatsuyuki: In general, anything that doesn't put a bound on T doesn't seem to work, as Rust doesn't allow any error to be spit out at monomorphization time. If the "silent fail" is deemed too annoying, then we might want to remove this check altogether.

aya/src/maps/ringbuf.rs
ishitatsuyuki: I added some newlines. I don't write explanation for most of the arguments because most of the time they are the only argument and the meaning is obvious, but if you have any feedback around how docs is written in this PR, let me know.

aya/src/maps/ringbuf.rs
ishitatsuyuki: We might be able to do that, but not with an `Iterator`, because then we get into the streaming iterator and self reference hell.

Return a non-Iterator handle struct that mutably borrows the RingBuf will probably work, and that can be used easily with a while loop like `while let Some(entry) = ringbuf.next()`. The reader pointer is incremented on `Drop`. What do you think?

ishitatsuyuki:
On async impl: For the callback model there's not much thing we can do inside aya to support an async model (see https://github.com/aya-rs/aya/pull/294#issuecomment-1143375108); though, if we move to an iterator-like model then it probably make sense to do that.

bpf/aya-bpf/src/maps/ringbuf.rs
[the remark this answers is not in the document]
Matthias247: I think in regular Rust you would want to borrow the actual RingBuffer in the `RingBufEntry` and have a lifetime on it, so that the entry can't outlive the datastructure. And either commit or discard on drop (probably discard because the value might be unset?).

One more thought: in regular Rust actually both options are not the safest, since one could `std::mem::forget` a RingBufEntry which would then make with questionable what happens with follow-up reservations - that's the lesson of all the scoped-thread and async stuff. A safer option to model the problem is by providing some kind of scoped-based API along:

```rust
fn with_reservation(&mut self, callback: impl FnOnce(&mut MaybeUninit<T>)
```

Seems like the BPF verifier might make some of those things unnecessary, but it it might still be good to follow Rust best-practices.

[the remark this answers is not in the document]
ishitatsuyuki: > I think in regular Rust you would want to borrow the actual RingBuffer in the RingBufEntry and have a lifetime on it, so that the entry can't outlive the datastructure.

That's a good point.

> A safer option to model the problem is by providing some kind of scoped-based API along...

IMO, callback based APIs gets annoying very quickly (particularly because they block the ergonomic use of `?` operator etc) and I think it might worth sacrificing a bit of verifier-pass "safety" in favor of ergonomics.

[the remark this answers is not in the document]
alessandrod: > I think in regular Rust you would want to borrow the actual RingBuffer in the `RingBufEntry` and have a lifetime on it, so that the entry can't outlive the datastructure. And either commit or discard on drop (probably discard because the value might be unset?).

See my comment above: I think getting a verifier error is probably better than a silent bug because you forgot to commit and we discarded on drop. Or in other words: the verifier will force you to either commit or discard - I think that's good dx we should keep. What I don't like is that the verifier error will be somewhat hard to correlate with a rust source code line, so if we could improve that, we should do so. But committing or discarding in drop seems like a bad idea.

As for `RingBufEntry` borrowing the map yes that's a good idea, although it should be an immutable borrow, or it won't be possible to reserve multiple entries to do batching.

> One more thought: in regular Rust actually both options are not the safest, since one could `std::mem::forget` a RingBufEntry which would then make with questionable what happens with follow-up reservations - that's the lesson of all the scoped-thread and async stuff. A safer option to model the problem is by providing some kind of scoped-based API along:
> 
> ```rust
> fn with_reservation(&mut self, callback: impl FnOnce(&mut MaybeUninit<T>)
> ```

`&mut` wouldn't work with batching. Also in this case, since the verifier will reject a forgotten entry anyway, I don't think it's worth sacrificing a better API (iterator), to avoid an unsoundness issue that in practice can not happen.

aya/src/maps/ringbuf.rs
alessandrod: It's not immediately clear to me why it would be self referential?

ishitatsuyuki: Oops, it's not a self reference, but since we can't use lifetime without GATs, the entry returned from the iterator will need something like `Rc` that references the main `RingBuf` to properly work.

bpf/aya-bpf/src/maps/ringbuf.rs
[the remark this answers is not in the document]
Matthias247: Is it actually possible to call `bpf_ringbuf_reserve` multiple times? The use-case to reserve multiple slots before comitting can also be fulfilled by calling it a single time and reserving more than 1 element. The downside of allowing multiple  `.reserve()` calls which do an immutable borrow is that dropping/submitting them will commit the next possible slot, even if the slot actually refers to something else. E.g.

So I think the mutable borrow is maybe still a good hoice, and it can be slightly modified to handle multiple entries. E.g.  `reserve` could be modified to take a count parameter and return a modified `RingBufEntry/RingBufView` that implements `Target = [MaybeUnit<T>]`. Then the caller can write multiple entries. It could also offer a `submit(n: usize`) method which allows to specify how many of the elements would be submitted. Although I'm not sure whether that needs const generic magic to actually work or whether a deref to an array of variable size actually works - probably needs some experimentation. 

Btw: 
> Also in this case, since the verifier will reject a forgotten entry anyway,

I'm curious on what base it actually does that? Does it really check whether reserve/submit calls are properly paired? The case of forgetting something isn't really one of the usual memory issues that those environments are meant to protect against.

vadorovsky:
@ishitatsuyuki Hi! Do you have time to address the remaining comments? If not, I would be happy to take over.

ishitatsuyuki:
I'll try to get a revision on this later today.

ishitatsuyuki:
Sorry, ran out of time today but will prioritize this tomorrow.

bpf/aya-bpf/src/maps/ringbuf.rs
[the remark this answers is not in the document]
ishitatsuyuki: > Is it actually possible to call bpf_ringbuf_reserve multiple times?

It's perfectly fine to do that. Whether a slot is committed or not is tracked by a flag placed right before the slot's memory.

> dropping/submitting them will commit the next possible slot, even if the slot actually refers to something else

I don't think that happens. `submit` takes the pointer to the slot you're committing.

> I'm curious on what base it actually does that? Does it really check whether reserve/submit calls are properly paired? 

I don't know, honestly. The verifier's message is "unreleased reference", so I'd say it's basically on the same ground as preventing pointers from escaping from a "safe" (non-probe) eBPF scope.

ishitatsuyuki:
I'm getting really busy with internship kicking in in August so I will prefer someone to take over this. Sorry for the delay and for missing the promise.

I investigated implementing an iterator API, but doing this naively involves more overhead since it doesn't allow caching the producer index and results in a redundant cache miss due to how cache coherency works. It's probably not hard to cache this though, but I haven't tried that yet.

vadorovsky:
No worries, I will try to take this over.

ishitatsuyuki:
I still don't have time to commit to this but I've made some corrections to the memory ordering used on the host side, which I've found while doing research on the correctness of SPSC queues. The justification can be found in the comments.

No change is expected for compiled code on x86 since loads are always compiled to ordinary loads.

ishitatsuyuki:
Just did a rebase to fix build errors. Since I keep getting nerdsniped by the atomics stuff (just figured out that SeqCst is only required for the test right before you sleep/return None), I'm thinking of finishing this incrementally in my free time. @vadorovsky I hope you don't mind?

I think the review comments that we definitely should address are:
- Have an iterator like interface. (It looks like the overhead will be negligible if you cache them, machine code wise this is going to be store forwarded or a L1 cache hit. https://github.com/rigtorp/SPSCQueue is a good reference point)
- Have an async iterator like interface. I think this should come in a follow-up PR, mostly because this PR is already getting too aged, but also because there might be additional design decisions to be made. One of them is batching during polling: for efficiency I find that sleeping for a small amount of time after getting the epoll notification greatly improves efficiency. But I'm not sure what kind of default behavior and what kind of configuration should be implemented.

alessandrod:
> I think the review comments that we definitely should address are:
> 
>     * Have an iterator like interface. (It looks like the overhead will be negligible if you cache them, machine code wise this is going to be store forwarded or a L1 cache hit. https://github.com/rigtorp/SPSCQueue is a good reference point)

Ah thanks for the link looks interesting

>     * Have an async iterator like interface. I think this should come in a follow-up PR, mostly because this PR is already getting too aged, but also because there might be additional design decisions to be made. One of them is batching during polling: for efficiency I find that sleeping for a small amount of time after getting the epoll notification greatly improves efficiency. But I'm not sure what kind of default behavior and what kind of configuration should be implemented.

Ack I'm perfectly ok with the async bit being done separately. We need to implement some kind of batching for perf event arrays too, I didn't do it at the time since I wasn't sure what the API to opt-in to batching would look like.

ishitatsuyuki:
Sorry, after a long period of silence, here it is. I replaced `process_ring()` completely with a iterator-like API. A usage example is in the doc.

I didn't go as far as implementing `Iterator` though, since it has to be streaming, although this might be feasible now that GATs are stabilized.

I did test this a while ago, so it probably should work. But looks like right now I need to downgrade my kernel due to BTF changes which I don't have time to do.

brevilo:
Just wanted to say that this is real great and *highly anticipated* work 🚀 What can be done to see this merged and released?

Keep it up!

aya/src/maps/ringbuf.rs
```diff
@@ -0,0 +1,338 @@
...
+/// # }
+/// # async {
+/// # let mut bpf = aya::Bpf::load(&[])?;
+/// use std::convert::{TryFrom, TryInto};
+///
+/// let mut ring = RingBuf::try_from(bpf.map_mut("EVENTS").unwrap())?;
+///
+/// // Poll would be a helper that takes an `AsRawFd` implementation and provides functionality
+/// // to wait for the file descriptor to be readable.
+/// let mut poll = Poll::new(ring);
```
vadorovsky, line 76: I don't think it's a good example of how using ring buffer with Tokio should look like. IMO we shouldn't require users to write or provide any wrapper structures just to consume events. Also, it's quite inconsistent with the API we have for perf buffers with `AsyncPerfEventArray`.

I was playing with that branch recently and I think that it would be nice to have a dedicated "async" struct, similar to `AsyncPerfEventArray` . My idea is to have some method which consumes a bytes buffer and fills it with data from ring buffer, without returning the `RingBufItem` - so the user would just get the data and the original ring buffer item would be dropped, therefore immediately updating producer/consumer positions.  I imagine that it could look somehow like:

```rust
use std::os::{fd::AsRawFd, unix::prelude::RawFd};

#[cfg(feature = "async_std")]
use async_io::Async;
use bytes::BytesMut;
use thiserror::Error;
#[cfg(feature = "async_tokio")]
use tokio::io::unix::AsyncFd;

use crate::maps::{ringbuf::RingBuf, MapData, MapError};

/// Async ring buffer error.
#[derive(Error, Debug)]
pub enum AsyncRingBufError {
    /// A map error.
    #[error("map error: {0}")]
    MapError(MapError),

    /// An IO error.
    #[error(transparent)]
    IOError(#[from] std::io::Error),
}

#[cfg(any(feature = "async_std", feature = "async_tokio"))]
pub struct AsyncRingBuf<T: AsRef<MapData>> {
    ring_buf: RingBuf<T>,

    #[cfg(feature = "async_std")]
    async_fd: Async<RawFd>,

    #[cfg(feature = "async_tokio")]
    async_fd: AsyncFd<RawFd>,
}

#[cfg(any(feature = "async_std", feature = "async_tokio"))]
impl<T: AsRef<MapData>> AsyncRingBuf<T> {
    pub(crate) fn new(map: T) -> Result<AsyncRingBuf<T>, MapError> {
        let ring_buf = RingBuf::new(map)?;
        #[cfg(any(feature = "async_std", feature = "async_tokio"))]
        let fd = ring_buf.as_raw_fd();
        Ok(AsyncRingBuf {
            ring_buf: ring_buf,
            #[cfg(feature = "async_std")]
            async_fd: Async::new(fd).unwrap(),
            #[cfg(feature = "async_tokio")]
            async_fd: AsyncFd::new(fd).unwrap(),
        })
    }
}

#[cfg(feature = "async_tokio")]
impl<T: AsRef<MapData>> AsyncRingBuf<T> {
    /// Try to take a new entry from the ringbuf.
    pub async fn read_item(&mut self, buffer: &mut BytesMut) -> Result<usize, AsyncRingBufError> {
        loop {
            let mut guard = self.async_fd.readable_mut().await?;
            match self.ring_buf.next().as_deref() {
                Some(item) => {
                    buffer.extend_from_slice(item);
                    return Ok(item.len());
                }
                None => {
                    guard.clear_ready();
                    continue;
                }
            }
        }
    }
}
```

And the userspace code consuming it could look like:

```rust
let mut ring = AsyncRingBuf::try_from(bpf.map_mut("EVENTS").unwrap()).unwrap();
let mut buf = BytesMut::with_capacity(1024);

loop {
    let _ = ring.read_item(&mut buf).await.unwrap();

    // Here I'm just using u64 for simplicity, but we could handle any type.
    let val = u64::from_ne_bytes(buf[..mem::size_of::<u64>()].try_into().unwrap());

    info!("{}", val);

    buf.clear();
}
```

(I didn't include `read_item` implementation for async-std in this example, but we should have it too)

/cc @alessandrod

vadorovsky (changes requested):
Thanks for working on this and sorry for very late review!

I played a bit with this branch. It works great and I'm very happy with how ring buffer is supported here from the low-level point of view. However, I have some remarks about the API. See the comment.

aya/src/maps/ringbuf.rs
vadorovsky: OK, actually my example is bad, because it always copies the item. I will try to come up with a better one which doesn't drop the item.

brevilo:
Great to see this being taken up again! It will be an important and highly anticipated addition to Aya's feature set!

For reference: [here](https://nakryiko.com/posts/bpf-ringbuf) is the upstream author's take on the ringbuf goals, it's similarities to the perfbuf implementation and its advantages over it. I think this should guide how things are fleshed out in Aya as well.

aya/src/maps/ringbuf.rs
brevilo: Just curious: would "leave out the async integration" mean that reading the ring buffer from within a dedicated tokio task (thread) wouldn't be supported?

ishitatsuyuki: No, it should work, just that putting that in a wrapper function happens to be tripping compile error for no reason. I do it like this:

```rust
    let ring_processor = async move {
        // Throttle polling to get batched and more CPU-efficient event processing
        let interval = Duration::from_micros(200);
        let mut next_poll = Instant::now() + interval;
        loop {
            let time_before_poll = Instant::now().saturating_duration_since(next_poll);
            if !time_before_poll.is_zero() {
                sleep(time_before_poll).await;
                next_poll = next_poll + interval;
            } else {
                next_poll = Instant::now() + interval;
            }
            let mut guard = ring.readable_mut().await?;
            while let Some(e) = guard.get_inner_mut().next() {
                // do something with e
            };
            guard.clear_ready();
        }
        Ok::<(), anyhow::Error>(())
    };

```

brevilo: Ok, thanks. I'll leave those details to you guys. Just wanted to chime in and say that being able to process a ring buffer's events in a dedicated/separate tokio task is a necessity, at least for my use case :)

ajwerner:
I'm interested in helping get this over the finish line. What's needed to complete this work?

alessandrod:
> I'm interested in helping get this over the finish line. What's needed to complete this work?

It looks like the sync API is in place but the async API still needs works

ajwerner:
@ishitatsuyuki did you ever have an example of a test program that demonstrated the epoll hang? I couldn't exactly follow your arguments enough to justify the fences in code, so I went with memory accesses like in libbpf for https://github.com/aya-rs/aya/pull/629. I'd love to understand when that fails and how.

ishitatsuyuki:
Make sure that the epoll fd is configured with edge trigger. Level trigger is why libbpf "accidentally" works (after a memory fence, it calls into a kernel function that tests the ring counter again).

ishitatsuyuki:
Just realized I previously created a repro as described [here](https://lore.kernel.org/bpf/CANqewP0ne0vmPydNFMY=4ProLr3Y_5j7Fnhk+m4V3sdsuofMJw@mail.gmail.com/). Please try it and let me know if this one gets stuck.

As I described at the time of investigation a SeqCst write is definitely necessary for correctness, when:
- The producer is trying to skip the notification, or
- The consumer is about to go into sleep.
Otherwise, cases like the example outlined in the email was fairly reproducible at the time I worked on the PR.

ishitatsuyuki:
As for threads, you don't need more than one producer thread to reproduce this; this "race" can happen between just one consumer and on producer.

ajwerner:
Your repro program was helpful, thanks for it. I was being saved on my side by using `AtomicU32::fetch_add(&self, Release)` that became `lock; addl`. Taking this insight, we can see that if one uses `xchg` (which we observe in rust with `AtomicU32::store(&self, SeqCst)` when writing the consumer position, then everything is good -- there's not risk that the kernel will see a stale consumer position. 

I don't have a sense of the cost of these various approaches, but in general my sense is that a complete fence before sleeping is probably more expensive. What do you think?

ajwerner:
I suppose that the reason the `SeqCst` store for the consumer offset works is because of the effective `SeqCst` store that occurs when the kernel clears the ready bit. If the write to the consumer offset came after the kernel had cleared the busy bit, then the consumer will see the newly produced message. If the write to the consumer offset came before the kernel had cleared the busy bit, then the read of the consumer position will see that the consumer is caught up.

ishitatsuyuki:
In general on x86 I consider any instruction with the `lock` prefix to have a similar "base" cost, as they serialize memory access by acting as a barrier. (That said, do note that contended atomics can be far more expensive.) Using `fetch_add` means that we pay that cost on every position update; while with a memory barrier before sleep, we only pay that cost once per sleep.

Aside, relying on atomic release RMW to provide SeqCst semantics is technically an ill assumption (iirc AcqRel RMW is actually equivalent to SeqCst even in the spec).

brevilo:
Seems like this PR is obsolete now that #629 got merged, no?

PR #297, opened 2022-06-03
aya-gen cli rework

davibe:
Addressing what's left of https://github.com/aya-rs/aya/issues/95

- rename command from `btw-types` to `generate`
- support `--header`
- allow passing-in additional `bindgen` cli params


aya-gen/src/generate.rs
```diff
@@ -0,0 +1,106 @@
...
+    let flags = bindgen.command_line_flags();
+    let bindgen_args: Vec<&OsStr> = bindgen_args
+        .iter()
+        .map(|s| OsStr::from_bytes(s.as_ref().as_bytes()))
+        .collect();
+
+    let output = Command::new("bindgen")
+        .arg(file_path)
+        .args(&flags)
+        .args(bindgen_args)
```
ishitatsuyuki, line 76: This doesn't work as expected. Bindgen passes any clang flags after `--`, and since `flags` contains `--` right now every user bindgen arg is interpreted as clang arg instead.

ishitatsuyuki (changes requested):

aya-gen/src/generate.rs
davibe: Understood, I am going to fix it. Thanks!

ishitatsuyuki (approved):
LGTM. I like how the split logic is properly separated

PR #312, opened 2022-06-08
aya-gen: allow passing custom --ctypes-prefix to bindgen

davibe:


aya-gen/src/generate.rs
```diff
@@ -104,14 +108,28 @@ fn c_header_from_btf(path: &Path) -> Result<String, Error> {
...
 
+fn strip_ctypes_prefix(s: &[String]) -> (Vec<String>, Option<String>) {
+    let mut it = s.splitn(2, |el| el == "--ctypes-prefix");
+    let mut prefix = None;
+    let mut flags = Vec::new();
+    flags.extend_from_slice(it.next().unwrap());
+
+    if let Some(after) = it.next() {
+        prefix = after.get(0).cloned();
+        flags.extend_from_slice(&after[1..]);
```
alessandrod, line 119: I think this will panic with "--foo bar --ctypes-prefix"? (no prefix value)

PR #316, opened 2022-06-09
aya-bpf: Implement set() for array maps

sbernauer:
Hi aya-team, thanks for your awesome project!

I was missing an option to set array values from my ebpf code so i added the needed functions

### DIFF
--- a/aya-gen/src/bin/aya-gen.rs
+++ b/aya-gen/src/bin/aya-gen.rs
@@ -40,12 +40,10 @@ fn try_main() -> Result<(), anyhow::Error> {
             names,
             bindgen_args,
         } => {
-            let bindings: String;
-            if let Some(header) = header {
-                bindings = generate(InputFile::Header(header), &names, &bindgen_args)?;
-            } else {
-                bindings = generate(InputFile::Btf(btf), &names, &bindgen_args)?;
-            }
+            let bindings = match header {
+                Some(header) => generate(InputFile::Header(header), &names, &bindgen_args)?,
+                None => generate(InputFile::Btf(btf), &names, &bindgen_args)?,
+            };
             println!("{}", bindings);
         }
     };
--- a/aya-gen/src/generate.rs
+++ b/aya-gen/src/generate.rs
@@ -132,7 +132,7 @@ mod test {
     use super::combine_flags;
 
     fn to_vec(s: &str) -> Vec<String> {
-        s.split(" ").map(|x| x.into()).collect()
+        s.split(' ').map(|x| x.into()).collect()
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


sbernauer:
Thanks for your fast response! I have to admit I didn't understand it fully.

Is the problem with `set()` or with `get_mut()` or with both? I honestly only care about `set()` and only implemented `get_mut()` for the sake of completeness (it exists in PerCpuArray but not in Array). I'm happy to remove `get_mut()` from Array again if you'd like.

If the `set()` is unsound my question would be how to properly update array elements from an ebpf program? I've only used `bpf_map_update_elem` from C code so far. (In my special case i don't care about data races but that's normally not the case ;)). I see you refering to a `UnsafeCell` but i have never used that and I' not sure how to use that to modify the Array.
Any help would be really appreciated.

dave-tucker:
Hey @sbernauer thanks for the contribution!

What @ishitatsuyuki is referring to is a discussion that's happening over in #290, where the API for maps is undergoing a large change.

`set()` is looks ok, but `get_mut()` would be replaced by `get_mut_ptr()`.

I think it makes sense to hold off reviewing this until after #290 lands, which won't be too much longer I hope.

ishitatsuyuki:
First, the thing about soundness is that you can't write to a memory location while another thread has a reference to it. Having a safe `set()` function makes that UB extremely easy to do, and making it safe is definitely a no-go.

On second thought, I think doing things correctly is extremely hard in eBPF, even with an UnsafeCell. The only safe way is to use atomics or bpf locks, but I don't know if aya actually supports those.

The way I recommend to do things with the current API is to wrap your struct with an UnsafeCell. First, rename your struct to `FooInner`, then wrap it like this:

```
struct Foo(UnsafeCell<FooInner>);
unsafe impl Sync for Foo {}
```

Then, do `FOO_MAP.get(index).0.get()` will get you a `*mut FooInner`. By dereferencing this pointer, you assert that this is the (globally) only reference to the element and it's race-free (which is only the case if the map is per-cpu). AFAIK, multi-thread maps are never race-free unless you use atomics to introduce proper happens-before relations.

sbernauer:
Well that was faster than expected. Thanks for merging https://github.com/aya-rs/aya/pull/290!
I updated my PR so that it only adds `set()`. It would be great if you could give it another round of review.

ishitatsuyuki:
As I have mentioned, the set() operation cannot be safe because that would create a soundness hole: it allows you to mutate an entry you might already be holding a reference to.

Now that `get_mut_ptr` is added, you should dereference the `*mut` pointer from that and write to it.

brevilo:
Related: https://github.com/aya-rs/aya/pull/402

sbernauer:
Setting array values can now be done via `Array::get_ptr_mut` (implemented in #402), this can be closed

PR #321, opened 2022-06-14
Fix linting issues

davibe:
Fixes [linting errors](https://github.com/aya-rs/aya/runs/6871782093?check_suite_focus=true) on `main`
A good part of those were caused by [clap changes in 3.2](https://epage.github.io/blog/2022/06/clap-32-last-call-before-40/)


dave-tucker (approved):
LGTM, Thanks @davibe

PR #322, opened 2022-06-17
aya: Add MapFd trait

dave-tucker:
This allows for programs to get a maps fd, like the program fd trait.


dave-tucker:
Per discord discussion, this isn't a good idea. Closing.

PR #323, opened 2022-06-18
aya: Add Map::fd() function to return a MapFd

dave-tucker:
A rework of #322 to not accomplish the same thing without a trait... in a similar way to how #282 does with `ProgramFd`

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -33,8 +33,15 @@
 //! versa. Because of that, all map values must be plain old data and therefore
 //! implement the [Pod] trait.
 use std::{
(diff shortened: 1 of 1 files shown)


aya/src/maps/mod.rs
```diff
@@ -295,6 +311,11 @@ impl Map {
         self.pinned = true;
         Ok(())
     }
+
+    /// Returns the file descriptor of the Map for use in APIs that require it.
```
alessandrod, line 315: I'd make this:

```
Returns the file descriptor of the map.

Can be converted to [`RawFd`] using [`AsRawFd`].
```

alessandrod (approved):
Looks good, minor doc nit.

PR #325, opened 2022-06-22
Add missing inline macro to xdp methods

NoneTirex:
I think these inline attributes are missing for XdpContext.


alessandrod (approved):

alessandrod:
Thank you! Will merge as soon as tests complete

PR #326, opened 2022-06-22
Change condititon for some bpf helpers

RealTirex:
As described #289, the problem is that negative return values appear as positive values in some kernels.
This change ensures that the described problem does not interfere with the changed functions.


alessandrod (approved):
Thank you!

alessandrod:
For context since part of the discussion on this happened on discord: we seem to be hitting a LLVM miscompilation where LLVM generates `s>` instead of `s>=`. We're going to look at that separately. Meanwhile, this PR is a workaround and generates better code anyway. Fixes https://github.com/aya-rs/aya/issues/288.

PR #327, opened 2022-06-22
aya: Add fuzz targets

dave-tucker:
This is an initial attempt to add fuzzing with cargo-fuzz.
2 targets seemed simple enough to add which are Bpf::load and
Btf::parse_file since these are both public APIs which take byte
streams. Other more structured fuzzing can be added later

### DIFF
--- a/aya/fuzz/.gitignore
+++ b/aya/fuzz/.gitignore
@@ -1,0 +1,3 @@
+target
+corpus
+artifacts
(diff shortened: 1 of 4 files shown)


dave-tucker:
Changing my mind on the best way to do this. I'm not familiar with libFuzzer and the interface isn't as nice as AFL.
Since AFL can be guided via input (i.e valid BPF and BTF sections) I think I'll try that instead.

PR #335, opened 2022-07-02
test: Replace RTF with Rust

dave-tucker:
This commit replaces the existing RTF test runner with a simple rust
binary package called - integration-test.

 integration-test depends on integration-ebpf, which contains test eBPF
code written in Rust and C. `cargo xtask build-integration-test-ebpf`
can be used to build this code and supress rust-analyzer warnings. It
does require `bpf-linker`, but that is highly likely to be available to
developers of Aya. It also requires a checkout of `libbpf` to extract
headers like bpf-helpers.h.

Since everything is compiled into a single binary, it can be run
be run locally using `cargo xtask integration-test` or remotely using
`./run.sh` which re-uses the bash script from the old test framework
 to spawn a VM in which to run the tests.


alessandrod (approved):
This is amazing!

PR #339, opened 2022-07-04
Improve symbol resolution API

Matthias247:
This change implements additional symbol resolving for stack traces, which
improve the support for userspace stack traces.

A challenge with the stack traces obtained via ebpf is that they just contain
a raw address inside the virtual memory space of the process, which
requires some extra steps to be translated into a function name. This change
provides most of the infrastructure to provide the translation via a `DefaultResolver`
type which can resolve both kernelspace and userspace functions based on
a PID and address. In addition in adds a `SymbolResolver` trait that allows to customize
resolving behavior and extend it further.

I originally looked into just using the [StackTrace::resolve](https://docs.rs/aya/0.11.0/aya/maps/stack_trace/struct.StackTrace.html#method.resolve)
function. However I noticed that the information that is at the moment stored
in the StackTrace struct is not sufficient for resolving, e.g. due to missing the PID
and being unable to account for the virtual -> physical memory address translation.
Therefore this change uses a new resolver infrastructure.

Usage via a suitable EBPF program that sends stack traces and process IDs:

```rust

/// The BPF program populates this Queue with the process ID, kernel-space and user-space
/// stack trace IDs.
/// The former can be obtained via `bpf_get_current_pid_tgid()`, the stack
/// traces via `aya_bpf::maps::StackTrace`.
let mut stacks = Queue::<_, [u64; 3]>::try_from(bpf.map_mut("STACKS")?)?;
let stack_traces = StackTraceMap::try_from(bpf.map_mut("STACK_TRACES")?)?;
let resolver = DefaultResolver::new().unwrap();

loop {
    match stacks.pop(0) {
        Ok([pid_tgid, ktrace_id, utrace_id]) => {
            let tgid = pid_tgid & 0xFFFFFFFF;

            if let Ok(trace) = stack_traces.get(&(ktrace_id as u32), 0) {
                for f in trace.frames() {
                    let mut symbol = SymbolInfo::unresolved_kernel(f.ip);
                    resolver.resolve(&mut symbol);
                    println!("Resolved kernel address: 0x{:x} to {:?}", f.ip, symbol);
                }
            }

            if let Ok(trace) = stack_traces.get(&(utrace_id as u32), 0) {
                for f in trace.frames() {
                    let mut symbol = SymbolInfo::unresolved_user(tgid as _, f.ip);
                    resolver.resolve(&mut symbol);
                    info!("Resolved pid {}, address: 0x{:x} to {:?}", tgid, f.ip, symbol);
                }
            }
        }
    }
}
```

BPF code:
```rust
static STACK_TRACES: StackTrace = StackTrace::with_max_entries(10, 0);

static STACKS: Queue<[u64; 3]> = Queue::with_max_entries(1024, 0);

pub fn testprobe(ctx: ProbeContext) -> u32 {
    unsafe {
        let pid_tgid = bpf_get_current_pid_tgid();
        let ustack = STACK_TRACES.get_stackid(&ctx, BPF_F_USER_STACK as _);
        let kstack = STACK_TRACES.get_stackid(&ctx, 0);

        match (kstack, ustack) {
            (Ok(kstack), Ok(ustack)) => {
                if let Err(e) = STACKS.push(&[pid_tgid, kstack as _, ustack as _], 0) {
                    info!(&ctx, "Error pushing stack: {}", e);
                }
            },
            _ => {}
        }

        0
    }
}
```

Output when fetching stack traces of a kprobe on `sendmmsg` on a test program:
```
Resolved kernel address: 0xffffffff81a62691 to SymbolInfo { virtual_address: 18446744071589734033, object_address: Some(18446744071589734033), process_id: None, function_name: Some("__sys_sendmmsg"), object_path: None }
Resolved kernel address: 0xffffffff81a62870 to SymbolInfo { virtual_address: 18446744071589734512, object_address: Some(18446744071589734512), process_id: None, function_name: Some("__x64_sys_sendmmsg"), object_path: None }
Resolved kernel address: 0xffffffff81da30a3 to SymbolInfo { virtual_address: 18446744071593144483, object_address: Some(18446744071593144483), process_id: None, function_name: Some("do_syscall_64"), object_path: None }
Resolved kernel address: 0xffffffff81e0007c to SymbolInfo { virtual_address: 18446744071593525372, object_address: Some(18446744071593525372), process_id: None, function_name: Some("entry_SYSCALL_64_after_hwframe"), object_path: None }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x7fccf3b2adee to SymbolInfo { virtual_address: 140518238629358, object_address: Some(1195502), process_id: Some(22158), function_name: None, object_path: Some("/usr/lib/x86_64-linux-gnu/libc-2.31.so") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570cd66cc to SymbolInfo { virtual_address: 94856245241548, object_address: Some(947916), process_id: Some(22158), function_name: Some("tokio::io::async_fd::AsyncFdReadyGuard<Inner>::try_io"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570cd43e6 to SymbolInfo { virtual_address: 94856245232614, object_address: Some(938982), process_id: Some(22158), function_name: Some("quinn_udp::imp::UdpSocket::poll_send"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570ccd31f to SymbolInfo { virtual_address: 94856245203743, object_address: Some(910111), process_id: Some(22158), function_name: Some("<quinn::endpoint::EndpointDriver as core::future::future::Future>::poll"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570cc2400 to SymbolInfo { virtual_address: 94856245158912, object_address: Some(865280), process_id: Some(22158), function_name: Some("<core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570cd0e8d to SymbolInfo { virtual_address: 94856245218957, object_address: Some(925325), process_id: Some(22158), function_name: Some("tokio::runtime::task::harness::poll_future"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570cd160a to SymbolInfo { virtual_address: 94856245220874, object_address: Some(927242), process_id: Some(22158), function_name: Some("tokio::runtime::task::harness::Harness<T,S>::poll"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c6ef25 to SymbolInfo { virtual_address: 94856244817701, object_address: Some(524069), process_id: Some(22158), function_name: Some("std::thread::local::LocalKey<T>::with"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c7e505 to SymbolInfo { virtual_address: 94856244880645, object_address: Some(587013), process_id: Some(22158), function_name: Some("tokio::runtime::basic_scheduler::Context::run_task"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c7b2fa to SymbolInfo { virtual_address: 94856244867834, object_address: Some(574202), process_id: Some(22158), function_name: Some("tokio::macros::scoped_tls::ScopedKey<T>::set"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c7de7c to SymbolInfo { virtual_address: 94856244878972, object_address: Some(585340), process_id: Some(22158), function_name: Some("tokio::runtime::basic_scheduler::BasicScheduler::block_on"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c75a3d to SymbolInfo { virtual_address: 94856244845117, object_address: Some(551485), process_id: Some(22158), function_name: Some("tokio::runtime::Runtime::block_on"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c5c597 to SymbolInfo { virtual_address: 94856244741527, object_address: Some(447895), process_id: Some(22158), function_name: Some("std::sys_common::backtrace::__rust_begin_short_backtrace"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570c59d44 to SymbolInfo { virtual_address: 94856244731204, object_address: Some(437572), process_id: Some(22158), function_name: Some("core::ops::function::FnOnce::call_once{{vtable.shim}}"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
02:46:16 [INFO] ayatest: [ayatest/src/main.rs:98] Resolved pid 22158, address: 0x564570ecc763 to SymbolInfo { virtual_address: 94856247297891, object_address: Some(3004259), process_id: Some(22158), function_name: Some("std::sys::unix::thread::Thread::new::thread_start"), object_path: Some("/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk") }
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/339)

### DIFF
--- a/aya/src/symbols.rs
+++ b/aya/src/symbols.rs
@@ -1,0 +1,688 @@
+//! Utilities for resolving symbols from addresses obtained via stack traces
+use std::{
+    cell::RefCell,
+    collections::BTreeMap,
+    fs::File,
+    io::{self, BufRead, BufReader},
+    path::{Path, PathBuf},
+};
+
+use addr2line::{
+    gimli::{EndianRcSlice, RunTimeEndian},
+    Context,
+};
+use lru_cache::LruCache;
+use thiserror::Error;
+
+use crate::util::kernel_symbols;
+
+/// The error typey returned when creating various [`SymbolResolver`]s.
+#[derive(Debug, Error)]
+pub enum CreateResolverError {
+    /// Failed to create a kernel symbol resolver
+    #[error("Error creating KernelSymbolResolver: {error}")]
+    KernelSymbolResolverBuildError {
+        #[source]
+        /// The original io::Error
+        error: io::Error,
+    },
+}
+
+/// Stores all information that was resolved for a specific symbol
+#[derive(Debug, Clone)]
+pub struct SymbolInfo {
+    /// Virtual memory address of a certain function
+    virtual_address: u64,
+    /// Address of a function inside an object. `None` if it was not resolved
+    object_address: Option<u64>,
+    /// The PID of the process in user space
+    /// `None` is the symbol if the address is a kernel function
+    process_id: Option<u32>,
+    /// The resolved function name
+    function_name: Option<String>,
+    /// Path to the object which defines the symbol
+    object_path: Option<PathBuf>,
+}
+
+impl SymbolInfo {
+    /// Creates a `SymbolInfo` instance for a kernel-space function which hasn't been resolved yet
+    pub fn unresolved_kernel(address: u64) -> Self {
+        Self {
+            virtual_address: address,
+            object_address: Some(address),
+            function_name: None,
+            object_path: None,
+            process_id: None,
+        }
+    }
+
+    /// Creates a `SymbolInfo` instance for a user-space function which hasn't been resolved yet
+    pub fn unresolved_user(process_id: u32, address: u64) -> Self {
+        Self {
+            virtual_address: address,
+            object_address: None,
+            function_name: None,
+            object_path: None,
+            process_id: Some(process_id),
+        }
+    }
+
+    /// Whether the symbol refers to a kernel-space function
+    pub fn is_kernel(&self) -> bool {
+        self.process_id.is_none()
+    }
+
+    /// Whether the symbol refers to a user-space function
+    pub fn is_user(&self) -> bool {
+        self.process_id.is_some()
+    }
+
+    /// Returns the ID of the process for which the symbol was recorded
+    pub fn process_id(&self) -> Option<u32> {
+        self.process_id
+    }
+
+    /// Virtual memory address of a certain function
+    pub fn virtual_address(&self) -> u64 {
+        self.virtual_address
+    }
+
+    /// Address of a function inside an object. `None` if it was not resolved
+    pub fn object_address(&self) -> Option<u64> {
+        self.object_address
+    }
+
+    /// Path to the object which defines the symbol
+    pub fn object_path(&self) -> Option<&Path> {
+        self.object_path.as_deref()
+    }
+
+    /// The resolved function name
+    pub fn function_name(&self) -> Option<&str> {
+        self.function_name.as_deref()
+    }
+}
+
+/// Resolves a symbol based on it's address.
+///
+/// See [`DefaultResolver`] for exemplaric usage.
+pub trait SymbolResolver {
+    /// Resolves a symbol based on it's address
+    fn resolve(&self, symbol: &mut SymbolInfo);
+}
+
+/// A resolver for kernel symbols
+pub struct KernelSymbolResolver {
+    symbols: BTreeMap<u64, String>,
+}
+
+impl KernelSymbolResolver {
+    /// Creates a new `KernelSymbolResolver`
+    ///
+    /// This will load all kernel symbols from `/proc/kallsyms`.
+    pub fn new() -> Result<Self, CreateResolverError> {
+        let symbols = kernel_symbols()
+            .map_err(|e| CreateResolverError::KernelSymbolResolverBuildError { error: e })?;
+        Ok(Self { symbols })
+    }
+}
+
+impl SymbolResolver for KernelSymbolResolver {
+    fn resolve(&self, symbol: &mut SymbolInfo) {
+        if symbol.process_id().is_some() || symbol.function_name.is_some() {
+            return;
+        }
+
+        symbol.function_name = self
+            .symbols
+            .range(..=symbol.virtual_address())
+            .next_back()
+            .map(|(_, s)| s.clone());
+    }
+}
+
+/// A SymbolResolver which uses the `addr2line` Rust library
+pub struct Addr2LineResolver {
+    state: RefCell<Addr2LineResolverState>,
+}
+
+/// Builder for a `Addr2LineResolver`
+impl Addr2LineResolver {
+    /// Returns `Addr2lineResolverBuilder` which allows to configure the `Addr2Resolver`
+    pub fn builder() -> Addr2lineResolverBuilder {
+        Addr2lineResolverBuilder {
+            proc_map_lru_capacity: 512,
+            object_resolver_capacity: 512,
+        }
+    }
+}
+
+/// A builder for `Addr2LineResolver` instances
+pub struct Addr2lineResolverBuilder {
+    proc_map_lru_capacity: usize,
+    object_resolver_capacity: usize,
+}
+
+impl Addr2lineResolverBuilder {
+    /// Configures the capacity of the LRU cache that is used to look up process memory maps
+    pub fn proc_map_lru_capacity(&mut self, capacity: usize) -> &mut Self {
+        self.proc_map_lru_capacity = capacity;
+        self
+    }
+
+    /// Configures the capacity of the LRU cache that is used to hold information
+    /// about libraries and executables in order to resolve function names.
+    pub fn object_resolver_capacity(&mut self, capacity: usize) -> &mut Self {
+        self.object_resolver_capacity = capacity;
+        self
+    }
+
+    /// Builds an `Addr2LineResolver` with the provided configuration
+    pub fn build(&self) -> Result<Addr2LineResolver, CreateResolverError> {
+        Ok(Addr2LineResolver {
+            state: RefCell::new(Addr2LineResolverState {
+                proc_map: LruCache::new(self.proc_map_lru_capacity),
+                object_resolvers: LruCache::new(self.object_resolver_capacity),
+            }),
+        })
+    }
+}
+
+struct Addr2LineResolverState {
+    proc_map: LruCache<u32, Option<ProcMemMap>>,
+    object_resolvers: LruCache<PathBuf, Option<Addr2LineObjectContext>>,
+}
+
+impl SymbolResolver for Addr2LineResolver {
+    fn resolve(&self, symbol: &mut SymbolInfo) {
+        // Resolving a userspace symbol from a virtual memory address is a 2 step process
+        // - First, we translate the address from a virtual memory address to the
+        //   address that is actually used in the object file. If the information is
+        //   already provided, we can skip this step.
+        // - Next, we resolve the actual symbol inside the object.
+
+        if symbol.object_path().is_none() || symbol.object_address().is_none() {
+            let pid = match symbol.process_id() {
+                Some(pid) => pid,
+                None => return,
+            };
+
+            let mut guard = self.state.borrow_mut();
+            let proc_map = match guard.proc_map.get_mut(&pid) {
+                Some(Some(proc_map)) => proc_map,
+                Some(None) => return,
+                None => {
+                    let proc_map = ProcMemMap::from_process_id(pid).ok();
+                    guard.proc_map.insert(pid, proc_map);
+
+                    match guard
+                        .proc_map
+                        .get_mut(&pid)
+                        .expect("Entry was just inserted")
+                        .as_ref()
+                    {
+                        Some(proc_map) => proc_map,
+                        None => return,
+                    }
+                }
+            };
+
+            match proc_map.lookup(symbol.virtual_address()) {
+                Some(lookup_result) => {
+                    symbol.object_address = Some(lookup_result.address());
+                    symbol.object_path = lookup_result.object_path().map(|path| path.to_path_buf());
+                }
+                None => return,
+            };
+        };
+
+        let (object_path, object_address) = match (symbol.object_path(), symbol.object_address()) {
+            (Some(object_path), Some(object_address)) => (object_path, object_address),
+            _ => return,
+        };
+
+        let mut guard = self.state.borrow_mut();
+        symbol.function_name = match guard.object_resolvers.get_mut(object_path) {
+            Some(Some(p)) => p.resolve(object_address),
+            Some(None) => return,
+            None => match Addr2LineObjectContext::from_object_path(object_path) {
+                Err(_) => {
+                    // Cache the error, in order to avoid retrying loading the file
+                    guard.object_resolvers.insert(object_path.to_owned(), None);
+                    return;
+                }
+                Ok(addr2line) => {
+                    let function_name = addr2line.resolve(object_address);
+                    guard
+                        .object_resolvers
+                        .insert(object_path.to_owned(), Some(addr2line));
+                    function_name
+                }
+            },
+        };
+    }
+}
+
+/// A [`SymbolResolver`] which uses the [`Addr2LineResolver`] for resolving
+/// userspace functions, and the [`KernelSymbolResolver`] for resolving kernel functions.
+struct CombinedResolver {
+    kernel: KernelSymbolResolver,
+    user: Addr2LineResolver,
+}
+
+impl CombinedResolver {
+    pub fn new() -> Result<Self, CreateResolverError> {
+        let kernel = KernelSymbolResolver::new()?;
+        let user = Addr2LineResolver::builder().build()?;
+        Ok(Self { kernel, user })
+    }
+}
+
+impl SymbolResolver for CombinedResolver {
+    fn resolve(&self, symbol: &mut SymbolInfo) {
+        match symbol.is_user() {
+            true => self.user.resolve(symbol),
+            false => self.kernel.resolve(symbol),
+        }
+    }
+}
+
+/// A default resolver which can resolve kernel and userspace symbols,
+/// and caches last observed symbol information in memory.
+///
+/// # Examples
+///
+/// ```no_run
+/// # use aya::symbols::{DefaultResolver, SymbolResolver, SymbolInfo};
+/// let resolver = DefaultResolver::new().unwrap();
+///
+/// // Resolve a kernel-space symbol based on an address obtained from a stack trace
+/// let mut kernel_symbol = SymbolInfo::unresolved_kernel(0x1234_5678);
+/// resolver.resolve(&mut kernel_symbol);
+/// println!("{:?}", kernel_symbol.function_name());
+///
+/// // Resolve a user-space symbol based on an a process-id and address obtained from a stack trace
+/// let mut user_symbol = SymbolInfo::unresolved_user(7654, 0x1234_5678);
+/// resolver.resolve(&mut user_symbol);
+/// println!("{:?}", user_symbol.function_name());
+/// ```
+pub struct DefaultResolver {
+    inner: CachingResolver<CombinedResolver>,
+}
+
+impl DefaultResolver {
+    /// Create a new DefaultResolver
+    pub fn new() -> Result<Self, CreateResolverError> {
+        let caching_resolver = CachingResolver::with_capacity(8192, CombinedResolver::new()?);
+        Ok(Self {
+            inner: caching_resolver,
+        })
+    }
+}
+
+impl SymbolResolver for DefaultResolver {
+    fn resolve(&self, symbol: &mut SymbolInfo) {
+        self.inner.resolve(symbol)
+    }
+}
+
+/// Hash key for the `CachingResolver`
+#[derive(Hash, PartialEq, Eq, Copy, Clone)]
+struct CachingResolverHashKey {
+    virtual_address: u64,
+    process_id: Option<u32>,
+}
+
+/// A resolver which caches observed symbols
+pub struct CachingResolver<T> {
+    cache: RefCell<LruCache<CachingResolverHashKey, SymbolInfo>>,
+    inner: T,
+}
+
+impl<T: SymbolResolver> CachingResolver<T> {
+    /// Creates a new CachingResolve using a cache that can hold up to `capacity` results
+    pub fn with_capacity(capacity: usize, inner: T) -> Self {
+        Self {
+            cache: RefCell::new(LruCache::new(capacity)),
+            inner,
+        }
+    }
+}
+
+impl<T: SymbolResolver> SymbolResolver for CachingResolver<T> {
+    fn resolve(&self, symbol: &mut SymbolInfo) {
+        let key = CachingResolverHashKey {
+            virtual_address: symbol.virtual_address(),
+            process_id: symbol.process_id(),
+        };
+
+        let mut guard = self.cache.borrow_mut();
+        match guard.get_mut(&key) {
+            Some(result) => {
+                *symbol = result.clone();
+            }
+            None => {
+                self.inner.resolve(symbol);
+                // TODO: Since we store potentially incomplete SymbolInfo here,
+                // we might miss out the chance to get later on additional information
+                guard.insert(key, symbol.clone());
+            }
+        }
+    }
+}
+
+/// The error type that is used when resolving symbols using the addr2line library
+#[derive(Debug, Error)]
+pub enum Addr2LineError {
+    /// Failed to read the context of the executable or library
+    #[error("Failed to read object data at path {path}")]
+    InvalidObjectPath {
+        /// The object path
+        path: PathBuf,
+        /// The original io::Error
+        source: io::Error,
+    },
+    /// Failed to parse data for an executable or library
+    #[error("Failed to read object data at path {path}")]
+    ParseError {
+        /// The object path
+        path: PathBuf,
+        /// The original error
+        source: addr2line::object::Error,
+    },
+    /// Failed to create the gimli resolver context
+    #[error("Failed to read object data at path {path}")]
+    CreateContextError {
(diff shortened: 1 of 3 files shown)


Matthias247:
@dave-tucker I noticed this has some overlap with #329 - at least both changes implement parsing of `/proc/[pid]/maps` - which can probably be in a single place. Maybe also with symbol resolving, since I think a part of that is also implemented in #329, but in a more low level fashion than here (this PR just asks `addr2line` instead of going more low level). I really just wanted to get some basic stack traces working, and used the tools that seemed to require the least effort for this. I don't have a ton of experience with ELF files and no strong opinion on how it should look like :-)

Matthias247:
Ping. I just rebased this on top of `main`. I will also add a few comments on my thoughts

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+        self.function_name.as_deref()
+    }
+}
+
+/// Resolves a symbol based on it's address.
+///
+/// See [`DefaultResolver`] for exemplaric usage.
+pub trait SymbolResolver {
+    /// Resolves a symbol based on it's address
+    fn resolve(&self, symbol: &mut SymbolInfo);
```
Matthias247, line 111: I chose this API with the intention of "you provide all the info you have, and the resolver fetches the rest". Alternatives could be:

```
fn resolve(&self, symbol: &SymbolInfo) -> Option<Symbolnfo>;
```
or
```
fn resolve(&self, symbol: &SymbolInfo) -> Vec<Symbolnfo>;
```
which don't mutate the initial info. But at the the former is also achievable with the current API, and cloning the `sybol` before.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+    /// The resolved function name
+    pub fn function_name(&self) -> Option<&str> {
+        self.function_name.as_deref()
+    }
+}
+
+/// Resolves a symbol based on it's address.
+///
+/// See [`DefaultResolver`] for exemplaric usage.
+pub trait SymbolResolver {
```
Matthias247, line 109: This should probably replace `StackTrace::resolve` in a backward incompatible API change. It's much more powerful.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+    /// Resolves a symbol based on it's address
+    fn resolve(&self, symbol: &mut SymbolInfo);
+}
+
+/// A resolver for kernel symbols
+pub struct KernelSymbolResolver {
```
Matthias247, line 115: Once `StackTrace::resolve` is removed, this resolver can also avoid `kernel_symbols()`  being a public function, since the resolver should be able to fulfill the whole job.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+/// println!("{:?}", user_symbol.function_name());
+/// ```
+pub struct DefaultResolver {
+    inner: CachingResolver<CombinedResolver>,
+}
+
+impl DefaultResolver {
+    /// Create a new DefaultResolver
+    pub fn new() -> Result<Self, CreateResolverError> {
+        let caching_resolver = CachingResolver::with_capacity(8192, CombinedResolver::new()?);
```
Matthias247, line 316: I have no strong opinion about any of the LRU capacities in this change. I just thought it would be a good idea to limit the amount of metadata that the process loads - if this is used for long-running profiling.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+
+                while let Ok(Some(frame)) = frames.next() {
+                    let frame = frame.function.and_then(|function_name| {
+                        function_name
+                            .demangle()
+                            .map(|demangled_name| demangled_name.to_string())
+                            .ok()
+                    });
+
+                    // Return the last frame in the stack of frames, since the inlined functions
```
Matthias247, line 441: Now this part is an interesting design choice: Should the aya resolver return just the outermost frame? Or potentially all the frames? The latter might provide a bit more data for profilers that run on top of that, or for debugging purposes. But there's also more data produced - and the `SymbolResolver` API would need to be changed to be able to return a stack of symbols. Not exactly sure how that API should look like, since just returning `Vec<SymbolInfo>` might then have wrongly populated virtual  addresses if we just copy&paste those. Maybe the API would need to be closer to `resolve(&SymbolInfo) -> Vec<ResolvedSymbol>` where the latter doesn't care about PIDs and virtual addresses anymore?

Besides that producing more data obviously also has a cost.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+    virtual_address: u64,
+    /// Address of a function inside an object. `None` if it was not resolved
+    object_address: Option<u64>,
+    /// The PID of the process in user space
+    /// `None` is the symbol if the address is a kernel function
+    process_id: Option<u32>,
+    /// The resolved function name
+    function_name: Option<String>,
+    /// Path to the object which defines the symbol
+    object_path: Option<PathBuf>,
```
Matthias247, line 44: those and potentially even the other fields will be shared a lot between symbols. Could be using `Arc`s? But then OTOH all those strings should be relatively short and there might be more atomic refcount operations, which probably negates the benefit a bit.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
+//! Utilities for resolving symbols from addresses obtained via stack traces
+use std::{
+    cell::RefCell,
+    collections::BTreeMap,
+    fs::File,
+    io::{self, BufRead, BufReader},
+    path::{Path, PathBuf},
+};
+
+use addr2line::{
```
alessandrod, line 10: As you've noticed in the USDT PR, we use the `object` crate to parse ELF and I think we should use it here too

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+    KernelSymbolResolverBuildError {
+        #[source]
+        /// The original io::Error
+        error: io::Error,
+    },
+}
+
+/// Stores all information that was resolved for a specific symbol
+#[derive(Debug, Clone)]
+pub struct SymbolInfo {
```
alessandrod, line 33: I'm thinking, shouldn't we have different types for kernel and user symbols? Eg `process_id`, `object_address` and `object_path` don't really apply to kernel symbols. Is there an advantage to having the same type?

aya/src/symbols.rs
alessandrod: I think it's probably too soon to think about this - ideally we would borrow from the underlying ELF parser.

aya/src/symbols.rs
alessandrod: We agree we need a more powerful API

aya/src/symbols.rs
alessandrod: I find this a bit confusing, wouldn't it be clearer if we had something like:

```

let user_resolver = FileResolver::new("/some/path")?;
let sym = user_resolver.resolve(123);

let user_resolver = PidResolver::new(42)?;
let sym = user_resolver.resolve(123);

let kernel_resolver = KernelResolver()?;
let sym = kernel_resolver.resolve(123)
```

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+        symbol.function_name = self
+            .symbols
+            .range(..=symbol.virtual_address())
+            .next_back()
+            .map(|(_, s)| s.clone());
+    }
+}
+
+/// A SymbolResolver which uses the `addr2line` Rust library
+pub struct Addr2LineResolver {
```
alessandrod, line 145: I think this is probably not the right level of abstraction. I'm not sure we should provide an API that can resolve from multiple pids and multiple object files. I think we should provide primitives and then let people could compose them as they wish. Eg, we should definitely provide the primitives to resolve from kernel, object files and pids - but should we provide a full blown API to implement a system wide profiler that can profile kernel and separate processes? I'm inclined to say no.

Generally speaking, I see aya as a high level rust wrapper for the primitives offered by the ebpf platform. Higher level abstractions that can be built using core primitives should ideally go in side crates. I do think that it would be really cool to provide a complete profiler API, but it should probably go in a `aya-profiler` crate or something.

aya/src/symbols.rs
```diff
@@ -0,0 +1,688 @@
...
+                }
+
+                result
+            }
+            Err(_) => None,
+        }
+    }
+}
+
+/// Parsed line for /proc/[pid]/maps
```
alessandrod, line 458: As you've noticed in the other PR we already have proc maps parsing code we use to attach uprobes, and dave just moved it so it can be used to implement USDT probes too.

alessandrod (changes requested):
Thanks for doing this and apologies again for the delay in reviewing it! 

We definitely need a better API to resolve symbols and once we settle on one we'll deprecate `StackTrace::resolve`. 

See my comments inline.

alessandrod:
@mstange @kakkoyun you've built profilers with aya, tagging you in case you have any inputs

aya/src/symbols.rs
Matthias247: Given `addr2line` is a tiny wrapper around `object`, what would be the win? It seems like it would just duplicate the functionality. Plus at least I don't have the domain knowledge about ELF files at the moment and time to dig into it, so if that's preferable someone else would need to look into it.

aya/src/symbols.rs
Matthias247: It's mostly there to have a common resolver interface. Also if someone wants to represent a full call stack covering kernel and application side, they could use `Vec<SymbolInfo>` or `Stack<SymbolInfo>` instead of having a hetereogenous list. 

But sure, seperate types would also work.

aya/src/symbols.rs
Matthias247: Sure. But that PR isn't merged yet - and it didn't even exist when I started writing this. My understanding is that this can be unified after whatever PR is merged first. Or even after both are merged in the form of a cleanup PR. I can remove the `pub` annotations to avoid this being a breaking API change.

aya/src/symbols.rs
alessandrod: Oh, I did not realize that :) It seems that it can do demangling too, and that's behind cargo features. I think we should default to having those features off so that we don't bring in more dependencies for people that are not going to implement profilers.

In any case we shouldn't re-export any symbols from `addr2line` in the same way that we don't re-export anything from any dependencies.

aya/src/symbols.rs
Matthias247: > . Eg, we should definitely provide the primitives to resolve from kernel, object files and pids

I agree - that's why all basic resolvers are also individually available

> but should we provide a full blown API to implement a system wide profiler that can profile kernel and separate processes? I'm inclined to say no.

I think yes - just because it requires a significant amount of domain knowledge to get a full stack trace for whatever values the BPF code returns otherwise. I certainly spent multiple hours on topics that are 100% unrelated to both eBPF and the application I wanted to look into  before I got any useful results. I think eBPF application developers will often want to get a readable stack trace, but only the minority might be interested in how the innards of that work.

Also consider that those functions are not only interesting for profiler authors that might have a bit of a better background around call stacks - but also for all others. E.g. I might want to write a program for debugging purposes which logs the time spent on particular syscalls - and we then want to figure out which codepath actually triggered high latency. A simple resolver with a default constructor that allows to act on the address/pid values that the BPF code submits can do that job - and exactly that is available via `DefaultResolver`. 

I assume other tools like `bpftrace` have the necessary code included for  a similar reason.

It's however a fair question to ask which "intermediate" layer tools should be provided, and e.g. whether the `struct Addr2LineResolver` with 2 layers of caching here really should be public - instead of being purely an implementation detail of  `DefaultResolver` (or DefaultUserspaceResolver` if you want to separate those).

If only building blocks and the high level resolver would be public, it would be:
- `KernelResolver`
- `ProcMemMap`
- `Addr2LineObjectContext`  (potentially to be renamed)
- `DefaultResolver` (potentially to be renamed) 

And since most of those don't have common input/result types anymore, it might not be worthwhile to have the `Resolver` trait anymore.

aya/src/symbols.rs
alessandrod: I don't think that the ebpf API can return complete user + stack traces. One would have to do some work to reassemble them, and at that point they could also convert to a unified interface.

aya/src/symbols.rs
alessandrod: We already have some proc maps parsing code here https://github.com/aya-rs/aya/blob/d968094b662be3449624420b76ea2dd239ef657b/aya/src/programs/uprobe.rs#L192. Whatever PR is merged first (I haven't reviewed the USDT PR yet), the code should be shared.

I haven't commented on the details of the API yet, but yes we would definitely not make the proc maps code public, nor the addr2line re-exports or addr2line specific errors. The API should be small and self contained.

aya/src/symbols.rs
Matthias247: See the comment at https://github.com/aya-rs/aya/pull/339/files#r933884841 on what could be public. That could certainly also be done, but it means there's no common trait (and thereby also builtin mechnaism for composition/caching/etc) anymore. Also the question what `sym` contains in those results. Is it always the same, or an individual result type per resolver?

In your example, I guess `PidResolver` will not resolve a symbol but just translate the address? It is `ProcMemMap`?

mstange:
Thanks for tagging me, I have a lot of opinions on this topic! :)

Symbol resolution can be quite a big topic, depending on how deep you want to go. I'm not sure if aya is the right place to offer a full symbol resolution API and choose which trade-offs to make; I'm thinking it might be better to offer just enough information so that absolute addresses can be mapped to relative address (along with the file path of the library), so that the user can call into a different crate to do the actual symbol resolution. I was already pondering whether I should make such a crate; I think this PR has convinced me that it would be worth doing.

The following design questions come to mind:

- When do you read proc maps? Whenever symbols are requested, or do you maintain a cached representation? Do you handle libraries being loaded and unloaded during program observation? Perf does the following: It reads proc maps once at the start of recording, and then sets `perf_event_attr.mmap2 = true` so that it receives `PERF_RECORD_MMAP2` records for any newly loaded libraries. Then it can map absolute addresses to the right library at any point during recording.
- Which type of "relative address" do you store? When you map an absolute address to a relative address, there are two types of relative addresses you might choose: file offsets or "stated virtual memory addresses" (SVMAs). The former is easy to compute from the proc maps info, but the latter is what addr2line wants. This is a source of bugs because, for many libraries, the two end up being the same, so you don't notice the discrepancy until you run into a library in which the two are different. For example, libxul.so (from Firefox) is one where the two are different; libxul.so contains an ELF LOAD command whose file offset and the virtual memory address are different. Perf stores relative addresses as file offsets (it calls them "rip", i.e. "relative instruction pointer") and then resolves those to SVMAs (which it calls "objdump addresses") only whenever it calls out to addr2line / objdump. [Samply](https://github.com/mstange/samply) stores SVMAs; it has to [look at the library's section information](https://github.com/mstange/samply/blob/da71321c597d4499730976923a7d0695147030b7/samply/src/linux_shared/mod.rs#L982-L990) before it can compute any relative addresses.
- Where do you look for symbol information? Only in the actual loaded binary, or do you search [other](https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html) [places](https://github.com/rust-lang/backtrace-rs/blob/b02ed04a7e915659eea6fb1607df469b84a30638/src/symbolize/gimli/elf.rs#L303-L423) as well? Do you support [split](https://gcc.gnu.org/wiki/DebugFission) [dwarf](https://github.com/rust-lang/backtrace-rs/blob/b02ed04a7e915659eea6fb1607df469b84a30638/src/symbolize/gimli/elf.rs#L53-L63)? Do you use [debuginfod](https://sourceware.org/elfutils/Debuginfod.html) to pull additional debug files from the web?
- How do you handle libraries which have incomplete symbol information? You may want to generate placeholder symbols based on function start addresses from unwind information. If you don't do this, you may end up with many misleading function names for unrelated addresses.
- Do you [handle proc mountinfo](https://github.com/koute/not-perf/blob/18bd8d3788e678becf09cfe585cb886587faa70b/src/mount_info.rs#L189-L207) information? I think this is needed when observing ubuntu snap builds such as the snap Firefox build. Those load many libraries from the snap file system, so you can't just trust the paths in the proc maps.

---

Personally, I think I'd like it aya to maintain a process mapping over time (one for each process), and have it tell me when libraries are loaded and unloaded, and when processes appear and disappear.

I'll close with one piece of advice: Watch out when resolving symbols for return addresses, i.e. any non-topmost address in a call stack. If you get file+line information from addr2line, or if you get inlines, you need to make sure that the "lookup address" is shifted into the "call" instruction. By default, addresses in stacks are usually "return addresses", i.e. the address of the instruction which will be executed *after* the call returns. The common remedy is to just subtract one byte from all return addresses, i.e. ask addr2line to resolve `return_address - 1`. If you don't do this, you can end up with wrong line numbers and wrong inline stacks.
Oh, also, please don't discard inline stack data. Yes, the callee-most function of an inline stack on its own is usually not very useful, but the outer function on its own is often not very useful either. For best results you really want to keep the entire inline stack for each address.

kakkoyun:
Thanks for tagging me as well. I'm still catching up with my review requests. I'll have an in-depth look at this.

dave-tucker:
As far as the #329 is concerned, the only similarity is that it requires more fields that are currently exposed from `/proc/$pid/maps` - we need the VMAs for `find_by_offset`, which yields the proc map entry for a given relative address, so you can compute an absolute address. All ELF parsing is related to the special USDT notes format, which contains the instruction pointer(s) used to attach to. No symbol resolution is required.

TL;DR: There should be no conflicts between this PR, and my one... and even if there were they should be trivial to resolve.

> Symbol resolution can be quite a big topic, depending on how deep you want to go. I'm not sure if aya is the right place to offer a full symbol resolution API and choose which trade-offs to make; I'm thinking it might be better to offer just enough information so that absolute addresses can be mapped to relative address (along with the file path of the library), so that the user can call into a different crate to do the actual symbol resolution. I was already pondering whether I should make such a crate; I think this PR has convinced me that it would be worth doing.

@mstange this :point_up: feels like the right approach to me.
I know very little about stack traces and symbol resolution, but given my recent experience with `/proc/$pid/maps` I'll weigh in on a few points below...

> Personally, I think I'd like aya to maintain each process's mappings over time, and have it tell me when libraries are loaded and unloaded, and when processes appear and disappear..

I certainly feel like the first item would be ~easy to do. I'd already considered a lazy-init cache for proc/maps, and using something (inotify?) to keep it updated until the proc is closed. Adding a public API to register a callback when the cache is updated for a given PID would hopefully fulfil the requirement.

I'm not sure how we'd handle "when process appear or disappear" without watching procfs for new pids... and how Aya would know which pids are interesting... :thinking: It might be easier to defer that to the user, and to let them register interesting PIDs with Aya via the same API as described above?

mstange:
When you call `perf_event_open`, you can request event records for new processes, i.e. "task" (fork) records. And you can also request "comm" records for process renames / execs. These records are written into the perf event's mmap buffer, but I'm not really sure how this buffer interacts with the buffer that has the PerfEventArray output - or if that's even the exact same buffer.
Here's some code which requests these records: https://github.com/mstange/samply/blob/da71321c597d4499730976923a7d0695147030b7/samply/src/linux/perf_event.rs#L313-L315
You can parse these records with https://docs.rs/linux-perf-event-reader/.

zz85:
I tried taking a stab at resolving symbols here https://github.com/zz85/profile-bee/blob/main/profile-bee/src/symbols.rs

The approach is similar to this PR but falls back to loading symbols from the ELF format. There's definitely still some rough edges (don't think it's handling relocations/offsets correctly) but putting it out here in case it's useful to anyone looking into symbol lookups

addisoncrump:
Bringing this back up with #705: what's the preferred implementation here? We all agree on the need for a `SymbolResolver` trait, but no consensus. Is the memory mapping monitor popular enough to justify a sample implementation?

dave-tucker:
Closing this since the consensus arrived at (see https://github.com/aya-rs/aya/pull/740) is that symbol resolution for stack traces should exist outside of aya.

PR #343, opened 2022-07-07
Small Pinning API improvements

dave-tucker:
1, Allow pinning from the program directly (e.g `TracePoint::pin("/path")`
2. Add `FdLink::pin("/path")`, which requires that you've called `take_link()` first, which actually isn't a bad thing at all :wink:

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -51,6 +51,7 @@ use thiserror::Error;
 use crate::{
     generated::bpf_map_type,
     obj,
+    pin::PinError,
     sys::{bpf_create_map, bpf_get_object, bpf_map_get_next_key, bpf_pin_object, kernel_version},
     util::nr_cpus,
     Pod,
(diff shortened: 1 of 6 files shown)


aya/src/programs/links.rs
```diff
@@ -111,6 +137,24 @@ impl FdLink {
...
         FdLink { fd }
     }
+
+    /// Pins the FdLink to a BPF filesystem.
+    ///
+    /// When a BPF object is pinned to a BPF filesystem it will remain attached after
+    /// Aya has detached the link.
+    /// To remove the attachment, the file on the BPF filesystem must be remove.
+    /// Any directories in the the path provided should have been created by the caller.
+    pub fn pin<P: AsRef<Path>>(&self, path: P) -> Result<(), LinkError> {
```
alessandrod, line 147: Why is this implemented on `FdLink`? Can other links not be pinned?

dave-tucker:
Comments addressed. All pin operations across Map, Program and Links now use `PinError`.
Link pinning is restricted to FdLink because, as far as I know, only BPF FDs can be pinned, which is not what is returned by other link types.

See: #329 which uses `bpf_link_create` for programs that used PerfLink, which would now allow for tracepoints, kprobes etc.. to be pinned.

aya/src/maps/mod.rs
```diff
@@ -315,10 +301,12 @@ impl Map {
                 })
             }
         };
```
alessandrod, line 303: MapError::InvalidPinPath should go

alessandrod (commented):
I think `MapError::PinError` and `ProgramError::PinError` shouldn't exist, and we should instead return `PinError` directly from the pin methods.

In general we need to stop adding variants to existing errors, we should be removing them instead. Imagine you have:

```
match bpf.map_something_that_doesnt_do_any_pinning() {
   MapError::SomethingINeedToHandle => ...,
   MapError::SomethingElseINeedToHandle => ...,
   MapError::PinError(_) => I shouldn't have to handle this i'm not pinning,
   ...
}
```

PR #348, opened 2022-07-24
bpf: add `bpf_printk!` helper

athre0z:
The `bpf_printk!` macro is a helper providing a convenient way to invoke the `bpf_trace_printk` and `bpf_trace_vprintk` BPF helpers. It is implemented as a macro because it requires variadic arguments.

### DIFF
--- a/bpf/aya-bpf/src/helpers.rs
+++ b/bpf/aya-bpf/src/helpers.rs
@@ -13,7 +13,7 @@ pub use aya_bpf_bindings::helpers as gen;
 #[doc(hidden)]
 pub use gen::*;
 
-use crate::cty::{c_long, c_void, c_char};
+use crate::cty::{c_char, c_long, c_void};
 
 /// Read bytes stored at `src` and store them as a `T`.
 ///
@@ -735,7 +735,9 @@ pub fn bpf_get_current_uid_gid() -> u64 {
 ///
 /// ```no_run
 /// # use aya_bpf::helpers::bpf_printk;
-/// bpf_printk!(b"hi there! dec: %d, hex: 0x%08X", 42, 0x1234);
+/// unsafe {
+///   bpf_printk!(b"hi there! dec: %d, hex: 0x%08X", 42, 0x1234);
+/// }
 /// ```
 ///
 /// [fmt]: https://www.kernel.org/doc/html/latest/core-api/printk-formats.html#printk-specifiers
@@ -743,22 +745,16 @@ pub fn bpf_get_current_uid_gid() -> u64 {
 macro_rules! bpf_printk {
     ($fmt:literal $(,)? $($arg:expr),* $(,)?) => {{
         use $crate::helpers::PrintkArg;
-        // Arguments must be placed on the stack to pass the verifier.
-        let fmt = unsafe {
-            // `concat_bytes!` is still unstable, so we have to emulate it manually.
-            let mut fmt = ::core::mem::MaybeUninit::<[u8; $fmt.len() + 1]>::uninit();
-            ::core::ptr::copy_nonoverlapping($fmt.as_ptr(), fmt.as_mut_ptr() as _, 1);
-            (*fmt.as_mut_ptr()).as_mut_ptr().add($fmt.len()).write(0);
-            fmt.assume_init()
-        };
+        const FMT: [u8; { $fmt.len() + 1 }] = $crate::helpers::zero_pad_array::<
+            { $fmt.len() }, { $fmt.len() + 1 }>(*$fmt);
         let data = [$(PrintkArg::from($arg)),*];
-        $crate::helpers::bpf_printk_impl(&fmt, &data)
+        $crate::helpers::bpf_printk_impl(&FMT, &data)
     }};
 }
 
 // Macros are always exported from the crate root. Also export it from `helpers`.
 #[doc(inline)]
-pub use bpf_printk as bpf_printk;
+pub use bpf_printk;
 
 /// Argument ready to be passed to `printk` BPF helper.
 #[repr(transparent)]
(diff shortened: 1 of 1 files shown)


bpf/aya-bpf/src/helpers.rs
```diff
@@ -701,3 +701,128 @@ pub fn bpf_get_current_pid_tgid() -> u64 {
...
+
+    match NUM_ARGS {
+        0 => printk(fmt_ptr, fmt_size),
+        1 => printk(fmt_ptr, fmt_size, args[0]),
+        2 => printk(fmt_ptr, fmt_size, args[0], args[1]),
+        3 => printk(fmt_ptr, fmt_size, args[0], args[1], args[2]),
+        4..=12 => {
+            gen::bpf_trace_vprintk(fmt_ptr, fmt_size, args.as_ptr() as _, (NUM_ARGS * 8) as _)
+        }
+        _ => panic!("the kernel limits printk to a maximum of 12 arguments"),
```
athre0z, line 826: Note that this case will actually result in the compiler inlining the panic handler here if `NUM_ARGS > 12`. If the suggested panic handler from the cargo-generate template is used (the one calling `core::hint::unreachable_unchecked), the compiler will optimize away all code paths [dominated by](https://en.wikipedia.org/wiki/Dominator_(graph_theory)) this macro invocation. In my test run this resulted in a funny situation where my whole tracepoint was simply omitted from the binary. Not sure how to best error out here: I wish Rust had proper static asserts.

athre0z: Switched to the "let the kernel deal with it" approach in [379c27c](https://github.com/aya-rs/aya/pull/348/commits/379c27c557bd4828484b76b69c19fae56845e2e1) -- I think it's the best for now. This way we also don't need to hard-code the limit here, in case the kernel folks ever change it.

alessandrod: Maybe we can use https://doc.rust-lang.org/stable/std/macro.compile_error.html?

bpf/aya-bpf/src/helpers.rs
```diff
@@ -701,3 +701,138 @@ pub fn bpf_get_current_pid_tgid() -> u64 {
...
+///
+/// The macro can read from arbitrary pointers, so it must be used in an `unsafe`
+/// scope in order to compile.
+///
+/// Invocations with less than 4 arguments call the `bpf_trace_printk` helper,
+/// otherwise the `bpf_trace_vprintk` function is called. The latter function
+/// requires a kernel version >= 5.16 whereas the former has been around since
+/// the dawn of eBPF.
+///
+/// Messages can be read by executing the following command in a second terminal:
```
alessandrod, line 720: I think we should also say how to enable trace_pipe if it's disabled

alessandrod (approved):
This looks great, thank you! See minor doc nit.

bpf/aya-bpf/src/helpers.rs
athre0z: Don't think so: since Rust doesn't have something like C++'s `if constexpr (...)` and we instead rely on the optimizer to get rid of the inactive match arms, I expect this to just fail unconditionally. Will try though! 

`assert!` can be const now (by writing `const _: () = assert!(...)`), but `const`s currently can't depend on generic arguments because they are effectively globals under the hood. All a bit unfortunate -- I hope this will be improved over time.

bpf/aya-bpf/src/helpers.rs
athre0z: I didn't realize it can be disabled -- that is good to know! Will figure out how and add a corresponding paragraph!

bpf/aya-bpf/src/helpers.rs
```diff
@@ -701,3 +701,138 @@ pub fn bpf_get_current_pid_tgid() -> u64 {
 pub fn bpf_get_current_uid_gid() -> u64 {
     unsafe { gen::bpf_get_current_uid_gid() }
 }
+
+/// Prints a debug message to the BPF debugging pipe.
```
athre0z, line 705: Note to self: also document the return value here.

bpf/aya-bpf/src/helpers.rs
athre0z: Yeah: as expected, this makes compilation fail unconditionally.

bpf/aya-bpf/src/helpers.rs
```diff
@@ -815,6 +809,27 @@ impl<T> From<*mut T> for PrintkArg {
...
 }
 
+/// Expands the given byte array to `DST_LEN`, right-padding it with zeros. If
+/// `DST_LEN` is smaller than `SRC_LEN`, the array is instead truncated.
+///
+/// This function serves as a helper for the [`bpf_printk!`] macro.
+#[doc(hidden)]
+pub const fn zero_pad_array<const SRC_LEN: usize, const DST_LEN: usize>(
+    src: [u8; SRC_LEN],
+) -> [u8; DST_LEN] {
```
athre0z, line 819: In case you are wondering why I went with a generic zero padding function instead of a "append_zero" function: a signature such as `const fn append_zero<const N: usize>(a: [u8; N]) -> [u8; N + 1]` will not compile on stable Rust, due to the arithmetic in the return value. It is however perfectly happy to do arithmetic on the call-site.

bpf/aya-bpf/src/helpers.rs
```diff
@@ -815,6 +809,27 @@ impl<T> From<*mut T> for PrintkArg {
...
+    let mut out: [u8; DST_LEN] = [0u8; DST_LEN];
+
+    // The `min` function is not `const`. Hand-roll it.
+    let mut i = if DST_LEN > SRC_LEN { SRC_LEN } else { DST_LEN };
```
athre0z, line 823: This isn't really required for what we do, but it didn't feel right to leave this edge-case unhandled in case we end up using this function somewhere else as well.

PR #355, opened 2022-08-02
aya: Remove MapError::InvalidPinPath

dave-tucker:

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -102,13 +102,6 @@ pub enum MapError {
         name: String,
     },
 
-    /// Pin path is invalid
-    #[error("invalid map path `{error}`")]
-    InvalidPinPath {
-        /// The error message
-        error: String,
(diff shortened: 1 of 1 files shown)


aya/src/maps/mod.rs
```diff
@@ -300,14 +293,7 @@ impl Map {
...
         let map_path = path.as_ref().join(name);
-        let path_string = match CString::new(map_path.to_str().unwrap()) {
-            Ok(s) => s,
-            Err(e) => {
-                return Err(MapError::InvalidPinPath {
-                    error: e.to_string(),
-                })
-            }
-        };
+        let path_string = CString::new(map_path.to_str().unwrap()).unwrap();
```
dave-tucker, line 296: The only possible error here is `NulError`. That would require the map name to end with a nul byte - which if it ever did have one was (hopefully) lost when the `object` lib converted the symbol name into a Rust string.

PR #359, opened 2022-08-04
Add support for setting priority for classifier programs

anfredette:
Addresses issue #358

The SchedClassifier attach function does not currently allow the priority (aka preference) of a classifier program to be explicitly set. Instead, the underlying netlink_qdisc_attach function defaults the priority to zero which tells the system to assign it the next highest priority for the given qdisc. It is useful to be able to explicitly control the order in which filters are executed and also to be able to insert new filters into an existing set at less than the highest priority.

This change adds support for setting the priority explicitly.

Note: aya-template will also need to be updated to add the new parameter.

### DIFF
--- a/aya/src/programs/tc.rs
+++ b/aya/src/programs/tc.rs
@@ -64,7 +64,7 @@ pub enum TcAttachType {
 ///
 /// let prog: &mut SchedClassifier = bpf.program_mut("redirect_ingress").unwrap().try_into()?;
 /// prog.load()?;
-/// prog.attach("eth0", TcAttachType::Ingress)?;
+/// prog.attach("eth0", TcAttachType::Ingress, 0)?;
 ///
 /// # Ok::<(), Error>(())
 /// ```
@@ -108,6 +108,10 @@ impl SchedClassifier {
 
     /// Attaches the program to the given `interface`.
     ///
+    /// Valid priority values range from 0 - 65535 with lower number = higher priority.
+    /// 0 means let the system choose the next highest priority, or 49152 if no filters exist yet.
+    /// All other values in the range are taken as an explicit priority setting (aka "preference").
+    ///
     /// The returned value can be used to detach, see [SchedClassifier::detach].
     ///
     /// # Errors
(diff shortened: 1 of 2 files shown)


aya/src/programs/tc.rs
```diff
@@ -120,12 +124,13 @@ impl SchedClassifier {
         &mut self,
         interface: &str,
         attach_type: TcAttachType,
+        priority: u32,
```
alessandrod, line 127: I think this should be `u16` in the API if the valid range is `0..u16::MAX`, then we should cast internally.

alessandrod (commented):
Thank you!

See minor nit and failing tests.

aya/src/programs/tc.rs
anfredette: @alessandrod thanks for the review.  I went back and forth on this myself.  Using a u16 requires more code changes and conversions back and forth, so I ended up keeping it as a u32 in the API.  However, making it a u16 is probably best for the user, so I'll make that change.

alessandrod:
Self reminder: this is a breaking change so requires a version bump

aya/src/programs/tc.rs
anfredette: The question I had was whether to just change priority to u16 in the SchedClassifier attach API, or everywhere in tc.rs and netlink.rs.  I chose everywhere to keep it consistent.  What do you think?

anfredette: FWIW, the iproute2 code has priority/preference as a u32 and does a range check:
https://github.com/shemminger/iproute2/blob/main/tc/tc_filter.c#L150

alessandrod: Do you know if there's a rationale for this? Is the range expected to be increased?

anfredette: Not that I know of.  They also do a <<16 [here](https://github.com/shemminger/iproute2/blob/main/tc/tc_filter.c#L186) like we're doing, so only the lower 16 bits are used anyway.  I think it's more user-friendly for us to make it a u16 at least in the API.

alessandrod: Yep agreed!

alessandrod (approved):

PR #360, opened 2022-08-05
Minor fix for `lpm_trie::key` docs

conectado:
Hi, this is a very minor fix to the docs of `lpm_trie::Key` I think the `prefix_len` represent the number of bits instead of the number of bytes to match against, this is needed to work with the whole range of CIDRs


alessandrod (approved):
Oops! Nice catch, thank you :)

PR #364, opened 2022-08-15
Allow SkBuffContext .store() without a mutable reference

zz85:
- there's no mutable reference to SkBuffContext
- there's interior mutability since skb is stored as a mutable reference


zz85:
Closing this as it's not needed, the immutable context can be declared or assigned mutable in order to use the mutable functions eg.

```
#[classifier(name = "tc_cls_ingress")]
pub fn tc_cls_ingress(ctx: SkBuffContext) -> i32 {
   let mut ctx = ctx;
   ctx.store(...)
}
```

PR #366, opened 2022-08-16
aya: Fix Link Pinning

dave-tucker:
1. Removes OwnedLink
2. Allows Links to be converted into FdLink
3. Introduces a PinnedLink type to handle wrap FdLink when pinned and
   support un-pinning


aya/src/programs/links.rs
```diff
@@ -106,17 +79,17 @@ impl<T: Link> Drop for LinkMap<T> {
...
+    pub(crate) fd: Option<RawFd>,
 }
 
 impl FdLink {
     pub(crate) fn new(fd: RawFd) -> FdLink {
-        FdLink { fd }
+        FdLink { fd: Some(fd) }
     }
 
     /// Pins the FdLink to a BPF filesystem.
```
alessandrod, line 95:
```
Pins the link to a BPF file system.

When a link is pinned to a BPF file system, it will remain attached even after the link instance is dropped, and will only be detached once the pinned file is removed. To unpin, see [PinnedFd::unpin].

The parent directories in the provided path must already exist before calling this method, and the path must point to a location on a BPF file system.
```

aya/src/programs/links.rs
```diff
@@ -125,18 +98,43 @@ impl FdLink {
...
+    /// #     #[error(transparent)]
+    /// #     Pin(#[from] aya::pin::PinError),
+    /// #     #[error(transparent)]
+    /// #     Program(#[from] aya::programs::ProgramError)
+    /// # }
+    /// # let mut bpf = aya::Bpf::load(&[])?;
+    /// # let prog: &mut Extension = bpf.program_mut("example").unwrap().try_into()?;
+    /// let link_id = prog.attach()?;
+    /// let owned_link = prog.take_link(link_id)?;
+    /// let fd_link : FdLink = owned_link.into();
```
alessandrod, line 119: extra white space between fd_link and :

aya/src/programs/links.rs
```diff
@@ -148,11 +146,49 @@ impl Link for FdLink {
...
+    fn new(path: PathBuf, fd: RawFd) -> Self {
+        PinnedLink {
+            inner: FdLink::new(fd),
+            path: Some(path),
+        }
+    }
+
+    /// Removes the pinned link from the filesystem and returns an [`FdLink`].
+    pub fn unpin(mut self) -> Result<FdLink, io::Error> {
+        if let Some(path) = self.path.take() {
```
alessandrod, line 185: unpin consumes self so you don't need the option? it can only be called once which means the path will always be there

aya/src/programs/links.rs
```diff
@@ -220,18 +256,32 @@ macro_rules! define_link_wrapper {
...
+            }
+        }
     };
 }
 
 pub(crate) use define_link_wrapper;
 
+#[derive(Error, Debug)]
+/// Errors from operations on Links.
+pub enum LinkError {
```
alessandrod, line 272: links lowercase

aya/src/programs/links.rs
```diff
@@ -220,18 +256,32 @@ macro_rules! define_link_wrapper {
...
+    /// link does not wrap an FdLink.
```
alessandrod, line 273:
```
/// Invalid link.
#[error("Invalid link")]
InvalidLink
```

alessandrod (commented):
Looks great! See comments

PR #370, opened 2022-08-26
aya-bpf/programs: Add `pull_data` method to SKB context

chenhengqi:
The pull_data method is used to ensure that all the required bytes
are available in the linear portion of the skb.

Signed-off-by: Hengqi Chen <chenhengqi@outlook.com>

### DIFF
--- a/bpf/aya-bpf/src/programs/sk_buff.rs
+++ b/bpf/aya-bpf/src/programs/sk_buff.rs
@@ -6,7 +6,8 @@ use core::{
 
 use aya_bpf_bindings::helpers::{
     bpf_clone_redirect, bpf_get_socket_uid, bpf_l3_csum_replace, bpf_l4_csum_replace,
-    bpf_skb_adjust_room, bpf_skb_change_type, bpf_skb_load_bytes, bpf_skb_store_bytes,
+    bpf_skb_adjust_room, bpf_skb_change_type, bpf_skb_load_bytes, bpf_skb_pull_data,
+    bpf_skb_store_bytes,
 };
 use aya_bpf_cty::c_long;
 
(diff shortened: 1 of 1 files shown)


vadorovsky (approved):

bpf/aya-bpf/src/programs/sk_buff.rs
```diff
@@ -224,6 +225,37 @@ impl SkBuffContext {
             Err(ret)
         }
     }
+
+    /// Pulls data to the SKB linear section, returning
```
alessandrod, line 229: This seems wrong? We return `()` on success.

I'd say:

```
Pulls in non-linear data in case the skb is non-linear.

Make len bytes from skb readable and writable. If a zero value is passed for `len`, then the whole length of the skb is pulled. This helper is only needed for reading and writing with direct packet access.
```

alessandrod (commented):
Thanks! Only a small doc nit

bpf/aya-bpf/src/programs/sk_buff.rs
chenhengqi: Update as requested.

bpf/aya-bpf/src/programs/sk_buff.rs
```diff
@@ -224,6 +225,39 @@ impl SkBuffContext {
             Err(ret)
         }
     }
+
+    /// Pulls in non-linear data in case the skb is non-linear.
```
alessandrod, line 229: Whitespace is significant in Markdown, so we need a newline after the initial short intro, and no newline after the periods in the next paragraph.

```
Pulls in non-linear data in case the skb is non-linear.

Make `len` bytes from skb readable and writable. If a zero value is passed for
`len`, then the whole length of the skb is pulled. This helper is only needed
for reading and writing with direct packet access.
```

alessandrod (commented):
Small doc formatting nit

bpf/aya-bpf/src/programs/sk_buff.rs
chenhengqi: OK, didn't realize the Rust doc comment format.

alessandrod (approved):

PR #371, opened 2022-08-28
aya: expose BPF verifier log level configuration

conectado:
Hi!

Simple change that allows configuring verifier log level using `BpfLoader`.

Reducing the log level is very useful when debugging since the last line of the verifier error is its cause, so when the buffer is out of space you can't see the cause.

For reference: change discussed [here](https://discord.com/channels/855676609003651072/855676609003651075/1012486817087500389)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -194,6 +194,7 @@ pub struct BpfLoader<'a> {
     globals: HashMap<&'a str, &'a [u8]>,
     features: Features,
     extensions: HashSet<&'a str>,
+    verifier_log_level: u32,
 }
 
 impl<'a> BpfLoader<'a> {
@@ -207,6 +208,7 @@ impl<'a> BpfLoader<'a> {
             globals: HashMap::new(),
             features,
             extensions: HashSet::new(),
+            verifier_log_level: 7,
         }
     }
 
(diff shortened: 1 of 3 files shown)


aya/src/bpf.rs
```diff
@@ -314,6 +316,24 @@ impl<'a> BpfLoader<'a> {
...
+    /// ```no_run
+    /// use aya::BpfLoader;
+    ///
+    /// let bpf = BpfLoader::new()
+    ///     .verifier_log_level(3)
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::BpfError>(())
+    /// ```
+    ///
+    pub fn verifier_log_level(&mut self, level: u32) -> &mut BpfLoader<'a> {
```
alessandrod, line 332: I think instead of `u32` we should probably have an enum:

```
#[non_exhaustive]
#[repr(u32)]
pub enum VerifierLogLevel {
    Default = 4,
    Verbose = 1,
    Debug = 7,
}
```

See https://github.com/torvalds/linux/blob/f16214c102f0f64b2f3546e989498525bd7b7708/include/linux/bpf_verifier.h#L419. The Default/Verbose value correspond to what libbpf does too.

alessandrod (commented):
Thank you for doing this! Look good, see enum comment

aya/src/bpf.rs
conectado: Updated it [here](https://github.com/aya-rs/aya/pull/371/commits/edd80397dce46f6e2a4cc96bd951562987721e55), changed the names and added some valid configurations.

Let me know if you still prefer the above variants, I just felt like this it was slightly clearer.

conectado: Could also change it up so that is similar to a bitmap, e.g. set it up like:

```rs
VerifierLogLevel::default().level_1(true).stats(true)
```

But might be too verbose for something like this.

alessandrod: > Updated it [here](https://github.com/aya-rs/aya/pull/371/commits/edd80397dce46f6e2a4cc96bd951562987721e55), changed the names and added some valid configurations.
> 
> Let me know if you still prefer the above variants, I just felt like this it was slightly clearer.

I don't think this is cleaner - in fact it took me a good 10 minutes yesterday to understand why those names were used in the kernel and how they work. The names are bad, and I don't think we should surface the fact that log level s are implemented as a bitfield in the kernel all the way up to the aya API.

alessandrod: We should also have a `Disable = 0` variant in addition to the ones I initially suggested tho

alessandrod (approved):
Looks great, thank you!

PR #376, opened 2022-09-01
Update `VerifierLogLevel` to use bitflags

conectado:
As talked with @alessandrod [here](https://discord.com/channels/855676609003651072/855676609003651075/1014213239762067486), using `bitflags!` to implement the `VerifierLogLevel`.

I think that using the `Default` trait might be better than having a `DEFAULT` value in the struct but not so sure on that.

Also setting `Level2` as `Level1 | 2` since in the kernel level 1 seems to imply level 2. Meaning, that when they check for log level they either check for [`BPF_LOG_LEVEL2`](https://github.com/torvalds/linux/blob/f16214c102f0f64b2f3546e989498525bd7b7708/include/linux/bpf_verifier.h#L419) (when it's exclusive to 2) or [`BPF_LOG_LEVEL`](https://github.com/torvalds/linux/blob/f16214c102f0f64b2f3546e989498525bd7b7708/include/linux/bpf_verifier.h#L422) when it also include level 1, So I didn't think being able to use 2 was very useful.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -205,7 +205,7 @@ bitflags! {
         const DEBUG = 1;
         /// Enables verbose verifier logging.
         const VERBOSE = 2 | Self::DEBUG.bits;
-        /// Enables verifier stats
+        /// Enables verifier stats.
         const STATS = 4;
     }
 }
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -196,22 +196,26 @@ pub struct BpfLoader<'a> {
     verifier_log_level: VerifierLogLevel,
 }
 
-/// Used to set the verifier log level in [BpfLoader](BpfLoader::verifier_log_level()).
-#[repr(u32)]
-#[non_exhaustive]
-#[derive(Debug, Clone, Copy)]
-pub enum VerifierLogLevel {
-    /// Disable all logging.
-    Disable = 0,
+bitflags! {
+    /// Used to set the verifier log level in [BpfLoader](BpfLoader::verifier_log_level()).
+    pub struct VerifierLogLevel: u32 {
+        /// Used to disable all logging.
+        const DISABLE = 0;
+        ///  Logs tracing with details level 1
+        const LEVEL1 = 1;
+        /// Log tracing with details level 2
+        const LEVEL2 = 2 | Self::LEVEL1.bits;
+        /// Logs eBPF verifier stats
+        const STATS = 4;
+    }
+}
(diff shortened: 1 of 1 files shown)


aya/src/bpf.rs
```diff
@@ -196,22 +196,26 @@ pub struct BpfLoader<'a> {
...
-
-    /// Prints verbose logs showing tracing.
-    Verbose = 1,
+bitflags! {
+    /// Used to set the verifier log level in [BpfLoader](BpfLoader::verifier_log_level()).
+    pub struct VerifierLogLevel: u32 {
+        /// Used to disable all logging.
+        const DISABLE = 0;
+        ///  Logs tracing with details level 1
+        const LEVEL1 = 1;
```
alessandrod, line 205: I think this should be `DEBUG`

aya/src/bpf.rs
```diff
@@ -196,22 +196,26 @@ pub struct BpfLoader<'a> {
...
+        /// Log tracing with details level 2
+        const LEVEL2 = 2 | Self::LEVEL1.bits;
```
alessandrod, line 207: I think this should be `VERBOSE`

alessandrod (changes requested):
Oops sorry I forgot to review this last week!

See comments - I still think that LEVEL1 and LEVEL2 are bad names (as you discovered, LEVEL2 implies LEVEL1 anyway). I think we should use `DEBUG` and `VERBOSE`.

conectado:
agreed! those are better names

alessandrod (approved):
Perfect, thank you!

aya/src/bpf.rs
```diff
@@ -196,22 +196,26 @@ pub struct BpfLoader<'a> {
...
+bitflags! {
+    /// Used to set the verifier log level flags in [BpfLoader](BpfLoader::verifier_log_level()).
+    pub struct VerifierLogLevel: u32 {
+        /// Sets no verifier logging.
+        const DISABLE = 0;
+        /// Enables debug verifier logging.
+        const DEBUG = 1;
+        /// Enables verbose verifier logging.
+        const VERBOSE = 2 | Self::DEBUG.bits;
+        /// Enables verifier stats
```
alessandrod, line 208: Missing period at the end of the sentence

alessandrod: @conectado oops just noticed this - could you fix it pls?

conectado: sorry didn't see this message, right away

PR #390, opened 2022-09-09
clippy: Fix latest nightly lints

dave-tucker:


dave-tucker:
@vadorovsky @alessandrod would like one of you to look at this since i had to fix a miri issue in `aya-log`... which i think is fine, but at least review would be nice :wink:

vadorovsky (approved):

PR #395, opened 2022-09-21
aya: Fixup BTF STRUCT

dave-tucker:
The kernel doesn't like Rust idents. Until we fix the kernel, we should instead remove them from the BTF that gets loaded into the kernel. As type-id's are constant, it's not too much bother to recover the name from the original BTF should you need the it.

### DIFF
--- a/aya/src/obj/btf/btf.rs
+++ b/aya/src/obj/btf/btf.rs
@@ -399,6 +399,18 @@ impl Btf {
                     fixed_ty.name_offset = 0;
                     types.types[i] = BtfType::Ptr(fixed_ty)
                 }
+                // Fixup STRUCT for Rust
+                // LLVM emits names that include characters that aren't valid C idents.
+                // Rather than replace all unsupported characters, just omit the name entirely.
+                BtfType::Struct(s) => {
+                    let name = self.string_at(s.name_offset)?.to_string();
+                    if name.contains(['<', '>', ' ', ':', ',']) {
+                        debug!("STRUCT name {} not supported. Omitted from BTF", name);
+                        let mut fixed_ty = s.clone();
+                        fixed_ty.name_offset = 0;
+                        types.types[i] = BtfType::Struct(fixed_ty)
+                    }
+                }
                 // Sanitize VAR if they are not supported
                 BtfType::Var(v) if !features.btf_datasec => {
                     types.types[i] = BtfType::Int(Int::new(v.name_offset, 1, IntEncoding::None, 0));


aya/src/obj/btf/btf.rs
```diff
@@ -399,6 +399,18 @@ impl Btf {
                     fixed_ty.name_offset = 0;
                     types.types[i] = BtfType::Ptr(fixed_ty)
                 }
+                // Fixup STRUCT for Rust
+                // LLVM emits names that include characters that aren't valid C idents.
+                // Rather than replace all unsupported characters, just omit the name entirely.
+                BtfType::Struct(s) => {
+                    let name = self.string_at(s.name_offset)?.to_string();
```
alessandrod, line 406: Why do we need this only for struct types and not all idents?

dave-tucker: Here's the constraints we're working under - https://elixir.bootlin.com/linux/v5.19.10/source/kernel/bpf/btf.c#L731
Which are basically "alphanumeric with spaces, underscores and possibly a dot"

Most variable, function and even struct names in rust follow this convention BUT not always.
e.g.

This is probably OK:
```
[19] STRUCT 'PerfEventByteArray' size=28 vlen=1
        'def' type_id=15 bits_offset=0
```

But `14` and `15` are not:
```
[14] STRUCT 'PerCpuArray<aya_log_ebpf::LogBuf>' size=28 vlen=2
        'def' type_id=15 bits_offset=0
        '_t' type_id=17 bits_offset=0
[15] STRUCT 'UnsafeCell<aya_bpf_bindings::bindings::bpf_map_def>' size=28 vlen=1
        'value' type_id=16 bits_offset=0
[16] STRUCT 'bpf_map_def' size=28 vlen=7
        'type_' type_id=3 bits_offset=0
        'key_size' type_id=3 bits_offset=32
        'value_size' type_id=3 bits_offset=64
        'max_entries' type_id=3 bits_offset=96
        'map_flags' type_id=3 bits_offset=128
        'id' type_id=3 bits_offset=160
        'pinning' type_id=3 bits_offset=192
```

It appears that anything that is `#[repr(C)]` does get correct BTF (see `16` and [here](https://github.com/aya-rs/aya/blob/6b1f47323bcb5f022cfda45c86be153b8cb68d8e/bpf/aya-bpf-bindings/src/lib.rs#L46))

For the most part, other idents are pretty well behaved... so I've left them alone.

I guess I should only sanitize the names that violate constraints.

alessandrod: Yes I understand and I think we should probably even use exactly the same code as `__btf_name_char_ok`, and probably even link to it for reference. 

My point is why only structs, what about functions with generics and methods with generics? Do those work?

dave-tucker: Ok, let me take another pass at this, fixing it for all types.

alessandrod: And maybe a better approach would be to mangle those names with the standard mangling scheme? https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html

PR #404, opened 2022-10-04
docs: Add labels for optional features

dave-tucker:
Following the lead of crates like tokio and nix, we now annotate APIs that require optional features. This helps in cases where a user wants to have an `AsyncPerfEventArray` which is documented on crates.io, but it's not obvious that you have to enable the `async` feature.

### DIFF
--- a/aya/src/lib.rs
+++ b/aya/src/lib.rs
@@ -36,6 +36,7 @@
     html_logo_url = "https://aya-rs.dev/assets/images/crabby.svg",
     html_favicon_url = "https://aya-rs.dev/assets/images/crabby.svg"
 )]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![deny(clippy::all, missing_docs)]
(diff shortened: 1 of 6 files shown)


dave-tucker:
image

aya/src/lib.rs
```diff
@@ -33,9 +33,10 @@
 //! [async-std]: https://docs.rs/async-std
 
 #![doc(
-    html_logo_url = "https://aya-rs.dev/assets/images/crabby.svg",
-    html_favicon_url = "https://aya-rs.dev/assets/images/crabby.svg"
+    html_logo_url = "https://aya-rs.dev/assets/images/crabby_dev.svg",
```
vadorovsky, line 36: Huh, shouldn't the userspace documentation have a normal crabee?

vadorovsky (approved):

aya/src/lib.rs
dave-tucker: Argh! Good catch.

vadorovsky (approved):

Issue #375, opened 2022-08-31
Program Name should come from the symbols table, not the ELF section


dave-tucker:
Since libbpf 1.0, section names got strict (actually this happened in an earlier point release).

aya-bpf sometimes includes the program name as part of the section name... meaning libbpf will now flat out refuse to load programs that used aya-bpf in that manner. We also presented an inconsistent UX to users...
Sometimes we use the section name as default - e.g `xdp` - other times we use the actual function name.

In order to be libbpf 1.0 compatible - and generally to make aya easier to use - we should:

1. Fix the `aya-bpf` macros to be consistent - we can continue to support `name` but it will be become shorthand for `export_name`
1. Use the ELF symbol table to find **multiple programs** in a the same ELF section.
1. Document that in the `Bpf::program` and  `Bpf::program_mut` APIs that programs are uniquely identified by their function name

PR #413, opened 2022-10-12
aya: fix names once and for all

dave-tucker:
:warning: WARNING :warning:

This will probably break the API for people, but at the cost of making the name used in `bpf.prog_mut("name")` more predictable since:

```c
SEC("tracepoint")
int foo {}
```
Name is `foo`

```c
SEC("tracepoint/foo")
int foo {}
```
Name is `foo`

```c
SEC("sk_skb/stream_parser/loads/of/other/stuff")
int foo {}
```
Name is `foo`

You get the idea :laughing: 

This will also allow for us to clean up the section-name parsing logic significantly 🙏 

# Migration Notes

As a user of Aya migration is simple:

- The use of `name` in all program `aya-bpf::macros` has been removed and will be caught by the compiler when you update. It is safe to remove this entirely.
- When using `Bpf::program_mut` you should ALWAYS use the name of the function from your BPF application code - i.e if the function you annotated with a `#[tracepoint]` macro is named `bpf_prog`, then you would use `bpf.programs_mut("bpf_prog")`
- It is now safe to use multiple instances of the same macro in a single BPF program. For an example please see `two_progs.rs` in the integration tests.

Fixes: #375


alessandrod:
This still leaves the `name` argument in program macros right? But now it gets ignored which doesn't seem great.

I think we can do this change without breaking existing programs, by changing the macros to override the provided item name (the function name), with the name provided as an argument, for instance:

```
#[xdp(name = "foo")]
fn bar(...) { ... }
```

the macro would output:

```
#[no_mangle]
#[link_section = "xdp"]
fn foo(...) { ... }
```

dave-tucker:
@alessandrod per discord discussion I went ahead and fixed up all the aya-ebpf-macros and removed name - it's a breaking change. In doing so I refactored the macros to be more easy to test and added loads of tests - probably 50% or more of total lines changed are tests 🤓

tamird:
@dave-tucker could you rebase this? I'd like to review it as well <3

dave-tucker:
@tamird @alessandrod rebased.

aya-obj/src/obj.rs
```diff
@@ -947,10 +970,20 @@ impl Object {
                 res?
             }
             BpfSectionKind::Program => {
-                let (program, function) = self.parse_program(&section)?;
-                self.functions.insert(program.function_key(), function);
-                self.programs
-                    .insert(program.section.name().to_owned(), program);
+                let symbols: HashMap<String, Symbol> = self
```
alessandrod, line 973: We now do this in 3 different places: here, in ::Maps and in ::BtfMaps. We
should do it only once, build `self.symbols_by_section: HashMap<SectionIndex, Vec<usize>>`
while we build `self.symbol_table`. Keys are section indexes, values are symbol
indexes used to index into `self.symbol_table`.

aya-obj/src/obj.rs
```diff
@@ -712,10 +712,30 @@ impl Object {
...
+        Ok(())
+    }
 
+    fn parse_program(
+        &self,
+        section: &Section,
+        name: &str,
+        symbol: &Symbol,
+    ) -> Result<(Program, Function), ParseError> {
+        let prog_sec = ProgramSection::from_str(section.name)?;
```
alessandrod, line 738: This is going to be the same for all programs in the section, so it could be
moved to the caller

aya-obj/src/obj.rs
```diff
@@ -712,10 +712,30 @@ impl Object {
...
-        let name = prog_sec.name().to_owned();
+    fn parse_programs(
+        &mut self,
+        section: &Section,
+        symbols: HashMap<String, Symbol>,
+    ) -> Result<(), ParseError> {
+        let progs = symbols
+            .iter()
+            .map(|(name, symbol)| self.parse_program(section, name, symbol))
+            .collect::<Result<Vec<(Program, Function)>, ParseError>>()?;
```
alessandrod, line 723: There's no need to collect here. Instead of using map, iterate on
symbols.iter(), and we save an extra allocation + an extra loop

aya-bpf-macros/src/sk_msg.rs
```diff
@@ -0,0 +1,33 @@
...
+            #[no_mangle]
+            #[link_section = "sk_msg"]
+            #fn_vis fn #fn_name(ctx: *mut ::aya_bpf::bindings::sk_msg_md) -> u32 {
+                return #fn_name(::aya_bpf::programs::SkMsgContext::new(ctx));
+
+                #item
+            }
+        })
+    }
+}
```
alessandrod, line 33: Forgot a test here? 😅

alessandrod (changes requested):
Amazing stuff! See comments, and also I think we're missing an integration test with multiple programs in the same section?

tamird (commented):
> Reviewed 6 of 6 files at r1, 9 of 32 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/413)* status: 14 of 37 files reviewed, 7 unresolved discussions (waiting on @alessandrod and @dave-tucker)

___
*[`-- commits` line 21 at r2](https://reviewable.io/reviews/aya-rs/aya/413#-N_TjG_t9DD5ql0ZR2K0):*
can you help me understand why this required removing all the shared code? many of these macros remain quite similar yet share nothing.
___
*[`aya-obj/src/obj.rs` line 1864 at r1](https://reviewable.io/reviews/aya-rs/aya/413#-N_Th0bu723QL9hmkkbL:-N_Th0bu723QL9hmkkbM:b-c4ma66) ([raw file](https://github.com/aya-rs/aya/blob/2608b4f1698de4d6299d06975ce6c1342da2ab2a/aya-obj/src/obj.rs#L1864)):*
> ```XML
>                 kernel_version: None,
>                 section: ProgramSection::KProbe { .. },
>                 ..} if license.to_string_lossy() == "GPL"
> ```

unusual formatting here

also: why do we need to_string_lossy here?
___
*[`aya-obj/src/obj.rs` line 723 at r2](https://reviewable.io/reviews/aya-rs/aya/413#gh-1264402487:-N_Tg_dR6zOgtn1eo87W:b-irfu55) ([raw file](https://github.com/aya-rs/aya/blob/82b0d459906d7923b1067abe3d0c3fe4a6dd7b1a/aya-obj/src/obj.rs#L723)):*
<details><summary><i>Previously, alessandrod (Alessandro Decina) wrote…</i></summary><blockquote>

There's no need to collect here. Instead of using map, iterate on
symbols.iter(), and we save an extra allocation + an extra loop
</blockquote></details>

looks like collect is being used to materialize errors before mutating `self`
___
*[`test/integration-test/src/tests/rbpf.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/413#-N_TgqEQ1FI7kyJ2tQkU:-N_TgqER4AguSWne5bHU:b5l3mfi) ([raw file](https://github.com/aya-rs/aya/blob/2608b4f1698de4d6299d06975ce6c1342da2ab2a/test/integration-test/src/tests/rbpf.rs#L38)):*
> ```XML
> 
>     assert_eq!(object.programs.len(), 1);
>     assert!(matches!(
> ```

why change this?

alessandrod:
> looks like collect is being used to materialize errors before mutating `self`

we're in an unrecoverable path (Object::parse) so mutating self is fine

aya-bpf-macros/src/sk_msg.rs
dave-tucker: Done

aya-obj/src/obj.rs
dave-tucker: Done

aya-obj/src/obj.rs
dave-tucker: Done.

aya-obj/src/obj.rs
dave-tucker: Done

aya-obj/src/obj.rs
dave-tucker: Done

aya-bpf-macros/src/sk_msg.rs
dave-tucker: Done.

dave-tucker:
> > Reviewed 6 of 6 files at r1, 9 of 32 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/413)_ status: 14 of 37 files reviewed, 7 unresolved discussions (waiting on @alessandrod and @dave-tucker)
> 
> _[`-- commits` line 21 at r2](https://reviewable.io/reviews/aya-rs/aya/413#-N_TjG_t9DD5ql0ZR2K0):_ can you help me understand why this required removing all the shared code? many of these macros remain quite similar yet share nothing.

No shared code has been removed - shared code (for argument handling) lives in `args.rs`.
Without this change `expand.rs` blew up to ~3000 lines and was a mess of parsing functions + expand functions + shared logic. Refactoring this way makes it much simpler to add new macros and/or debug/test existing macros.

> 
> > ```
> >                 kernel_version: None,
> >                 section: ProgramSection::KProbe { .. },
> >                 ..} if license.to_string_lossy() == "GPL"
> > ```
> 
> unusual formatting here
> 
> also: why do we need to_string_lossy here?

That's down to cargo fmt, not me I think.
`.to_string_lossy` is required because license is a CString and we're comparing to a `str`.

> > ```
> >     assert_eq!(object.programs.len(), 1);
> >     assert!(matches!(
> > ```
> 
> why change this?

Bad rebase. Fixed

dave-tucker:
> > Reviewed 2 of 32 files at r3, 1 of 31 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/413)_ status: 7 of 37 files reviewed, 12 unresolved discussions (waiting on @alessandrod and @dave-tucker)
> 
> _[`aya-bpf-macros/src/socket_filter.rs` line 11 at r4](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZITg65zxI0Gw07_xy:-N_ZITg65zxI0Gw07_xz:b-xykkdk) ([raw file](https://github.com/aya-rs/aya/blob/34a20b1cc6ab131e5be1d6557b759c68d7235e9a/aya-bpf-macros/src/socket_filter.rs#L11)):_
> 
> > ```
> > impl SocketFilter {
> >     pub(crate) fn parse(attrs: TokenStream, item: TokenStream) -> Result<Self> {
> > ```
> 
> elaborating on what i saying about shared code. consider this file and tc.rs; the difference between these files is 3 tokens:
> 
> SocketFilter -> SchedClassifier "socket" -> "classifier" SkBuffContext -> TcContext
> 
> I haven't looked at the other files in enough detail but this is the kind of thing I'm asking about.

Right, none of this was shared code before though.
Take `SchedClassifier` for example:
- Where `parse` was here: https://github.com/aya-rs/aya/blob/main/aya-bpf-macros/src/lib.rs#L76-L85
- And `expand` was here: https://github.com/aya-rs/aya/blob/main/aya-bpf-macros/src/expand.rs#L246-L277

Macros that take 0 attributes are all quite similar yes, and you could probably reduce duplication using generics.
That said, the duplication doesn't bother me in that it enforces convention and makes it easier to write tests etc...

> _[`aya-obj/src/obj.rs` line 1864 at r1](https://reviewable.io/reviews/aya-rs/aya/413#-N_Th0bu723QL9hmkkbL:-N_ZHT3w-xIeqPRcPfzo:b-r76ua7) ([raw file](https://github.com/aya-rs/aya/blob/2608b4f1698de4d6299d06975ce6c1342da2ab2a/aya-obj/src/obj.rs#L1864)):_
> 
> Previously, tamird (Tamir Duberstein) wrote…
> rustfmt often gives up on formatting macros, sadly
> 
> clarifying the question about to_string_lossy: can we use .to_str.unwrap()? we expect these strings to be UTF8 in these tests, do we not?

This is test code. There's little appreciable difference in a panic due to .to_str.unwrap vs the assertion that an empty or non-UTF-8 string != GPL IMHO. If it bothers you that much I'll change it.

> 
> _[`aya-obj/src/obj.rs` line 727 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZDvAi1PFnn1sY_jKJ:-N_ZDvAi1PFnn1sY_jKK:bq2k2ou) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L727)):_
> 
> > ```
> >         for symbol_index in self.symbols_by_section.get(&section.index).unwrap().iter() {
> >             let symbol = self.symbol_table.get(symbol_index).unwrap();
> >             let name = symbol.name.as_ref().unwrap();
> > ```
> 
> can you help me understand why these unwrap while other fallible calls return errors?

Because in both cases they are practically infallible.
symbol.table.get() can't fail for any entry in symbols_by_section
similarly, symbol.name can't be empty either since unnamed symbols aren't in symbols_by_section.

> 
> _[`aya-obj/src/obj.rs` line 887 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZE0nv2Ogfqh4fdUre:-N_ZE0nv2Ogfqh4fdUrf:b-yd5u4i) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L887)):_
> 
> > ```
> >             .symbols_by_section
> >             .get(&section.index)
> >             .unwrap()
> > ```
> 
> same question here and a few lines down

As above

> 
> _[`aya-obj/src/obj.rs` line 977 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZI4I1DDaxgpAfrj2f:-N_ZI4I1DDaxgpAfrj2g:bmfmk7t) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L977)):_
> 
> > ```
> >     fn parse_section(&mut self, section: Section) -> Result<(), ParseError> {
> >         let mut parts = section.name.rsplitn(2, '/').collect::<Vec<_>>();
> > ```
> 
> this doesn't seem to be used. should it?

It's not part of my PR but looks like you've taken care of it already in #665 

> _[`aya-obj/src/obj.rs` line 1006 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZEEToDGhuv_NvLX9E:-N_ZEEToDGhuv_NvLX9F:bpgwxg1) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L1006)):_
> 
> > ```
> >                 // extract the symbols for the .maps section, we'll need them
> >                 // during parsing
> >                 let symbols = self.symbols_by_section.get(&section.index).unwrap().iter();
> > ```
> 
> here as well

We should never see a `.maps` section with no symbols in it.

> _[`test/integration-test/src/tests/rbpf.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/413#-N_TgqEQ1FI7kyJ2tQkU:-N_ZHfiC6YjEnH6NfyqN:b-19h7px) ([raw file](https://github.com/aya-rs/aya/blob/2608b4f1698de4d6299d06975ce6c1342da2ab2a/test/integration-test/src/tests/rbpf.rs#L38)):_
> 
> Previously, tamird (Tamir Duberstein) wrote…
> looks like this wasn't changed in the right commit? and there's a debug print two lines above in the first commit

Oops. I'll remove the debug print. Had to change it in the first commit to get the integration tests passing after adding `symbols_by_section` from Alessandro's comments

tamird (commented):
> Reviewed 1 of 31 files at r6.
> Dismissed @tamird from a discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/413)* status: 6 of 37 files reviewed, 11 unresolved discussions (waiting on @alessandrod and @dave-tucker)

___
*[`aya-bpf-macros/src/socket_filter.rs` line 11 at r4](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZITg65zxI0Gw07_xy:-N_Zq30q8a9kyBx5Jk5r:bxyps2p) ([raw file](https://github.com/aya-rs/aya/blob/34a20b1cc6ab131e5be1d6557b759c68d7235e9a/aya-bpf-macros/src/socket_filter.rs#L11)):*

Right, none of this was shared code before though.
Take `SchedClassifier` for example:
- Where `parse` was here: https://github.com/aya-rs/aya/blob/main/aya-bpf-macros/src/lib.rs#L76-L85
- And `expand` was here: https://github.com/aya-rs/aya/blob/main/aya-bpf-macros/src/expand.rs#L246-L277

Macros that take 0 attributes are all quite similar yes, and you could probably reduce duplication using generics.
That said, the duplication doesn't bother me in that it enforces convention and makes it easier to write tests etc...

</blockquote></details>

Okay. I'm happy to let more knowledgeable folks like @alessandrod and yourself continue this discussion if necessary.

Small note that even the duplicated code isn't consistent everywhere, for instance some variable definitions are reordered, etc; and it might be good to make identical code 100% identical to avoid the appearance of logical differences.
___
*[`aya-obj/src/obj.rs` line 1864 at r1](https://reviewable.io/reviews/aya-rs/aya/413#-N_Th0bu723QL9hmkkbL:-N_ZpaRm2Ld-FchtcxAc:bttsihi) ([raw file](https://github.com/aya-rs/aya/blob/2608b4f1698de4d6299d06975ce6c1342da2ab2a/aya-obj/src/obj.rs#L1864)):*

This is test code. There's little appreciable difference in a panic due to .to_str.unwrap vs the assertion that an empty or non-UTF-8 string != GPL IMHO. If it bothers you that much I'll change it.
</blockquote></details>

I think stronger assertions are better than weaker ones, that's all.
___
*[`aya-obj/src/obj.rs` line 727 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZDvAi1PFnn1sY_jKJ:-N_ZpucLAlf0wSQYvSEj:bihx20f) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L727)):*

Because in both cases they are practically infallible.
symbol.table.get() can't fail for any entry in symbols_by_section
similarly, symbol.name can't be empty either since unnamed symbols aren't in symbols_by_section.

</blockquote></details>

seems reasonable. should we use `expect` to document the reasoning?
___
*[`aya-obj/src/obj.rs` line 887 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZE0nv2Ogfqh4fdUre:-N_ZqNAW20q3WUlKE_O_:bb77ucv) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L887)):*

As above
</blockquote></details>

Yep.
___
*[`aya-obj/src/obj.rs` line 977 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZI4I1DDaxgpAfrj2f:-N_ZpoA88w9LW9N9QPI2:b-5kdgtr) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L977)):*

It's not part of my PR but looks like you've taken care of it already in #665 
</blockquote></details>

Yeah, answered my own question.
___
*[`aya-obj/src/obj.rs` line 1006 at r3](https://reviewable.io/reviews/aya-rs/aya/413#-N_ZEEToDGhuv_NvLX9E:-N_ZpggIBqbnAoyACAYM:bzht5m5) ([raw file](https://github.com/aya-rs/aya/blob/e5d27b6b3139fba0e6f9ccdb0247c5a7a0bea50e/aya-obj/src/obj.rs#L1006)):*

We should never see a `.maps` section with no symbols in it.
</blockquote></details>

Seems reasonable, do we want these unwraps to be `expect`s?
___
*[`test/integration-test/src/tests/rbpf.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/413#-N_TgqEQ1FI7kyJ2tQkU:-N_ZpOG15mnhcm-x38Q_:byn8xhx) ([raw file](https://github.com/aya-rs/aya/blob/2608b4f1698de4d6299d06975ce6c1342da2ab2a/test/integration-test/src/tests/rbpf.rs#L38)):*

Oops. I'll remove the debug print. Had to change it in the first commit to get the integration tests passing after adding `symbols_by_section` from Alessandro's comments
</blockquote></details>

Ack. Can you remove this change in the first commit please? Currently first commit changes assert_matches -> assert(matches( and then second commit changes it back.

aya-obj/src/obj.rs
```diff
@@ -711,10 +719,29 @@ impl Object {
         Ok(())
     }
 
-    fn parse_program(&self, section: &Section) -> Result<(Program, Function), ParseError> {
-        let prog_sec = ProgramSection::from_str(section.name)?;
-        let name = prog_sec.name().to_owned();
+    fn parse_programs(&mut self, section: &Section) -> Result<(), ParseError> {
+        let program_section = ProgramSection::from_str(section.name)?;
+        for symbol_index in self.symbols_by_section.get(&section.index).unwrap().iter() {
```
alessandrod, line 724: I don't think we should unwrap here? It's valid for a section to not have any
symbols in the symbol table

aya-obj/src/obj.rs
```diff
@@ -711,10 +719,29 @@ impl Object {
...
+            self.functions.insert(key, f);
+        }
+        Ok(())
+    }
 
+    fn parse_program(
+        &self,
+        section: &Section,
+        program_section: &ProgramSection,
+        name: &str,
```
alessandrod, line 742: This should be String and the caller should clone if necessary

aya-obj/src/obj.rs
```diff
@@ -711,10 +719,29 @@ impl Object {
...
+            self.programs.insert(f.name.clone(), p);
+            self.functions.insert(key, f);
+        }
+        Ok(())
+    }
 
+    fn parse_program(
+        &self,
+        section: &Section,
+        program_section: &ProgramSection,
```
alessandrod, line 741: Same here, this should be ProgramSection and the caller should clone

aya-obj/src/obj.rs
```diff
@@ -845,15 +875,29 @@ impl Object {
...
         }
         let btf = self.btf.as_ref().unwrap();
+        let maps: HashMap<&String, usize> = self
+            .symbols_by_section
+            .get(&section.index)
+            .unwrap()
+            .iter()
+            .filter_map(|s| {
+                let symbol = self.symbol_table.get(s).unwrap();
+                if let Some(name) = symbol.name.as_ref() {
```
alessandrod, line 890: This is Option::and_then

aya-obj/src/obj.rs
```diff
@@ -898,47 +983,25 @@ impl Object {
...
                 let mut maps = mem::take(&mut self.maps);
 
                 // extract the symbols for the .maps section, we'll need them
                 // during parsing
-                let symbols = self.symbol_table.values().filter(|s| {
-                    s.section_index
-                        .map(|idx| idx == section.index.0)
-                        .unwrap_or(false)
-                });
+                let symbols = self.symbols_by_section.get(&section.index).unwrap().iter();
```
alessandrod, line 994: We shouldn't unwrap here

Say I'm bpfd and load a bunch of user provided .o files

Say I'm an attacker and I manage to put a specially crafted .o file in the bpfd
load path

I can strip .maps symbols from the symbol table to crash the whole bpfd process. 

In general, it's ok to unwrap() on things that can't fail by construction, for instance
the symbol.name.unwrap() you do where symbols with None names are pre-filtered
(although in that case expect() is better than unwrap()). It's not ok to unwrap
on invariants that depend on external inputs.

alessandrod (commented):
See comments. If you replace some unwrap()s with expect()s it'll help you/us understand where we're unwrapping on assumptions we can't actually guarantee internally

alessandrod:
Also before landing this it would be good to have something somewhere that explains what breaks and how to migrate?

dave-tucker:
@alessandrod latest iteration is a lot less `unwrap`y - everything is now either an expect, or has more robust error handling.

aya-obj/src/obj.rs
```diff
@@ -711,10 +719,36 @@ impl Object {
         Ok(())
     }
 
-    fn parse_program(&self, section: &Section) -> Result<(Program, Function), ParseError> {
-        let prog_sec = ProgramSection::from_str(section.name)?;
-        let name = prog_sec.name().to_owned();
+    fn parse_programs(&mut self, section: &Section) -> Result<(), ParseError> {
+        let program_section = ProgramSection::from_str(section.name)?;
+        if let Some(syms) = self.symbols_by_section.get(&section.index) {
+            for symbol_index in syms.iter() {
```
alessandrod, line 725: nit: I think iter() here is redundant?

aya-obj/src/obj.rs
```diff
@@ -711,10 +719,36 @@ impl Object {
...
+                let symbol = self
+                    .symbol_table
+                    .get(symbol_index)
+                    .expect("all symbols in symbols_by_section are also in symbol_table");
+                let empty_name = String::new();
+                let name = symbol.name.as_ref().unwrap_or(&empty_name);
```
alessandrod, line 731: let Some(name) = symbol.name.as_ref() else { continue }

aya-obj/src/obj.rs
```diff
@@ -711,10 +719,36 @@ impl Object {
         Ok(())
     }
 
-    fn parse_program(&self, section: &Section) -> Result<(Program, Function), ParseError> {
-        let prog_sec = ProgramSection::from_str(section.name)?;
-        let name = prog_sec.name().to_owned();
+    fn parse_programs(&mut self, section: &Section) -> Result<(), ParseError> {
+        let program_section = ProgramSection::from_str(section.name)?;
+        if let Some(syms) = self.symbols_by_section.get(&section.index) {
```
alessandrod, line 724: Not sure what we should do here. For maps we return an error, but here we just
do nothing and skip the section. We should probably rename
NoSymbolsForMapsSection to MissingSectionSymbols and return it here too?

dave-tucker: In your last review you said it's valid for sections that to have no symbols...
Sections that contain BPF programs always have symbols though - which is why this patch works 🤔

For maps (and BTF maps) we know that both sections always have symbols, and we're able to derive the section type from the name - "maps" or ".maps".

Our detection logic for `ProgramSection` is basically a catch-all - i.e it wasn't one of the other enum variants and it's a  Text section with length > 0.

If we're 100% confident that will only match sections that contain BPF programs (and therefore have symbols) we can error out here too.

I'm not sure that's always the case though, so skipping the section seems sensible for now - we could always emit a warn log to help debug if that's easier.

aya-obj/src/obj.rs
dave-tucker: Done

alessandrod (approved):
Welp, this took forever 😅 Thank you SO MUCH for doing this work and enduring so long 🥳❤️

PR #431, opened 2022-11-01
aya: use impl Borrow<T> instead of T for maps

### LINKED ISSUE #427: API for updating maps with large Pod
Although Rust is memory safe, it does not have any stack overflow protection. The default Rust stack is 2MB which is plenty for most applications. However, when dealing with Pod with large arrays embedded, it's possible to overflow the user space Rust stack.

Here's an example:

```rust
#[derive(Clone, Copy)]
pub struct Data {
    arr: [u8; 2 << 10 << 10],
}
```

Now, in order to prevent this, we need to `Box` the struct. However, since `Box` can't be `Copy`. So such structs are unable to implement `aya::Pod`.

0b01:
This PR changes map related APIs to use reference to Pod instead of Pod. This should fix #427

### DIFF
--- a/aya/src/maps/array/array.rs
+++ b/aya/src/maps/array/array.rs
@@ -1,6 +1,7 @@
 use std::{
+    borrow::Borrow,
     convert::{AsMut, AsRef},
-    marker::PhantomData, borrow::Borrow,
+    marker::PhantomData,
 };
 
 use crate::{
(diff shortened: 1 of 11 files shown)


alessandrod:
Thanks for doing this! Instead of taking references, I think we should take `impl AsRef<V>`, so that then both are possible

alessandrod (approved):

alessandrod:
Thank you, great work 😊

PR #444, opened 2022-11-17
aya: Add PinnedProgram

dave-tucker:
This commit adds PinnedProgram which allows the creation of a Program from a path on bpffs. This is useful to be able to call `attach` or other APIs for programs that are already loaded to the kernel. Unfortunately figuring out the concrete type of program is hard given the information in the kernel so this may not work for all program types.


vadorovsky (approved):

alessandrod:
> . Unfortunately figuring out the concrete type of program is hard given the information in the kernel so this may not work for all program types.

This makes me think that adding an extra `PinnedProgram` is probably not the correct API? If we did `ConcreteProgram::from_pinned(...)` then we could have function arguments to fill in the gaps in info we can't infer from the pin.

astoycos:
Shouldn't this be closed now that #496 is merged?

PR #460, opened 2022-12-07
maps: add missing TryFrom<Map> for HashMap, PerCpuHashMap and LpmTrie

Tuetuopay:


aya/src/maps/mod.rs
```diff
@@ -412,8 +412,19 @@ macro_rules! impl_try_from_map_generic_key_and_value {
...
                         },
                         _ => Err(MapError::InvalidMapType{ map_type: map.map_type()}),
+                    }
                 }
             }
+
+            impl<V: Pod, K: Pod> TryFrom<Map> for $ty<MapData, V, K> {
+                type Error = MapError;
+
+                fn try_from(map: Map) -> Result<Self, Self::Error> {
```
astoycos, line 422:
```suggestion
                fn try_from(map: Map) -> Result<$ty<MapData, V, K>, MapError> {
```

astoycos:
Hey @Tuetuopay Thanks so much for finding this!!! Just one small change I think

Issue #412, opened 2022-10-11
Add integration tests for BTF relocations


alessandrod:
We should add integration tests for BTF relocations. These could be one small program written in C for each relocation type (struct field, array field, bitfield, etc). 

Each test should define and load its "target BTF", with the final layout for the thing being relocated and tested. Then it should have a local definition for the thing being relocated with an incompatible layout. The test should check that after relocation, the layout matches the target BTF.

shaneutt:
I'm very interested in giving this a try. My only caveat is that I'm a few weeks out from having time for it as my life is wrapped up in Kubecon right now, so if there's anyone else who needs to take it in the meantime please do. Otherwise my intention will be to have a go at it after Kubecon.

tamird:
Done, I think.

PR #467, opened 2022-12-16
Add integration tests for BTF relocations

MatteoNardi:
I've started to work on https://github.com/aya-rs/aya/issues/412
It's not ready for review yet, but I need some feedback before going on.

For now, I have a test for a moved field inside a struct and an enum change. The integration test works by trying to read a relocatable field from a raw array of bytes. The read data is then sent to userspace using a BPF_MAP_TYPE_ARRAY. The integration_test then asserts the output is consistent with the target BTF instead of the local definition.
Originally I was planning to send the relocatable data structure from userspace into eBPF using a global variable. I discarded that option because I felt it was making things more complicated and harder to understand than just reading the raw bytes.

I thought it would be best to have a test generator (`RelocationTest`). This is different from the pre-existing integration-test way of doing things: instead of adding a `.bpf.c` program for each relocation type, we compile the ebpf directly inside the test. The rationale is:
- We'll have many tests like these, adding a new eBPF program (and btf definition) for each is verbose
- I dislike having the final `assert_eq!` expectation in a different file from the one where everything is defined.

I think it would make the test cases easier to understand if we included a "run without BTF and make sure the results are different". This is why I added a `RelocationTestRunner::run_no_btf()`. Unfortunately all these tests are currently failing with a:
```
Option::unwrap()` on a `None` value', aya/src/obj/btf/relocation.rs:888:33
```

Let me know what you think.

### DIFF
--- a/test/integration-test/src/tests/relocations.rs
+++ b/test/integration-test/src/tests/relocations.rs
@@ -54,6 +54,25 @@ fn relocate_enum() {
     assert_eq!(test.run_no_btf().unwrap(), 0);
 }
 
+#[integration_test]
+fn relocate_pointer() {
+    let test = RelocationTest {
+        local_definition: r#"
+            struct foo {};
+            struct bar { struct foo *f; };
+        "#,
+        target_btf: r#""#,
+        relocation_code: r#"
+            __u8 memory[] = {0, 0, 0, 0, 0, 0, 0, 42};
+            struct foo *f = BPF_CORE_READ((struct bar *)&memory, f);
+            __u32 value = (f == (struct foo *) 42);
+        "#,
+    }
+    .build()
+    .unwrap();
+    assert_eq!(test.run().unwrap(), 1);
+}
+
 /// Utility code for running relocation tests:
 /// - Generates the eBPF program using probided local definition and relocation code
 /// - Generates the BTF from the target btf code
--- a/test/integration-test/src/tests/relocations.rs
+++ b/test/integration-test/src/tests/relocations.rs
@@ -145,7 +145,7 @@ impl RelocationTest {
 
     /// - Generate the target BTF source with a mock main()
     /// - Compile it with clang
-    /// - Extract the BTF with pahole
+    /// - Extract the BTF with llvm-objcopy
     fn build_btf(&self) -> Result<Btf> {
         let target_btf = self.target_btf;
         let relocation_code = self.relocation_code;
@@ -171,12 +171,12 @@ impl RelocationTest {
             "#
         ))
         .context("Failed to compile BTF")?;
-        Command::new("pahole")
+        Command::new("llvm-objcopy")
             .current_dir(tmp_dir.path())
-            .arg("--btf_encode_detached=target.btf")
+            .args(["--dump-section", ".BTF=target.btf"])
             .arg(compiled_file)
             .status()
-            .context("Failed to run pahole")?
+            .context("Failed to run llvm-objcopy")?
             .success()
             .then_some(())
             .context("Failed to extract BTF")?;
--- a/test/integration-test/src/tests/relocations.rs
+++ b/test/integration-test/src/tests/relocations.rs
@@ -14,7 +14,7 @@ fn relocate_field() {
               __u8 b;
               __u8 c;
               __u8 d;
-            } __attribute__((preserve_access_index));
+            };
         "#,
         target_btf: r#"
             struct foo {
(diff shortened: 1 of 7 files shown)


xtask/src/build_ebpf.rs
```diff
@@ -103,6 +103,7 @@ fn build_c_ebpf(opts: &BuildEbpfOptions) -> anyhow::Result<()> {
 
     let include_path = out_path.join("include");
     get_libbpf_headers(&opts.libbpf_dir, &include_path)?;
+    std::env::set_var("LIBBPF_INCLUDE", &include_path);
```
MatteoNardi, line 106: This is  a hack to get libbpf headers inside the RelocationTest builder. I'll gather feedback on the general approach before thinking about a more appropriate solution!

alessandrod:
> I've started to work on #412 It's not ready for review yet, but I need some feedback before going on.

Sorry it took a while - holidays and all.

I just skimmed through this and it looks amazing!

> Unfortunately all these tests are currently failing with a: ```Option::unwrap() on a `None` value', aya/src/obj/btf/relocation.rs:888:33```

I'm happy to look into this

MatteoNardi:
Marking it as ready for review!
I've never seen some of these tests passing, so take care: some might be wrong.

Many more relocation types are missing, but I think it's better to get these in and fix the issues which are causing them to fail before adding new ones. Hopefully adding new ones is trivial for those working on the actual relocation code.

alessandrod:
> > Unfortunately all these tests are currently failing with a: `` Option::unwrap() on a `None` value', aya/src/obj/btf/relocation.rs:888:33 ``
> 
> I'm happy to look into this

Update: I've debugged this, and it happens invoking `run_no_btf()`. The reason is that in that case we don't have the target BTF, we only have the local BTF). While some relocation kinds like `TypeIdLocal` don't require the target BTF, most relocation kinds do require it and should error out gracefully when it's missing. The fix is simple, I'll push it tomorrow.

test/integration-test/src/tests/relocations.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+#[integration_test]
+fn relocate_pointer() {
+    let test = RelocationTest {
+        local_definition: r#"
+            struct foo {};
+            struct bar { struct foo *f; };
+        "#,
+        target_btf: r#""#,
```
alessandrod, line 64: This is missing the target definition

test/integration-test/src/tests/relocations.rs
```diff
@@ -0,0 +1,247 @@
...
+        // $ pahole --btf_encode_detached=target.btf -V target.o
+        // $ bpftool btf dump file ./target.btf  format c
+        let btf_file = tmp_dir.path().join("target_btf.c");
+        std::fs::write(
+            &btf_file,
+            format!(
+                r#"
+                    #include <linux/types.h>
+                    {target_btf}
+                    int main() {{ return 0; }}
```
alessandrod, line 178: As this is now, the compiler isn't guaranteed to emit BTF info for the target btf, since it's unused and could be dead-code eliminated. In fact this happens with the pointer test (once you add the target declarations, which are now missing). 

One workaround is to compile the tracepoint program even for the target definition. Otherwise we need to ensure that all target definitions are actually definitions (they create globals or something), instead of just being declarations.

test/integration-test/src/tests/relocations.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Run without loading btf
+    fn run_no_btf(&self) -> Result<u32> {
+        self.run_internal(false)
+    }
+
+    fn run_internal(&self, with_relocations: bool) -> Result<u32> {
+        let mut loader = BpfLoader::new();
+        if with_relocations {
+            loader.btf(Some(&self.btf));
+        }
```
alessandrod, line 227: by default the loader will load BTF info from sysfs if `btf()` is invoked, so here we need an else branch that does `loader.btf(None)`

test/integration-test/src/tests/relocations.rs
```diff
@@ -0,0 +1,247 @@
...
+        target_btf: r#""#,
+        relocation_code: r#"
+            __u8 memory[] = {0, 0, 0, 0, 0, 0, 0, 42};
+            struct foo *f = BPF_CORE_READ((struct bar *)&memory, f);
+            __u32 value = (f == (struct foo *) 42);
+        "#,
+    }
+    .build()
+    .unwrap();
+    assert_eq!(test.run().unwrap(), 1);
```
alessandrod, line 73: this is missing a `run_no_btf()` run

alessandrod (changes requested):
Some more comments after playing around with this :)

test/integration-test/src/tests/relocations.rs
MatteoNardi: Interesting, thanks!

test/integration-test/src/tests/relocations.rs
MatteoNardi: Fixed, except that I don't really know what we should test here. Basically we're just testing loading ends with no error.
Let me know if you have a better idea or if we can keep it as is.

MatteoNardi:
Thanks for the review and the fixes! The tests are passing now.

vadorovsky (approved):

alessandrod (approved):

alessandrod:
Amazing work thank you! 😊

PR #470, opened 2022-12-21
Add type Enum64

levigross:
Enum64 was added in the 6.x kernels

src: https://github.com/torvalds/linux/blob/master/tools/include/uapi/linux/btf.h#L190


alessandrod:
Hey, thanks for your contribution!

We already have a PR for this at https://github.com/aya-rs/aya/pull/453. Supporting enum64 requires relocation work too.

levigross:
I now see this is a duplicate of https://github.com/aya-rs/aya/pull/453 (at least in attempt of adding the feature). I am happy to close it if you are looking to merge that one. Sorry for the noise.

PR #471, opened 2022-12-22
add tracefs mount point select function

banditopazzo:
This PR would to improve the access to `tracefs`.

This is implemented with a single check to determine if the `tracefs` is mounted or not. It checks in two known locations, in order of priority:

1. `/sys/kernel/tracing`
2. `/sys/kernel/debug/tracing`

Most of the systems have both `tracefs` and `debugfs` mounted, but in some hardened systems the kernel is compiled without `debugfs` or simply not mounted. Android documentation for example has a warning for debugfs in user builds.

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -207,9 +207,9 @@ pub enum ProgramError {
         name: String,
     },
 
-    /// TraceFS not found.
-    #[error("tracefs non found")]
-    TraceFsNotFound,
+    /// An error occurred while working with IO.
+    #[error(transparent)]
+    IOError(#[from] io::Error),
 }
 
 /// A [`Program`] file descriptor.
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -206,6 +206,10 @@ pub enum ProgramError {
         /// program name
         name: String,
     },
+
+    /// TraceFS not found.
+    #[error("tracefs non found")]
+    TraceFsNotFound,
 }
 
 /// A [`Program`] file descriptor.
--- a/aya/src/programs/trace_point.rs
+++ b/aya/src/programs/trace_point.rs
@@ -1,5 +1,5 @@
 //! Tracepoint programs.
-use std::{fs, io};
+use std::{fs, io, path::Path};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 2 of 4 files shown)


aya/src/programs/utils.rs
```diff
@@ -22,3 +22,25 @@ pub(crate) fn attach_raw_tracepoint<T: Link + From<FdLink>>(
 
     program_data.links.insert(FdLink::new(pfd).into())
 }
+
+// Get tracefs filesystem 
+pub(crate) fn get_tracefs() -> Result<&'static Path, ProgramError> {
```
alessandrod, line 27: I think this should be call `find_tracefs_path` or something

aya/src/programs/utils.rs
```diff
@@ -22,3 +22,25 @@ pub(crate) fn attach_raw_tracepoint<T: Link + From<FdLink>>(
...
+                Path::new("/sys/kernel/debug/tracing"),
+            ];
+
+            for mount in mounts {
+                if mount.exists() {
+                    return Some(mount);
+                }
+            }
+            None
+
```
alessandrod, line 41: remove this empty line, if you want to add vertical padding the newline should be before None

aya/src/programs/mod.rs
```diff
@@ -206,6 +206,10 @@ pub enum ProgramError {
         /// program name
         name: String,
     },
+
+    /// TraceFS not found.
+    #[error("tracefs non found")]
+    TraceFsNotFound,
```
alessandrod, line 212: I think that this is too specific - it's not something that will likely ever be handled programmatically so it doesn't deserve its own variant. We should probably add a `ProgramError::IOError(#[from] IOError)` and create a `Error::new(ErrorKind::Other, "tracefs not found")` error for this. Then move some of the other fringe variants into it (not as part of this PR, but at some point in the future)

aya/src/programs/trace_point.rs
```diff
@@ -12,6 +12,8 @@ use crate::{
     sys::perf_event_open_trace_point,
 };
 
+use super::utils::get_tracefs;
```
alessandrod, line 15: please merge this use statement inside the one above

alessandrod (commented):
Thanks for working on this! Just a few minor comments

aya/src/programs/mod.rs
banditopazzo: sounds good

vadorovsky (approved):
Thanks!

@alessandrod are you happy with the current state of this PR?

alessandrod (commented):
Can you please rebase on main so we can get CI green?

alessandrod (approved):

Issue #469, opened 2022-12-21
Regression: unexpected BTF type id when relocating pointer


MatteoNardi:
The current main branch has a regression compared to 0.11 which causes load-time fails for eBPF programs trying to relocate pointers.
This issue is reproduced with an integration test in #467 (I'm opening up an issue to avoid forgetting about this one in case the PR won't get merged)

Trying to `BPF_CORE_READ(foo, pointer_field);` will fail with `Unexpected BTF type id xyz`, with `xyz` which changes depending on the eBPF code.

I stumbled on this while [porting pulsar](https://github.com/Exein-io/pulsar/pull/121) to latest aya commit. This is a shortened version of the code:
```
#include <linux/bpf.h>

#include <bpf/bpf_core_read.h>
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_tracing.h>

struct sockaddr {};
struct sock {};
struct socket {
  struct sock *sk;
};

SEC("lsm/socket_bind")
int BPF_PROG(socket_bind, struct socket *sock, struct sockaddr *address,
             int addrlen, int ret) {
  struct sock_common *sk = BPF_CORE_READ(sock, sk);
  return ret;
}

char _license[] SEC("license") = "GPL";
```

alessandrod:
Could you try the commit before https://github.com/aya-rs/aya/pull/285 was merged? Should be 7b143199fb61edd168f3efc860a8e8c1d4cd9136

alessandrod:
This was indeed a regression caused by #285. Fix coming soon.

PR #480, opened 2022-12-31
btf: fix regression computing pointer sizes

alessandrod:
Computing pointer sizes was broken in #285


alessandrod:
This fixes https://github.com/aya-rs/aya/issues/469

PR #496, opened 2023-01-19
aya: Add from_pin for Programs

dave-tucker:
This commit adds from_pin() which allows the creation of a Program from a path on bpffs. This is useful to be able to call `attach` or other APIs for programs that are already loaded to the kernel.

This differs from #444 since it implements this on the concrete program type, not the Program enum, allowing the user to pass in any additional context that isn't available from bpf_prog_info.

### DIFF
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -2,6 +2,7 @@
 use std::{
     hash::Hash,
     os::unix::prelude::{AsRawFd, RawFd},
+    path::Path,
 };
 
 use crate::{
@@ -137,6 +138,23 @@ impl CgroupSkb {
     /// See [CgroupSkb::attach].
     pub fn detach(&mut self, link_id: CgroupSkbLinkId) -> Result<(), ProgramError> {
         self.data.links.remove(link_id)
+    }
+
+    /// Creates a program from a pinned entry on a bpffs.
+    ///
+    /// Existing links will not be populated. To work with existing links you should use [`crate::programs::links::PinnedLink`].
+    ///
+    /// On drop, any managed links are detached and the program is unloaded. This will not result in
+    /// the program being unloaded from the kernel if it is still pinned.
+    pub fn from_pin<P: AsRef<Path>>(
+        path: P,
+        expected_attach_type: CgroupSkbAttachType,
+    ) -> Result<Self, ProgramError> {
+        let data = ProgramData::from_pinned_path(path)?;
+        Ok(Self {
+            data,
+            expected_attach_type: Some(expected_attach_type),
+        })
     }
 }
 
--- a/aya/src/programs/cgroup_sock.rs
+++ b/aya/src/programs/cgroup_sock.rs
@@ -4,6 +4,7 @@ pub use aya_obj::programs::CgroupSockAttachType;
 use std::{
     hash::Hash,
     os::unix::prelude::{AsRawFd, RawFd},
+    path::Path,
 };
 
 use crate::{
@@ -112,6 +113,20 @@ impl CgroupSock {
     /// See [CgroupSock::attach].
     pub fn detach(&mut self, link_id: CgroupSockLinkId) -> Result<(), ProgramError> {
         self.data.links.remove(link_id)
+    }
+
+    /// Creates a program from a pinned entry on a bpffs.
+    ///
+    /// Existing links will not be populated. To work with existing links you should use [`crate::programs::links::PinnedLink`].
+    ///
+    /// On drop, any managed links are detached and the program is unloaded. This will not result in
+    /// the program being unloaded from the kernel if it is still pinned.
+    pub fn from_pin<P: AsRef<Path>>(
+        path: P,
+        attach_type: CgroupSockAttachType,
+    ) -> Result<Self, ProgramError> {
+        let data = ProgramData::from_pinned_path(path)?;
+        Ok(Self { data, attach_type })
     }
 }
 
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -69,7 +69,7 @@ use std::{
     ffi::CString,
     io,
     os::unix::io::{AsRawFd, RawFd},
-    path::Path,
+    path::{Path, PathBuf},
 };
 use thiserror::Error;
 
@@ -108,8 +108,9 @@ use crate::{
     obj::{self, btf::BtfError, Function, KernelVersion},
     pin::PinError,
     sys::{
-        bpf_get_object, bpf_load_program, bpf_pin_object, bpf_prog_get_fd_by_id,
-        bpf_prog_get_info_by_fd, bpf_prog_query, retry_with_verifier_logs, BpfLoadProgramAttrs,
+        bpf_btf_get_fd_by_id, bpf_get_object, bpf_load_program, bpf_pin_object,
+        bpf_prog_get_fd_by_id, bpf_prog_get_info_by_fd, bpf_prog_query, retry_with_verifier_logs,
+        BpfLoadProgramAttrs,
     },
     util::VerifierLog,
 };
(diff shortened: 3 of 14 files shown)


aya/src/programs/cgroup_skb.rs
```diff
@@ -56,6 +58,7 @@ use crate::{
 pub struct CgroupSkb {
     pub(crate) data: ProgramData<CgroupSkbLink>,
     pub(crate) expected_attach_type: Option<CgroupSkbAttachType>,
+    pub(crate) path: Option<PathBuf>,
```
alessandrod, line 61: It seems you've added this to all the program structs - doesn't it belong to
ProgramData then? (ProgramData is used for fields common to all structs)

aya/src/programs/cgroup_sock.rs
```diff
@@ -113,6 +121,24 @@ impl CgroupSock {
...
+
+    /// Creates a program from a pinned entry on a bpffs.
+    ///
+    /// Existing links will not be populated. To work with existing links you should use [`PinnedLink`].
+    ///
+    /// On drop, any managed links are detached and the program is unloaded. This will not result in
+    /// the program being unloaded from the kernel if it is still pinned.
+    pub fn from_pin<P: AsRef<Path>>(
+        path: P,
+        attach_type: Option<CgroupSockAttachType>,
```
alessandrod, line 133: Why can this be optional?

aya/src/programs/mod.rs
```diff
@@ -692,6 +738,31 @@ impl_fd!(
     CgroupDevice,
 );
 
+macro_rules! program_data_from_path {
```
alessandrod, line 741: Couldn't this be a function?

aya/src/programs/lirc_mode2.rs
```diff
@@ -49,6 +52,7 @@ use libc::{close, dup};
 #[doc(alias = "BPF_PROG_TYPE_LIRC_MODE2")]
 pub struct LircMode2 {
     pub(crate) data: ProgramData<LircLink>,
+    pub(crate) path: Option<PathBuf>,
```
alessandrod, line 55: You forgot to add from_pin here?

aya/src/programs/perf_event.rs
```diff
@@ -119,6 +121,7 @@ pub enum PerfEventScope {
 #[doc(alias = "BPF_PROG_TYPE_PERF_EVENT")]
 pub struct PerfEvent {
     pub(crate) data: ProgramData<PerfLink>,
+    pub(crate) path: Option<PathBuf>,
```
alessandrod, line 124: You forgot from_pin here?

aya/src/programs/socket_filter.rs
```diff
@@ -61,6 +62,7 @@ pub enum SocketFilterError {
 #[doc(alias = "BPF_PROG_TYPE_SOCKET_FILTER")]
 pub struct SocketFilter {
     pub(crate) data: ProgramData<SocketFilterLink>,
+    pub(crate) path: Option<PathBuf>,
```
alessandrod, line 65: Can SocketFilters get pinned?

alessandrod (commented):
This seems largely ok! See comments

aya/src/programs/socket_filter.rs
dave-tucker: Done

aya/src/programs/perf_event.rs
dave-tucker: Fixed.

aya/src/programs/lirc_mode2.rs
dave-tucker: Also Fixed.

aya/src/programs/mod.rs
dave-tucker: It can. Fixed.

aya/src/programs/cgroup_sock.rs
dave-tucker: It can't. Fixed.

aya/src/programs/cgroup_skb.rs
dave-tucker: Done.

dave-tucker:
@alessandrod changes addressed. 
- `Path` was moved to `ProgramData` and `program_data_from_path` was made a function instead of a macro
- `from_pin` was added using a macro `impl_from_pin` for simple cases and manually otherwise

dave-tucker:
Updated to fix CI...
I had to move the `Drop` trait from the `Program` enum to the concrete program type (i.e XDP) for this to behave as expected. For example:
```rust
{
    let prog = Xdp::from_pin("/path/to/prog").unwrap();
    prog.attach("eth0");
}
// Link should get detached here and program fd should be closed
```
When I moved `Drop`, clippy was kind enough to point out that we have both `Program::unload` and `Xdp::unload` - the latter was `pub` where the former was `pub (crate)` and only called via the drop implementation. I went ahead and made `Program::unload` public since to me it makes sense to have it there as a convenience.

alessandrod:
> Updated to fix CI... I had to move the `Drop` trait from the `Program` enum to the concrete program type (i.e XDP) for this to behave as expected. For example:
> 
> ```rust
> {
>     let prog = Xdp::from_pin("/path/to/prog").unwrap();
>     prog.attach("eth0");
> }
> // Link should get detached here and program fd should be closed
> ```
> 
> When I moved `Drop`, clippy was kind enough to point out that we have both `Program::unload` and `Xdp::unload` - the latter was `pub` where the former was `pub (crate)` and only called via the drop implementation. I went ahead and made `Program::unload` public since to me it makes sense to have it there as a convenience.

This now will not drop programs unless you access them through concrete program types tho right? Eg if you Bpf::load and never create a concrete program

aya/src/programs/mod.rs
```diff
@@ -210,6 +211,10 @@ pub enum ProgramError {
     /// An error occurred while working with IO.
     #[error(transparent)]
     IOError(#[from] io::Error),
+
+    /// The program defintion is incomplete.
+    #[error("incomplete program defintion. {0}")]
+    IncompleteProgramDefinition(String),
```
alessandrod, line 217: This can be removed now

aya/src/programs/mod.rs
```diff
@@ -439,8 +439,70 @@ impl<T: Link> ProgramData<T> {
...
+            btf_fd: None,
+            verifier_log_level: 0,
+            path: Some(path.to_path_buf()),
+        })
+    }
+
+    pub(crate) fn from_pinned_path<P: AsRef<Path>>(
+        path: P,
+    ) -> Result<ProgramData<T>, ProgramError> {
+        let path_string = std::ffi::CString::new(path.as_ref().to_str().unwrap()).unwrap();
```
alessandrod, line 487: I think we should use to_string_lossy here

alessandrod: also import types, don't use full qualified paths

aya/src/programs/mod.rs
```diff
@@ -439,8 +439,70 @@ impl<T: Link> ProgramData<T> {
...
+        let fd = crate::sys::bpf_get_object(&path_string).map_err(|(_, io_error)| {
+            ProgramError::SyscallError {
+                call: "bpf_obj_get".to_owned(),
+                io_error,
+            }
+        })? as std::os::fd::RawFd;
+
+        let info = crate::sys::bpf_prog_get_info_by_fd(fd).map_err(|io_error| {
```
alessandrod, line 495: import bpf_get_object

aya/src/programs/mod.rs
```diff
@@ -703,8 +776,18 @@ macro_rules! impl_program_pin{
                 /// To remove the program, the file on the BPF filesystem must be removed.
                 /// Any directories in the the path provided should have been created by the caller.
                 pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
+                    self.data.path = Some(PathBuf::from(path.as_ref().to_owned()));
```
alessandrod, line 779: I think this is path.to_path_buf()?

aya/src/programs/mod.rs
```diff
@@ -703,8 +776,18 @@ macro_rules! impl_program_pin{
...
                     pin_program(&mut self.data, path)
                 }
+
+                /// Removes the pinned link from the filesystem.
+                pub fn unpin(mut self) -> Result<(), io::Error> {
+                    let path = self.data.path.take();
```
alessandrod, line 785: if let Some(path) = self.data.path.take() {
	
}

test/integration-test/src/tests/load.rs
```diff
@@ -192,16 +192,28 @@ fn pin_lifecycle() {
...
         prog.load().unwrap();
+        prog.pin("/sys/fs/bpf/aya-xdp-test-prog").unwrap();
+    }
+
+    // should still be loaded since prog was pinned
+    assert_loaded!("pass", true);
+
+    // 2. Load program from bpffs
+    {
+        let mut prog = Xdp::from_pin("/sys/fs/bpf/aya-xdp-test-prog").unwrap();
```
alessandrod, line 203: I think we're missing a test where a Xdp::from_pin program goes out of scope
without links and without being unpinned. We should assert that the program is
still loaded after.

alessandrod (commented):
Almost there!

aya/src/programs/mod.rs
dave-tucker: Ugh, so Path -> Cstring isn't exactly easy, but this is what I settled on.
```rust
CString::new(path.as_ref().as_os_str().to_string_lossy().as_bytes())
```

aya/src/programs/mod.rs
dave-tucker: Done

aya/src/programs/mod.rs
dave-tucker: yep. fixed.

aya/src/programs/mod.rs
dave-tucker: done.

test/integration-test/src/tests/load.rs
dave-tucker: added, thanks.

alessandrod (approved):
LGTM! Feel free to merge on green

PR #497, opened 2023-01-20
BTF fixes

astoycos:
1. Make sure we support LLVM generated names which aren't valid C idents. 

2. Also, fix a bug which resulted in failure to load maps->  

```
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_PERF_EVENT_ARRAY, key_size=4, value_size=4, max_entries=24, map_flags=0, inner_map_fd=0, map_name="ingress_node_fi", map_ifindex=0, btf_fd=11, btf_key_type_id=6, btf_value_type_id=6, btf_vmlinux_value_type_id=0, map_extra=0}, 144) = -1 ENOTSUPP (Unknown error 524)
```
Specifically it is the same fix deployed by libbpf, more information can be found here https://github.com/libbpf/libbpf/issues/355

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -425,6 +425,18 @@ impl Btf {
                     fixed_ty.name_offset = 0;
                     types.types[i] = BtfType::Ptr(fixed_ty)
                 }
+                // Fixup STRUCT for Rust
+                // LLVM emits names that include characters that aren't valid C idents.
+                // Rather than replace all unsupported characters, just omit the name entirely.
+                BtfType::Struct(s) => {
+                    let name = self.string_at(s.name_offset)?.to_string();
+                    if name.contains(['<', '>', ' ', ':', ',']) {
+                        debug!("STRUCT name {} not supported. Omitted from BTF", name);
+                        let mut fixed_ty = s.clone();
+                        fixed_ty.name_offset = 0;
+                        types.types[i] = BtfType::Struct(fixed_ty)
+                    }
+                }
                 // Sanitize VAR if they are not supported
                 BtfType::Var(v) if !features.btf_datasec => {
                     types.types[i] = BtfType::Int(Int::new(v.name_offset, 1, IntEncoding::None, 0));


astoycos:
@dave-tucker  I picked up a commit you had from https://github.com/dave-tucker/aya/tree/program-from-pinned

aya-obj/src/btf/btf.rs
```diff
@@ -425,6 +425,18 @@ impl Btf {
                     fixed_ty.name_offset = 0;
                     types.types[i] = BtfType::Ptr(fixed_ty)
                 }
+                // Fixup STRUCT for Rust
+                // LLVM emits names that include characters that aren't valid C idents.
+                // Rather than replace all unsupported characters, just omit the name entirely.
+                BtfType::Struct(s) => {
+                    let name = self.string_at(s.name_offset)?.to_string();
+                    if name.contains(['<', '>', ' ', ':', ',']) {
```
alessandrod, line 433: Shouldn't this be applied to everything that uses generics? functions, enums, etc not only structs

astoycos: Good question, I personally had only run into this with Struct's but I guess it could apply to the following (maybe I'm missing some) 

```
#define BTF_KIND_PTR            2       /* Pointer      */
#define BTF_KIND_STRUCT         4       /* Struct       */
#define BTF_KIND_UNION          5       /* Union        */
#define BTF_KIND_ENUM           6       /* Enumeration up to 32-bit values */
#define BTF_KIND_FUNC           12      /* Function     */
#define BTF_KIND_FUNC_PROTO     13      /* Function Proto       */
#define BTF_KIND_VAR            14      /* Variable     */
#define BTF_KIND_ENUM64         19      /* Enumeration up to 64-bit values */
```

alessandrod: The way I would go about this is:

- create a rust program that uses all items that can include generics (presumably you already have a program that hits the Struct case, so you could extend that)
- generate BTF for it
- have an integration test that loads the program

This should give you all the possible errors until you fix all of them 😅

astoycos: Will do!

astoycos:
@alessandrod I broke out the other commit into it's own PR, I will still work on this one ^^ but the other is more pressing ATM 

see -> https://github.com/aya-rs/aya/pull/512

dave-tucker:
@astoycos I'm going to close this one since it only applies to Rust programs with BTF using my ancient branch of bpf-linker. @vadorovsky has a branch of bpf-linker which actually fixes the LLVM debug info, which is the better approach IMHO.

PR #501, opened 2023-01-26
aya: btf: fix relocations for signed enums (32 bits)

alessandrod:
Once upon a time relocations were always signed. That was changed and now enums include a sign bit in the info flags. Adjust the code accordingly.

### DIFF
--- a/test/integration-test/src/tests/relocations.rs
+++ b/test/integration-test/src/tests/relocations.rs
@@ -28,9 +28,7 @@ fn relocate_field() {
         relocation_code: r#"
             __u8 memory[] = {1, 2, 3, 4};
             struct foo *ptr = (struct foo *) &memory;
-            bpf_probe_read_kernel(&value,
-                                  sizeof(__u8),
-                                  __builtin_preserve_access_index(&ptr->c));
+            value = __builtin_preserve_access_index(ptr->c);
         "#,
     }
     .build()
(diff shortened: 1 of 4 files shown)


test/integration-test/src/tests/relocations.rs
```diff
@@ -55,8 +55,28 @@ fn relocate_enum() {
...
+        "#,
+        relocation_code: r#"
+            #define BPF_ENUMVAL_VALUE 1
+            value = __builtin_preserve_enum_value(*(typeof(enum foo) *)D, BPF_ENUMVAL_VALUE);
+        "#,
+    }
+    .build()
+    .unwrap();
+    assert_eq!(dbg!(test.run().unwrap()) as i64, -0x7BBBBBBBi64);
+    assert_eq!(dbg!(test.run_no_btf().unwrap()) as i64, -0x7AAAAAAAi64);
```
shaneutt, line 79: Do you think it would be helpful to put these hex values in constants and document them with comments so that future code readers who find their way in here would have some breadcrumbs to help them understand their origin and significance?

alessandrod: Added the following comment:
```
// In the tests below we often use values like 0xAAAAAAAA or -0x7AAAAAAA. Those values have no
// special meaning, they just have "nice" bit patterns that can be helpful while debugging.
```

Issue #516, opened 2023-02-08
What's missing for (pure) xdp.frags support?


brevilo:
The section name `xdp.frags` is currently [not supported](https://github.com/aya-rs/aya/blob/cfa693bc3b4442a8c97cfcd24551ea6439f25e50/aya-obj/src/obj.rs#L163). What's missing to change that? AFAIK, `xdp.frags` does not depend on anything else. Yes, `xdp.frags/cpumap` and `xdp.frags/devmap` probably depend on the new maps themselves being [implemented](https://github.com/aya-rs/aya/pull/292) but `xdp.frags` itself doesn't depend on those, if I'm not mistaken.

Note: XDP multi-buffer (fragment) support is required e.g. when using jumbo frames.

dave-tucker:
Correct. We'd just need to add some logic to translate `xdp.frags` into a an `Xdp` program with the flag `BPF_F_XDP_HAS_FRAGS` set so it gets used at load time.

brevilo:
Thanks, that's what I figured. I might give this a try but it'll take some time as I'm still coming to grips with the Aya internals (and Rust for that matter).

dave-tucker:
No worries! I took care of this in #519 if you want to try it out. It handles the userspace side as well as kernel side.

PR #519, opened 2023-02-09
XDP Multi-Buffer Support

dave-tucker:
Fixes: #516

### DIFF
--- a/aya-bpf-macros/src/expand.rs
+++ b/aya-bpf-macros/src/expand.rs
@@ -200,20 +200,33 @@ impl SkMsg {
 pub struct Xdp {
     item: ItemFn,
     name: Option<String>,
+    frags: bool,
 }
 
 impl Xdp {
     pub fn from_syn(mut args: Args, item: ItemFn) -> Result<Xdp> {
-        let name = name_arg(&mut args)?;
-
-        Ok(Xdp { item, name })
+        let name = pop_arg(&mut args, "name");
+        let mut frags = false;
+        if let Some(s) = pop_arg(&mut args, "frags") {
+            if let Ok(m) = s.parse() {
+                frags = m
+            } else {
+                return Err(Error::new_spanned(
+                    "mutlibuffer",
+                    "invalid value. should be 'true' or 'false'",
+                ));
+            }
+        }
+        err_on_unknown_args(&args)?;
+        Ok(Xdp { item, name, frags })
     }
 
     pub fn expand(&self) -> Result<TokenStream> {
+        let section_prefix = if self.frags { "xdp.frags" } else { "xdp" };
         let section_name = if let Some(name) = &self.name {
-            format!("xdp/{name}")
+            format!("{section_prefix}/{name}")
         } else {
-            "xdp".to_owned()
+            section_prefix.to_string()
         };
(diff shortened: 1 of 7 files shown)


brevilo:
Awesome, that was fast! Looks good from my point of view 👍 

One minor comment: I wonder if `multibuffer` is the ideal attribute name. Yes, technically it's the correct term but devs coming from the original C implementation (docs, examples, kernel) might expect `frags` due to the original program section and symbol names used.

test/integration-ebpf/src/map_test.rs
```diff
@@ -14,7 +14,7 @@ static FOO: Array<u32> = Array::<u32>::with_max_entries(10, 0);
 #[map(name = "BAR")]
 static BAZ: Array<u32> = Array::<u32>::with_max_entries(10, 0);
 
-#[xdp]
+#[xdp(multibuffer = "true")]
```
alessandrod, line 17: I think we shouldn't switch all tests to use frags. It's unlikely but just in case we end up breaking the non-frags case.

dave-tucker: ack. we still have 2 aya xdp programs in the integration tests that will ensure we support the non-frags case too... and any of the C progs used in integration tests are just plain old xdp

aya-bpf-macros/src/lib.rs
```diff
@@ -73,6 +73,17 @@ pub fn xdp(attrs: TokenStream, item: TokenStream) -> TokenStream {
         .into()
 }
 
+#[proc_macro_attribute]
+pub fn xdp_frags(attrs: TokenStream, item: TokenStream) -> TokenStream {
```
alessandrod, line 77: is this meant to be an alias for `#[xdp(multibuffer=true)]`? I don't think we should have two ways to do the same thing

aya-bpf-macros/src/expand.rs
```diff
@@ -200,20 +200,37 @@ impl SkMsg {
...
 }
 
 impl Xdp {
     pub fn from_syn(mut args: Args, item: ItemFn) -> Result<Xdp> {
-        let name = name_arg(&mut args)?;
-
-        Ok(Xdp { item, name })
+        let name = pop_arg(&mut args, "name");
+        let mut multibuffer = false;
+        if let Some(s) = pop_arg(&mut args, "multibuffer") {
```
alessandrod, line 210: I agree with the other comment that this should be called "frags"

alessandrod (commented):
Just a couple of nits, otherwise amazing thanks so much!

aya-bpf-macros/src/lib.rs
dave-tucker: ooops. that's not supposed to be there - nothing to see here lol

aya-bpf-macros/src/expand.rs
dave-tucker: Renamed

dave-tucker:
@alessandrod fixed all comments except for the note re: tests since we've still got good coverage on integration tests for the frags and non-frags cases.

alessandrod (approved):

brevilo:
Thanks guys, you rock! 👏

PR #520, opened 2023-02-10
Add Unsupported Map type

astoycos:
Just because aya doesn't support working with some map types doesn't mean we should't allow them to be loaded.

Add a new `Unsupported` map type to facilitate this, also add a debug log to let the user know we just loaded and unsupported map type.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -13,7 +13,7 @@ use aya_obj::{
     relocation::BpfRelocationError,
     BpfSectionKind, Features,
 };
-use log::debug;
+use log::{debug, warn};
 use thiserror::Error;
 
 use crate::{
@@ -123,6 +123,7 @@ pub struct BpfLoader<'a> {
     max_entries: HashMap<&'a str, u32>,
     extensions: HashSet<&'a str>,
     verifier_log_level: VerifierLogLevel,
+    allow_unsupported_maps: bool,
 }
 
 bitflags! {
@@ -156,6 +157,7 @@ impl<'a> BpfLoader<'a> {
             max_entries: HashMap::new(),
             extensions: HashSet::new(),
             verifier_log_level: VerifierLogLevel::default(),
+            allow_unsupported_maps: false,
         }
     }
 
@@ -178,6 +180,29 @@ impl<'a> BpfLoader<'a> {
     /// ```
     pub fn btf(&mut self, btf: Option<&'a Btf>) -> &mut BpfLoader<'a> {
         self.btf = btf.map(Cow::Borrowed);
+        self
+    }
+
+    /// Allows bytecode containing maps unsupported by Aya to be loaded.
+    ///
+    /// By default, programs containing maps unsupported by Aya will not be loaded.
+    /// This function changes the default behavior, allowing programs to be loaded.
+    /// This should only be used in cases where you do not require access to eBPF
+    /// maps from this crate.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use aya::BpfLoader;
+    ///
+    /// let bpf = BpfLoader::new()
+    ///     .allow_unsupported_maps()
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::BpfError>(())
+    /// ```
+    ///
+    pub fn allow_unsupported_maps(&mut self) -> &mut BpfLoader<'a> {
+        self.allow_unsupported_maps = true;
         self
     }
 
@@ -599,12 +624,22 @@ impl<'a> BpfLoader<'a> {
                 (name, program)
             })
             .collect();
-        let maps: Result<HashMap<String, Map>, BpfError> = maps.drain().map(parse_map).collect();
+        let maps = maps
+            .drain()
+            .map(parse_map)
+            .collect::<Result<HashMap<String, Map>, BpfError>>()?;
 
-        Ok(Bpf {
-            maps: maps?,
-            programs,
-        })
+        if !self.allow_unsupported_maps
+            && maps
+                .iter()
+                .filter(|(_, x)| matches!(x, Map::Unsupported(_)))
+                .count()
+                != 0
+        {
+            return Err(BpfError::UnsupportedMap);
+        }
+
+        Ok(Bpf { maps, programs })
     }
 }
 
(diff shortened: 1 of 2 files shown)


aya/src/maps/mod.rs
```diff
@@ -262,6 +262,8 @@ pub enum Map {
     StackTraceMap(MapData),
     /// A [`Queue`] map
     Queue(MapData),
+    /// A [`Unsupported`] map type
```
dave-tucker, line 265:
```suggestion
    /// An unsupported map type
```

astoycos: Fixed thanks!

dave-tucker (changes requested):
Hey @astoycos thanks for the PR!

I do think we need this BUT I also think that we need to protect users from footguns.
Therefore could we add a new option to `BpfLoader` called `allow_unsupported_maps()` that will enable this?

Then in the normal `load()` path we'd be able to iter the maps list and return a nice error to users. For example:

```
The map ${map name} is of type ${map_type} which is currently unsupported in Aya
```

astoycos:
Yep I agree, let me work on adding this!

astoycos:
Ok I think this should do it @dave-tucker.  

The way I had to re-write the `parse_map` function as an `impl` of `bpfLoader` in order to get access to the new `allow_unsupported_maps` field within it was a bit tricky....so If you want me to redo that just let me know 

Also are there existing tests for these bpfLoader impl functions? I couldn't find any.

aya/src/bpf.rs
```diff
@@ -580,44 +605,58 @@ impl<'a> BpfLoader<'a> {
                 (name, program)
             })
             .collect();
-        let maps: Result<HashMap<String, Map>, BpfError> = maps.drain().map(parse_map).collect();
+        let maps: Result<HashMap<String, Map>, BpfError> =
+            maps.drain().map(self.parse_map()).collect();
```
dave-tucker, line 609: Why not keep parse_maps as it is, and then just check for unsupported maps afterwards.

```suggestion
        let maps: Result<HashMap<String, Map>, BpfError> = maps.drain().map(parse_map).collect();
        if !self.allow_unsupported_maps && maps.iter().filter(|(_, x)| matches!(x, Map::Unsupported(_))).count() != 0 {
            return Err(BpfError::UnsupportedMap)
        }
```

dave-tucker (approved):
LGTM

aya/src/bpf.rs
```diff
@@ -189,6 +191,26 @@ impl<'a> BpfLoader<'a> {
         self
     }
 
+    /// Sets the allow_unsupported_maps field to true.
+    ///
+    /// Programs which have maps that are not supported via the
+    /// aya map api will still be allowed to load.
+    /// # Example
```
dave-tucker, line 198:
```suggestion
    /// Allows bytecode containing maps unsupported by Aya to be loaded.
    ///
    /// By default, programs containing maps unsupported by Aya will not be loaded.
    /// This function changes the default behaviour, allowing programs to be loaded.
    /// This should only be used in cases where you do not require access to eBPF maps from this crate.
    ///
    /// # Example
```

dave-tucker:
@alessandrod for the API changes:
- `BpfLoader::allow_unsupported_maps()`
And the functional change, which is:
- If you attempt to load bytecode which uses an unimplemented map type, you get `BpfError::UnsupportedMap`
- If you set `BpfLoader::allow_unsupported_maps()` then we allow the bytecode to be loaded anyway, even if the map can't be interacted with from the Aya code. This is to support usecases like bpfd, where we're loading code on behalf of others, but aren't interacting with their maps.

aya/src/bpf.rs
astoycos: Done :)

astoycos:
Shoot that won't work and I still can't assign folks here :(

aya/src/bpf.rs
```diff
@@ -189,6 +191,29 @@ impl<'a> BpfLoader<'a> {
         self
     }
 
+    /// Allows bytecode containing maps unsupported by Aya to be loaded.
+    ///
+    /// By default, programs containing maps unsupported by Aya will not be loaded.
+    /// This function changes the default behavior, allowing programs to be loaded.
+    /// This should only be used in cases where you do not require access to eBPF
```
alessandrod, line 198: Allows programs containing unsupported maps to be loaded.

By default programs containing unsupported maps will fail to load. This method
can be used to configure the loader so that unsupported maps will be loaded, but
won't be accessible from userspace. Can be useful when using unsupported
maps that are only accessed from eBPF code and don't require any userspace
interaction.

aya/src/bpf.rs
```diff
@@ -580,12 +605,22 @@ impl<'a> BpfLoader<'a> {
...
+        let maps = maps
+            .drain()
+            .map(parse_map)
+            .collect::<Result<HashMap<String, Map>, BpfError>>()?;
+
+        if !self.allow_unsupported_maps
+            && maps
+                .iter()
+                .filter(|(_, x)| matches!(x, Map::Unsupported(_)))
+                .count()
```
alessandrod, line 617: This should use `any`

aya/src/bpf.rs
```diff
@@ -580,12 +605,22 @@ impl<'a> BpfLoader<'a> {
...
+                != 0
+        {
+            return Err(BpfError::UnsupportedMap);
```
alessandrod, line 620: This should be BpfError::MapError(MapError::Unsupported)

aya/src/bpf.rs
```diff
@@ -876,6 +912,10 @@ pub enum BpfError {
     #[error("program error: {0}")]
     /// A program error
     ProgramError(#[from] ProgramError),
+
+    /// Unsupported Map type
+    #[error("Unsupported map types found")]
+    UnsupportedMap,
```
alessandrod, line 918: This should be `MapError::Unsupported { map_type: u32 }`

alessandrod (changes requested):
Thanks and sorry it took so long to review

alessandrod:
> Shoot that won't work and I still can't assign folks here :(

Just sent you an invite

aya/src/bpf.rs
astoycos: So I ended up changing this logic a bit in order to support `MapError::Unsupported { map_type: u32 }`

More specifically we only search our maps if allow_unsupported_maps is set.

aya/src/bpf.rs
astoycos: Done

aya/src/bpf.rs
astoycos: Done

alessandrod (approved):

PR #522, opened 2023-02-10
Use bpf_link for perf_attach programs (Kprobe/Uprobe/Tracepoint etc...)

dave-tucker:
This adds support for using bpf_link to perf_attach programs.

In addition, it also makes the feature detection a lazy-static, which while not strictly necessary should allow applications to do feature detection only once when using Aya vs. every time a `Bpf` object is created.
This also allows for features to be used in other parts of the codebase that might require feature discovery to have happened already - I think I ran into this in the USDT probe PR which is why I implemented this in the first place!

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -35,8 +35,8 @@ use crate::{
     sys::{
         bpf_load_btf, bpf_map_freeze, bpf_map_update_elem_ptr, is_btf_datasec_supported,
         is_btf_decl_tag_supported, is_btf_float_supported, is_btf_func_global_supported,
-        is_btf_func_supported, is_btf_supported, is_btf_type_tag_supported, is_prog_name_supported,
-        retry_with_verifier_logs,
+        is_btf_func_supported, is_btf_supported, is_btf_type_tag_supported, is_perf_link_supported,
+        is_prog_name_supported, retry_with_verifier_logs,
     },
     util::{bytes_of, possible_cpus, VerifierLog, POSSIBLE_CPUS},
 };
@@ -73,6 +73,7 @@ lazy_static! {
 #[derive(Default, Debug)]
 pub(crate) struct Features {
     pub bpf_name: bool,
+    pub bpf_perf_link: bool,
     pub btf: Option<BtfFeatures>,
 }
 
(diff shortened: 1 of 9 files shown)


aya/src/bpf.rs
```diff
@@ -360,8 +345,8 @@ impl<'a> BpfLoader<'a> {
         let mut obj = Object::parse(data)?;
         obj.patch_map_data(self.globals.clone())?;
 
-        let btf_fd = if let Some(ref btf) = self.features.btf {
-            if let Some(btf) = obj.fixup_and_sanitize_btf(btf)? {
+        let btf_fd = if let Some(ref features) = FEATURES.btf {
```
alessandrod, line 348:
```suggestion
        let btf_fd = if let Some(features) = &FEATURES.btf {
```

aya/src/sys/bpf.rs
```diff
@@ -599,6 +599,37 @@ pub(crate) fn is_prog_name_supported() -> bool {
...
+    let insns = copy_instructions(prog).unwrap();
+    u.insn_cnt = insns.len() as u32;
+    u.insns = insns.as_ptr() as u64;
+    u.prog_type = bpf_prog_type::BPF_PROG_TYPE_TRACEPOINT as u32;
+
+    if let Ok(fd) = sys_bpf(bpf_cmd::BPF_PROG_LOAD, &attr) {
+        if let Err((code, _)) =
+            bpf_link_create(fd as i32, -1, bpf_attach_type::BPF_PERF_EVENT, None, 0)
+        {
+            if code == (-libc::EBADF).into() {
```
alessandrod, line 623: This should be:

// Returns XXX if unsupported. EBADF if supported.
let res = code == (-libc::EBADF).into();
libc::close(fd)

alessandrod (approved):
Great PR! Just a couple of smol nits

aya/src/bpf.rs
dave-tucker: Fixed. Thanks.

aya/src/sys/bpf.rs
dave-tucker: Done!

alessandrod (approved):

PR #524, opened 2023-02-13
aya: Add loaded_programs() API to list all loaded programs

dave-tucker:
This uses a Programs iterator to yield all loaded bpf programs using bpf_prog_get_next_id.

Example program [here](https://github.com/dave-tucker/prog-lister)

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -109,8 +109,8 @@ use crate::{
     pin::PinError,
     sys::{
         bpf_btf_get_fd_by_id, bpf_get_object, bpf_load_program, bpf_pin_object,
-        bpf_prog_get_fd_by_id, bpf_prog_get_info_by_fd, bpf_prog_query, retry_with_verifier_logs,
-        BpfLoadProgramAttrs,
+        bpf_prog_get_fd_by_id, bpf_prog_get_info_by_fd, bpf_prog_get_next_id, bpf_prog_query,
+        retry_with_verifier_logs, BpfLoadProgramAttrs,
     },
     util::VerifierLog,
 };
@@ -912,6 +912,7 @@ impl_try_from_program!(
 );
 
 /// Provides information about a loaded program, like name, id and statistics
+#[derive(Debug)]
 pub struct ProgramInfo(bpf_prog_info);
 
 impl ProgramInfo {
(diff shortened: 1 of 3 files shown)


aya/src/programs/mod.rs
```diff
@@ -971,3 +972,43 @@ impl ProgramInfo {
...
+                        }),
+                )
+            }
+            None => None,
+        }
+    }
+}
+
+/// Returns an iterator over all loaded bpf programs
+pub fn programs() -> Programs {
```
alessandrod, line 1012: I think this should be called `all_loaded_programs` or something to differentiate it from `Bpf::programs`, and the documentation should explain the difference between the two

dave-tucker: I was about to ask you about that! I was thinking maybe re-export this fn at the top level of the crate and call it `iter_programs()`. Better docs + an example/doctest is obviously required - i'll get on that now 😆

dave-tucker:
I went with `iter_programs()` and re-exported at the root of the crate for convenience.
Docs have been added also... and hopefully all missing punctuation has been added 😆

aya/src/programs/mod.rs
```diff
@@ -971,3 +972,81 @@ impl ProgramInfo {
         Ok(ProgramInfo(info))
     }
 }
+
+/// Programs is an Iterator over loaded eBPF programs.
+pub struct Programs {
+    current: u32,
+    done: bool,
```
alessandrod, line 979: I think this should be called `error` instead. You're not setting it when the
syscall returns ENOENT (and there's no need to)

aya/src/programs/mod.rs
```diff
@@ -971,3 +972,81 @@ impl ProgramInfo {
...
+                }))
+            }
+        }
+    }
+}
+
+/// Returns an iterator over all loaded bpf programs.
+///
+/// This differs from [`crate::Bpf::programs`] since it will return all programs
+/// listed on the host system and not only programs loaded by Aya.
```
alessandrod, line 1027:
```suggestion
/// listed on the host system and not only programs loaded by a specific [`crate::Bpf`] instance.
```

aya/src/programs/mod.rs
```diff
@@ -971,3 +972,81 @@ impl ProgramInfo {
...
+/// }
+/// ```
+///
+/// # Errors
+///
+/// Returns [`ProgramError::SyscallError`] if any of the syscalls required to either get
+/// next program id, get the program fd, or the [`ProgramInfo`] fail. In cases where
+/// iteration can't be performed, for example the caller does not have the necessary privileges,
+/// a single item will be yielded containing the error that occurred.
+pub fn iter_programs() -> Programs {
```
alessandrod, line 1047: I think that `loaded_programs` would be a better name for this. I don't think we
need to prepend `iter_` in the same way we don't for `HashMap::keys` etc

alessandrod (commented):
Just a few nits otherwise looks good

dave-tucker:
Nits fixed. Merging

PR #525, opened 2023-02-14
aya: MapData should be Borrow, not AsRef

dave-tucker:
We don't ever do ref-to-ref conversion for MapData so Borrow should suffice.


dave-tucker:
@alessandrod @vadorovsky this doesn't touch API... and the last comment from alessandro on discord was:

> We should have used Borrow to begin with

Could one of you take a look?

alessandrod (approved):
lgtm. (This does change API since it changes public from impls, in case you want to add a label or something)

PR #530, opened 2023-02-23
Added helper module for easier BTF object manipublation

qjerome:
Created a new module under btf module. The goal of this module is to provide higher level APIs to query BTF objects.
For the moment, APIs provide access to:

    search a structure inside BTF and return a helper object
    StructHelper object provides APIs to:
        Get the size of a BTF structure (aligned and unaligned)
        Query MemberHelper objects
    MemberHelper provide APIs to:
        Access the relative offset (from structure) where to find the member
        Get the size of the structure member (aligned and unaligned)

An example scenario of the implemented feature would be to give an easy access to information needed to implement CO-RE (whenever builtin CO-RE is not possible) manually by passing structure(s)/member(s) information down to eBPF.

### DIFF
--- a/aya-obj/src/btf/helpers.rs
+++ b/aya-obj/src/btf/helpers.rs
@@ -1,3 +1,6 @@
+//! Helper module containing higher level APIs to query
+//! BTF objects.
+
 use super::*;
 use alloc::{
     borrow::Cow,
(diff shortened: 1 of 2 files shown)


tamird:
> An example scenario of the implemented feature would be to give an easy access to information needed to implement CO-RE (whenever builtin CO-RE is not possible) manually by passing structure(s)/member(s) information down to eBPF.

Does this PR provide any value until the rest of this functionality is built?

tamird:
Can we close this until such time as we have a use for it?

PR #531, opened 2023-02-23
aya: Make FEATURES public

dave-tucker:
This commit adds a new probe for bpf_attach_cookie, which would be used to implement USDT probes. Since USDT probes aren't currently supported, this triggers a dead_code warning in clippy.

There are cases where exposing FEATURES - our lazy static - is actually helpful to users of the library. For example, they may wish to choose to load a different version of their bytecode based on current features. Or, in the case of an orchestrator like bpfd, we might want to allow users to describe which features their program needs and return nice error message is one or more nodes in their cluster doesn't support the necessary feature set.

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -166,12 +166,68 @@ pub enum BtfError {
 #[derive(Default, Debug)]
 #[allow(missing_docs)]
 pub struct BtfFeatures {
-    pub btf_func: bool,
-    pub btf_func_global: bool,
-    pub btf_datasec: bool,
-    pub btf_float: bool,
-    pub btf_decl_tag: bool,
-    pub btf_type_tag: bool,
+    btf_func: bool,
+    btf_func_global: bool,
+    btf_datasec: bool,
+    btf_float: bool,
+    btf_decl_tag: bool,
+    btf_type_tag: bool,
+}
+
+impl BtfFeatures {
+    #[doc(hidden)]
+    pub fn new(
+        btf_func: bool,
+        btf_func_global: bool,
+        btf_datasec: bool,
+        btf_float: bool,
+        btf_decl_tag: bool,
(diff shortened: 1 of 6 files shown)


alessandrod:
We can't expose the static because it leaks `lazy_static` API (you need to call `.get()` or deref), and we can't expose the `Features` type as it is as it would be the ultimate API breaking type: every new feature breaks API.

What we can do is to add a `features() -> &'static Features` util function somewhere, and then add pub method accessors for all the still private fields in `Features`. That would allow us to add features without breaking API.

alessandrod (changes requested):
See other comment

aya-obj/src/btf/btf.rs
```diff
@@ -166,12 +166,68 @@ pub enum BtfError {
...
+    btf_func_global: bool,
+    btf_datasec: bool,
+    btf_float: bool,
+    btf_decl_tag: bool,
+    btf_type_tag: bool,
+}
+
+impl BtfFeatures {
+    /// Creates a new BtfFeatures struct.
+    pub fn new(
```
alessandrod, line 179: I think this should probably be #[doc(hidden)]

aya-obj/src/obj.rs
```diff
@@ -42,11 +42,63 @@ const KERNEL_VERSION_ANY: u32 = 0xFFFF_FFFE;
...
+    bpf_probe_read_kernel: bool,
+    bpf_perf_link: bool,
+    bpf_global_data: bool,
+    bpf_cookie: bool,
+    btf: Option<BtfFeatures>,
+}
+
+impl Features {
+    /// Creates a new Features struct.
+    pub fn new(
```
alessandrod, line 55: And this too #[doc(hidden)]

alessandrod (approved):
Thanks! See doc nits, then it's ready to go

PR #539, opened 2023-03-10
aya: Fix MapData Clone implementation

marysaka:
The Clone implementation of MapData was previously not storing the result of the dup operation.


vadorovsky (approved):

marysaka:
Addressed lint errors.

alessandrod (approved):
Oops, nice catch and thanks for the fix!

PR #545, opened 2023-03-17
Add `BPF_F_SLEEPABLE` flag support to LSM

epompeii:
This adds support for the `BPF_F_SLEEPABLE` flag for LSM : https://lore.kernel.org/netdev/20200827220114.69225-3-alexei.starovoitov@gmail.com/

It follows the same tack as XDP multi-buffer support: https://github.com/aya-rs/aya/pull/519

Internally, this option follows the `libbpf` convention and is represented as `lsm.s`: https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/libbpf.c#L8560

In the public API, this is exposed via the eBPF `lsm` macro as an optional `sleepable = "bool"` argument.

```
#[lsm(name = "lsm_bprm_check_security", sleepable = "true")]
pub fn bprm_check_security(ctx: LsmContext) -> i32 {
    match try_bprm_check_security(ctx) {
        Ok(ret) => ret,
        Err(ret) => ret,
    }
}
```

### DIFF
--- a/aya-bpf-macros/src/expand.rs
+++ b/aya-bpf-macros/src/expand.rs
@@ -613,18 +613,39 @@ impl RawTracePoint {
 
 pub struct Lsm {
     item: ItemFn,
-    name: String,
+    name: Option<String>,
+    sleepable: bool,
 }
 
 impl Lsm {
     pub fn from_syn(mut args: Args, item: ItemFn) -> Result<Lsm> {
-        let name = name_arg(&mut args)?.unwrap_or_else(|| item.sig.ident.to_string());
-
-        Ok(Lsm { item, name })
+        let name = pop_arg(&mut args, "name");
+        let mut sleepable = false;
+        if let Some(s) = pop_arg(&mut args, "sleepable") {
+            if let Ok(m) = s.parse() {
+                sleepable = m
+            } else {
+                return Err(Error::new_spanned(
+                    "mutlibuffer",
+                    "invalid value. should be 'true' or 'false'",
+                ));
+            }
+        }
+        err_on_unknown_args(&args)?;
+        Ok(Lsm {
+            item,
+            name,
+            sleepable,
+        })
     }
 
     pub fn expand(&self) -> Result<TokenStream> {
-        let section_name = format!("lsm/{}", self.name);
+        let section_prefix = if self.sleepable { "lsm.s" } else { "lsm" };
+        let section_name = if let Some(name) = &self.name {
+            format!("{section_prefix}/{name}")
+        } else {
+            section_prefix.to_string()
+        };
         let fn_vis = &self.item.vis;
         let fn_name = &self.item.sig.ident;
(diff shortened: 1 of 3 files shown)


aya-bpf-macros/src/expand.rs
```diff
@@ -613,18 +613,39 @@ impl RawTracePoint {
...
-
-        Ok(Lsm { item, name })
+        let name = pop_arg(&mut args, "name");
+        let mut sleepable = false;
+        if let Some(s) = pop_arg(&mut args, "sleepable") {
+            if let Ok(m) = s.parse() {
+                sleepable = m
+            } else {
+                return Err(Error::new_spanned(
+                    "mutlibuffer",
```
alessandrod, line 629: bad copypasta, including typo

aya-bpf-macros/src/expand.rs
```diff
@@ -613,18 +613,39 @@ impl RawTracePoint {
...
 }
 
 impl Lsm {
     pub fn from_syn(mut args: Args, item: ItemFn) -> Result<Lsm> {
-        let name = name_arg(&mut args)?.unwrap_or_else(|| item.sig.ident.to_string());
-
-        Ok(Lsm { item, name })
+        let name = pop_arg(&mut args, "name");
+        let mut sleepable = false;
+        if let Some(s) = pop_arg(&mut args, "sleepable") {
```
alessandrod, line 624: I think this should be a boolean argument. I'd be ok leaving this as is for now
and then having a followup PR fix it to be a boolean both here and in the xdp macro

aya-obj/src/obj.rs
```diff
@@ -231,6 +231,7 @@ pub enum ProgramSection {
     },
     Lsm {
         name: String,
+        sleepable_supported: bool,
```
alessandrod, line 234: I think this should be just `sleepable`?

aya-obj/src/obj.rs
```diff
@@ -1926,6 +1934,30 @@ mod tests {
...
+                "lsm.s/foo",
+                bytes_of(&fake_ins())
+            )),
+            Ok(())
+        );
+        assert_matches!(
+            obj.programs.get("foo"),
+            Some(Program {
+                section: ProgramSection::Lsm {
+                    sleepable_supported: true,
```
alessandrod, line 1953: In the test above (without .s), you should check that sleepable: false

alessandrod (changes requested):
Thanks for doing this! See comments

aya-obj/src/obj.rs
epompeii: I totally agree. I was trying to keep it consistent with XDP which uses `frags_supported`.

Maybe change both?

alessandrod (approved):
Thank you! Lgtm. Can  you please rebase? I've fixed the CI failure on main.

PR #548, opened 2023-03-20
aya: Allow to attach XDP probe by interface index

kriomant:


aya/src/programs/xdp.rs
```diff
@@ -98,14 +98,29 @@ impl Xdp {
...
+        let if_index = unsafe { if_nametoindex(c_interface.as_ptr()) };
         if if_index == 0 {
             return Err(ProgramError::UnknownInterface {
                 name: interface.to_string(),
             });
         }
+        self.attach_by_ifindex(if_index, flags)
+    }
+
+    /// Attaches the program to the given `interface`.
```
alessandrod, line 111:
```suggestion
    /// Attaches the program to the given interface index.
```

aya/src/programs/xdp.rs
```diff
@@ -98,14 +98,29 @@ impl Xdp {
...
+    ///
+    /// The returned value can be used to detach, see [Xdp::detach].
+    ///
+    /// # Errors
+    ///
+    /// When attaching fails, [`ProgramError::SyscallError`] is returned for
+    /// kernels `>= 5.9.0`, and instead
+    /// [`XdpError::NetlinkError`] is returned for older
+    /// kernels.
+    pub fn attach_by_ifindex(&mut self, if_index: libc::c_uint, flags: XdpFlags) -> Result<XdpLinkId, ProgramError> {
```
alessandrod, line 121:
```suggestion
    pub fn attach_to_if_index(&mut self, if_index: u32, flags: XdpFlags) -> Result<XdpLinkId, ProgramError> {
```

I think this should be called attach_to_if_index for consistency with other attach_to_* methods, and we shouldn't re-export libc types but use u32

alessandrod (commented):
Thanks! Having this makes sense, see minor nits

aya/src/programs/xdp.rs
```diff
@@ -98,14 +98,29 @@ impl Xdp {
...
+    ///
+    /// The returned value can be used to detach, see [Xdp::detach].
+    ///
+    /// # Errors
+    ///
+    /// When attaching fails, [`ProgramError::SyscallError`] is returned for
+    /// kernels `>= 5.9.0`, and instead
+    /// [`XdpError::NetlinkError`] is returned for older
+    /// kernels.
+    pub fn attach_to_if_index(&mut self, if_index: libc::c_uint, flags: XdpFlags) -> Result<XdpLinkId, ProgramError> {
```
alessandrod, line 121: if_index should be u32, so that we don't leak libc API to our crate

alessandrod (approved):

Issue #552, opened 2023-03-21
Bug in bpf_probe_read_.*_str_bytes helpers


qjerome:
It seems the returned buffer includes NULL byte, while the documentation claims it does not.
```rust
pub unsafe fn bpf_probe_read_user_str_bytes(
    src: *const u8,
    dest: &mut [u8],
) -> Result<&[u8], c_long> {
    let len = gen::bpf_probe_read_user_str(
        dest.as_mut_ptr() as *mut c_void,
        dest.len() as u32,
        src as *const c_void,
    );
    if len < 0 {
        return Err(-1);
    }

    let len = len as usize;
    if len >= dest.len() {
        // this can never happen, it's needed to tell the verifier that len is
        // bounded
        return Err(-1);
    }

    Ok(&dest[..len]) // ERROR HERE !!!! the len returned by bpf_probe_read_user_str includes NULL byte
}
```

qjerome:
It seems same error has been propagated to `bpf_probe_read_kernel_str_bytes`

qjerome:
On success bpf_probe_read_str returns a **strictly positive length of the string** (according to documentation) so it should be safe just to return Ok(&dest[..len-1])

alessandrod:
Nice catch! Want to make a PR for it?

Where we currently do 

```
    if len >= dest.len() {
        // this can never happen, it's needed to tell the verifier that len is
        // bounded
        return Err(-1);
    }
```

We'll also have to check for `len == 0`. The documentation says it returns a strictly positive length but the verifier doesn't know that.

qjerome:
I can do a PR for it for sure however I think it will be mixed up with my current fork pending PR (https://github.com/aya-rs/aya/pull/530). Is it possible to make a different PR without doing another fork of the project (sorry if that's a silly question I've never been in such a situation) ? You can ping me on the Discord channel if you want to help me out with that aspect.

PR #554, opened 2023-03-21
fix issue #552

qjerome:
Fixing issue #552
The fix has been tested in a previously (not working as intended) probe. So the fix passes the verifier.


alessandrod (approved):
Thank you!

PR #555, opened 2023-03-23
aya: Add as_bytes() to Pod trait

dave-tucker:
This adds as_bytes() to the Pod trait with a default implementation. The set_global API changes to use a Vec<u8> (and the hashmap it inserts into is also a Vec<u8>).
As such, users may either pass raw bytes they have constructed themselves - i.e `vec![1u8, 2u8, 3u8]`
Or any `Pod` type by calling `.as_bytes` on it.

```rust
use aya::BpfLoader;
use aya::{BpfLoader, Pod};

#[repr(C)]
#[derive(Copy,Clone)]
struct MyThing {
    foo: u64
}

impl Pod for MyThing {}

let thing1 = MyThing{foo:42};
let bpf = BpfLoader::new()
    .set_global("VERSION", unsafe { 2.as_bytes() })
    .set_global("MY_THING", unsafe { thing1.as_bytes() })
    .set_global("REVISION", vec![1u8])
    .load_file("file.o")?;
```

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -12,6 +12,7 @@ use aya_obj::{
     generated::BPF_F_XDP_HAS_FRAGS,
     relocation::BpfRelocationError,
 };
+use bytes::BufMut;
 use log::debug;
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


anfredette:
I tested this change in [bpfd](https://github.com/redhat-et/bpfd) including our existing use of set_global() as well as new support for setting it in the programs we load as seen in the above commit.  It works well.

aya/src/bpf.rs
```diff
@@ -234,18 +254,16 @@ impl<'a> BpfLoader<'a> {
     /// # Example
```
anfredette, line 254: Personally, I think it would be helpful to show the examples from your commit message for thing1 and vec![1u8] because I think these are likely to be real-world use cases, but I usually make my examples more complicated than people want, so I'll leave it up to the rest of you.

anfredette (approved):
lgtm

alessandrod:
> Since values are now moved into `set_global`, we get around lifetime issues that are introduced with the Pod trait being `static`. The function owns the data coming in, stores it in a hashmap, and applies it at map creation time. Subsequent changes to `thing1` are not allowed since it was moved.

I think this is wrong. Conceptually, there's no reason for BpfLoader to own the data and the data is already guaranteed to be immutable since it's borrowed as an immutable reference.

This PR makes the data owned to workaround an issue somewhere else in the API: the fact that Pod is 'static. This seems wrong for two reasons: it leaks an implementation wart of another bit of the API, and it forces all set_global() callers to do a new allocation.

PR #558, opened 2023-03-25
aya: make it possible to use set_global() with slices of Pod(s)

alessandrod:
Alternative to https://github.com/aya-rs/aya/pull/555

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -38,7 +38,7 @@ use crate::{
         is_btf_func_supported, is_btf_supported, is_btf_type_tag_supported, is_perf_link_supported,
         is_prog_name_supported, retry_with_verifier_logs,
     },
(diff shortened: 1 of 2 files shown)


alessandrod:
@dave-tucker @anfredette I played with this a bit and think that this is the best API. It makes set_global work with both Pod and &[Pod] without changing the Pod trait nor introducing allocations. I don't love having to introduce a new type but because of `Into` you don't actually ever have to use it directly.

dave-tucker (approved):
LGTM, but let's hold of merging until @anfredette tries this out in bpfd.

anfredette (approved):
@alessandrod, thanks for the help with this!  It looks good to me.

I also updated https://github.com/redhat-et/bpfd/pull/371 to use this new API and it's working fine.

PR #560, opened 2023-03-27
Implement FdLink conversions

astoycos:
Implement TryFrom functions to convert Fdlink to PerfLink/TracePointLink/ KprobeLink, and UprobeLink and vice-versa.

This allows us to pin taken links for perf programs, ultimately ensuring the link isn't dropped when the loading process exits.

Required for https://github.com/redhat-et/bpfd/issues/306

TODO: It'd be nice to add some documentation around this API somewhere...

```
    let mut bpf = BpfLoader::new()
    // load pinned maps from /sys/fs/bpf/my-program
    .map_pin_path("/sys/fs/bpf")
    // finally load the code
    .load_file("/home/astoycos/go/src/github.com/redhat-et/bpfd/examples/go-tracepoint-counter/bpf_bpfel.o")?;
    let program: &mut TracePoint = bpf.program_mut("tracepoint_kill_recorder").unwrap().try_into()?;
    program.load()?;
    let link_id = program.attach("syscalls", "sys_enter_kill")?;
    program.pin("/sys/fs/bpf/tracepoint_kill_recorder_pin")?;
    let owned_link: TracePointLink = program.take_link(link_id)?;
    let fd_link: FdLink = owned_link.try_into()?;
    fd_link.pin("/sys/fs/bpf/tracepoint_kill_recorder_link")?;
```

### DIFF
--- a/aya/src/programs/kprobe.rs
+++ b/aya/src/programs/kprobe.rs
@@ -3,13 +3,14 @@ use std::{io, path::Path};
 use thiserror::Error;
 
 use crate::{
-    generated::bpf_prog_type::BPF_PROG_TYPE_KPROBE,
+    generated::{bpf_link_type, bpf_prog_type::BPF_PROG_TYPE_KPROBE},
     programs::{
         define_link_wrapper, load_program,
         perf_attach::{PerfLinkIdInner, PerfLinkInner},
         probe::{attach, ProbeKind},
-        ProgramData, ProgramError,
+        FdLink, LinkError, ProgramData, ProgramError,
     },
+    sys::bpf_link_get_info_by_fd,
     VerifierLogLevel,
 };
 
@@ -118,4 +119,33 @@ pub enum KProbeError {
         #[source]
         io_error: io::Error,
     },
+}
+
+impl TryFrom<KProbeLink> for FdLink {
+    type Error = LinkError;
+
+    fn try_from(value: KProbeLink) -> Result<Self, Self::Error> {
+        if let PerfLinkInner::FdLink(fd) = value.into_inner() {
+            Ok(fd)
+        } else {
+            Err(LinkError::InvalidLink)
+        }
+    }
+}
+
+impl TryFrom<FdLink> for KProbeLink {
+    type Error = LinkError;
+
+    fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
+        let info =
+            bpf_link_get_info_by_fd(fd_link.fd).map_err(|io_error| LinkError::SyscallError {
+                call: "BPF_OBJ_GET_INFO_BY_FD",
+                code: 0,
+                io_error,
+            })?;
+        if info.type_ == (bpf_link_type::BPF_LINK_TYPE_KPROBE_MULTI as u32) {
+            return Ok(KProbeLink::new(PerfLinkInner::FdLink(fd_link)));
+        }
+        Err(LinkError::InvalidLink)
+    }
 }
--- a/aya/src/programs/links.rs
+++ b/aya/src/programs/links.rs
@@ -17,6 +17,10 @@ use crate::{
     sys::{bpf_get_object, bpf_pin_object, bpf_prog_detach},
 };
 
+// for docs link
+#[allow(unused)]
+use crate::programs::cgroup_skb::CgroupSkb;
+
 /// A Link.
 pub trait Link: std::fmt::Debug + 'static {
     /// Unique Id
@@ -82,6 +86,30 @@ impl<T: Link> Drop for LinkMap<T> {
 pub struct FdLinkId(pub(crate) RawFd);
 
 /// A file descriptor link.
+///
+/// Fd links are returned directly when attaching some program types (for
+/// instance [`CgroupSkb`]), or can be obtained by converting other link
+/// types (see the `TryFrom` implementations).
+///
+/// An important property of fd links is that they can be pinned. Pinning
+/// can be used keep a link attached "in background" even after the program
+/// that has created the link terminates.
+///
+/// # Example
+///
+///```no_run
+/// # let mut bpf = Bpf::load_file("ebpf_programs.o")?;
+/// use aya::{Bpf, programs::{links::FdLink, KProbe}};
+///
+/// let program: &mut KProbe = bpf.program_mut("intercept_wakeups").unwrap().try_into()?;
+/// program.load()?;
+/// let link_id = program.attach("try_to_wake_up", 0)?;
+/// let link = program.take_link(link_id).unwrap();
+/// let fd_link: FdLink = link.try_into().unwrap();
+/// fd_link.pin("/sys/fs/bpf/intercept_wakeups_link").unwrap();
+///
+/// # Ok::<(), aya::BpfError>(())
+/// ```
 #[derive(Debug)]
 pub struct FdLink {
     pub(crate) fd: RawFd,
--- a/aya/src/programs/perf_event.rs
+++ b/aya/src/programs/perf_event.rs
@@ -6,6 +6,7 @@ pub use crate::generated::{
 
 use crate::{
     generated::{
+        bpf_link_type,
         bpf_prog_type::BPF_PROG_TYPE_PERF_EVENT,
         perf_type_id::{
             PERF_TYPE_BREAKPOINT, PERF_TYPE_HARDWARE, PERF_TYPE_HW_CACHE, PERF_TYPE_RAW,
@@ -16,9 +17,9 @@ use crate::{
         links::define_link_wrapper,
         load_program, perf_attach,
         perf_attach::{PerfLinkIdInner, PerfLinkInner},
-        ProgramData, ProgramError,
+        FdLink, LinkError, ProgramData, ProgramError,
     },
-    sys::perf_event_open,
+    sys::{bpf_link_get_info_by_fd, perf_event_open},
 };
 
 /// The type of perf event
@@ -186,6 +187,35 @@ impl PerfEvent {
     /// for managing its lifetime.
     pub fn take_link(&mut self, link_id: PerfEventLinkId) -> Result<PerfEventLink, ProgramError> {
         self.data.take_link(link_id)
+    }
+}
+
+impl TryFrom<PerfEventLink> for FdLink {
+    type Error = LinkError;
+
+    fn try_from(value: PerfEventLink) -> Result<Self, Self::Error> {
+        if let PerfLinkInner::FdLink(fd) = value.into_inner() {
+            Ok(fd)
+        } else {
+            Err(LinkError::InvalidLink)
+        }
+    }
+}
+
+impl TryFrom<FdLink> for PerfEventLink {
+    type Error = LinkError;
+
+    fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
+        let info =
+            bpf_link_get_info_by_fd(fd_link.fd).map_err(|io_error| LinkError::SyscallError {
+                call: "BPF_OBJ_GET_INFO_BY_FD",
+                code: 0,
+                io_error,
+            })?;
+        if info.type_ == (bpf_link_type::BPF_LINK_TYPE_PERF_EVENT as u32) {
+            return Ok(PerfEventLink::new(PerfLinkInner::FdLink(fd_link)));
+        }
+        Err(LinkError::InvalidLink)
     }
 }
 
--- a/aya/src/programs/trace_point.rs
+++ b/aya/src/programs/trace_point.rs
@@ -3,14 +3,14 @@ use std::{fs, io, path::Path};
 use thiserror::Error;
 
 use crate::{
-    generated::bpf_prog_type::BPF_PROG_TYPE_TRACEPOINT,
+    generated::{bpf_link_type, bpf_prog_type::BPF_PROG_TYPE_TRACEPOINT},
     programs::{
         define_link_wrapper, load_program,
         perf_attach::{perf_attach, PerfLinkIdInner, PerfLinkInner},
         utils::find_tracefs_path,
-        ProgramData, ProgramError,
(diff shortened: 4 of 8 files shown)


astoycos:
Hrm I don't know if this error is related: 

```
Error: e: 19:33:45 [ERROR] llvm: Unknown attribute kind (86) (Producer: 'LLVM16.0.0-rust-1.70.0-nightly' Reader: 'LLVM 15.0.7')
          error: failure linking module /home/fedora/aya/test/integration-ebpf/../../target/bpfel-unknown-none/debug/deps/map_test-a9de5ef8344e854d.3hyxcigylsh430vj.rcgu.o
          19:33:45 [ WARN] ignoring file "/tmp/rustcIDSCW0/symbols.o": no embedded bitcode
          

error: could not compile `integration-ebpf` (bin "map_test") due to previous error
```

aya/src/programs/kprobe.rs
```diff
@@ -118,3 +119,33 @@ pub enum KProbeError {
...
+
+    fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
+        // unwrap of fd_link.fd will not panic since it's only None when being dropped.
+        let info =
+            bpf_link_get_info_by_fd(fd_link.fd).map_err(|io_error| LinkError::SyscallError {
+                call: "BPF_OBJ_GET_INFO_BY_FD".to_string(),
+                code: 0,
+                io_error,
+            })?;
+        if info.type_ == (bpf_link_type::BPF_LINK_TYPE_XDP as u32) {
```
alessandrod, line 146: here and everywhere else: bad copypasta, you need to check the correct type

alessandrod (changes requested):
Thanks! 

I agree that this should be documented. We could add docs here https://docs.aya-rs.dev/user/aya/programs/links/struct.fdlink? Explain that FdLink allows you to pin, and that many links can be converted to FdLink by using either into() or try_into(), giving examples of both.

aya/src/programs/kprobe.rs
alessandrod: also, this clearly shows that we could use integration tests for this :P

astoycos: Done!

astoycos:
Sorry It's been so long @alessandrod I fixed the issues and added integration testing for `TracePoint`, `Uprobe` and `Kprobe` programs.  I'm happy to do it for perf_event progs here or in a follow up I'm just not 100% sure how to ATM and this is what I got to today.

@dave-tucker PTAL at how I also changed the integration test tooling to verify that a program is loaded AND linked in certain cases.

astoycos:
>  We could add docs here https://docs.aya-rs.dev/user/aya/programs/links/struct.fdlink? Explain that FdLink allows you to pin, and that many links can be converted to FdLink by using either into() or try_into(), giving examples of both.

Shoot and I still need to do this :)

astoycos:
> Shoot and I still need to do this.

done

aya/src/programs/links.rs
```diff
@@ -81,7 +81,12 @@ impl<T: Link> Drop for LinkMap<T> {
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub struct FdLinkId(pub(crate) RawFd);
 
-/// A file descriptor link.
+/// FdLink is a generic file descriptor link.
+///
+/// FdLinks can be pinned in order to keep probes attached and running after the loading
+/// process closes. Additionally many program specific links can be converted into
+/// FdLinks by using either into() or try_into() implementations. For examples of such functions
+/// see [`TracePointLink::try_from()`](crate::programs::trace_point::TracePointLink::try_from()) and [`FdLink::try_from()`].
```
astoycos, line 89: I was trying to link directly to the TracePoint Try_From implementation but `crate::programs::trace_point::FdLink::try_from()` won't work for some reason

astoycos:
Poke @alessandrod :)

astoycos:
Rebased 👍

ajwerner:
@tamird and I are working on redoing the integration test infrastructure such that the tests themselves do not rely on userspace to execute. Today the cycle time of compiling and running tests in the VM is less than wonderful. I chatted with @alessandrod about it [here](https://discord.com/channels/855676609003651072/1126903316497059941/1126908645607755796). 

The preferred solution would be to have aya provide the relevant APIs for reading these links. If it's easy to implement the functionality you need in aya as opposed to shelling out, would you be open to it?

astoycos:
Hiya @ajwerner and yeah I totally agree, we need a `loaded_links()` API for [bpfd](https://bpfd.netlify.app/) regardless to go along with the work I did here: https://github.com/aya-rs/aya/pull/637

I'd rather go ahead and get this bit in for now since I've been carrying it for quite some time 😄 , but in the meantime I'll start working on a basic `loaded_links()` API.

aya/src/programs/links.rs
```diff
@@ -81,7 +81,12 @@ impl<T: Link> Drop for LinkMap<T> {
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub struct FdLinkId(pub(crate) RawFd);
 
-/// A file descriptor link.
+/// FdLink is a generic file descriptor link.
```
alessandrod, line 84: "A file descriptor link." seems better to me for consistency with the rest of the docs. We don't repeat the type name, eg:

```
// A kernel probe.
...
struct KProbe {
```

I'd say:

```
/// A file descriptor link.
/// 
/// Fd links are returned directly when attaching some program types (for
/// instance [`CgroupSkb`]), or can be obtained by converting other link
/// types (see the `TryFrom` implementations).
/// 
/// An important property of fd links is that they can be pinned. Pinning
/// can be used keep a link attached "in background" even after the program
/// that has created the link terminates.
///
/// # Example
/// ...
```

tamird (changes requested):
> Reviewed 7 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/560)* status: 7 of 8 files reviewed, 8 unresolved discussions (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`aya/src/programs/kprobe.rs` line 140 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9ts3v9fmUiwDmd6vb:-N_9ts3v9fmUiwDmd6vc:b99qyho) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/aya/src/programs/kprobe.rs#L140)):*
> ```XML
> 
>     fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
>         // unwrap of fd_link.fd will not panic since it's only None when being dropped.
> ```

there's no unwrap here?
___
*[`aya/src/programs/links.rs` line 89 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9u6Ff2z-tkq6dfRBn:-N_9u6Ff2z-tkq6dfRBo:bsrp4qm) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/aya/src/programs/links.rs#L89)):*
> ```XML
> /// process closes. Additionally many program specific links can be converted into
> /// FdLinks by using either into() or try_into() implementations. For examples of such functions
> /// see [`TracePointLink::try_from()`](crate::programs::trace_point::TracePointLink::try_from()) and [`FdLink::try_from()`].
> ```

nit: long line
___
*[`aya/src/programs/perf_event.rs` line 209 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9u8FX9uBoU8emkn26:-N_9u8FX9uBoU8emkn27:b-rqxvnq) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/aya/src/programs/perf_event.rs#L209)):*
> ```XML
> 
>     fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
>         // unwrap of fd_link.fd will not panic since it's only None when being dropped.
> ```

ditto?
___
*[`aya/src/programs/trace_point.rs` line 135 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9u9kM2VLyj3MvL_d4:-N_9u9kM2VLyj3MvL_d5:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/aya/src/programs/trace_point.rs#L135)):*
> ```XML
> 
>     fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
>         // unwrap of fd_link.fd will not panic since it's only None when being dropped.
> ```

ditto
___
*[`aya/src/programs/uprobe.rs` line 180 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9uBE49Yzcj3N4t22M:-N_9uBE49Yzcj3N4t22N:bb7hf91) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/aya/src/programs/uprobe.rs#L180)):*
> ```XML
> 
>     fn try_from(fd_link: FdLink) -> Result<Self, Self::Error> {
>         // unwrap of fd_link.fd will not panic since it's only None when being dropped.
> ```

here
___
*[`test/integration-test/tests/load.rs` line 67 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9uLUY3XdoiQCgng73:-N_9uLUY3XdoiQCgng74:b-xi4d8s) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/test/integration-test/tests/load.rs#L67)):*
> ```XML
> }
> 
> macro_rules! assert_loaded_and_linked {
> ```

can this just be `assert_loaded` followed by assert linked? i don't think we need to retry the bpftool invocation separately from waiting for loadedness

test/integration-test/tests/load.rs
```diff
@@ -73,60 +110,108 @@ macro_rules! assert_loaded {
 fn unload_xdp() {
     let bytes = include_bytes_aligned!("../../../target/bpfel-unknown-none/release/test");
     let mut bpf = Bpf::load(bytes).unwrap();
-    let prog: &mut Xdp = bpf
-        .program_mut("test_unload_xdp")
```
alessandrod, line 77: why rename this? truncation? if so, how did it work before?

tamird (changes requested):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/560)* status: all files reviewed, 9 unresolved discussions (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`test/integration-ebpf/src/test.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/560#-N_9v6a71sjetOHImM8A:-N_9v6a71sjetOHImM8B:b2gilff) ([raw file](https://github.com/aya-rs/aya/blob/8eef908f97359cd2ccc4d644b5efba2db2125a6f/test/integration-ebpf/src/test.rs#L23)):*
> ```XML
> 
> #[kprobe]
> // truncated name to match bpftool output
> ```

the names were previously truncated to 15 characters, this comment is detritus

test/integration-test/tests/load.rs
astoycos: https://discord.com/channels/855676609003651072/855676609003651075/1128725406313353216 :)

aya/src/programs/links.rs
astoycos: Done!

astoycos:
> can this just be assert_loaded followed by assert linked? i don't think we need to retry the bpftool invocation separately from waiting for loadedness

This is the only thing I didn't get to, simply because these functions are built as macros and if we were to reuse them I'd need to return the program id from `assert_loaded()`.  This will change anyways with #645 so I think it's alright for now.

test/integration-test/tests/load.rs
```diff
@@ -94,7 +95,10 @@ macro_rules! assert_loaded_and_linked {
 macro_rules! assert_loaded {
     ($name:literal, $loaded:expr) => {
         for i in 0..(MAX_RETRIES + 1) {
-            let state = loaded_programs().any(|prog| prog.unwrap().name() == $name.as_bytes());
+            let state = loaded_programs()
+                .filter(|prog| prog.is_ok())
```
tamird, line 99: instead of filter->unwrap can you use filter_map(prog.ok()) and add a comment?

tamird (approved):
:lgtm:

> Reviewed 8 of 8 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/560#-:-N_AU-b57nicaF3CbbCp:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`test/integration-test/tests/load.rs` line 59 at r3](https://reviewable.io/reviews/aya-rs/aya/560#-N_ATm1iB6Bx_YpwsKU9:-N_ATm1iB6Bx_YpwsKUA:bito7gb) ([raw file](https://github.com/aya-rs/aya/blob/9ce22837ed3e5c35c14781a39b80de1108789e74/test/integration-test/tests/load.rs#L59)):*
> ```XML
> fn is_linked(prog_id: &u32) -> bool {
>     let output = Command::new("bpftool").args(["link"]).output();
>     let output = match output {
> ```

i think this is output.expect(...)

test/integration-test/tests/load.rs
astoycos: Done

tamird (approved):
> Reviewed 1 of 1 files at r5, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/560)* status: all files reviewed, 3 unresolved discussions (waiting on @alessandrod and @dave-tucker)

alessandrod (approved):

astoycos:
Whelp ended up having to rebase again :/

PR #527, opened 2023-02-20
Implement XDP map types

Tuetuopay:
This PR is an update of @dave-tucker’s #292 updated for the current Aya main.

It adds implementations for:
- xskmap
- cpumap
- devmap
- devmap_hash

All of those can be used to redirect XDP packets to various notations. Note that the userspace part (umem ans socket) of xskmap (i.e. AF_XDP) is not handled by this PR and requires other libraries (like libbpf, xsk-rs, …). For this, see #507.

All map types have been tested working.

TODO:
- [x] integration tests
- [x] ~add bindings to `struct xdp_sock` to implement `aya::maps::XskMap::get`~
  Cancelled: contrary to what [the kernel documentation](https://docs.kernel.org/bpf/map_xskmap.html#id1) say, `XskMap` does not support element lookup from the userspace (https://elixir.bootlin.com/linux/v6.2/source/net/xdp/xskmap.c#L148).
- [x] frags support (`xdp.frags/devmap` et al)

Closes: #193 #194 #195 #199

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/527)

PR #561, opened 2023-03-29
replace lo interface with dummy interface

murex971:
Fixes: #422

### DIFF
--- a/test/integration-test/src/tests/load.rs
+++ b/test/integration-test/src/tests/load.rs
@@ -11,7 +11,7 @@ use aya::{
 };
 use log::warn;
 
-use crate::tests::kernel_version;
+use crate::tests::{kernel_version, util::DummyInterface};
 
 use super::{integration_test, IntegrationTest};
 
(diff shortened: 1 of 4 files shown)


Tuetuopay:
Hi!

Just to let you know, #527 contains another fix fo #422. It still keeps using loopbacks, but runs tests in dedicated namespaces. (See https://github.com/aya-rs/aya/pull/527/commits/9da1dfecd977ed86199fca5d80ac6d35dbce443f)

Using loopbacks has advantages compared to dummy interfaces: they loop back the traffic. In other words, any packet transmitted on the loopback is recieved on the same loopback. This is very useful for testing, especially XDP which only works in rx.

To get the same feature level, you’d need veth pairs. However that’s another can of worms, as IP configuration is required, and configuration that does not conflicts with the host machine. You could use vrfs or netns, but it’s gonna be complicated.

So yeah, for the current tests where we don’t test at all by sending packets it’s fine. But as soon as tests start to do it, dummy won’t cut it.

PR #563, opened 2023-03-31
aya-obj: Make relocations less strict

### LINKED ISSUE #533: Relocations are too strict
I'm trying to make use of struct "flavors" but this cannot work with aya current design.

The following eBPF doesn't load with aya (full reproducer [here](https://github.com/marysaka/aya_struct_flavor_reloc_repro)):
```c
#include "vmlinux.h"
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_tracing.h>
#include <bpf/bpf_core_read.h>

struct task_struct_rh___redhat {
    u64 self_exec_id;
};

struct task_struct___redhat {
    struct task_struct_rh___redhat *task_struct_rh;
};

#define printk(fmt, ...) ({                                    \
        char __fmt[] = fmt;                                    \
        bpf_trace_printk(__fmt, sizeof(__fmt), ##__VA_ARGS__); \
})

SEC("tp/bpf_prog")
int
bpf_prog(
    void *ctx
) {
    struct task_struct *task;
    u64 exec_id;

    task = (struct task_struct *)bpf_get_current_task();
    if (!task) {
        return 0;
    }

    // On CentOS 7 kernel, self_exec_id is part of task_struct_rh.
    if (bpf_core_field_exists(((struct task_struct___redhat *)task)->task_struct_rh)) {
        exec_id = BPF_CORE_READ((struct task_struct___redhat *)task, task_struct_rh, self_exec_id);
    } else {
        exec_id = BPF_CORE_READ(task, self_exec_id);
    }

    printk("bpf_prog exec_id: %lx", exec_id);

    return 0;
}

char LICENSE[] SEC("license") = "GPL";
```

At the time of writing, `aya-obj` will generate a ``MissingTargetDefinition`` error on the relocation for ``task_struct_rh`` in the ``BPF_CORE_READ``.

This behavior differ with what [libbpf does](https://github.com/libbpf/libbpf/blob/5b795f7b3028b125bb0ad1db7dc809297b28094f/src/relo_core.c#L1398-L1408).

Missing relocations at load time shouldn't cause an error in `aya-obj` but instead poison related instructions.

If any of those code-path are taken, the verifier will refuse to load the eBPF.

marysaka:
Missing relocations at load time shouldn't cause an error in `aya-obj` but instead poison related instructions.

This makes struct flavors work.

Close #533

### DIFF
--- a/aya-obj/src/btf/relocation.rs
+++ b/aya-obj/src/btf/relocation.rs
@@ -14,8 +14,8 @@ use crate::{
         IntEncoding, Struct, Union, MAX_SPEC_LEN,
     },
     generated::{
-        bpf_core_relo, bpf_core_relo_kind::*, bpf_insn, BPF_ALU, BPF_ALU64, BPF_B, BPF_DW, BPF_H,
-        BPF_K, BPF_LD, BPF_LDX, BPF_ST, BPF_STX, BPF_W, BTF_INT_SIGNED,
+        bpf_core_relo, bpf_core_relo_kind::*, bpf_insn, BPF_ALU, BPF_ALU64, BPF_B, BPF_CALL,
+        BPF_DW, BPF_H, BPF_JMP, BPF_K, BPF_LD, BPF_LDX, BPF_ST, BPF_STX, BPF_W, BTF_INT_SIGNED,
     },
     util::HashMap,
     Object, Program, ProgramSection,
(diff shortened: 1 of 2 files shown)


vadorovsky (commented):
In general looks good, just one comment. Thanks for your work!

Do you have an example of a project which didn't work before this change? If so, it would be great to add an integration test. I know it's specific to the old kernel versions, but I still think we could add it. @alessandrod wdyt?

qjerome:
This PR has been tested on my side and everything works as expected.

aya-obj/src/btf/relocation.rs
```diff
@@ -862,9 +878,30 @@ impl ComputedRelocation {
...
+                poison_insn(next_ins);
+            }
+
+            return Ok(());
+        }
+
         let class = (ins.code & 0x07) as u32;
 
-        let target_value = self.target.value;
+        let target = self.target.as_ref().unwrap();
```
alessandrod, line 903: instead of doing if is_none() above and then unwrap, use a match statement and
return early in the None case

aya-obj/src/btf/relocation.rs
```diff
@@ -312,13 +312,21 @@ fn relocate_btf_program<'target>(
             // same value, else the relocation is ambiguous and can't be applied
             let conflicts = matches
                 .filter_map(|(cand_name, cand_spec, cand_comp_rel)| {
-                    if cand_spec.bit_offset != target_spec.bit_offset
-                        || cand_comp_rel.target.value != target_comp_rel.target.value
-                    {
-                        Some(cand_name)
-                    } else {
-                        None
+                    if cand_spec.bit_offset != target_spec.bit_offset {
```
alessandrod, line 315: This seems wrong? Previous condition was `||` now it's `&&`

alessandrod (changes requested):
Thanks and sorry it took so long to review this, see comments

aya-obj/src/btf/relocation.rs
alessandrod: Also, would be good having an integration test for this

aya-obj/src/btf/relocation.rs
marysaka: Nice catch oops

alessandrod (approved):

PR #581, opened 2023-04-17
aya-obj: fix DATASEC to STRUCT conversion

marysaka:
This fix the following issues:
- Previously the DATASEC name wasn't sanitized resulting on "Invalid name" returned by old kernels.
- The newly created BTF struct had a size of 0 making old kernels refuse it.

This was tested on Debian 10 with kernel 4.19.0-21.

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -432,6 +432,19 @@ impl Btf {
                 // Sanitize DATASEC if they are not supported
                 BtfType::DataSec(d) if !features.btf_datasec => {
                     debug!("{}: not supported. replacing with STRUCT", kind);
+
+                    // STRUCT aren't allowed to have "." in their name, fixup this if needed.
+                    let mut name_offset = t.name_offset();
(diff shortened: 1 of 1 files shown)


dave-tucker:
It's been a while since I was last in this part of the code, but since this was borrowed from libbpf, do you know what does `libbpf` does in this case? It would be nice to be consistent.
I would assume that if it knows the name would fail a validity check it would just omit the name entirely 🤔

marysaka:
> It's been a while since I was last in this part of the code, but since this was borrowed from libbpf, do you know what does `libbpf` does in this case? It would be nice to be consistent. I would assume that if it knows the name would fail a validity check it would just omit the name entirely thinking

I do follow what libbpf does here https://github.com/libbpf/libbpf/blob/378357716118bba6ead1b0dec3562a8a2f141108/src/libbpf.c#L2701-L2724

dave-tucker (approved):
LGTM thanks!

alessandrod (approved):

PR #583, opened 2023-04-19
Fix builtin linkage

qjerome:
- fixing memset return value for proper BTF validation
- fixing memset + memcpy linkage type when loading BPF program

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -544,7 +544,12 @@ impl Btf {
                         let typedef_type =
                             BtfType::Typedef(Typedef::new(ty.name_offset, ty.btf_type));
                         types.types[i] = typedef_type;
-                    } else if !features.btf_func_global || name == "memset" || name == "memcpy" {
+                    } else if !features.btf_func_global
+                        || name == "memset"
+                        || name == "memcpy"
+                        || name == "memmove"
+                        || name == "memcmp"
+                    {
                         // Sanitize BTF_FUNC_GLOBAL and memset, memcpy
                         let mut fixed_ty = ty.clone();
                         if ty.linkage() == FuncLinkage::Global {
@@ -1359,7 +1364,7 @@ mod tests {
     }
 
     #[test]
-    fn test_sanitize_memset_memcpy() {
+    fn test_sanitize_mem_builtins() {
         let mut btf = Btf::new();
         let name_offset = btf.add_string("int".to_string());
         let int_type_id = btf.add_type(BtfType::Int(Int::new(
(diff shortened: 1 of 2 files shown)


bpf/aya-bpf/src/lib.rs
```diff
@@ -57,11 +57,12 @@ pub trait BpfContext {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn memset(s: *mut u8, c: c_int, n: usize) {
+pub unsafe extern "C" fn memset(s: *mut u8, c: c_int, n: usize) -> u8 {
```
alessandrod, line 60: This is not needed now that we're making these functions static I think?

We should still fix our memory builtins to 1) be faster 2) work as globals, but we should do that in a separate PR and for all of them

aya-obj/src/btf/btf.rs
```diff
@@ -536,22 +536,30 @@ impl Btf {
...
-                        fixed_ty.set_linkage(FuncLinkage::Static);
+                BtfType::Func(ty) => {
+                    let name = self.string_at(ty.name_offset)?;
+                    // Sanitize FUNC
+                    if !features.btf_func {
+                        debug!("{}: not supported. replacing with TYPEDEF", kind);
+                        let typedef_type =
+                            BtfType::Typedef(Typedef::new(ty.name_offset, ty.btf_type));
+                        types.types[i] = typedef_type;
+                    } else if !features.btf_func_global || name == "memset" || name == "memcpy" {
```
alessandrod, line 547: this needs to be done for memmove and memcmp too

alessandrod (changes requested):
Thanks for working on this! See comments

bpf/aya-bpf/src/lib.rs
qjerome: You are right ! It works without. I will change that

aya-obj/src/btf/btf.rs
qjerome: Done

bpf/aya-bpf/src/lib.rs
qjerome: Done

aya-obj/src/btf/btf.rs
```diff
@@ -1350,6 +1363,61 @@ mod tests {
...
+            },
+            BtfParam {
+                name_offset: btf.add_string("b".to_string()),
+                btf_type: int_type_id,
+            },
+        ];
+        let func_proto_type_id =
+            btf.add_type(BtfType::FuncProto(FuncProto::new(params, int_type_id)));
+
+        ["memset", "memcpy", "memcmp", "memmove"]
```
alessandrod, line 1390: Please make this a for loop

aya-obj/src/btf/btf.rs
```diff
@@ -536,22 +536,35 @@ impl Btf {
...
+                        let typedef_type =
+                            BtfType::Typedef(Typedef::new(ty.name_offset, ty.btf_type));
+                        types.types[i] = typedef_type;
+                    } else if !features.btf_func_global
+                        || name == "memset"
+                        || name == "memcpy"
+                        || name == "memmove"
+                        || name == "memcmp"
+                    {
+                        // Sanitize BTF_FUNC_GLOBAL and memset, memcpy
```
alessandrod, line 553: The comment here should say:

```
Sanitize BTF_FUNC_GLOBAL when not supported and ensure that memory builtins are marked as static. Globals are type checked and verified separately from their callers, while instead we want tracking info (eg bound checks) to be propagated to the memory builtins.  
```

alessandrod (commented):
Thanks, almost there!

aya-obj/src/btf/btf.rs
qjerome: Done

aya-obj/src/btf/btf.rs
qjerome: Done

PR #584, opened 2023-04-19
aya: Ignore embedded BTF error if not truely required

marysaka:
This fix BpfLoader::load logic by following what libbpf does.

Required for Debian 10 support.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -389,9 +389,51 @@ impl<'a> BpfLoader<'a> {
 
         let btf_fd = if let Some(features) = &FEATURES.btf() {
             if let Some(btf) = obj.fixup_and_sanitize_btf(features)? {
-                // load btf to the kernel
-                let btf = load_btf(btf.to_bytes(), *verifier_log_level)?;
-                Some(btf)
+                match load_btf(btf.to_bytes(), *verifier_log_level) {
+                    Ok(btf_fd) => Some(btf_fd),
+                    // Only report an error here if the BTF is truely needed, otherwise proceed without.
+                    Err(err) => {
+                        for program in obj.programs.values() {
+                            match program.section {
+                                ProgramSection::Extension { .. }
+                                | ProgramSection::FEntry { .. }
+                                | ProgramSection::FExit { .. }
+                                | ProgramSection::Lsm { .. }
+                                | ProgramSection::BtfTracePoint { .. } => {
+                                    return Err(BpfError::BtfError(err))
+                                }
+                                ProgramSection::KRetProbe { .. }
+                                | ProgramSection::KProbe { .. }
+                                | ProgramSection::UProbe { .. }
+                                | ProgramSection::URetProbe { .. }
+                                | ProgramSection::TracePoint { .. }
+                                | ProgramSection::SocketFilter { .. }
+                                | ProgramSection::Xdp { .. }
+                                | ProgramSection::SkMsg { .. }
+                                | ProgramSection::SkSkbStreamParser { .. }
+                                | ProgramSection::SkSkbStreamVerdict { .. }
+                                | ProgramSection::SockOps { .. }
+                                | ProgramSection::SchedClassifier { .. }
+                                | ProgramSection::CgroupSkb { .. }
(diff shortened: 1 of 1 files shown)


aya/src/bpf.rs
```diff
@@ -345,8 +345,12 @@ impl<'a> BpfLoader<'a> {
 
         let btf_fd = if let Some(features) = &FEATURES.btf {
             if let Some(btf) = obj.fixup_and_sanitize_btf(features)? {
-                // load btf to the kernel
-                Some(load_btf(btf.to_bytes())?)
+                match load_btf(btf.to_bytes()) {
+                    Ok(btf_fd) => Some(btf_fd),
+                    // Only report an error here if the BTF is truely needed, otherwise proceed without.
+                    Err(err) if obj.kernel_needs_btf() => return Err(BpfError::BtfError(err)),
```
alessandrod, line 351: I think it would be better if we iterated obj.program here, looking at
program.section and deciding based on that whether we actually need BTF or not.
That way we don't need to set new state (`has_struct_ops`) and we don't need
`kernel_needs_btf` either, which doesn't seem like something that belongs to
`aya-obj`. And we need to check for all programs that require BTF, not only
struct ops. 

Also, I think we should only ignore the error if the user hasn't explicitly
called `BpfLoader::btf()` - if they explicitly asked for BTF to be loaded, they
might want BTF for other reasons (eg line info), and we shouldn't silently fail
to load it. In that case we should return the error and let callers decide
whether they want to try loading again with `btf(None)`.

alessandrod (changes requested):

aya-obj/src/btf/btf.rs
```diff
@@ -621,6 +621,11 @@ impl Object {
             Ok(None)
         }
     }
+
+    /// Indicates if the [Object] needs its [Btf] to be able to load by current kernel.
+    pub fn kernel_needs_btf(&self) -> bool {
+        self.has_struct_ops
```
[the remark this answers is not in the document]
marysaka: I did try to follow libbpf logic here, but you are right, I will change that, this is way smarter :+1:

aya/src/bpf.rs
```diff
@@ -389,9 +389,24 @@ impl<'a> BpfLoader<'a> {
...
+                    // Only report an error here if the BTF is truely needed, otherwise proceed without.
+                    Err(err) => {
+                        for program in obj.programs.values() {
+                            match program.section {
+                                ProgramSection::Extension { .. }
+                                | ProgramSection::FEntry { .. }
+                                | ProgramSection::FExit { .. } => {
+                                    return Err(BpfError::BtfError(err))
+                                }
+                                _ => {}
```
alessandrod, line 403: we should log a warning here

marysaka: Added the warning outside of the loop.

aya/src/bpf.rs
```diff
@@ -389,9 +389,26 @@ impl<'a> BpfLoader<'a> {
...
```
alessandrod, line 403: I think instead of using `_` here, we should handle all the program sections explicitly, so that when we add new program types it'll fail to compile if this code isn't updated

aya/src/bpf.rs
```diff
@@ -389,9 +389,51 @@ impl<'a> BpfLoader<'a> {
...
+                                | ProgramSection::CgroupSkbIngress { .. }
+                                | ProgramSection::CgroupSkbEgress { .. }
+                                | ProgramSection::CgroupSockAddr { .. }
+                                | ProgramSection::CgroupSysctl { .. }
+                                | ProgramSection::CgroupSockopt { .. }
+                                | ProgramSection::LircMode2 { .. }
+                                | ProgramSection::PerfEvent { .. }
+                                | ProgramSection::RawTracePoint { .. }
+                                | ProgramSection::Lsm { .. }
+                                | ProgramSection::BtfTracePoint { .. }
```
alessandrod, line 425: Surely this needs BTF? 😋 And I think Lsm does too?

marysaka: Fair point :thinking:

alessandrod (approved):

PR #590, opened 2023-04-25
aya-bpf: Add helper `change_proto`

Kab1r:
This patch adds a new helper function, `change_proto`, to the `SkBuff` struct in the `aya-bpf` crate.

The `change_proto` function allows changing the protocol of a socket buffer (`skb`) in BPF programs.
This patch also adds a corresponding wrapper for the `change_proto` function in the `TcContext` struct in the `aya-bpf` crate, making it accessible from TC (traffic control) programs as well.

### DIFF
--- a/bpf/aya-bpf/src/programs/sk_buff.rs
+++ b/bpf/aya-bpf/src/programs/sk_buff.rs
@@ -6,8 +6,8 @@ use core::{
 
 use aya_bpf_bindings::helpers::{
     bpf_clone_redirect, bpf_get_socket_uid, bpf_l3_csum_replace, bpf_l4_csum_replace,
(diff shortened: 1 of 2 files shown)


alessandrod (approved):
Thank you!

vadorovsky (approved):
Awesome, thanks!

@Kab1r do you have an example where you're using it? I'm just thinking that it would be nice to have some integration tests for TC and related helpers (including this one) in future.

Kab1r:
> do you have an example where you're using it?

@vadorovsky I am looking at implementing NAT64 in aya-ebpf, but I don't currently have anything concrete.

The manpage (`bpf-helpers(7)`) explicitly states that "The main for this helper is to perform NAT64 operations out of an eBPF program".

PR #592, opened 2023-04-27
build(deps): update bitflags requirement from 1.2.1 to 2.2.1

tamird:
Putting dependabot out of its misery (see #589, #571, #549, #541).


alessandrod (approved):
🙏

PR #598, opened 2023-05-01
xtask: codegen: generate bindings for user_regs_struct

nak3:
This patch adds `user_regs_struct`.

riscv provides struct `user_regs_struct` instead of struct `pt_regs` to userspace.
After bindings generates the code, adding the riscv support in `bpf/aya-bpf/src/args.rs`
then aya-bpf could be built for riscv.


vadorovsky (approved):

vadorovsky (approved):

nak3:
Thank you so much for merging this!
Could anyone trigger the action https://github.com/aya-rs/aya/actions/workflows/gen.yml ? I don't have the permission.

PR #605, opened 2023-05-10
aya: Do not create data maps on kernel without global data support

marysaka:
Fix map creation failure when a BPF have a data section on older kernel. (< 5.2)

If the BPF uses that section, relocation will fail accordingly and report an error.


alessandrod (approved):
I just checked this off my todo list after more than 2 years, thank you! 🤗

PR #608, opened 2023-05-23
Fix load errors for empty (but existent) BTF/BTF.ext sections

rc-andres:
Compiling a program using an older version of clang (tested w/ clang 8) for a non-CORE program creates a `.BTF` and a `.BTF.ext` section that aya fails to load.

* `.BTF.ext` error: the `.BTF.ext` section is from an older header and `aya` isn't checking the `hdr_len` and instead assumes that the size of its pregenerated struct will work no matter what. This PR fixes this by first loading the data into a "minimal header" and then only loads as much data into the full fledged header as it is allowed to per the spec. 
* `.BTF` error: The created section has no types which returns an error at load type saying "type not found". This PR uses the existing "sanitize" method to check if there are any types and if not then it skips the load of the `.BTF` section as there is no gain in trying to load it.

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -688,11 +688,12 @@ impl BtfExt {
             return Err(BtfError::InvalidHeader);
         }
 
-        // Safety: btf_ext_header and MinimalHeader are POD so read_unaligned is safe
-        let header = unsafe {
+        let header = {
             // first find the actual size of the header by converting into the minimal valid header
-            let minimal_header =
-                ptr::read_unaligned::<MinimalHeader>(data.as_ptr() as *const MinimalHeader);
+            // Safety: MinimalHeader is POD so read_unaligned is safe
+            let minimal_header = unsafe {
+                ptr::read_unaligned::<MinimalHeader>(data.as_ptr() as *const MinimalHeader)
+            };
 
             let len_to_read = minimal_header.hdr_len as usize;
 
@@ -709,14 +710,22 @@ impl BtfExt {
             // now create our full-fledge header; but start with it
             // zeroed out so unavailable fields stay as zero on older
             // BTF.ext sections
-            let mut header: btf_ext_header = std::mem::MaybeUninit::zeroed().assume_init();
+            // Safety: btf_ext_header is POD so a zero'd out struct is safe
+            let mut header: btf_ext_header =
+                unsafe { std::mem::MaybeUninit::zeroed().assume_init() };
 
             // now copy `data` onto our `header` but only up to
             // hdr_len bytes
-            let header_as_slice: &mut [u8] = std::slice::from_raw_parts_mut(
-                &mut header as *mut btf_ext_header as *mut u8,
-                len_to_read,
-            );
+
+            // Safety: len_to_read is at most btf_ext_header's size
+            // and btf_ext_header is a POD so converting it to a slice
+            // of bytes is afe
+            let header_as_slice: &mut [u8] = unsafe {
+                std::slice::from_raw_parts_mut(
+                    &mut header as *mut btf_ext_header as *mut u8,
+                    len_to_read,
+                )
+            };
             header_as_slice.copy_from_slice(&data[0..len_to_read]);
             header
         };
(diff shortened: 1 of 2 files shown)


aya-obj/src/btf/btf.rs
```diff
@@ -667,13 +675,58 @@ impl BtfExt {
...
+            pub flags: u8,
+            pub hdr_len: i32,
+        }
+
+        if data.len() < std::mem::size_of::<MinimalHeader>() {
+            return Err(BtfError::InvalidHeader);
+        }
+
+        // Safety: btf_ext_header and MinimalHeader are POD so read_unaligned is safe
         let header = unsafe {
```
alessandrod, line 692: please don't use such a large unsafe section, it makes it harder to reason about
what's actually unsafe

aya-obj/src/btf/btf.rs
```diff
@@ -667,13 +675,58 @@ impl BtfExt {
...
         btf: &Btf,
     ) -> Result<BtfExt, BtfError> {
-        // Safety: btf_ext_header is POD so read_unaligned is safe
+        #[repr(C)]
+        #[derive(Debug, Copy, Clone)]
+        struct MinimalHeader {
+            pub magic: u16,
+            pub version: u8,
+            pub flags: u8,
+            pub hdr_len: i32,
```
alessandrod, line 684: I think this should be u32?

aya-obj/src/btf/btf.rs
```diff
@@ -667,13 +675,58 @@ impl BtfExt {
...
+
+        // Safety: btf_ext_header and MinimalHeader are POD so read_unaligned is safe
         let header = unsafe {
-            ptr::read_unaligned::<btf_ext_header>(data.as_ptr() as *const btf_ext_header)
+            // first find the actual size of the header by converting into the minimal valid header
+            let minimal_header =
+                ptr::read_unaligned::<MinimalHeader>(data.as_ptr() as *const MinimalHeader);
+
+            // prevent invalid input from causing UB
+            if data.len() < minimal_header.hdr_len as usize {
```
alessandrod, line 698: You also need to check that `minimal_header.hdr_len <= mem::size_of<btf_ext_header>()` right?

aya-obj/src/btf/btf.rs
```diff
@@ -667,13 +675,58 @@ impl BtfExt {
...
+            }
+
+            // now create our full-fledge header; but start with it
+            // zeroed out so unavailable fields stay as zero on older
+            // BTF.ext sections
+            let mut header: btf_ext_header = std::mem::MaybeUninit::zeroed().assume_init();
+
+            // now copy `data` onto our `header` but only up to
+            // hdr_len bytes
+            let header_as_slice: &mut [u8] = std::slice::from_raw_parts_mut(
```
alessandrod, line 709: instead of doing assume_init() and then creating a slice, I think it would be
better to use ptr::copy on header.as_mut_ptr()?

alessandrod (changes requested):
Thank you! See comments

aya-obj/src/btf/btf.rs
rc-andres: Instead of checking I decided to use the minimum of the two so that the current header struct still works if newer programs have a bigger btf.ext header.

aya-obj/src/btf/btf.rs
rc-andres: I am not quite sure what you mean by this. Would I still create a zero'd out header instance first (through `std::mem::MaybeUninit::zeroed().assume_init()` and then use `ptr::copy` on it somehow?

alessandrod: See comment below

aya-obj/src/btf/btf.rs
```diff
@@ -667,13 +675,74 @@ impl BtfExt {
...
+            // Safety: len_to_read is at most btf_ext_header's size
+            // and btf_ext_header is a POD so converting it to a slice
+            // of bytes is afe
+            let header_as_slice: &mut [u8] = unsafe {
+                std::slice::from_raw_parts_mut(
+                    &mut header as *mut btf_ext_header as *mut u8,
+                    len_to_read,
+                )
+            };
+            header_as_slice.copy_from_slice(&data[0..len_to_read]);
```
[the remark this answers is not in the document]
rc-andres: Ah that makes sense now, thanks!

alessandrod (approved):
Looks good, thank you! I'm going to merge once CI passes

rc-andres:
> Looks good, thank you! I'm going to merge once CI passes

Thanks for the quick feedback, I've been really enjoying Aya!

Issue #612, opened 2023-05-26
aya: maps + programs using RawFd could leak to fd leaking


nrxus:
As of Rust v1.63.0 new types `OwnedFd` and `BorrowedFd` were added so that rust can handle the lifetime of file descriptors using normal RAII and borrow rules. I think it'd benefit `aya` to move towards using these types as it can prevent some unforeseen errors or file descriptor leakage either by aya or by its users. 

As a side note on this we may have to remove the implementation of `Clone` on some types, at the very least on `Map` and change it to a `try_clone()`. The current implementation is faulty in that it both ignores errors and the cloned file descriptor is not set to close on exec which afaict should generally be set. While in practice this may not be an issue 99% of the time it is still not good practice to ignore errors during duplication of a file descriptor. Rust's `OwnedFd` does the right thing on its `try_clone()` method. 

I was contemplating making working on this and making a PR but I wanted to make sure this is something that the maintainers of aya are ok merging before I start on it since it'd be a pretty hefty PR and it might have some breaking changes, including having to up the MSRV to v1.63.

alessandrod:
> I was contemplating working on this and making a PR but I wanted to make sure this is something that the maintainers of aya are ok merging before I start on it since it'd be a pretty hefty PR and it might have some breaking changes, including having to up the MSRV to v1.63.

We've discussed this in the past and is definitely something we want, we just need someone to drive the effort. I think MSRV 1.63 would be fine.

tamird:
This is almost done. We have two instances of held `RawFd` left:

https://github.com/aya-rs/aya/blob/0b6ea313ded3240715d1b30d3b247e2bc983659e/aya/src/programs/socket_filter.rs#L126-L127


https://github.com/aya-rs/aya/blob/0b6ea313ded3240715d1b30d3b247e2bc983659e/aya/src/programs/xdp.rs#L234

I'm not exactly sure what the semantics there are supposed to be. @dave-tucker ?

dave-tucker:
> This is almost done. We have two instances of held `RawFd` left:
> 
> https://github.com/aya-rs/aya/blob/0b6ea313ded3240715d1b30d3b247e2bc983659e/aya/src/programs/socket_filter.rs#L126-L127
> 
> https://github.com/aya-rs/aya/blob/0b6ea313ded3240715d1b30d3b247e2bc983659e/aya/src/programs/xdp.rs#L234
>
> I'm not exactly sure what the semantics there are supposed to be. @dave-tucker ?

[...]

PR #614, opened 2023-05-28
Safer file descriptor handling (i.e., using `OwnedFd` where possible)

nrxus:
Attempts to resolve https://github.com/aya-rs/aya/issues/612. 

Currently still a WIP, so far it just changes the following to `OwnedFd`:

* `MapData::fd`
* `FdLink.0`
* `ProgramData.fd`

These are changed in tandem because changing just the first out of them trickled down to the other two. `//TODO (AM)` is left in many spots that will be addressed in this PR once it's ready for a full review. Putting up a draft PR to start conversation on potential blockers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/614)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -3,7 +3,11 @@ use std::{
     collections::{HashMap, HashSet},
     ffi::CString,
     fs, io,
-    os::{fd::AsRawFd, raw::c_int, unix::io::RawFd},
+    os::{
+        fd::{AsFd, AsRawFd},
+        raw::c_int,
+        unix::io::RawFd,
+    },
     path::{Path, PathBuf},
 };
 
@@ -421,23 +425,16 @@ impl<'a> BpfLoader<'a> {
                 PinningType::None => map.create(&name)?,
             };
             // OK: we just loaded the map
-            let fd = map.fd.as_ref().unwrap();
+            let fd = map.fd.as_ref().unwrap().as_fd();
             if !map.obj.data().is_empty() && map.obj.section_kind() != BpfSectionKind::Bss {
-                // TODO (AM)
-                bpf_map_update_elem_ptr(
-                    fd.as_raw_fd(),
-                    &0 as *const _,
-                    map.obj.data_mut().as_mut_ptr(),
-                    0,
-                )
-                .map_err(|(_, io_error)| MapError::SyscallError {
-                    call: "bpf_map_update_elem".to_owned(),
-                    io_error,
-                })?;
+                bpf_map_update_elem_ptr(fd, &0 as *const _, map.obj.data_mut().as_mut_ptr(), 0)
+                    .map_err(|(_, io_error)| MapError::SyscallError {
+                        call: "bpf_map_update_elem".to_owned(),
+                        io_error,
+                    })?;
             }
             if map.obj.section_kind() == BpfSectionKind::Rodata {
-                // TODO (AM)
-                bpf_map_freeze(fd.as_raw_fd()).map_err(|(_, io_error)| MapError::SyscallError {
+                bpf_map_freeze(fd).map_err(|(_, io_error)| MapError::SyscallError {
                     call: "bpf_map_freeze".to_owned(),
                     io_error,
                 })?;
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -3,7 +3,7 @@ use std::{
     collections::{HashMap, HashSet},
     ffi::CString,
     fs, io,
-    os::{raw::c_int, unix::io::RawFd},
+    os::{fd::AsRawFd, raw::c_int, unix::io::RawFd},
     path::{Path, PathBuf},
 };
 
@@ -398,7 +398,7 @@ impl<'a> BpfLoader<'a> {
                 pinned: false,
                 btf_fd,
             };
-            let fd = match map.obj.pinning() {
+            match map.obj.pinning() {
                 PinningType::ByName => {
                     let path = match &self.map_pin_path {
                         Some(p) => p,
@@ -406,31 +406,38 @@ impl<'a> BpfLoader<'a> {
                     };
                     // try to open map in case it's already pinned
                     match map.open_pinned(&name, path) {
-                        Ok(fd) => {
+                        Ok(_) => {
                             map.pinned = true;
-                            fd as RawFd
                         }
                         Err(_) => {
-                            let fd = map.create(&name)?;
+                            map.create(&name)?;
                             map.pin(&name, path).map_err(|error| MapError::PinError {
                                 name: Some(name.to_string()),
                                 error,
                             })?;
-                            fd
                         }
                     }
                 }
                 PinningType::None => map.create(&name)?,
             };
+            // OK: we just loaded the map
+            let fd = map.fd.as_ref().unwrap();
             if !map.obj.data().is_empty() && map.obj.section_kind() != BpfSectionKind::Bss {
-                bpf_map_update_elem_ptr(fd, &0 as *const _, map.obj.data_mut().as_mut_ptr(), 0)
-                    .map_err(|(_, io_error)| MapError::SyscallError {
-                        call: "bpf_map_update_elem".to_owned(),
-                        io_error,
-                    })?;
+                // TODO (AM)
+                bpf_map_update_elem_ptr(
+                    fd.as_raw_fd(),
+                    &0 as *const _,
+                    map.obj.data_mut().as_mut_ptr(),
+                    0,
+                )
+                .map_err(|(_, io_error)| MapError::SyscallError {
+                    call: "bpf_map_update_elem".to_owned(),
+                    io_error,
+                })?;
             }
             if map.obj.section_kind() == BpfSectionKind::Rodata {
-                bpf_map_freeze(fd).map_err(|(_, io_error)| MapError::SyscallError {
+                // TODO (AM)
+                bpf_map_freeze(fd.as_raw_fd()).map_err(|(_, io_error)| MapError::SyscallError {
                     call: "bpf_map_freeze".to_owned(),
                     io_error,
                 })?;
@@ -446,7 +453,14 @@ impl<'a> BpfLoader<'a> {
 
         obj.relocate_maps(
             maps.iter()
-                .map(|(s, data)| (s.as_str(), data.fd, &data.obj)),
+                // TODO (AM)
+                .map(|(s, data)| {
+                    (
+                        s.as_str(),
+                        data.fd.as_ref().map(|f| f.as_raw_fd()),
+                        &data.obj,
+                    )
+                }),
             &text_sections,
         )?;
         obj.relocate_calls(&text_sections)?;
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -4,9 +4,8 @@ use std::{
     ffi::CString,
     fs, io,
     os::{
-        fd::{AsFd, AsRawFd},
+        fd::{AsFd, AsRawFd, OwnedFd},
         raw::c_int,
-        unix::io::RawFd,
     },
     path::{Path, PathBuf},
 };
@@ -368,6 +367,9 @@ impl<'a> BpfLoader<'a> {
             None
         };
 
+        // TODO (AM)
+        let raw_btf_fd = btf_fd.as_ref().map(|f| f.as_raw_fd());
+
         if let Some(btf) = &self.btf {
             obj.relocate_btf(btf)?;
         }
@@ -400,7 +402,7 @@ impl<'a> BpfLoader<'a> {
                 obj,
                 fd: None,
                 pinned: false,
-                btf_fd,
+                btf_fd: raw_btf_fd,
             };
             match map.obj.pinning() {
                 PinningType::ByName => {
@@ -479,70 +481,90 @@ impl<'a> BpfLoader<'a> {
 
                 let program = if self.extensions.contains(name.as_str()) {
                     Program::Extension(Extension {
-                        data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                        data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                     })
                 } else {
                     match &section {
                         ProgramSection::KProbe { .. } => Program::KProbe(KProbe {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                             kind: ProbeKind::KProbe,
                         }),
                         ProgramSection::KRetProbe { .. } => Program::KProbe(KProbe {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                             kind: ProbeKind::KRetProbe,
                         }),
                         ProgramSection::UProbe { .. } => Program::UProbe(UProbe {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                             kind: ProbeKind::UProbe,
                         }),
                         ProgramSection::URetProbe { .. } => Program::UProbe(UProbe {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                             kind: ProbeKind::URetProbe,
                         }),
                         ProgramSection::TracePoint { .. } => Program::TracePoint(TracePoint {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                         }),
                         ProgramSection::SocketFilter { .. } => {
                             Program::SocketFilter(SocketFilter {
-                                data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                                data: ProgramData::new(
+                                    prog_name,
+                                    obj,
+                                    raw_btf_fd,
+                                    verifier_log_level,
+                                ),
                             })
                         }
                         ProgramSection::Xdp { frags, .. } => {
                             let mut data =
-                                ProgramData::new(prog_name, obj, btf_fd, verifier_log_level);
+                                ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level);
                             if *frags {
                                 data.flags = BPF_F_XDP_HAS_FRAGS;
                             }
                             Program::Xdp(Xdp { data })
                         }
                         ProgramSection::SkMsg { .. } => Program::SkMsg(SkMsg {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                         }),
                         ProgramSection::CgroupSysctl { .. } => {
                             Program::CgroupSysctl(CgroupSysctl {
-                                data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                                data: ProgramData::new(
+                                    prog_name,
+                                    obj,
+                                    raw_btf_fd,
+                                    verifier_log_level,
+                                ),
                             })
                         }
                         ProgramSection::CgroupSockopt { attach_type, .. } => {
                             Program::CgroupSockopt(CgroupSockopt {
-                                data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                                data: ProgramData::new(
+                                    prog_name,
+                                    obj,
+                                    raw_btf_fd,
+                                    verifier_log_level,
+                                ),
                                 attach_type: *attach_type,
                             })
                         }
                         ProgramSection::SkSkbStreamParser { .. } => Program::SkSkb(SkSkb {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                             kind: SkSkbKind::StreamParser,
                         }),
                         ProgramSection::SkSkbStreamVerdict { .. } => Program::SkSkb(SkSkb {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                             kind: SkSkbKind::StreamVerdict,
                         }),
                         ProgramSection::SockOps { .. } => Program::SockOps(SockOps {
-                            data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                            data: ProgramData::new(prog_name, obj, raw_btf_fd, verifier_log_level),
                         }),
                         ProgramSection::SchedClassifier { .. } => {
                             Program::SchedClassifier(SchedClassifier {
-                                data: ProgramData::new(prog_name, obj, btf_fd, verifier_log_level),
+                                data: ProgramData::new(
+                                    prog_name,
+                                    obj,
+                                    raw_btf_fd,
+                                    verifier_log_level,
+                                ),
                                 name: unsafe {
                                     CString::from_vec_unchecked(Vec::from(name.clone()))
                                         .into_boxed_c_str()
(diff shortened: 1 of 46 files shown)


aya/src/maps/bloom_filter.rs
```diff
@@ -274,4 +276,11 @@ mod tests {
             Err(MapError::ElementNotFound)
         ));
     }
+
+    fn create_fd() -> OwnedFd {
+        let dir = env::temp_dir();
+        File::create(dir.join("f1"))
+            .expect("unable to create file in tmpdir")
+            .into()
+    }
```
nrxus, line 285: if we don't like this more manual approach we could bring in `tempfile` as a dev dependency or perhaps use memfds. This was just the easiest way to get the tests to pass for now.

aya/src/maps/mod.rs
```diff
@@ -185,11 +188,17 @@ pub enum MapError {
 }
 
 /// A map file descriptor.
-pub struct MapFd(RawFd);
+pub struct MapFd<'f>(BorrowedFd<'f>);
```
nrxus, line 191: this is a breaking change in that prior to this PR `MapFd` would not have a lifetime associated to the map it came from whereas now it does. Having a lifetime associated with it is more "correct" though in that before a map could be dropped (closing the file descriptor) but a `MapFd` could still hang around. We could make `MapFd` have a cloned file descriptor or use `Arc<OwnedFd>` in maps but neither of those felt quite right so I left it at this "dumb" way for now. Feedback here on what feels more appropriate for maintainers would be appreciated.

aya/src/maps/mod.rs
```diff
@@ -474,15 +483,15 @@ pub(crate) fn check_v_size<V>(map: &MapData) -> Result<(), MapError> {
...
+    pub(crate) fd: Option<OwnedFd>,
     pub(crate) btf_fd: Option<RawFd>,
     /// Indicates if this map has been pinned to bpffs
     pub pinned: bool,
 }
 
 impl MapData {
     /// Creates a new map with the provided `name`
-    pub fn create(&mut self, name: &str) -> Result<RawFd, MapError> {
+    pub fn create(&mut self, name: &str) -> Result<(), MapError> {
```
nrxus, line 494: Another breaking change. I initially tried to make this return a `BorrowedFd<'_>` to stay similar to the old API but that caused some borrowing issues later since the lifetime of the *mutable* self borrow would be the lifetime of the returned file desriptor which made it awkward to work with. I think this is *okay* since one can ask for the internal file descriptor easy enough.

aya/src/maps/mod.rs
```diff
@@ -604,28 +622,32 @@ impl MapData {
...
+// TODO (AM): DELETE?
 impl Clone for MapData {
     fn clone(&self) -> MapData {
-        MapData {
-            obj: self.obj.clone(),
-            fd: self.fd.map(|fd| unsafe { libc::dup(fd) }),
-            btf_fd: self.btf_fd,
-            pinned: self.pinned,
-        }
+        self.try_clone().unwrap()
```
nrxus, line 650: I think we should delete this as implicit panics on a clone is weird and clone can definitely fail but I left it here so the choice is up to the maintainers on what feels best here.

aya/src/maps/mod.rs
```diff
@@ -604,28 +622,32 @@ impl MapData {
...
-    fn drop(&mut self) {
-        // TODO: Replace this with an OwnedFd once that is stabilized.
-        if let Some(fd) = self.fd.take() {
-            unsafe { libc::close(fd) };
-        }
+impl MapData {
+    /// Attempts to duplicate MapData
+    ///
+    /// Fails if if the inner file descriptor could not be cloned
+    // TODO (AM): should we return MapError? New variant? Maybe make MapError non_exhaustive?
```
nrxus, line 634: The current enum is exhaustive so adding any variant would be a breaking change, I think we probably should make it non exhaustive and add a new variant, unless there is an existing one that people think already fits this? nothing jumped out to me.

aya/src/programs/xdp.rs
```diff
@@ -209,6 +224,7 @@ impl Xdp {
     }
 }
 
+// TODO (AM): implement Drop to force detach?
```
nrxus, line 227: perhaps a dumb question but is there a reason we don't implement `Drop` on `NdLink` to force detach? I think that'd simplify some of the stuff that wrap `NlLink` that have to remember to call for `detach` manually during drop and optionally `forget`.

aya/src/programs/extension.rs
```diff
@@ -42,9 +42,11 @@ pub enum ExtensionError {
...
 ///
 /// let prog_fd = prog.fd().unwrap();
-/// let ext: &mut Extension = bpf.program_mut("extension").unwrap().try_into()?;
-/// ext.load(prog_fd, "function_to_replace")?;
-/// ext.attach()?;
+/// // TODO (AM): can't borrow here becauase prog_fd is holding a reference already
+/// // need to confirm what the API wants to be
+/// // let ext: &mut Extension = bpf.program_mut("extension").unwrap().try_into()?;
+/// // ext.load(prog_fd, "function_to_replace")?;
+/// // ext.attach()?;
```
nrxus, line 49: this test no longer compiles because `ProgramFd` now has a lifetime associated with the program it came from. This makes sense intuitively but it means things like this where you need to pass the `ProgramFd`  from one program to another in the same `Bpf` object fail the borrow checker because the checker doesn't know that these two programs modify different parts of the `Bpf` instance so it's okay to have two mutable references to them. 

There are probably many ways around this but off the top of my head what came to mind is:
* Using `Arc<OwnedFd>` inside `ProgramData` and clone that reference when making a `ProgramFd` thus disentangling the lifetimes at compile time. I don't really love this because we are  adding a cost (using `Arc`) for all users to solve a specific problem just because it makes lifetimes easier. That being said this may end up being the most ergonomic. 
* Cloning the `OwnedFd` when creating `ProgramFd` (thus making it fallible). This pushes the cost only to those that need a `ProgramFd` so that's the plus, but the cons are that it makes making a `ProgramFd` fallible, less "for free" as we now have to dup the file descriptor (although that's probably fairly cheap), and users of `ProgramFd` for cases like the ones above we'll probably have to keep the `ProgramFd` alive separately because if it drops the duplicated file descriptor will be closed which they may not have wanted (not what it used to do). 
* Having a way for users to take ownership of programs (and potentially putting it back later). We already have a `take_map` so if there was a similar `take_program` then we resolve the lifetime issue because the moment we take the program we disentangle from the lifetime of the `Bpf` instance so we can get other programs without the double mutable reference issue. We'll may also want to add a way to put it back though, so that it can go back to being managed by `Bpf`. The only quirk is how to make it clear that when you put a program back you are not saying "load it" or anything like that, it is literally just saying "I don't want to handle this separately anymore put it in your rust-side table of programs". This is I think my preferred option especially since we already have `take_map`.

aya/src/programs/mod.rs
```diff
@@ -215,11 +218,17 @@ pub enum ProgramError {
 
 /// A [`Program`] file descriptor.
 #[derive(Copy, Clone)]
-pub struct ProgramFd(RawFd);
+pub struct ProgramFd<'f>(BorrowedFd<'f>);
```
nrxus, line 221: As explained in another comment for the breaking doctest this change in behavior can cause borrow checker issues if we need to use one program's fd in a different program in the same bpf object.

aya/src/programs/lirc_mode2.rs
```diff
@@ -101,13 +107,13 @@ impl LircMode2 {
...
-            prog_fds.push(fd as RawFd);
+            prog_fds.push(fd);
         }
 
-        Ok(prog_fds
+        prog_fds
             .into_iter()
-            .map(|prog_fd| LircLink::new(prog_fd, target_fd.as_raw_fd()))
-            .collect())
+            .map(|prog_fd| Ok(LircLink::new(prog_fd, target_fd.try_clone_to_owned()?)))
```
nrxus, line 115: prior to this PR we would leak the `prog_fd` as it was created by `bpf_prog_get_fd_by_id` but never closed.

aya/src/programs/lirc_mode2.rs
```diff
@@ -60,9 +61,9 @@ impl LircMode2 {
     /// Attaches the program to the given lirc device.
     ///
     /// The returned value can be used to detach, see [LircMode2::detach].
-    pub fn attach<T: AsRawFd>(&mut self, lircdev: T) -> Result<LircLinkId, ProgramError> {
+    pub fn attach<T: AsFd>(&mut self, lircdev: T) -> Result<LircLinkId, ProgramError> {
```
nrxus, line 64: I wonder if this should actually just take an `OwnedFd`, which would remove our need to clone it and it'd be more obvious that we are taking ownership of the file descriptor whereas now it is not clear IMO who owns the file descriptor when calling for `attach` (if we do it here we may also want/need to do in other attach functions to be consistent though).

aya/src/bpf.rs
```diff
@@ -615,6 +662,7 @@ impl<'a> BpfLoader<'a> {
         Ok(Bpf {
             maps: maps?,
             programs,
+            _btf_fd: btf_fd,
```
nrxus, line 665: This was previously being leaked as we would create it but no one would own it/close it. My gut feeling is that `ProgramData` and `MapData` should stop storing the raw fd and instead *use* a a borrowedfd when loading and then drop it. This does change the API slightly as re-loading maps/programs would require the btf fd being passed though.

aya/src/programs/lirc_mode2.rs
```diff
@@ -71,7 +72,11 @@ impl LircMode2 {
             }
         })?;
 
-        self.data.links.insert(LircLink::new(prog_fd, lircdev_fd))
+        // TODO (AM)
+        self.data.links.insert(LircLink::new(
+            prog_fd.try_clone_to_owned()?,
```
nrxus, line 77: I am not sure if cloning the fd is the best but the `query` function below has an `OwnedFd` from `query(target_fd, BPF_LIRC_MODE2, 0, &mut None)` so that was the easiest thing to do. We could change the API for `query` and use `BorrowedFd<'_>` in `LircLink` but that that may be more awkward due to lifetimes.

nrxus:
@alessandrod whenever you have time I'd love some feedback on this PR. The latest push moved some of the file descriptors to be wrapped in Arcs to avoid lifetime issues but I am not sure if that's optimal or not. There are a few cases of leaked file descriptors this PR is already fixing though

alessandrod:
Hey! Thanks a lot for doing this! I have been traveling the past 3 weeks so time has been limited, I'll try to do a first pass sometime this week 😊

tamird:
Could you resolve the merge conflicts please?

nrxus:
> Could you resolve the merge conflicts please?

I've addressed all the merge conflicts.

tamird (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 39 of 39 files at r3.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/614)* status: 3 of 46 files reviewed, 26 unresolved discussions (waiting on @nrxus)

*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208424753:-NZxA9Xf7s0K_NkWYsXZ:b62ke7z) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

if we don't like this more manual approach we could bring in `tempfile` as a dev dependency or perhaps use memfds. This was just the easiest way to get the tests to pass for now. 
</blockquote></details>

Yes, let's do something that doesn't muck with the filesystem.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208428302:-NZxAMlDE2n9MOxh7dNQ:besge3o) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

this is a breaking change in that prior to this PR `MapFd` would not have a lifetime associated to the map it came from whereas now it does. Having a lifetime associated with it is more "correct" though in that before a map could be dropped (closing the file descriptor) but a `MapFd` could still hang around. We could make `MapFd` have a cloned file descriptor or use `Arc<OwnedFd>` in maps but neither of those felt quite right so I left it at this "dumb" way for now. Feedback here on what feels more appropriate for maintainers would be appreciated. 
</blockquote></details>

This seems preferable to various hacks.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208428735:-NZxAaXDDpNeVuLCdRZ3:b-hmkp49) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

Breaking change. I initially tried to make this return a `BorrowedFd<'_>` to stay similar to the old API but that caused some borrowing issues later since the lifetime of the *mutable* self borrow would be the lifetime of the returned file desriptor which made it awkward to work with. I think this is *okay* since one can ask for the internal file descriptor easy enough. 
</blockquote></details>

This suggests that the implementation here is ... weird. Do you think it'd make sense for MapData::create to return a new type (perhaps borrowed from `self`) rather that mutating `self`?

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208429467:-NZxAtZM-l457l1S_d4o:brutuod) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

I think we should delete this as implicit panics on a clone is weird and clone can definitely fail but I left it here so the choice is up to the maintainers on what feels best here. 
</blockquote></details>

Yes, delete.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208429709:-NZxAwf1AQtDRFUbmbf0:b-t2cist) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

The current enum is exhaustive so adding any variant would be a breaking change, I think we probably should make it non exhaustive and add a new variant, unless there is an existing one that people think already fits this? nothing jumped out to me. 
</blockquote></details>

What would be the benefit of shoehorning this into MapError?

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208444076:-NZxB6tLAW4-c-P0CTQk:bmzh52s) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

perhaps a dumb question but is there a reason we don't implement `Drop` on `NdLink` to force detach? I think that'd simplify some of the stuff that wrap `NlLink` that have to remember to call for `detach` manually during drop and optionally `forget`. 
</blockquote></details>

Let's implement it and see where we end up. Perhaps that can be a separate PR that comes ahead of this one.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1208451414:-NZxBbYi9mdXkTP-J0_N:buhwhpq) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

this test no longer compiles because `ProgramFd` now has a lifetime associated with the program it came from. This makes sense intuitively but it means things like this where you need to pass the `ProgramFd`  from one program to another in the same `Bpf` object fail the borrow checker because the checker doesn't know that these two programs modify different parts of the `Bpf` instance so it's okay to have two mutable references to them. 

There are probably many ways around this but off the top of my head what came to mind is:
* Using `Arc<OwnedFd>` inside `ProgramData` and clone that reference when making a `ProgramFd` thus disentangling the lifetimes at compile time. I don't really love this because we are  adding a cost (using `Arc`) for all users to solve a specific problem just because it makes lifetimes easier. That being said this may end up being the most ergonomic. 
* Cloning the `OwnedFd` when creating `ProgramFd` (thus making it fallible). This pushes the cost only to those that need a `ProgramFd` so that's the plus, but the cons are that it makes making a `ProgramFd` fallible, less "for free" as we now have to dup the file descriptor (although that's probably fairly cheap), and users of `ProgramFd` for cases like the ones above we'll probably have to keep the `ProgramFd` alive separately because if it drops the duplicated file descriptor will be closed which they may not have wanted (not what it used to do). 
* Having a way for users to take ownership of programs (and potentially putting it back later). We already have a `take_map` so if there was a similar `take_program` then we resolve the lifetime issue because the moment we take the program we disentangle from the lifetime of the `Bpf` instance so we can get other programs without the double mutable reference issue. We'll may also want to add a way to put it back though, so that it can go back to being managed by `Bpf`. The only quirk is how to make it clear that when you put a program back you are not saying "load it" or anything like that, it is literally just saying "I don't want to handle this separately anymore put it in your rust-side table of programs". This is I think my preferred option especially since we already have `take_map`. 
</blockquote></details>

I guess this was resolved in later commits? There's a lot of whiplash in reviewing this PR, could you rationalize the commit history please? I will stop my review here for the moment.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1212476192:-NZxCHYlBwRFfhLM_XBz:b-c2a6vg) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

I wonder if this should actually just take an `OwnedFd` (or `Into<OwnedFd>`), which would remove our need to clone it if the caller doesn't need the file descriptor anyway (i.e., an opened file that is not needed anymore)  (if we do it here we may also want/need to do in other attach functions to be consistent though). 
</blockquote></details>

Yes, that seems better.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1212478503:-NZxCP0YDZ3fCO0xuAZ_:bvi6q22) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

This was previously being leaked as we would create it but no one would own it/close it. My gut feeling is that `ProgramData` and `MapData` should stop storing the raw fd and instead *use* a a borrowedfd when loading and then drop it. This does change the API slightly as re-loading maps/programs would require the btf fd being passed though. 
</blockquote></details>

API changes of this nature seem desirable to me.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/614#gh-1212506953:-NZxCa1a4okOv4o0K6yK:b-z9pane) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

I am not sure if cloning the fd is the best but the `query` function below has an `OwnedFd` from `query(target_fd, BPF_LIRC_MODE2, 0, &mut None)` so that was the easiest thing to do. We could change the API for `query` and use `BorrowedFd<'_>` in `LircLink` but that that may be more awkward due to lifetimes. 
</blockquote></details>

I guess this has been resolved?

___
___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/aya-rs/aya/614#-NZx4QjeA3hQwnvnPUMD):*
Why this in its own commit? It'd be better to put this in the same commit that makes use of whatever was made available in 1.63.
___
*[`aya/src/bpf.rs` line 6 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx5D1TD-Gps-_0fiym:-NZx5D1TD-Gps-_0fiyn:b3rm3q) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/bpf.rs#L6)):*
> ```XML
>     ffi::CString,
>     fs, io,
>     os::{fd::AsRawFd, raw::c_int, unix::io::RawFd},
> ```

can this be `AsRawFd as _`? we only need the trait functions, not the trait identifier itself
___
*[`aya/src/bpf.rs` line 423 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx4iZA0rulhMLKhpY-:-NZx4iZB46X1LqBNMwJF:b-am7b1q) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/bpf.rs#L423)):*
> ```XML
>                 PinningType::None => map.create(&name)?,
>             };
>             // OK: we just loaded the map
> ```

this seems strictly worse. can you help me understand why the old code path isn't what we want?
___
*[`aya/src/bpf.rs` line 429 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx4xlh00qlPIGCO4tX:-NZx4xlh00qlPIGCO4tY:b2q1dcb) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/bpf.rs#L429)):*
> ```XML
>                 bpf_map_update_elem_ptr(
>                     fd.as_raw_fd(),
>                     &0 as *const _,
> ```

should this be core::ptr::null()?
___
*[`aya/src/maps/bloom_filter.rs` line 2 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx5MUi0j1SLtFggXq8:-NZx5MUi0j1SLtFggXq9:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/bloom_filter.rs#L2)):*
> ```XML
> //! A Bloom Filter.
> use std::{borrow::Borrow, marker::PhantomData, os::fd::AsRawFd};
> ```

ditto
___
*[`aya/src/maps/bloom_filter.rs` line 55 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx5PzoE9u4KCGncqxD:-NZx5PzoE9u4KCGncqxE:bdkkyir) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/bloom_filter.rs#L55)):*
> ```XML
>         let fd = self.inner.borrow().fd_or_err()?;
> 
>         // TODO (AM)
> ```

can you help me understand why these TODOs are scattered about? it would seem to me that we want these map functions to take anything that implements `AsRawFd`.
___
*[`aya/src/maps/bloom_filter.rs` line 280 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx5vtg8jZ7Rt38NYLc:-NZx5vth51IdbmfKm_cq:bcvwqzk) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/bloom_filter.rs#L280)):*
> ```XML
>     }
> 
>     fn create_fd() -> OwnedFd {
> ```

can we open /dev/null or something rather than polluting the filesystem?
___
*[`aya/src/maps/lpm_trie.rs` line 463 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx6Qh9Bg9RmmYQARDb:-NZx6Qh9Bg9RmmYQARDc:b-28izvj) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/lpm_trie.rs#L463)):*
> ```XML
>     }
> 
>     fn create_fd() -> OwnedFd {
> ```

please put this helper in a shared place rather than replicating it around.
___
*[`aya/src/maps/mod.rs` line 484 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx6d39AMVANTPKCquD:-NZx6d39AMVANTPKCquE:b9cs83o) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/mod.rs#L484)):*
> ```XML
>     pub(crate) obj: obj::Map,
>     pub(crate) fd: Option<OwnedFd>,
>     pub(crate) btf_fd: Option<RawFd>,
> ```

is this owned or unowned?
___
*[`aya/src/maps/mod.rs` line 513 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx6ipg2FhSMe734Uqk:-NZx6ipg2FhSMe734Uql:b-4femll) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/mod.rs#L513)):*
> ```XML
>         self.fd = Some(fd);
> 
>         Ok(())
> ```

can you help me understand why we don't want to return a reference to the fd and save a lookup in various code paths?
___
*[`aya/src/maps/mod.rs` line 630 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx78EjEEUCroV8rpad:-NZx78EjEEUCroV8rpae:b1tnovw) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/mod.rs#L630)):*
> ```XML
>     /// Attempts to duplicate MapData
>     ///
>     /// Fails if if the inner file descriptor could not be cloned
> ```

please avoid documenting this implementation detail.
___
*[`aya/src/maps/mod.rs` line 631 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx7FyWD3s3CRmaqBWb:-NZx7FyWD3s3CRmaqBWc:btbib55) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/mod.rs#L631)):*
> ```XML
>     ///
>     /// Fails if if the inner file descriptor could not be cloned
>     // TODO (AM): should we return MapError? New variant? Maybe make MapError non_exhaustive?
> ```

please either file an issue or remove this TODO
___
*[`aya/src/maps/mod.rs` line 644 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx7KmW22L4GEf3fQBC:-NZx7KmW22L4GEf3fQBD:b-64fp6m) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/mod.rs#L644)):*
> ```XML
> }
> 
> // TODO (AM): DELETE?
> ```

yes, delete. clone should never panic.
___
*[`aya/src/maps/mod.rs` line 922 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx7V310HUugHSJ_X-Q:-NZx7V310HUugHSJ_X-R:bu14znq) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/mod.rs#L922)):*
> ```XML
>             assert_eq!(io_error.raw_os_error(), Some(EFAULT));
>         }
>         assert!(map.fd.is_none());
> ```

why is this better?
___
*[`aya/src/maps/hash_map/hash_map.rs` line 701 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx7fx4-GfIii7XfMdJ:-NZx7fx4-GfIii7XfMdK:b-jcamef) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/maps/hash_map/hash_map.rs#L701)):*
> ```XML
>     }
> 
>     fn create_fd() -> OwnedFd {
> ```

please don't duplicate this around
___
*[`aya/src/sys/bpf.rs` line 719 at r3](https://reviewable.io/reviews/aya-rs/aya/614#-NZx9oPVBPo5lXHJAUt_:-NZx9oPVBPo5lXHJAUta:bej7y9a) ([raw file](https://github.com/aya-rs/aya/blob/a0cbb5428f5c33bb77edb09451a2d68287f1c1de/aya/src/sys/bpf.rs#L719)):*
> ```XML
> 
>     // OK: we just created this fd
>     let map_fd = map_data.fd.as_ref().unwrap();
> ```

seems strictly worse

nrxus:
I think this PR has gotten pretty rotten due to it staying stale and a lot of recent changes. Would it be easier if I just closed this and started from scratch in a new PR?

tamird:
Sure. It'd be helpful (to me) if you kept changes grouped into logical commits.

nrxus:
Closing to redo this in a separate PR since there's been a lot of changes to upstream lately.

PR #617, opened 2023-05-31
aya-bpf: helpers::read_{kernel,user}_str_bytes additional len check

tomaszjonak:
Without additional check verifier fails swearing len lower bound is a negative value. Observed on v5.19/ubuntu 22.04.
[verifier error](https://github.com/aya-rs/aya/files/11611925/verifier-on-latest.txt)

### DIFF
--- a/bpf/aya-bpf/src/helpers.rs
+++ b/bpf/aya-bpf/src/helpers.rs
@@ -430,6 +430,11 @@ pub unsafe fn bpf_probe_read_user_str_bytes(
         return Err(-1);
     }
 
+    // verifier complains if this check is missing on v5.19
+    if len < 1 {
+        return Err(-1);
+    }
+
     // len includes NULL byte
     Ok(&dest[..len - 1])
 }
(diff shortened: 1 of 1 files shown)


tomaszjonak:
Whole flow is driven by `bpf_probe_read_kernel_str` behavior.

```
Description
                     Copy a NUL terminated string from an unsafe kernel
                     address unsafe_ptr to dst. Same semantics as with
                     bpf_probe_read_user_str() apply.

Return
                     On success, the strictly positive length of the
                     string, including the trailing NUL character. On
                     error, a negative value.
```

Per `bpf_probe_read_user_str` documentation.

> On success, the strictly positive length of the string, including the trailing NUL character. On error, a negative value.

If I read this right 0 length can never happen as empty string would result in solo NULL character written to buffer with 1 returned as length. If that interpretation is correct then current flow looks valid.

tomaszjonak:
Superseded by https://github.com/aya-rs/aya/pull/634

PR #625, opened 2023-06-06
aya-utils: add syscall_prefix and syscall_fnname_add_prefix

FedericoPonzi:
These two functions are needed because kernel symbols representing syscalls have architecture-specific prefixes.

These are the equivalent of bcc's get_syscall_fnname and get_syscall_prefix.

Solves: #534

Inspired by https://github.com/iovisor/bcc/blob/a078cbc7cfbdcee88c9cf3e2b76c2c8c1517e478/src/python/bcc/__init__.py#L804

### DIFF
--- a/aya/src/util.rs
+++ b/aya/src/util.rs
@@ -95,6 +95,39 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
     Ok(syms)
 }
 
+/// Returns the prefix used by syscalls.
+///
+/// # Example
+///
+/// ```no_run
+/// use aya::util::syscall_prefix;
+/// let prefix = syscall_prefix().unwrap();
+/// let syscall_fname = format!("{prefix}exec");
+/// ```
+///
+/// # Errors
+///
+/// Returns [`std::io::ErrorKind::NotFound`] if the prefix can't be guessed. Returns other [`std::io::Error`] kinds if `/proc/kallsyms` can't be opened or is somehow invalid.
+pub fn syscall_prefix() -> Result<&'static str, io::Error> {
+    const PREFIXES: [&str; 7] = [
+        "sys_",
+        "__x64_sys_",
+        "__x32_compat_sys_",
+        "__ia32_compat_sys_",
+        "__arm64_sys_",
+        "__s390x_sys_",
+        "__s390_sys_",
+    ];
+    let ksym = kernel_symbols()?;
+    for p in PREFIXES {
+        let prefixed_syscall = format!("{}bpf", p);
+        if ksym.values().any(|el| *el == prefixed_syscall) {
+            return Ok(p);
+        }
+    }
+    Err(io::ErrorKind::NotFound.into())
+}
+
 pub(crate) fn ifindex_from_ifname(if_name: &str) -> Result<u32, io::Error> {
     let c_str_if_name = CString::new(if_name)?;
     let c_if_name = c_str_if_name.as_ptr();


aya/src/util.rs
```diff
@@ -95,6 +95,36 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
     Ok(syms)
 }
 
+/// Find current system's syscall prefix by testing on the BPF syscall.
```
alessandrod, line 98: This should return a io::ErrorKind::Other or maybe even io::ErrorKind::NotFound if a prefix can't be found. Also the docs should be something like:

```
/// Returns the prefix used by syscalls.
///
/// # Example
///
/// ...example that shows how to use format to prefix a syscall with this prefix...
///
/// # Errors
///
/// Returns an [`std::io::Error`] if kernel symbols can't be found or the prefix can't be determined.
```

aya/src/util.rs
```diff
@@ -95,6 +95,36 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
...
+            return Ok(p);
+        }
+    }
+    return Ok(prefixes[0])
+}
+
+/// Given a name, it finds and append the system's syscall prefix to it.
+/// This function doesn't check if the name is for an existing syscall.
+/// For example, given "clone" the helper would return "sys_clone" or "__x64_sys_clone".
+pub fn syscall_fnname_add_prefix(name: &str) -> Result<String, io::Error>{
```
alessandrod, line 124: I don't think we should provide this. It's essentially a more complex name for format!(), and it encourages bad perf eg:

```
let a = syscall_fnname_add_prefix(...)?;
let b = syscall_fnname_add_prefix(...)?;
let c = syscall_fnname_add_prefix(...)?;
```

Will parse and iterate over /proc/kallsyms three times, when the following doesn't:

```
let prefix = syscall_prefix()?;
let a = format!("{prefix}foo");
let b = format!("{prefix}bar");
let c = format!("{prefix}baz");
```

alessandrod (changes requested):
Thanks for the PR! See comments

aya/src/util.rs
FedericoPonzi: Agree, the python and go implementations seem to be using a cache: https://github.com/iovisor/bcc/blob/a078cbc7cfbdcee88c9cf3e2b76c2c8c1517e478/src/cc/bcc_syms.cc#L600 
I've removed this function, it might be added later if the kernel symbols cache gets added

aya/src/util.rs
```diff
@@ -95,6 +95,34 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
     Ok(syms)
 }
 
+/// Returns the prefix used by syscalls.
```
alessandrod, line 98: Please add a new line after this line

aya/src/util.rs
```diff
@@ -95,6 +95,34 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
...
+/// ## Example
```
alessandrod, line 99: This should be # not ##, and add a newline after

aya/src/util.rs
```diff
@@ -95,6 +95,34 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
...
+/// ```no_run
+/// let prefix = syscall_prefix()?;
+/// let syscall_fname = format!("{prefix}exec");
+/// ```
+/// ## Errors
```
alessandrod, line 104: Newline before and after please, also # not ##

Also the doc is wrong, it returns other error kinds as well. It should say:

```
Returns [`std::io::ErrorKind::NotFound`] if the prefix can't be guessed. Returns other [`std::io::Error`] kinds if `/proc/kallsyms` can't be opened or is somehow invalid.
```

aya/src/util.rs
```diff
@@ -95,6 +95,39 @@ fn parse_kernel_symbols(reader: impl BufRead) -> Result<BTreeMap<u64, String>, i
...
+        "sys_",
+        "__x64_sys_",
+        "__x32_compat_sys_",
+        "__ia32_compat_sys_",
+        "__arm64_sys_",
+        "__s390x_sys_",
+        "__s390_sys_",
+    ];
+    let ksym = kernel_symbols()?;
+    let values = ksym.into_values().collect::<Vec<_>>();
```
alessandrod, line 122: I don't think we should collect here, you can use `ksym.values()` multiple times

FedericoPonzi: because there is no contains in values(), I've replaced contains() with `any()`

alessandrod (approved):

PR #632, opened 2023-06-28
aya: allow global value to be optional

marysaka:
This allow to not error out when a global symbol is missing from the object.

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -203,6 +203,7 @@ impl<'a> BpfLoader<'a> {
     }
 
     /// Sets the value of a global variable.
+    ///
     /// If the `must_exist` argument is `true`, [`BpfLoader::load`] will fail with [`ParseError::SymbolNotFound`] if the loaded object code does not contain the variable.
     ///
     /// From Rust eBPF, a global variable can be defined as follows:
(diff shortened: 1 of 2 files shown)


aya/src/bpf.rs
```diff
@@ -233,8 +233,8 @@ impl<'a> BpfLoader<'a> {
     /// use aya::BpfLoader;
```
alessandrod, line 233: Here before the # Example section we should add

```
If the `must_exist` argument is `true`, [`BpfLoader::load`] will fail with [`ParseError::SymbolNotFound`] if the loaded object code does not contain the variable.
```

You may need to tweak the rustdoc refs, one day I'll learn how they work

aya/src/bpf.rs
```diff
@@ -243,8 +243,9 @@ impl<'a> BpfLoader<'a> {
         &mut self,
         name: &'a str,
         value: T,
+        is_required: bool,
```
alessandrod, line 246: I think `must_exist` would be a better name for this

alessandrod (changes requested):
Looks good, just a couple of minor nits!

aya/src/bpf.rs
```diff
@@ -203,6 +203,7 @@ impl<'a> BpfLoader<'a> {
     }
 
     /// Sets the value of a global variable.
+    /// If the `must_exist` argument is `true`, [`BpfLoader::load`] will fail with [`ParseError::SymbolNotFound`] if the loaded object code does not contain the variable.
```
alessandrod, line 206:
```suggestion
    ///
    /// If the `must_exist` argument is `true`, [`BpfLoader::load`] will fail with [`ParseError::SymbolNotFound`] if the loaded object code does not contain the variable.
```

PR #638, opened 2023-07-06
integration-test: Implement running on VMs

ajwerner:


ajwerner:
Probably we should get this into CI. It'll need a way to mark certain tests as not working in this environment or a way to plumb in the necessary dependencies.

ajwerner (commented):
I'm going to leave subsuming `run.sh` for later when I understand the world better. We'll need https://github.com/aya-rs/aya/pull/639 and we'll need to rework the relocation tests to build their probes at compile-time instead of at runtime. I'll tackle that next. 

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638#-:-NZl-JZ1981H442kBSCM:b-jz3rra)* status: 6 of 8 files reviewed, 5 unresolved discussions (waiting on @tamird)

___
*[`kerneltest/.gitkeep` line  at r2](https://reviewable.io/reviews/aya-rs/aya/638#-NZh7dJzE3yQ51AyFHUb:-NZl0nIuAqp_RRynUzR4:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/65079fe29b4d191bd5d2694c0d78efc31020ca93/kerneltest/.gitkeep#L)):*

remove? only need the deepest one
</blockquote></details>

Done.
___
*[`xtask/Cargo.toml` line 17 at r2](https://reviewable.io/reviews/aya-rs/aya/638#-NZh7y7h2j03gqihiUkp:-NZkyIFV61-qSj6QgIrk:bo2vwb4) ([raw file](https://github.com/aya-rs/aya/blob/65079fe29b4d191bd5d2694c0d78efc31020ca93/xtask/Cargo.toml#L17)):*

maybe go all the way and newline every attribute?

foo  = {
  bar = baz,
  features = [
    ...
  ],
}
</blockquote></details>

Seems like that's not legal toml. 
___
*[`xtask/src/integration_test.rs` line 44 at r1](https://reviewable.io/reviews/aya-rs/aya/638#-NZh6cBn6Do_IjU9bbL2:-NZkyHEeEVmw7AqJvB4G:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/5188b6d07ff3d91e8b2bad2e4d7e9562c44a7150/xtask/src/integration_test.rs#L44)):*

can we make `Options::libbpf_dir` a `PathBuf`?
</blockquote></details>

Done.
___
*[`xtask/src/integration_test.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/638#-NZh6Mor8AFb3VTo0n-8:-NZl-5tk97WCkOpXcRKB:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/5188b6d07ff3d91e8b2bad2e4d7e9562c44a7150/xtask/src/integration_test.rs#L52)):*

This field is never not None in this commit. Remove?
</blockquote></details>

Done.
___
*[`xtask/src/integration_test.rs` line 78 at r1](https://reviewable.io/reviews/aya-rs/aya/638#-NZh77qzDj2QbnWHN3KL:-NZkwsWf6BAFDPa45Vpq:b-igaklc) ([raw file](https://github.com/aya-rs/aya/blob/5188b6d07ff3d91e8b2bad2e4d7e9562c44a7150/xtask/src/integration_test.rs#L78)):*

attempting to live in String and PathBuf at the same time is a recipe for pain. See the fallible conversion you now have to do about 10 lines down.

Either everything is PathBuf (and by implication OsString) or everything is String.
</blockquote></details>

Okay, `String` for all the things.

tamird (commented):
Agreed, but I think we should treat that work as a prerequisite of this work. We should limit the number of ways we run VMs to 1.

> Reviewed 7 of 7 files at r3, 5 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638#-:-NZlDv5N5sJdKvZa29PZ:b-yagvub)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @ajwerner)

___
*[`xtask/Cargo.toml` line 8 at r4](https://reviewable.io/reviews/aya-rs/aya/638#-NZlE9hhCjMGUMezFlDy:-NZlE9hhCjMGUMezFlDz:bdr454a) ([raw file](https://github.com/aya-rs/aya/blob/7b0216e4e513c0d64712e78a274a386631a5a115/xtask/Cargo.toml#L8)):*
> ```TOML
> 
> [dependencies]
> reqwest = { version = "0.11.18", default-features = false, features = [
> ```

why did this move?
___
*[`xtask/src/integration_test.rs` line 78 at r1](https://reviewable.io/reviews/aya-rs/aya/638#-NZh77qzDj2QbnWHN3KL:-NZlD__XEVDi7h4ON2Qi:b-33p9sk) ([raw file](https://github.com/aya-rs/aya/blob/5188b6d07ff3d91e8b2bad2e4d7e9562c44a7150/xtask/src/integration_test.rs#L78)):*

Okay, `String` for all the things.
</blockquote></details>

Per discussion, let's go the other way. See https://github.com/aya-rs/aya/pull/640.

tamird (commented):
> Reviewed 47 of 47 files at r15, 47 of 47 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638)* status: 49 of 56 files reviewed, 11 unresolved discussions (waiting on @ajwerner)

___
*[`.github/workflows/ci.yml` line 222 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqtLoXD5rfwzy-HUUm:-NaqtLoXD5rfwzy-HUUn:b-w1k3z5) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/.github/workflows/ci.yml#L222)):*
> ```YAML
>             linux-${{ matrix.kernel-version }}.bz
> 
>       - name: Pull image file 
> ```

let's use the script here
___
*[`test/README.md` line 41 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqsseO8E9TKu4aCD72:-NaqsseO8E9TKu4aCD73:b-l6mwxj) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/README.md#L41)):*
> ```Markdown
> ``
> 
> mkdir -p integration-test-binaries
> ```

remove
___
*[`test/test.sh` line 1 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqqywSETSX74d_H1V6:-NaqqywSETSX74d_H1V7:b3hlryn) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L1)):*
> ```Shell
> #!/bin/bash
> ```

usr/bin/env bash
___
*[`test/test.sh` line 5 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqqepuCBtpipflp_Z8:-NaqqepuCBtpipflp_Z9:bqcrsf5) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L5)):*
> ```Shell
> set -euo pipefail
> 
> # Helper function to print the usage
> ```

comments end in periods plz
___
*[`test/test.sh` line 7 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqshnV0ZwJ6I9Xjk3G:-NaqshnV0ZwJ6I9Xjk3H:b-uq7tjn) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L7)):*
> ```Shell
> # Helper function to print the usage
> print_usage() {
>     echo "Usage: $0 [SUBCOMMAND] [OPTIONS]"
> ```

nit: single quotes where double not needed
___
*[`test/test.sh` line 12 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqqlki4XVfiP5B1axG:-Naqqlki4XVfiP5B1axH:b-kl4lrg) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L12)):*
> ```Shell
>     echo "                  Options:"
>     echo "                      --output <path>       Path to save the built image. (default: ./initrd_image.cpio)"
>     echo "                      --cargo-arg <arg>     Argument for the cargo command (can be repeated)."
> ```

". (can be repeated)" for consistency with previous

but also do we want to deal with this? maybe we say anything other than output is passthrough to cargo xtask build-integration-test?
___
*[`test/test.sh` line 14 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqqd7H7GNGSNZb8VJj:-Naqqd7H7GNGSNZb8VJk:b-3umkum) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L14)):*
> ```Shell
>     echo "                      --cargo-arg <arg>     Argument for the cargo command (can be repeated)."
>     echo
>     echo "  download-kernel  Download a kernel of a specified version."
> ```

misaligned
___
*[`test/test.sh` line 27 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqsjnx8XjJiIwMMmuu:-Naqsjnx8XjJiIwMMmuv:b9o4rpb) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L27)):*
> ```Shell
> # Subcommand: build
> build() {
>     local output="./initrd_image.cpio"
> ```

nit: no quotes where not needed
___
*[`test/test.sh` line 40 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqrlz-Ay-ERFImiHN5:-Naqrlz-Ay-ERFImiHN6:b-wmbo23) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L40)):*
> ```Shell
> 
>     cargo xtask build-integration-test "${cargo_args[@]}" \
>         | xargs -I {} bluebox -e {}:"--test-threads=1" -o "$output"
> ```

knowledge being encoded at vast distance - we should have a comment here and the same one in the xtask code that does the same
___
*[`test/test.sh` line 90 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqsRKy3FvdFvGZvU8S:-NaqsRKy3FvdFvGZvU8T:bgz11y5) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L90)):*
> ```Shell
>     fi
> 
>     log_file="$logs/$(basename "$initrd").log"
> ```

let's just use a randomly generated name (and skip the directory)
___
*[`test/test.sh` line 122 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqsfuqB7NY0nQs_P2C:-NaqsfuqB7NY0nQs_P2D:b-1m9ux7) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L122)):*
> ```Shell
>     run) shift; run "$@" ;;
>     *) echo "Unknown subcommand: $1"; print_usage; exit 1 ;;
> esac
> ```

missing newline

ajwerner (commented):
TFTR

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638#-:-Nar44ltDF3BIe-mLMKO:bb73zsl)* status: 48 of 56 files reviewed, 9 unresolved discussions (waiting on @tamird)

___
*[`.github/workflows/ci.yml` line 222 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqtLoXD5rfwzy-HUUm:-Naqzc__8Gkerw4ytYBw:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/.github/workflows/ci.yml#L222)):*

let's use the script here
</blockquote></details>

Done.
___
*[`test/README.md` line 41 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqsseO8E9TKu4aCD72:-Nar17L3BhVHMK2netK6:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/README.md#L41)):*

remove
</blockquote></details>

Done.
___
*[`test/test.sh` line 1 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqqywSETSX74d_H1V6:-Nar1BHe6jNDXLufRfKF:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L1)):*

usr/bin/env bash
</blockquote></details>

Done.
___
*[`test/test.sh` line 5 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqqepuCBtpipflp_Z8:-Nar1G4N0cnMUm3xyx3D:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L5)):*

comments end in periods plz
</blockquote></details>

Done.
___
*[`test/test.sh` line 12 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqqlki4XVfiP5B1axG:-Nar1TemCBUyf8SGlJ1M:bz1ziau) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L12)):*

". (can be repeated)" for consistency with previous

but also do we want to deal with this? maybe we say anything other than output is passthrough to cargo xtask build-integration-test?
</blockquote></details>

meh, I like that it's strict. If the use case comes up, we can make it more flexible. 
___
*[`test/test.sh` line 14 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqqd7H7GNGSNZb8VJj:-Nar1YnSALbLhbXOGYvA:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L14)):*

misaligned
</blockquote></details>

Done.
___
*[`test/test.sh` line 27 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqsjnx8XjJiIwMMmuu:-Naqzhp6-Hidf4X6E8q2:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L27)):*

nit: no quotes where not needed
</blockquote></details>

Done.
___
*[`test/test.sh` line 40 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-Naqrlz-Ay-ERFImiHN5:-Nar106lBSW7YKOnagUO:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L40)):*

knowledge being encoded at vast distance - we should have a comment here and the same one in the xtask code that does the same
</blockquote></details>

Done.
___
*[`test/test.sh` line 90 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqsRKy3FvdFvGZvU8S:-Nar1-q77ssMnI8gllOu:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L90)):*

let's just use a randomly generated name (and skip the directory)
</blockquote></details>

Done.
___
*[`test/test.sh` line 122 at r16](https://reviewable.io/reviews/aya-rs/aya/638#-NaqsfuqB7NY0nQs_P2C:-Nar1aBCDPR2glkKiKaC:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4da4a1c73e156cfdef6983d4eefb017c52caea59/test/test.sh#L122)):*

missing newline
</blockquote></details>

Done.
___
*[`xtask/Cargo.toml` line 8 at r4](https://reviewable.io/reviews/aya-rs/aya/638#-NZlE9hhCjMGUMezFlDy:-Nar13ZL7AjsbSD4hTp-:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7b0216e4e513c0d64712e78a274a386631a5a115/xtask/Cargo.toml#L8)):*

why did this move?
</blockquote></details>

Done.

tamird (changes requested):
> Reviewed 1 of 1 files at r17, 6 of 6 files at r18, 6 of 6 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638)* status: 50 of 56 files reviewed, 8 unresolved discussions (waiting on @ajwerner)

___
*[`.github/workflows/ci.yml` line 187 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NarfwjWDWhr2bMO8i4k:-NarfwjWDWhr2bMO8i4l:bdms0k1) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L187)):*
> ```YAML
>           set -euxo pipefail
>           mkdir -p integration-test-images
>           mkdir -p integration-test-images
> ```

duplicate line?
___
*[`.github/workflows/ci.yml` line 197 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-Narg5Q-E4ANopvcS9Uq:-Narg5Q-E4ANopvcS9Ur:b-3etzl) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L197)):*
> ```YAML
>             echo "::endgroup::"
>           done
>           for profile in dev release; do
> ```

duplicate stanza
___
*[`.github/workflows/ci.yml` line 228 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargB8M-n18ClDy8GH4:-NargB8M-n18ClDy8GH5:bjslgvx) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L228)):*
> ```YAML
>           sparse-checkout: test/test.sh
> 
>       - name: Download Kernel Image
> ```

nit: can we pick a consistent capitalization? below there is "Run integration tests" and above there is "Install Pre-requisites". I think just capitalizing the first word is good. Perhaps "Installing prerequisites" above as well while you're at it, since that's a word.
___
*[`.github/workflows/ci.yml` line 230 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargPnJ0lPMRvvoFqjd:-NargPnJ0lPMRvvoFqje:bqgdn45) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L230)):*
> ```YAML
>       - name: Download Kernel Image
>         run: |
>           mkdir -p kernel-images
> ```

can you add a `set -euxo pipefail` while you're making this multi-statement?
___
*[`.github/workflows/ci.yml` line 241 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargeSH2dTpk8nZkm2W:-NargeSH2dTpk8nZkm2X:bjdorzj) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L241)):*
> ```YAML
> 
>       - name: Run integration tests
>         run: |
> ```

`set -euxo pipefail`
___
*[`.github/workflows/ci.yml` line 250 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargdK00IahogsP0yrN:-NargdK00IahogsP0yrO:boaffr2) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L250)):*
> ```YAML
>             echo "::endgroup::"
>           done
>           for image in integration-test-images/*; do
> ```

duplicated
___
*[`test/test.sh` line 41 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-Narh0zI0WwnrKjIl2X6:-Narh0zI0WwnrKjIl2X7:bmifl59) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/test/test.sh#L41)):*
> ```Shell
>     if [[ -z "$output" ]]; then
>         mkdir -p .tmp
>         output=./.tmp/initrd_image.cpio
> ```

what happens when i run this script from not the directory that contains it? e.g. `test/test.sh` as described in the README.
___
*[`test/test.sh` line 73 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NarhCFQBEsUR9oQFd74:-NarhCFQBEsUR9oQFd75:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/test/test.sh#L73)):*
> ```Shell
> 
>     if [[ -z "$output" ]]; then
>         mkdir -p .tmp
> ```

ditto

test/test.sh
```diff
@@ -0,0 +1,134 @@
...
+    qemu-system-x86_64 \
+        -no-reboot \
+        -append 'printk.devkmsg=on kernel.panic=-1 crashkernel=256M noapic' \
+        -kernel "$kernel" \
+        -initrd "$initrd" \
+        -nographic \
+        -append 'console=ttyS0' \
+        -m '512M' \
+    | tee -a "$log_file"
+    awk '
```
dave-tucker, line 112: This won't do the right thing on non-x86 systems.
I'm also skeptical that it will use hardware assisted virtualization if available - this was all formerly baked into the old script.

dave-tucker (changes requested):
Can you please add some more information to the PR description:

- What's deficient with the existing test harness
- What you aim to get by replacing it

You say "simplify" but cloud-init + cloud vendor images are a know quantity to me.
This new-fangled bluebox thing looks interesting, but how widely used is it? and how well supported?

Similarly with cilium-ci/kernels. What's the advantage to testing against Cilium's testing kernels, built with their own custom Kconfigs vs. using distro cloud images?
It seems they are following LTS kernels, but we're pretty likely to want to test things that are not in LTS kernels yet - what's the strategy there?

I would imagine that the testing goal of Parca - cited as an inspiration -  is "does my program pass the verifier on this kernel version".  As a library we're more concerned with the BPF features that are enabled/disabled via Kconfig and/or whether or not certain features are actually available because they were backported making our kernel version checks bogus.

This could (probably) be achieved by overriding feature detection in our testing - to ensure older codepaths are followed.

The other concern is whether or not this might limit our ability to test certain things in the integration tests - i.e things like uprobes or usdts which might have expectations around libc, ldsocache etc.. being available.

ajwerner (commented):
> What's deficient with the existing test harness

My goals when starting this were:
1) to be able to test probes across kernel versions
2) to rapidly iterate on tests without the cycle time required in the current virtualized flow

The existing flow is slow and heavy. I aim to get super rapid iteration and then validation across kernel versions.

I feel like the ultimate right answer is to set up a mechanism where we can use an unmodified filesystem image as userland and then shove the tests and probes into it. For some use cases that's better, but for others it just means downloading a gigabyte to throw it away and doing a much longer boot process. I believe that between what's in this PR and what it's replacing we're going to find a happy medium.

> You say "simplify" but cloud-init + cloud vendor images are a know quantity to me.

That's somewhat legit. The bluebox thing wins in its simplicity more than anything. There are literally no variables in play because there's no filesystem and no binaries -- just you and the kernel. What it might lose in notoriety it wins in being a legitimately tiny amount of code. 

>Similarly with cilium-ci/kernels. What's the advantage to testing against Cilium's testing kernels, built with their own custom Kconfigs vs. using distro cloud images?

I'm hardly wedded to `cillium-ci/kernels`, they were just an easy thing to reach for to begin. I also downloaded and used some ubuntu kernels and the infrastructure here all works. It's quite nice to be able to just swap out the kernel image without having to swap out the entire disk image too. For example, we could just as easily download from, say, https://kernel.ubuntu.com/~kernel-ppa/mainline/v6.4/amd64/ or I'm sure there's similar servers for other distros. If anything, this approach makes it much easier to test against a full matrix of Kconfigs.

> The other concern is whether or not this might limit our ability to test certain things in the integration tests - i.e things like uprobes or usdts which might have expectations around libc, ldsocache etc.. being available.

This is a legitimate concern. For a lot of the libc and ldsocache interactions, you don't need to exercise the bpf at the same time, so they don't need to be integration tests. I can, however, imagine wanting some degree of testing of interactions between all of these moving pieces, and this doesn't tell a good story there.

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638#-:-NarpF7F1S-BgUuTCviS:bhagqvz)* status: 48 of 56 files reviewed, 10 unresolved discussions (waiting on @dave-tucker and @tamird)

___
*[`.github/workflows/ci.yml` line 181 at r17](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282108296:-Narr3_l6KgZ3RR0DKKg:bk2y1hq) ([raw file](https://github.com/aya-rs/aya/blob/83079ce02c6ff7c881f3639ec55033a4b6beee53/.github/workflows/ci.yml#L181)):*

Is this actively maintained? Are many other people using it?
The lack of a shell seems that it could be problematic in the long term.
What's the plan to test things like USDTs and/or symbol resolution for uprobe programs?
</blockquote></details>

See main comment thread.
___
*[`.github/workflows/ci.yml` line 220 at r17](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282120044:-Narr-U0EwggI5oHFPWa:bk2y1hq) ([raw file](https://github.com/aya-rs/aya/blob/83079ce02c6ff7c881f3639ec55033a4b6beee53/.github/workflows/ci.yml#L220)):*

Regarding cilium ci-kernels, what's the advantage to testing on kernels created from a custom kernel config vs. real kernels used in real distros. It's generally the distro vendors decision to backport (or not) or include certain kernel config flags (or not) that is the source of issues - combined with our assumptions about what is or is not supported in these kernel versions.
</blockquote></details>

See main comment thread.
___
*[`.github/workflows/ci.yml` line 187 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NarfwjWDWhr2bMO8i4k:-NarkVhZ4BW8YoBsP4xD:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L187)):*

duplicate line?
</blockquote></details>

Done.
___
*[`.github/workflows/ci.yml` line 197 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-Narg5Q-E4ANopvcS9Uq:-NarkW5F4YRx6H93NBDg:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L197)):*

duplicate stanza
</blockquote></details>

Done.
___
*[`.github/workflows/ci.yml` line 228 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargB8M-n18ClDy8GH4:-NarkdHi9fHhK6Vx5twm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L228)):*

nit: can we pick a consistent capitalization? below there is "Run integration tests" and above there is "Install Pre-requisites". I think just capitalizing the first word is good. Perhaps "Installing prerequisites" above as well while you're at it, since that's a word.
</blockquote></details>

Done.
___
*[`.github/workflows/ci.yml` line 230 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargPnJ0lPMRvvoFqjd:-NarkdwW9XR8NhBggG9j:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L230)):*

can you add a `set -euxo pipefail` while you're making this multi-statement?
</blockquote></details>

Done.
___
*[`.github/workflows/ci.yml` line 241 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargeSH2dTpk8nZkm2W:-NarkfRe8bmFT4ygac0A:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L241)):*

`set -euxo pipefail`
</blockquote></details>

Done.
___
*[`.github/workflows/ci.yml` line 250 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NargdK00IahogsP0yrN:-NarkgcID3MbqKXB4pjg:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/.github/workflows/ci.yml#L250)):*

duplicated
</blockquote></details>

Done.
___
*[`test/test.sh` line 41 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-Narh0zI0WwnrKjIl2X6:-NarmNon-hu7Fcmntg3W:b-lx4fj3) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/test/test.sh#L41)):*

what happens when i run this script from not the directory that contains it? e.g. `test/test.sh` as described in the README.
</blockquote></details>

Fixed. 
___
*[`test/test.sh` line 73 at r19](https://reviewable.io/reviews/aya-rs/aya/638#-NarhCFQBEsUR9oQFd74:-NarmOrSDWlaK-nIvBUx:bt2vr1b) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/test/test.sh#L73)):*

ditto
</blockquote></details>

Fixed.
___
*[`test/test.sh` line 112 at r19](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282343262:-Narr5wO4eOUmn4prOLg:bq7l4r) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/test/test.sh#L112)):*

This won't do the right thing on non-x86 systems.
I'm also skeptical that it will use hardware assisted virtualization if available - this was all formerly baked into the old script.
</blockquote></details>

Good callout, @tamird will do something about this.

test/README.md
```diff
@@ -17,7 +17,8 @@ To run locally all you need is:
 1. A POSIX shell
 1. `rustup target add x86_64-unknown-linux-musl`
 1. `cargo install bpf-linker`
-1. Install `qemu` and `cloud-init-utils` package - or any package that provides `cloud-localds`
+1. `go install github.com/florianl/bluebox@latest`
```
dave-tucker, line 20: I don't love that we need a go toolchain now :(
Is there a way we can use binary installs?

In fact even better would be for `xtask` to manage the version installed in somewhere like `xtask/bin`

test/README.md
```diff
@@ -32,9 +33,11 @@ cargo xtask integration-test
 ### Virtualized
```
dave-tucker, line 33: While we're here, can we make Virtualized the default way of running the tests - via xtask of course.
This has long been on the TODO list since we know the tests don't always clean up after themselves.

test/test.sh
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env bash
```
dave-tucker, line 1: Can we get rid of this completely and instead use xtask?

test/test.sh
```diff
@@ -0,0 +1,200 @@
...
+            machine+=',accel=hvf'
+        elif [[ -c /dev/kvm ]]; then
+            machine+=',accel=kvm'
+        fi
+    fi
+
+    log_file=$(mktemp)
+    trap 'rm -rf "$log_file"' EXIT
+
+    qemu-system-"$host_arch" \
```
dave-tucker, line 155: This should be `qemu-system-"$guest_arch"` otherwise on ARM you're trying to run an x86_64 guest with `qemu-system-x86_64` which won't work.

.github/workflows/ci.yml
```diff
@@ -174,53 +175,72 @@ jobs:
...
       - name: Build
         run: |
           set -euxo pipefail
-          mkdir -p integration-test-binaries
+          mkdir -p integration-test-images
           # See https://doc.rust-lang.org/cargo/reference/profiles.html for the
           # names of the builtin profiles. Note that dev builds "debug" targets.
-          cargo xtask build-integration-test --cargo-arg=--profile=dev | xargs -I % cp % integration-test-binaries/dev
-          cargo xtask build-integration-test --cargo-arg=--profile=release | xargs -I % cp % integration-test-binaries/release
+          for profile in dev release; do
```
dave-tucker, line 189: Wouldn't it be simpler to add `profile` to the `matrix`?

.github/workflows/ci.yml
```diff
@@ -174,53 +175,72 @@ jobs:
...
-          path: integration-test-binaries
+          name: integration-test-images
+          path: integration-test-images
 
       - name: Run integration tests
         run: |
           set -euxo pipefail
-          find integration-test-binaries -type f -exec chmod +x {} \;
-          test/run.sh integration-test-binaries
+          for image in integration-test-images/*; do
```
dave-tucker, line 236: As above, this can be simplified by moving `profile` into the matrix

tamird (commented):
> Reviewed 2 of 2 files at r20, 1 of 1 files at r21, 6 of 6 files at r22, 6 of 6 files at r23, all commit messages.
> Dismissed @dave-tucker from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638)* status: 50 of 56 files reviewed, 9 unresolved discussions (waiting on @ajwerner and @dave-tucker)

___
*[`.github/workflows/ci.yml` line 189 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1283008863:-Nav_q8t5vjdRcYHSjhf:b-4syngw) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/.github/workflows/ci.yml#L189)):*

Wouldn't it be simpler to add `profile` to the `matrix`?
</blockquote></details>

It might be, but two things:
- the overhead of setting up the task is non-trivial relative to the work we're actually doing here
- i'm going to rework this a bit so that we have just a single ramdisk that runs both dev and release test binaries one after the other -- this will ensure that our tests are properly cleaning up after themselves.
___
*[`test/README.md` line 20 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282987047:-NavaBbFAhH92XwEK3Np:bj48ups) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/README.md#L20)):*

I don't love that we need a go toolchain now :(
Is there a way we can use binary installs?

In fact even better would be for `xtask` to manage the version installed in somewhere like `xtask/bin`
</blockquote></details>

Unfortunately, no. Even if we shipped around a binary of the bluebox command, it internally generates a Go program and uses a Go toolchain to compile it, so we need one on hand. I think I'll implement a graceful fallback to run-on-not-VM in `xtask integration-test` if the user doesn't have a Go toolchain, though.
___
*[`test/README.md` line 33 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282989632:-NavaRqx8RZ7rS4EVPSC:b-oyp2ys) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/README.md#L33)):*

While we're here, can we make Virtualized the default way of running the tests - via xtask of course.
This has long been on the TODO list since we know the tests don't always clean up after themselves.
</blockquote></details>

Yes! I'll do this.
___
*[`test/test.sh` line 112 at r19](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282343262:-Naryx2yCOlnTW5EigGw:b6nmnec) ([raw file](https://github.com/aya-rs/aya/blob/4c985ad74d3584a0015294ac55a828787a5c98d5/test/test.sh#L112)):*

Good callout, @tamird will do something about this. 
</blockquote></details>

Done; now uses acceleration. You pointed out it's the wrong binary, there's another comment thread for that.
___
*[`test/test.sh` line 1 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282993401:-Nav_kpMAjuLsZTRwwz1:b1p8d7p) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/test.sh#L1)):*

Can we get rid of this completely and instead use xtask?
</blockquote></details>

Good idea! Along with the comment to use the VM by default I think this will shake out pretty nice. I will do that today.
___
*[`test/test.sh` line 155 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1283001324:-NavaV68BXvbMvUodtcp:b-76wbio) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/test.sh#L155)):*

This should be `qemu-system-"$guest_arch"` otherwise on ARM you're trying to run an x86_64 guest with `qemu-system-x86_64` which won't work.
</blockquote></details>

Ack. Will fix today.

tamird (commented):
> Dismissed @dave-tucker from a discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638)* status: 50 of 56 files reviewed, 6 unresolved discussions (waiting on @dave-tucker)

___
*[`.github/workflows/ci.yml` line 236 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1283009630:-NavaqYqDGMK8IvhNSJ0:bjdiqfh) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/.github/workflows/ci.yml#L236)):*

As above, this can be simplified by moving `profile` into the matrix
</blockquote></details>

Resolving; let's discuss up there.

dave-tucker:
> Unfortunately, no. Even if we shipped around a binary of the bluebox command, it internally generates a Go program and uses a Go toolchain to compile it, so we need one on hand. I think I'll implement a graceful fallback to run-on-not-VM in xtask integration-test if the user doesn't have a Go toolchain, though.

That seems overly complicated - while I also have to use Go for $dayjob it would be nice not to need it for the tests.
I don't think you need to find an alternative now, but finding one soon might be a decent idea.

I think busybox might work pretty well for this, as used by Rust-for-Linux:
See: https://github.com/Rust-for-Linux/linux/blob/rust/.github/workflows/qemu-initramfs.desc
Which can be easily adapted to add our test binaries

And assembled using scripts that are maintained as part of the kernel:
https://github.com/Rust-for-Linux/linux/blob/18b7491480025420896e0c8b73c98475c3806c6f/.github/workflows/ci.yaml#L237-L240

/usr/gen_init_cpio.c can be compiled very quickly - and we already require clang.

tamird (commented):
> Reviewed 8 of 8 files at r24.
> Dismissed @dave-tucker from 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638)* status: 46 of 64 files reviewed, all discussions resolved (waiting on @dave-tucker)

___
*[`test/README.md` line 20 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282987047:-NbGpXEF0YwMvIpvkBiI:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/README.md#L20)):*

Unfortunately, no. Even if we shipped around a binary of the bluebox command, it internally generates a Go program and uses a Go toolchain to compile it, so we need one on hand. I think I'll implement a graceful fallback to run-on-not-VM in `xtask integration-test` if the user doesn't have a Go toolchain, though.
</blockquote></details>

Done.
___
*[`test/README.md` line 33 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282989632:-NbGpYT02DjRr6sWuh1A:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/README.md#L33)):*

Yes! I'll do this.
</blockquote></details>

Done.
___
*[`test/test.sh` line 1 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1282993401:-NbGpan80FG2BXw8AnFe:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/test.sh#L1)):*

Good idea! Along with the comment to use the VM by default I think this will shake out pretty nice. I will do that today.
</blockquote></details>

Done.
___
*[`test/test.sh` line 155 at r23](https://reviewable.io/reviews/aya-rs/aya/638#gh-1283001324:-NbGpcDV4mOQbzx0JF7Q:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2bc93af95d3f858eb7214dc7c5fca6915abe9451/test/test.sh#L155)):*

Ack. Will fix today.
</blockquote></details>

Done.

ajwerner (commented):
Took a look, seems like it's getting close. 

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/638#-:-NbKNDIm95l2s5E2nykV:b-e5hc0f)* status: 46 of 64 files reviewed, 1 unresolved discussion (waiting on @dave-tucker and @tamird)

___
*[`xtask/src/qemu-initramfs.desc` line 1 at r27](https://reviewable.io/reviews/aya-rs/aya/638#-NbKLxKK3Hrgj8xq1thU:-NbKLxKK3Hrgj8xq1thV:bz0f92v) ([raw file](https://github.com/aya-rs/aya/blob/f9550fa5724b9d504e233b2565b81299922008f6/xtask/src/qemu-initramfs.desc#L1)):*
> ```
> dir     /bin                                          0755 0 0
> ```

This file doesn't seem to be used anymore.

ajwerner:
This has been through too many iterations to be merged in this PR. @tamird to open a new PR.

tamird:
#733.

PR #639, opened 2023-07-06
Remove dependency on bpftool in integration tests

tamird:


xtask/src/build_ebpf.rs
```diff
@@ -77,9 +77,12 @@ fn build_rust_ebpf(opts: &BuildEbpfOptions) -> anyhow::Result<()> {
 fn get_libbpf_headers<P: AsRef<Path>>(libbpf_dir: P, include_path: P) -> anyhow::Result<()> {
     let dir = include_path.as_ref();
     fs::create_dir_all(dir)?;
+    let mut includedir = std::ffi::OsString::new();
```
alessandrod, line 80: here and everywhere else - `use` instead of using full paths?

alessandrod (commented):
Looks great, just minor nit about full paths

xtask/src/build_ebpf.rs
tamird: done.

alessandrod (approved):

PR #643, opened 2023-07-11
Remove procfs dependency

tamird:
- Remove verifier log special case
- Remove procfs dependency
- Rewrite kernel version logic

This restores and enhances the logic originally added in #579.

### DIFF
--- a/aya/src/util.rs
+++ b/aya/src/util.rs
@@ -3,8 +3,10 @@ use std::{
     collections::BTreeMap,
     ffi::CString,
     fs::{self, File},
-    io::{self, BufReader},
-    mem, slice,
+    io::{self, BufRead, BufReader},
+    mem,
+    num::ParseIntError,
+    slice,
     str::FromStr,
 };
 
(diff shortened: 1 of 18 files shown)


aya/src/util.rs
```diff
@@ -13,9 +13,130 @@ use crate::{
...
+    pub fn new(major: u8, minor: u8, patch: u16) -> Self {
+        Self {
+            major,
+            minor,
+            patch,
+        }
+    }
+
+    /// Returns the kernel version of the currently running kernel.
+    pub fn current() -> Result<Self, String> {
```
alessandrod, line 38: I think we should return a `Box<dyn Error>` instead of String here

tamird: Can you help me understand how that's better? Currently aya does not depend on anyhow, so other than using strings I can't propagate context.

alessandrod: This is now a public API, and returning String as Err is frowned upon (i think there's even a clippy lint for it)

> Currently aya does not depend on anyhow, so other than using strings I can't propagate context.

libraries should depend on thiserror, apps on anyhow. You can define a thiserror derived error, give it one variant for each string you're returning now, and type erase it when you return it. The result is the same as using a string (Error implements Display), but it's not a String, gives you a backtrace on nightly if you want one etc

tamird: Done.

alessandrod (approved):

PR #646, opened 2023-07-12
use debugfs only in kernel version less than 4.17.0

shandowc:
https://github.com/aya-rs/aya/pull/641/files#diff-a5e39230544964f896876377e6830ed97cbf0754b083d180ba17435a44ab10b1R53

a small error made by this PR


shandowc:
duplicated as https://github.com/aya-rs/aya/pull/643 has fixed it

PR #647, opened 2023-07-12
Use procfs crate commit import a bug: kernel vesion must be smaller t…

WindySha:
In this commit : [Use procfs crate for kernel version parsing](https://github.com/aya-rs/aya/commit/b611038d5b41a45ca70553550dbdef9aa1fd117c)

Before this commit, in probe.rs, the code is  k_ver < 4.17.0
```
    let k_ver = kernel_version().unwrap();
    if k_ver < (4, 17, 0) {
```
but after this commit, the code became  KernelVersion >= 4.17.0
```
    if KernelVersion::current().unwrap() >= KernelVersion::new(4, 17, 0) {
```

This will cause this kind of error on android device:
```
Error: `/sys/kernel/tracing/uprobe_events`
Caused by:
```


tamird:
Fixed in #643, thanks!

PR #648, opened 2023-07-12
Clippy over tests and integration-ebpf

tamird:
Replace all `assert!(matches!(..))` with `assert_matches!(..)`.

Remove the now-unused build-integration-test xtask command whose logic
doesn't match that of the build-and-run command.


bpf/aya-bpf/src/helpers.rs
```diff
@@ -438,7 +422,12 @@ fn read_str_bytes(len: i64, dest: &mut [u8]) -> Result<&[u8], c_long> {
     // len includes the NULL terminator but not for b"\0" for which the kernel
     // returns len=0. So we do a saturating sub and for b"\0" we return the
     // empty slice, for all other cases we omit the terminator.
-    Ok(&dest[..(len as usize).saturating_sub(1)])
+    let len = usize::try_from(len).map_err(|core::num::TryFromIntError { .. }| -1)?;
+    let len = len.saturating_sub(1);
+    match dest.get(..len) {
```
alessandrod, line 427: this is ok_or?

bpf/aya-bpf/src/lib.rs
```diff
@@ -58,18 +59,17 @@ pub trait BpfContext {
 
 #[no_mangle]
 pub unsafe extern "C" fn memset(s: *mut u8, c: c_int, n: usize) {
-    let base = s as usize;
+    #[allow(clippy::cast_sign_loss)]
+    let b = c as u8;
     for i in 0..n {
-        *((base + i) as *mut u8) = c as u8;
+        *s.add(i) = b;
```
alessandrod, line 65: I'm not sure we can use add in mem builtins, see the allocated object remark here https://doc.rust-lang.org/stable/std/primitive.pointer.html#safety-7

bpf/aya-bpf/src/helpers.rs
tamird: it is, done

bpf/aya-bpf/src/lib.rs
tamird: I'm not sure what to make of that remark, what do you think the behavior change would be? Tests seem happy.

tamird: Where do we go from here? Shall we land and find out?

alessandrod (approved):

PR #650, opened 2023-07-13
Remove async feature; misc test cleanup

tamird:
See individual commits.

### DIFF
--- a/aya-log/src/lib.rs
+++ b/aya-log/src/lib.rs
@@ -68,7 +68,7 @@ use thiserror::Error;
 
 use aya::{
     maps::{
-        perf::{AsyncPerfEventArray, PerfBufferError},
+        perf::{AsyncPerfEventArray, Events, PerfBufferError},
         MapError,
     },
     util::online_cpus,
@@ -121,12 +121,10 @@ impl BpfLogger {
                 let mut buffers = vec![BytesMut::with_capacity(LOG_BUF_CAPACITY); 10];
 
                 loop {
-                    let events = buf.read_events(&mut buffers).await.unwrap();
+                    let Events { read, lost: _ } = buf.read_events(&mut buffers).await.unwrap();
 
-                    #[allow(clippy::needless_range_loop)]
-                    for i in 0..events.read {
-                        let buf = &mut buffers[i];
-                        log_buf(buf, &*log).unwrap();
+                    for buf in buffers.iter().take(read) {
+                        log_buf(buf.as_ref(), &*log).unwrap();
                     }
                 }
             });
--- a/aya/src/maps/perf/async_perf_event_array.rs
+++ b/aya/src/maps/perf/async_perf_event_array.rs
@@ -4,6 +4,10 @@ use std::{
     os::fd::{AsRawFd, RawFd},
 };
 
+// See https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features.
+//
+// We should eventually split async functionality out into separate crates "aya-async-tokio" and
+// "async-async-std". Presently we arbitrarily choose tokio over async-std when both are requested.
 #[cfg(all(not(feature = "async_tokio"), feature = "async_std"))]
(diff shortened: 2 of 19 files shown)


ajwerner (approved):
:LGTM:

> Reviewed 9 of 9 files at r1, 2 of 2 files at r2, 5 of 5 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/650#-:-N_Eku0F3lM-3vz6PmzP:bmxilld)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tamird)

aya-log/src/lib.rs
```diff
@@ -121,12 +121,10 @@ impl BpfLogger {
...
 
                 loop {
-                    let events = buf.read_events(&mut buffers).await.unwrap();
+                    let Events { read, lost: _ } = buf.read_events(&mut buffers).await.unwrap();
 
-                    #[allow(clippy::needless_range_loop)]
-                    for i in 0..events.read {
-                        let buf = &mut buffers[i];
-                        log_buf(buf, &*log).unwrap();
+                    for buf in buffers.iter().take(read) {
```
alessandrod, line 126: not all buffers are guaranteed to be filled

aya/src/maps/perf/async_perf_event_array.rs
```diff
@@ -98,16 +98,17 @@ impl<T: BorrowMut<MapData> + Borrow<MapData>> AsyncPerfEventArray<T> {
         index: u32,
         page_count: Option<usize>,
     ) -> Result<AsyncPerfEventArrayBuffer<T>, PerfBufferError> {
-        let buf = self.perf_map.open(index, page_count)?;
+        let Self { perf_map } = self;
```
alessandrod, line 101: you love a destructuring don't you 😅 In this case, how is it better than just doing self.perf_map? It's the only field used and it's used only once

tamird: I sure do! It's just an assertion that no fields have been added and don't need to be dealt with. It's a bit of paranoia that protects this code from future changes that affect semantics.

aya/src/maps/perf/async_perf_event_array.rs
```diff
@@ -98,16 +98,17 @@ impl<T: BorrowMut<MapData> + Borrow<MapData>> AsyncPerfEventArray<T> {
...
+        let buf = perf_map.open(index, page_count)?;
         let fd = buf.as_raw_fd();
         Ok(AsyncPerfEventArrayBuffer {
             buf,
 
             #[cfg(feature = "async_tokio")]
-            async_fd: AsyncFd::new(fd)?,
+            async_tokio_fd: AsyncFd::new(fd)?,
```
alessandrod, line 108: I'm not sure I understand - is the purpose of this change to make cargo hack
happy? This code (pre PR) is messy and I'm glad it's getting improved, but now
it'll let you compile with both features on but won't do the right thing at
runtime. I think we should compile_error or something if both features are on?

Also I'm not sure that merging the `read_events` implementations is an
improvement. The code seems harder to read and understand to me. Not having to
write docs twice is an improvement, but that can probably be achieved in some
other way.

tamird: We can compile error, but we're in somewhat unsupported territory because mutually exclusive features aren't really a thing in cargo.

We can also do something approximating the old code: pick tokio If both are enabled.

A third option is to crash at runtime. Not clear to me what's best here.

alessandrod: I think we should keep the current behaviour and use tokio when both are enabled https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

tamird: Done.

alessandrod (approved):

PR #652, opened 2023-07-13
xtask: remove assumptions from docs command

tamird:
This slightly changes the site layout: crate documentation is now flat
rather than being nested under "user" and  "bpf".

- Run `cargo clean --doc` before generating docs to ensure hermiticity.
- Generate header.html into a temporary directory.
- Remove "site" on each run to ensure hermiticity.
- Invoke cargo only once.
- Avoid editing sources.


xtask/src/docs/mod.rs
```diff
@@ -22,95 +17,92 @@ pub fn exec(cmd: &mut Command) -> Result<()> {
...
-    let current_dir = PathBuf::from(".");
-    let header_path = current_dir.join("header.html");
-    let mut header = fs::File::create(&header_path).expect("can't create header.html");
-    header
-        .write_all(r#"<meta name="robots" content="noindex">"#.as_bytes())
-        .expect("can't write header.html contents");
-    header.flush().expect("couldn't flush contents");
-    let abs_header_path = fs::canonicalize(&header_path).unwrap();
+    const PACKAGE_TO_DESCRIPTION: &[(&str, &str)] =
+        &[("aya", "User-space"), ("aya-bpf", "Kernel-space")];
```
alessandrod, line 21: Now that we moved it to the main repo, I think we should also do aya-log?

tamird: Can that be a separate PR? I think that'd be a nice demonstration of the generality this enables.

alessandrod: sure

alessandrod (approved):

PR #656, opened 2023-07-14
Handle WSL kernel version strings

tamird:


aya/src/util.rs
```diff
@@ -132,8 +128,22 @@ impl KernelVersion {
         // terminated by a null byte ('\0').
         let p = unsafe { CStr::from_ptr(info.release.as_ptr()) };
         let p = p.to_str()?;
+        Self::parse_kernel_version_string(p)
+    }
+
+    fn parse_kernel_version_string(p: &str) -> Result<Self, CurrentKernelVersionError> {
         // Unlike sscanf, text_io::try_scan! does not stop at the first non-matching character.
-        let p = match p.split_once(|c: char| c != '.' && !c.is_ascii_digit()) {
+        let mut dots = 0;
```
alessandrod, line 136: This seems a pretty involved way to implement this function. Something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=651301bf503224ff671756d69cccbaaf ?

tamird: you skipped the hardest part: all the errors :)

alessandrod: Uh? Which errors did I skip?

tamird: The part where your implementation has to return an error type that isn't `()`

alessandrod: My implementation doesn't call text_io so it doesn't need to proxy any of its errors? `_ => CurrentKernelVersionError::ParseError`

tamird: Right, it just doesn't produce helpful errors at all - I contend that we probably do want helpful errors.

Anyway, done.

alessandrod: The error can't be handled programmatically, and will never be seen by anyone unless there's a bug in aya. Besides some of the errors produced by text_io aren't really applicable to this case (eg the utf8 error one).

I think we should remove the other uses of text_io too: in the case of debian we're already doing a split on 'Debian ' and we can match the rest with parse_kernel_version. In the case of ubuntu we can do rfind(' ') and then use parse_kernel_version

tamird: The error is super useful when there's a bug in aya, though.

Anyway, removed text_io.

alessandrod (approved):

PR #660, opened 2023-07-16
Make logging test output more legible

tamird:
Presently when this test fails you just get "5 items instead of 6" but
not which item was lost.

See failures in https://github.com/aya-rs/bpf-linker/pull/72.

### DIFF
--- a/test/integration-test/src/tests/log.rs
+++ b/test/integration-test/src/tests/log.rs
@@ -1,73 +1,21 @@
-use std::sync::{Arc, LockResult, Mutex, MutexGuard};
+use std::{
+    borrow::Cow,
+    sync::{Arc, Mutex},
+};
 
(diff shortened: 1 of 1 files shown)


test/integration-test/src/tests/log.rs
```diff
@@ -101,37 +63,72 @@ async fn log() {
...
+
+    let mut records = records.iter();
+
+    assert_matches!(
+        records.next(),
+        Some(CapturedLog {
+            body,
+            level: Level::Debug,
+            target,
+         }) if body == "Hello from eBPF!" && target == "log"
```
alessandrod, line 85: couldn't these be assert_eq?

tamird: done

alessandrod (approved):

PR #661, opened 2023-07-16
integration-tests: run in release also

ajwerner:
In release, the trigger functions were being optimized out and the tests did not work. Use core::hint::black_box to ensure that the functions are not optimized out. Also, run these integration tests in CI to ensure that we don't regress.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/661)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,14 @@ jobs:
 
   integration-test:
     runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # See https://doc.rust-lang.org/cargo/reference/profiles.html for the names
+        # of the builtin profiles. Note that dev builds "debug" targets.
+        profile:
+          - release
+          - dev
     steps:
       - uses: actions/checkout@v3
         with:
@@ -154,7 +162,7 @@ jobs:
           key: tmp-files-${{ hashFiles('test/run.sh') }}
 
       - name: Run integration tests
-        run: test/run.sh
+        run: test/run.sh --cargo-arg=--profile=${{ matrix.profile }}
 
   # Provides a single status check for the entire build workflow.
   # This is used for merge automation, like Mergify, since GH actions
--- a/test/integration-test/src/tests/bpf_probe_read.rs
+++ b/test/integration-test/src/tests/bpf_probe_read.rs
@@ -116,8 +116,12 @@ fn load_and_attach_uprobe(prog_name: &str, func_name: &str, bytes: &[u8]) -> Bpf
 
 #[no_mangle]
 #[inline(never)]
-pub extern "C" fn trigger_bpf_probe_read_user(_string: *const u8, _len: usize) {}
+pub extern "C" fn trigger_bpf_probe_read_user(string: *const u8, len: usize) {
+    core::hint::black_box((string, len));
+}
 
 #[no_mangle]
 #[inline(never)]
-pub extern "C" fn trigger_bpf_probe_read_kernel(_len: usize) {}
+pub extern "C" fn trigger_bpf_probe_read_kernel(len: usize) {
+    core::hint::black_box(len);
+}
--- a/test/integration-test/src/tests/log.rs
+++ b/test/integration-test/src/tests/log.rs
@@ -9,7 +9,9 @@ use log::{Level, Log, Record};
 
 #[no_mangle]
 #[inline(never)]
-pub extern "C" fn trigger_ebpf_program() {}
+pub extern "C" fn trigger_ebpf_program() {
+    core::hint::black_box(trigger_ebpf_program);
+}
 
 struct TestingLogger<F> {
     log: F,
--- a/test/integration-test/src/tests/relocations.rs
+++ b/test/integration-test/src/tests/relocations.rs
@@ -49,4 +49,6 @@ fn load_and_attach(name: &str, bytes: &[u8]) -> Bpf {
 
 #[no_mangle]
 #[inline(never)]
-pub extern "C" fn trigger_relocations_program() {}
+pub extern "C" fn trigger_relocations_program() {
+    core::hint::black_box(trigger_relocations_program);
+}
--- a/test/run.sh
+++ b/test/run.sh
@@ -242,7 +242,7 @@ trap cleanup_vm EXIT
 exec_vm "rm -rf aya/*"
 rsync_vm "--exclude=target --exclude=.tmp $AYA_SOURCE_DIR"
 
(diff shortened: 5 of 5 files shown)


alessandrod:
I swear I thought I removed the `--release` argument but maybe I did it in some other project. The BPF target isn't really meant to be used with anything less than -O2 so I think we should remove the argument and always run in release mode. 

Also the fact that `#[no_mangle]` items are being optimized out seems fishy - no_mangle is supposed to mean "don't mangle _and_ export the symbol". Maybe it's accidentally not being exported since we're outputting binaries instead of libraries? Who knows

ajwerner:
> I swear I thought I removed the --release argument but maybe I did it in some other project. The BPF target isn't really meant to be used with anything less than -O2 so I think we should remove the argument and always run in release mode.

These days, the bpf is always built in release (see [here](https://github.com/aya-rs/aya/blob/af54a819a364015bb42671e6892681bcb8f7b928/test/integration-test/build.rs#L161)). This flag controls whether the integration test rust (and, well, xtask) are built using release. I think we should ensure that that code remains functional in both debug and release. 

> Also the fact that #[no_mangle] items are being optimized out seems fishy - no_mangle is supposed to mean "don't mangle and export the symbol". Maybe it's accidentally not being exported since we're outputting binaries instead of libraries? Who knows

I agree, but I promise it's true. If I don't have this black_box hint, the entire function gets optimized out. My googling around suggests that the `#[no_mangle]` and `#[inline(never)]` prevent the compiler from messing with the function, but then the linker can come along and throw it away.

tamird (approved):

tamird (approved):
> Reviewed 1 of 1 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/661)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ajwerner)

alessandrod:
Can we run only in release mode in CI? It's not clear to me what the upside of running both is, and the downside is 2x CI time so the upside would have to be very compelling

tamird:
Could we run them in release and debug in parallel? btw https://github.com/aya-rs/aya/pull/680 significantly increased CI time by serializing lint -> test -> integration. We should make all these parallel again.

alessandrod:
> Could we run them in release and debug in parallel? btw #680 significantly increased CI time by serializing lint -> test -> integration. We should make all these parallel again.

Yeah absolutely parallel all the things

tamird (commented):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/661)* status: all files reviewed, 2 unresolved discussions (waiting on @ajwerner)

___
*[`.github/workflows/ci.yml` line 167 at r3](https://reviewable.io/reviews/aya-rs/aya/661#-NaHaggn1IFXdvZvmgW2:-NaHaggn1IFXdvZvmgW3:b-4gyf5j) ([raw file](https://github.com/aya-rs/aya/blob/bef6a438e1775838e2e7fb347719a4f49ec90782/.github/workflows/ci.yml#L167)):*
> ```YAML
>         # The AYAY_TEST_CARGO_FLAGS env var is used to pass flags to cargo in the
>         # test script.
>         run: AYA_TEST_CARGO_ARGS='--profile=${{ matrix.profile }}' test/run.sh
> ```

rather than an env variable, should we pass all arguments to run.sh through to cargo xtash integration-test? then we'd have the --cargo-arg= in here with fewer layers of translation.
___
*[`test/run.sh` line 257 at r3](https://reviewable.io/reviews/aya-rs/aya/661#-NaHasCq47q1Mv6qlMnb:-NaHasCr-kHgLoNWefyN:b-7mgug4) ([raw file](https://github.com/aya-rs/aya/blob/bef6a438e1775838e2e7fb347719a4f49ec90782/test/run.sh#L257)):*
> ```Shell
> rsync_vm "--exclude=target --exclude=.tmp $AYA_SOURCE_DIR"
> 
> set -x
> ```

seems useful enough to do for the whole script

ajwerner (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/661)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @tamird)

___
*[`.github/workflows/ci.yml` line 167 at r3](https://reviewable.io/reviews/aya-rs/aya/661#-NaHaggn1IFXdvZvmgW2:-NaHcket8_aLeMA6qh55:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/bef6a438e1775838e2e7fb347719a4f49ec90782/.github/workflows/ci.yml#L167)):*

rather than an env variable, should we pass all arguments to run.sh through to cargo xtash integration-test? then we'd have the --cargo-arg= in here with fewer layers of translation.
</blockquote></details>

Done.
___
*[`test/run.sh` line 257 at r3](https://reviewable.io/reviews/aya-rs/aya/661#-NaHasCq47q1Mv6qlMnb:-NaHclUV-bjnIO9LDs-l:b7yug6b) ([raw file](https://github.com/aya-rs/aya/blob/bef6a438e1775838e2e7fb347719a4f49ec90782/test/run.sh#L257)):*

seems useful enough to do for the whole script
</blockquote></details>

Decided not to.

tamird (approved):
:lgtm:

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/661#-:-NaHdc1j2fmSpesPDuwl:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @ajwerner)

___
*[`.github/workflows/ci.yml` line 165 at r4](https://reviewable.io/reviews/aya-rs/aya/661#-NaHdefdEROs8x_uXdAD:-NaHdefdEROs8x_uXdAE:b-9m9rgf) ([raw file](https://github.com/aya-rs/aya/blob/807090841fded16cab6e00ff014df0b70aa4e890/.github/workflows/ci.yml#L165)):*
> ```YAML
> 
>       - name: Run integration tests
>         # The AYAY_TEST_CARGO_FLAGS env var is used to pass flags to cargo in the
> ```

this comment is out of date.

ajwerner:
@alessandrod this now runs the release build in parallel, does that please?

tamird (approved):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/661)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ajwerner)

dave-tucker (approved):

ajwerner:
TFTR!

PR #662, opened 2023-07-17
Use `OwnedFd` for BTF's file descriptor

nrxus:
This PR is meant as the first step towards using `OwnedFd` where possible instead of `RawFd` to be more foolproof and safer to use. This PR tackles the BTF file descriptor that is loaded when the object has a BTF section and fixes an existing file descriptor leak as that file descriptor was never being closed. Since we are moving to use `OwnedFd` this close will now happen by default. 

The returned file descriptor will be wrapped in an `Arc` so that it can be shared across programs and maps of the loaded BPF file with no user-visible impact (no breaking changes). 

As an add-on I also changed the `Cargo/toml` to be explicit about the MSRV. This had been previously bumped by https://github.com/aya-rs/aya/pull/633 by using the [`std::os::fd`](https://doc.rust-lang.org/stable/std/os/fd/index.html) module which was only introduced in v1.66.0. This PR only really needs MSRV v1.63.0 to use `OwnedFd`, 

Related to https://github.com/aya-rs/aya/issues/612

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -3,8 +3,12 @@ use std::{
     collections::{HashMap, HashSet},
     ffi::CString,
     fs, io,
-    os::{fd::RawFd, raw::c_int},
+    os::{
+        fd::{OwnedFd, RawFd},
+        raw::c_int,
+    },
     path::{Path, PathBuf},
+    sync::Arc,
 };
 
 use aya_obj::{
@@ -390,7 +394,7 @@ impl<'a> BpfLoader<'a> {
         let btf_fd = if let Some(features) = &FEATURES.btf() {
             if let Some(btf) = obj.fixup_and_sanitize_btf(features)? {
                 match load_btf(btf.to_bytes(), *verifier_log_level) {
-                    Ok(btf_fd) => Some(btf_fd),
+                    Ok(btf_fd) => Some(Arc::new(btf_fd)),
                     // Only report an error here if the BTF is truely needed, otherwise proceed without.
                     Err(err) => {
                         for program in obj.programs.values() {
@@ -473,7 +477,7 @@ impl<'a> BpfLoader<'a> {
                 obj,
                 fd: None,
                 pinned: false,
-                btf_fd,
+                btf_fd: btf_fd.as_ref().map(Arc::clone),
             };
             let fd = match map.obj.pinning() {
                 PinningType::ByName => {
@@ -543,6 +547,7 @@ impl<'a> BpfLoader<'a> {
                 let section = prog_obj.section.clone();
                 let obj = (prog_obj, function_obj);
 
+                let btf_fd = btf_fd.as_ref().map(Arc::clone);
                 let program = if extensions.contains(name.as_str()) {
                     Program::Extension(Extension {
                         data: ProgramData::new(prog_name, obj, btf_fd, *verifier_log_level),
@@ -993,17 +998,14 @@ pub enum BpfError {
     ProgramError(#[from] ProgramError),
 }
 
-fn load_btf(raw_btf: Vec<u8>, verifier_log_level: VerifierLogLevel) -> Result<RawFd, BtfError> {
+fn load_btf(raw_btf: Vec<u8>, verifier_log_level: VerifierLogLevel) -> Result<OwnedFd, BtfError> {
     let (ret, verifier_log) = retry_with_verifier_logs(10, |logger| {
         bpf_load_btf(raw_btf.as_slice(), logger, verifier_log_level)
     });
-    match ret {
-        Ok(fd) => Ok(fd as RawFd),
-        Err((_, io_error)) => Err(BtfError::LoadError {
-            io_error,
-            verifier_log,
-        }),
-    }
+    ret.map_err(|(_, io_error)| BtfError::LoadError {
+        io_error,
+        verifier_log,
+    })
 }
 
 /// Global data that can be exported to eBPF programs before they are loaded.
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -3,7 +3,7 @@ use std::{
     collections::{HashMap, HashSet},
     ffi::CString,
     fs, io,
-    os::{raw::c_int, unix::io::RawFd},
+    os::{fd::RawFd, raw::c_int},
     path::{Path, PathBuf},
 };
 
--- a/aya/Cargo.toml
+++ b/aya/Cargo.toml
@@ -9,6 +9,7 @@ repository = "https://github.com/aya-rs/aya"
 readme = "README.md"
 documentation = "https://docs.rs/aya"
 edition = "2021"
+rust-version = "1.66"
 
 [dependencies]
 async-io = { workspace = true, optional = true }
--- a/aya/src/maps/bloom_filter.rs
+++ b/aya/src/maps/bloom_filter.rs
@@ -90,7 +90,7 @@ mod tests {
     };
     use assert_matches::assert_matches;
     use libc::{EFAULT, ENOENT};
-    use std::io;
+    use std::{ffi::c_long, io};
 
     fn new_obj_map() -> obj::Map {
         obj::Map::Legacy(LegacyMap {
@@ -108,7 +108,7 @@ mod tests {
         })
     }
 
-    fn sys_error(value: i32) -> SysResult {
+    fn sys_error(value: i32) -> SysResult<c_long> {
         Err((-1, io::Error::from_raw_os_error(value)))
     }
 
--- a/aya/src/maps/hash_map/hash_map.rs
+++ b/aya/src/maps/hash_map/hash_map.rs
@@ -105,7 +105,7 @@ impl<T: Borrow<MapData>, K: Pod, V: Pod> IterableMap<K, V> for HashMap<T, K, V> 
 
 #[cfg(test)]
 mod tests {
-    use std::io;
+    use std::{ffi::c_long, io};
 
     use assert_matches::assert_matches;
     use libc::{EFAULT, ENOENT};
@@ -139,7 +139,7 @@ mod tests {
         })
     }
 
-    fn sys_error(value: i32) -> SysResult {
+    fn sys_error(value: i32) -> SysResult<c_long> {
         Err((-1, io::Error::from_raw_os_error(value)))
     }
 
@@ -451,7 +451,7 @@ mod tests {
         assert_matches!(keys, Ok(ks) if ks.is_empty())
     }
 
-    fn get_next_key(attr: &bpf_attr) -> SysResult {
+    fn get_next_key(attr: &bpf_attr) -> SysResult<c_long> {
         match bpf_key(attr) {
             None => set_next_key(attr, 10),
             Some(10) => set_next_key(attr, 20),
@@ -463,7 +463,7 @@ mod tests {
         Ok(1)
     }
 
-    fn lookup_elem(attr: &bpf_attr) -> SysResult {
+    fn lookup_elem(attr: &bpf_attr) -> SysResult<c_long> {
         match bpf_key(attr) {
             Some(10) => set_ret(attr, 100),
             Some(20) => set_ret(attr, 200),
--- a/aya/src/maps/lpm_trie.rs
+++ b/aya/src/maps/lpm_trie.rs
@@ -209,7 +209,7 @@ mod tests {
     };
     use assert_matches::assert_matches;
     use libc::{EFAULT, ENOENT};
-    use std::{io, mem, net::Ipv4Addr};
+    use std::{ffi::c_long, io, mem, net::Ipv4Addr};
 
     fn new_obj_map() -> obj::Map {
         obj::Map::Legacy(LegacyMap {
@@ -227,7 +227,7 @@ mod tests {
         })
     }
 
-    fn sys_error(value: i32) -> SysResult {
+    fn sys_error(value: i32) -> SysResult<c_long> {
         Err((-1, io::Error::from_raw_os_error(value)))
     }
 
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -42,9 +42,10 @@ use std::{
     marker::PhantomData,
     mem,
     ops::Deref,
-    os::fd::{AsRawFd, RawFd},
+    os::fd::{AsFd, AsRawFd, OwnedFd, RawFd},
     path::Path,
     ptr,
+    sync::Arc,
 };
 
 use crate::util::KernelVersion;
@@ -486,7 +487,7 @@ pub(crate) fn check_v_size<V>(map: &MapData) -> Result<(), MapError> {
 pub struct MapData {
     pub(crate) obj: obj::Map,
     pub(crate) fd: Option<RawFd>,
-    pub(crate) btf_fd: Option<RawFd>,
+    pub(crate) btf_fd: Option<Arc<OwnedFd>>,
     /// Indicates if this map has been pinned to bpffs
     pub pinned: bool,
 }
@@ -504,19 +505,23 @@ impl MapData {
         let kernel_version = KernelVersion::current().unwrap();
         #[cfg(test)]
         let kernel_version = KernelVersion::new(0xff, 0xff, 0xff);
-        let fd = bpf_create_map(&c_name, &self.obj, self.btf_fd, kernel_version).map_err(
-            |(code, io_error)| {
-                if kernel_version < KernelVersion::new(5, 11, 0) {
-                    maybe_warn_rlimit();
-                }
+        let fd = bpf_create_map(
+            &c_name,
+            &self.obj,
+            self.btf_fd.as_ref().map(|f| f.as_fd()),
+            kernel_version,
+        )
+        .map_err(|(code, io_error)| {
+            if kernel_version < KernelVersion::new(5, 11, 0) {
+                maybe_warn_rlimit();
+            }
 
-                MapError::CreateError {
-                    name: name.into(),
-                    code,
-                    io_error,
-                }
-            },
-        )? as RawFd;
+            MapError::CreateError {
+                name: name.into(),
+                code,
+                io_error,
+            }
+        })? as RawFd;
 
         self.fd = Some(fd);
 
@@ -639,7 +644,7 @@ impl Clone for MapData {
         MapData {
             obj: self.obj.clone(),
             fd: self.fd.map(|fd| unsafe { libc::dup(fd) }),
-            btf_fd: self.btf_fd,
+            btf_fd: self.btf_fd.as_ref().map(Arc::clone),
             pinned: self.pinned,
         }
     }
--- a/aya/src/maps/perf/perf_buffer.rs
+++ b/aya/src/maps/perf/perf_buffer.rs
@@ -1,7 +1,7 @@
 use std::{
     ffi::c_void,
     io, mem,
-    os::unix::io::{AsRawFd, RawFd},
+    os::fd::{AsRawFd, RawFd},
     ptr, slice,
     sync::atomic::{self, AtomicPtr, Ordering},
 };
--- a/aya/src/maps/perf/perf_event_array.rs
+++ b/aya/src/maps/perf/perf_event_array.rs
@@ -4,7 +4,7 @@
 use std::{
     borrow::{Borrow, BorrowMut},
     ops::Deref,
-    os::unix::io::{AsRawFd, RawFd},
+    os::fd::{AsRawFd, RawFd},
     sync::Arc,
 };
 
@@ -64,7 +64,7 @@ impl<T: BorrowMut<MapData> + Borrow<MapData>> AsRawFd for PerfEventArrayBuffer<T
 /// A map that can be used to receive events from eBPF programs using the linux [`perf`] API.
 ///
 /// Each element of a [`PerfEventArray`] is a separate [`PerfEventArrayBuffer`] which can be used
-/// to receive events sent by eBPF programs that use `bpf_perf_event_output()`.    
+/// to receive events sent by eBPF programs that use `bpf_perf_event_output()`.
 ///
 /// To receive events you need to:
 /// * call [`PerfEventArray::open`]
@@ -138,7 +138,7 @@ impl<T: BorrowMut<MapData> + Borrow<MapData>> AsRawFd for PerfEventArrayBuffer<T
 ///
 /// In the example above the implementation of `poll_buffers()` and `poll.poll_readable()` is not
 /// given. [`PerfEventArrayBuffer`] implements the [`AsRawFd`] trait, so you can implement polling
-/// using any crate that can poll file descriptors, like [epoll], [mio] etc.  
+/// using any crate that can poll file descriptors, like [epoll], [mio] etc.
 ///
 /// Perf buffers are internally implemented as ring buffers. If your eBPF programs produce large
 /// amounts of data, in order not to lose events you might want to process each
--- a/aya/src/maps/sock/mod.rs
+++ b/aya/src/maps/sock/mod.rs
@@ -5,7 +5,7 @@ mod sock_map;
 pub use sock_hash::SockHash;
 pub use sock_map::SockMap;
 
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::fd::{AsRawFd, RawFd};
 
 /// A socket map file descriptor.
 #[derive(Copy, Clone)]
--- a/aya/src/maps/sock/sock_hash.rs
+++ b/aya/src/maps/sock/sock_hash.rs
@@ -1,7 +1,7 @@
 use std::{
     borrow::{Borrow, BorrowMut},
     marker::PhantomData,
-    os::unix::io::{AsRawFd, RawFd},
+    os::fd::{AsRawFd, RawFd},
 };
 
 use crate::{
@@ -42,7 +42,7 @@ use crate::{
 /// # let mut bpf = aya::Bpf::load(&[])?;
 /// use std::io::Write;
 /// use std::net::TcpStream;
-/// use std::os::unix::io::AsRawFd;
+/// use std::os::fd::AsRawFd;
 /// use aya::maps::SockHash;
 /// use aya::programs::SkMsg;
 ///
--- a/aya/src/programs/links.rs
+++ b/aya/src/programs/links.rs
@@ -354,7 +354,7 @@ pub enum LinkError {
 #[cfg(test)]
 mod tests {
     use assert_matches::assert_matches;
-    use std::{cell::RefCell, fs::File, mem, os::unix::io::AsRawFd, rc::Rc};
+    use std::{cell::RefCell, fs::File, mem, os::fd::AsRawFd, rc::Rc};
     use tempfile::tempdir;
 
     use crate::{programs::ProgramError, sys::override_syscall};
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -69,8 +69,9 @@ use libc::ENOSPC;
 use std::{
     ffi::CString,
     io,
-    os::fd::{AsRawFd, IntoRawFd as _, RawFd},
+    os::fd::{AsFd, AsRawFd, IntoRawFd as _, OwnedFd, RawFd},
     path::{Path, PathBuf},
+    sync::Arc,
 };
 use thiserror::Error;
 
@@ -413,7 +414,7 @@ pub(crate) struct ProgramData<T: Link> {
     pub(crate) attach_btf_obj_fd: Option<u32>,
     pub(crate) attach_btf_id: Option<u32>,
     pub(crate) attach_prog_fd: Option<RawFd>,
-    pub(crate) btf_fd: Option<RawFd>,
+    pub(crate) btf_fd: Option<Arc<OwnedFd>>,
     pub(crate) verifier_log_level: VerifierLogLevel,
     pub(crate) path: Option<PathBuf>,
     pub(crate) flags: u32,
@@ -423,7 +424,7 @@ impl<T: Link> ProgramData<T> {
     pub(crate) fn new(
         name: Option<String>,
         obj: (obj::Program, obj::Function),
-        btf_fd: Option<RawFd>,
+        btf_fd: Option<Arc<OwnedFd>>,
         verifier_log_level: VerifierLogLevel,
     ) -> ProgramData<T> {
         ProgramData {
@@ -613,7 +614,7 @@ fn load_program<T: Link>(
         license,
         kernel_version: target_kernel_version,
         expected_attach_type: *expected_attach_type,
-        prog_btf_fd: *btf_fd,
+        prog_btf_fd: btf_fd.as_ref().map(|f| f.as_fd()),
         attach_btf_obj_fd: *attach_btf_obj_fd,
         attach_btf_id: *attach_btf_id,
         attach_prog_fd: *attach_prog_fd,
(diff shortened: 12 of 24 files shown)


tamird (commented):
> Reviewed 17 of 17 files at r1, 35 of 35 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 5 unresolved discussions (waiting on @nrxus)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSHW24M14g2WUE10x):*
- capitalize please
- s/leaked/leak
- period please
___
*[`-- commits` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSk1CCjoDtWLZY-DL):*
can these be exposed via accessors rather than by being made public?
___
*[`aya/src/maps/mod.rs` line 495 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cT37U8L9D0lwCuGZf:-N_cT37U8L9D0lwCuGZg:b-ek5nmv) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/maps/mod.rs#L495)):*
> ```XML
> impl MapData {
>     /// Creates a new map with the provided `name`
>     pub fn create(&mut self, name: &str, btf_fd: Option<impl AsFd>) -> Result<RawFd, MapError> {
> ```

could we pass BorrowedFd<'_>? i'm wearing of generics introducing more monomorphization and slower compilation as a result

here and everywhere else
___
*[`aya/src/maps/mod.rs` line 891 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cTJ-x1xv-K7D56M9z:-N_cTJ-x1xv-K7D56MA-:b-6ys1ag) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/maps/mod.rs#L891)):*
> ```XML
> 
>         let mut map = new_map();
>         assert_matches!(map.create("foo", Option::<BorrowedFd>::None), Ok(42));
> ```

I think if you remove the generic you won't need all this type annotation

here and a few other places
___
*[`aya/src/sys/mod.rs` line 20 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cUD7_8CeQ8blP2xq-:-N_cUD7aEebHZD2hNX7Q:b-z4vr2j) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/sys/mod.rs#L20)):*
> ```XML
> use crate::generated::{bpf_attr, bpf_cmd, perf_event_attr};
> 
> pub(crate) type SysResult<T = c_long> = Result<T, (c_long, io::Error)>;
> ```

do we need the default?

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 5 unresolved discussions (waiting on @tamird)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSHW24M14g2WUE10x):*

- capitalize please
- s/leaked/leak
- period please
</blockquote></details>

To reword this commit should I just rebase and force push on this branch? I am not sure if that might mess with reviewable or with your own tracking of what you've already seen or not seen but I am not sure how else to reword commits that doesn't involve a rebase + force push. 
___
*[`-- commits` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSk1CCjoDtWLZY-DL):*

can these be exposed via accessors rather than by being made public?
</blockquote></details>

Unfortunately not. If it was an accessor like `fn(&mut self) -> &mut HashMap` then the lifetime of the returned hashmap is aliasing the entire `&mut self` as there is no way to do partial borrows in functions signatures. By exposing these fields I allow the compiler to be smart enough to detect that when I am mutably borrowing programs I am not borrowing neither btf file descriptor nor the maps so I can still borrow either of them (mutably or not) without issues. Borrowing of both the programs and the file descriptor is needed because loading of the programs needs to mutably borrow a program and (non mutably) borrow the btf file descriptor. 
___
*[`aya/src/maps/mod.rs` line 495 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cT37U8L9D0lwCuGZf:-N_gOVmc0S6tRL20bQjo:bvn6qmq) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/maps/mod.rs#L495)):*

could we pass BorrowedFd<'_>? i'm wearing of generics introducing more monomorphization and slower compilation as a result

here and everywhere else
</blockquote></details>

They become awkward to use if I use `Option<BorrowedFd<'_>>` because then usage of it when having a `Option<OwnedFd>` becomes something like: `my_owned_fd.as_ref().map(|f| f.as_fd())`. I could use something like `Option<&OwnedFd>` but that is significantly less flexible because then we *need* to have a reference to an owned fd. In our case we do always have an `OwnedFd` but I didn't know if that was too limiting going forwards. 
___
*[`aya/src/sys/mod.rs` line 20 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cUD7_8CeQ8blP2xq-:-N_gP50b6n2Rm-0Fc-cQ:b1t7ju) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/sys/mod.rs#L20)):*

do we need the default?
</blockquote></details>

We don't and I can remove the default if we don't want to deal with that but since so many functions already returned `SysResult` without any generics specified I figured adding a default would touch the least amount of code.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 4 unresolved discussions (waiting on @nrxus)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSHW24M14g2WUE10x):*

To reword this commit should I just rebase and force push on this branch? I am not sure if that might mess with reviewable or with your own tracking of what you've already seen or not seen but I am not sure how else to reword commits that doesn't involve a rebase + force push. 
</blockquote></details>

Yes, force push is the way to go. You'll need to rebase to resolve some of the clippy failures, and looks like CI is failing because all the tests need updating.
___
*[`-- commits` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSk1CCjoDtWLZY-DL):*

Unfortunately not. If it was an accessor like `fn(&mut self) -> &mut HashMap` then the lifetime of the returned hashmap is aliasing the entire `&mut self` as there is no way to do partial borrows in functions signatures. By exposing these fields I allow the compiler to be smart enough to detect that when I am mutably borrowing programs I am not borrowing neither btf file descriptor nor the maps so I can still borrow either of them (mutably or not) without issues. Borrowing of both the programs and the file descriptor is needed because loading of the programs needs to mutably borrow a program and (non mutably) borrow the btf file descriptor. 
</blockquote></details>

Makes sense. We still have accessors for maps and programs; should we remove or deprecate those?
___
*[`aya/src/maps/mod.rs` line 495 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cT37U8L9D0lwCuGZf:-N_h_1SM8LsarZ6lLIR3:b-u5kwjj) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/maps/mod.rs#L495)):*

They become awkward to use if I use `Option<BorrowedFd<'_>>` because then usage of it when having a `Option<OwnedFd>` becomes something like: `my_owned_fd.as_ref().map(|f| f.as_fd())`. I could use something like `Option<&OwnedFd>` but that is significantly less flexible because then we *need* to have a reference to an owned fd. In our case we do always have an `OwnedFd` but I didn't know if that was too limiting going forwards. 
</blockquote></details>

Ok, let's leave this as is.
___
*[`aya/src/sys/mod.rs` line 20 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cUD7_8CeQ8blP2xq-:-N_hZTR_96qf9lafopBk:b-z3dm9j) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/sys/mod.rs#L20)):*

We don't and I can remove the default if we don't want to deal with that but since so many functions already returned `SysResult` without any generics specified I figured adding a default would touch the least amount of code. 
</blockquote></details>

Let's just add the generic in a parent commit (without a default, please)

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: 7 of 50 files reviewed, 4 unresolved discussions (waiting on @tamird)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSHW24M14g2WUE10x):*

Yes, force push is the way to go. You'll need to rebase to resolve some of the clippy failures, and looks like CI is failing because all the tests need updating.
</blockquote></details>

I believe I've fixed the build failures in the integrations tests (sorry I didn't catch those earlier, I usually just run `cargo clippy --workspace` which wasn't enough in this case. However, the clippy failures seem bogus to me. It looks like they are happening mostly due to a lint that is not yet available in stable and the lint seems to be poorly implemented so far (hence why it's not in stable yet I am guessing): https://github.com/rust-lang/rust-clippy/issues/5546#issuecomment-1555963013. Since the clippy errors don't seem to be about this PR but rather code that was here before that's only now starting to alert I am hesitant on addressing them for a nightly lint that is still a WIP. For example it's alerting about using a mutable reference to mapdata where we only need a reference for functions that update the bpf map because it doesn't know that we are modifying something under the hood using the file descriptor. I tried to disregard the warning but still pass clippy by adding `#[allow(...)]` but it didn't seem to do anything, perhaps because it's not a lint that is in stable yet I am not sure. 
___
*[`-- commits` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSk1CCjoDtWLZY-DL):*

Makes sense. We still have accessors for maps and programs; should we remove or deprecate those?
</blockquote></details>

I'd be in favor of just removing them since using the fields already gives more flexibility but I leave that decision up to you all. Let me know what you'd like me to do. 

I do think that having a method in the `Bpf` object to load programs and have it pass the btf file descriptor is probably warranted since it's such a common use case. The problem though is that having a method like `Bpf::load_program(&mut self, name: &str)` doesn't work for all program types because some of them require more information than just the name (i.e., lsm, BTF trace point, fentry, fexit, and extension). This could be left as a future improvement in another PR though, the focus right now is to fix the file descriptor leak. 
___
*[`aya/src/maps/mod.rs` line 891 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cTJ-x1xv-K7D56M9z:-N_lb9b31hQz1TdQzPZr:bxfxvs8) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/maps/mod.rs#L891)):*

I think if you remove the generic you won't need all this type annotation

here and a few other places
</blockquote></details>

Since we've decided to keep this generic I am marking this as done
___
*[`aya/src/sys/mod.rs` line 20 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cUD7_8CeQ8blP2xq-:-N_mEdgi7LtHPOwv1Xym:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/9f88ed37c36541a75fee5849a490b0f60a49c4c7/aya/src/sys/mod.rs#L20)):*

Let's just add the generic in a parent commit (without a default, please)
</blockquote></details>

Done.

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: 4 of 50 files reviewed, 4 unresolved discussions (waiting on @tamird)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSHW24M14g2WUE10x):*

I believe I've fixed the build failures in the integrations tests (sorry I didn't catch those earlier, I usually just run `cargo clippy --workspace` which wasn't enough in this case. However, the clippy failures seem bogus to me. It looks like they are happening mostly due to a lint that is not yet available in stable and the lint seems to be poorly implemented so far (hence why it's not in stable yet I am guessing): https://github.com/rust-lang/rust-clippy/issues/5546#issuecomment-1555963013. Since the clippy errors don't seem to be about this PR but rather code that was here before that's only now starting to alert I am hesitant on addressing them for a nightly lint that is still a WIP. For example it's alerting about using a mutable reference to mapdata where we only need a reference for functions that update the bpf map because it doesn't know that we are modifying something under the hood using the file descriptor. I tried to disregard the warning but still pass clippy by adding `#[allow(...)]` but it didn't seem to do anything, perhaps because it's not a lint that is in stable yet I am not sure. 
</blockquote></details>

And just after I sent the last message I see that you all fixed it in a separate PR. I don't quite agree with making some of those `&mut` be `&` just because clipper said so but since it's not really relevant to this PR I just rebased to the latest main and the nightly clippy seems happy (on my machine at least).

tamird (changes requested):
> Reviewed 38 of 38 files at r3, 7 of 7 files at r4, 40 of 40 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod and @nrxus)

___
*[`-- commits` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/662#-N_cSk1CCjoDtWLZY-DL):*

I'd be in favor of just removing them since using the fields already gives more flexibility but I leave that decision up to you all. Let me know what you'd like me to do. 

I do think that having a method in the `Bpf` object to load programs and have it pass the btf file descriptor is probably warranted since it's such a common use case. The problem though is that having a method like `Bpf::load_program(&mut self, name: &str)` doesn't work for all program types because some of them require more information than just the name (i.e., lsm, BTF trace point, fentry, fexit, and extension). This could be left as a future improvement in another PR though, the focus right now is to fix the file descriptor leak. 
</blockquote></details>

I'm in favor of removing as well. I know @alessandrod cares about this sort of thing. @alessandrod can you please weigh in?
___
*[`test/integration-test/src/tests/bpf_probe_read.rs` line 108 at r5](https://reviewable.io/reviews/aya-rs/aya/662#-N_nysq73znj_GlzICr-:-N_nysq73znj_GlzICr0:bbce4cy) ([raw file](https://github.com/aya-rs/aya/blob/7456d2bb0d66b9390f0a40b34933be6466e6d8f5/test/integration-test/src/tests/bpf_probe_read.rs#L108)):*
> ```XML
>     let mut bpf = Bpf::load(bytes).unwrap();
> 
>     let prog: &mut UProbe = bpf.programs.get_mut(prog_name).unwrap().try_into().unwrap();
> ```

instead of these point accesses, could we write

```
let Bpf { programs, btf_fd, .. } = &mut bpf;
```

that makes it clear that these things are borrowed from the same structure. here and everywhere please.

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: 7 of 50 files reviewed, 2 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`test/integration-test/src/tests/bpf_probe_read.rs` line 108 at r5](https://reviewable.io/reviews/aya-rs/aya/662#-N_nysq73znj_GlzICr-:-N_qiTmT8Rf0AMXNEvfe:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7456d2bb0d66b9390f0a40b34933be6466e6d8f5/test/integration-test/src/tests/bpf_probe_read.rs#L108)):*

instead of these point accesses, could we write

```
let Bpf { programs, btf_fd, .. } = &mut bpf;
```

that makes it clear that these things are borrowed from the same structure. here and everywhere please.
</blockquote></details>

Done.

alessandrod (commented):
I haven't done a full review, just commenting on the load() change: I don't think this is an acceptable change API wise.

BTF is largely an implementation detail. As a user, in 99% of the cases, I don't even know that it exists. There's BTF from the kernel, there's BTF in object files produced by the toolchain, they're used by the kernel to do useful things, but as a user I never interact with it. I never create it, I never lookup things inside it, nothing.

With this change, you're forcing _everyone_ to have to be aware of BTF, by having to shuttle this opaque btf fd thing around. Something that is an implementation detail, suddenly becomes a central thing in the API. Seems wrong to me.

tamird:
WDYT about introducing a type that wraps `Program` + `BorrowedFd`, and making that the return from `Bpf::program` and `Bpf::program_mut`? That type could [`std::ops::Deref`](https://doc.rust-lang.org/stable/std/ops/trait.Deref.html) and [`std::ops::DerefMut`](https://doc.rust-lang.org/stable/std/ops/trait.DerefMut.html) with `Target = Program`, and the current `Program::load` function would move to that type. Then the user would never the see the BTF file descriptor.

tamird:
@nrxus does that sound reasonable? If you like, I can pick this up. We're desperately in need of fixing the FD ownership problems that are all over the codebase.

nrxus:
> @nrxus does that sound reasonable? If you like, I can pick this up. We're desperately in need of fixing the FD ownership problems that are all over the codebase.

Ah my bad I thought your last question was for @alessandrod. I can try to implement that tonight and send it to you all for review. I also don't mind if you want to take over this PR, I also just want the file descriptors ownership fixed, it doesn't matter to me much by who 😅

tamird:
Go for it! I'd rather cultivate more contributors :)

nrxus:
> Go for it! I'd rather cultivate more contributors :)

I started down the road of making the intermediary struct (`ProgramMutRef`) but it gets pretty awkward since the current workflow is:

 `Bpf` -> `&mut Program `-> `&mut SpecificProgramType` -> load
 
Because of this we would have to have a wrapper struct for each individual program type essentially (since not all programs are loaded with the same parameters). Before I start down that rabbit hole I think it may be worth mentioning alternatives:

* Use `Arc<OwnedFd>` inside `ProgramData`. This means we keep the same interface as we do now where the BTF file descriptor is a hidden detail inside `ProgramData`. The con is that we now have the "cost" of using `Arc`. I don't think this is a *huge* deal because `Arc` is pretty cheap to store and the creation cost is just at load time of `Bpf` where it'd be minimal compared to the rest

* Add a `Bpf::load_program` method and a `ProgramLoad` trait that'd look something like:

```rust
trait ProgramLoad {
    type Data;
    
    fn load(&mut self, btf_fd: Option<impl AsFd>,data: Self::Data) -> Result<(), ProgramError>;
}

impl Bpf {
    /// Loads a program with default data - useful when the programs don't need any extra information (e.g., kprobes, tracepoints)
    fn load_program<P: ProgramLoad>(&mut self, name: &str) -> Result<P, ProgramError>
    where P::Data : Default {
        self.load_program_with(name, Data::Default)
    }
    
    /// Loads a program that needs extra information (e.g., LSM)
    fn load_program_with<P: ProgramLoad>(&mut self, name: &str, data: P::Data) -> Result<P, ProgramError> {
        let program = self.programs.get_mut(name).ok_or(ProgramError::NotFound)?:
        let program: &mut P = program.try_into()?;
        program.load(self.btf_fd.as_ref(), data)
    }
}
```

Then each program could implement `ProgramLoad` and define the kind of program-specific data they need. This has the benefit that the `Bpf` object is still the clear owner of the BTF descriptor (the program only gets it at load time) but the con is that it is a breaking change from the current way programs are load (from `bpf.program_mut("my_probe").unwrap().load()` to `bpf.load_program("my_probe")`. It is also more code to write but 🤷🏾 . 

I am curious to hear what you all think or if you all have a different idea.

tamird:
> > Go for it! I'd rather cultivate more contributors :)
> 
> I started down the road of making the intermediary struct (`ProgramMutRef`) but it gets pretty awkward since the current workflow is:
> 
> `Bpf` -> `&mut Program `-> `&mut SpecificProgramType` -> load
> 
> Because of this we would have to have a wrapper struct for each individual program type essentially (since not all programs are loaded with the same parameters). Before I start down that rabbit hole I think it may be worth mentioning alternatives:

Are you sure you need a separate type for each one? I think you just need a separate impl block for each program type.

```rust
struct WithBtf<'a, T> {
  program: &'a mut T,
  btf: BorrowedFd<'a>,
}

impl WithBtf<'_, Xdp> {
  fn load(...) {
    // this is the existing load method on Xdp
  }
}
```

> 
> * Use `Arc<OwnedFd>` inside `ProgramData`. This means we keep the same interface as we do now where the BTF file descriptor is a hidden detail inside `ProgramData`. The con is that we now have the "cost" of using `Arc`. I don't think this is a _huge_ deal because `Arc` is pretty cheap to store and the creation cost is just at load time of `Bpf` where it'd be minimal compared to the rest. Not really a con nor a plus but it is worth noting that in this approach the mental model of the btf file descriptor is that it isn't owned by anyone but rather just shared by all the programs and maps.

I'd rather not change performance characteristics or introduce multiple ownership.

> * Add a `Bpf::load_program` method and a `ProgramLoad` trait that'd look something like:
> 
> ```rust
> trait ProgramLoad {
>     type Data;
>     
>     fn load(&mut self, btf_fd: Option<impl AsFd>,data: Self::Data) -> Result<(), ProgramError>;
> }
> 
> impl Bpf {
>     /// Loads a program with default data - useful when the programs don't need any extra information (e.g., kprobes, tracepoints)
>     fn load_program<P: ProgramLoad>(&mut self, name: &str) -> Result<P, ProgramError>
>     where P::Data : Default {
>         self.load_program_with(name, Data::Default)
>     }
>     
>     /// Loads a program that needs extra information (e.g., LSM)
>     fn load_program_with<P: ProgramLoad>(&mut self, name: &str, data: P::Data) -> Result<P, ProgramError> {
>         let program = self.programs.get_mut(name).ok_or(ProgramError::NotFound)?:
>         let program: &mut P = program.try_into()?;
>         program.load(self.btf_fd.as_ref(), data)
>     }
> }
> ```
> 
> Then each program could implement `ProgramLoad` and define the kind of program-specific data they need. This has the benefit that the `Bpf` object is the clear owner of the BTF descriptor (the program and maps only gets it at load time) but the con is that it is a breaking change from the current way programs are load (from `bpf.program_mut("my_probe").unwrap().load()` to `bpf.load_program("my_probe")`. It is also more code to write but 🤷🏾 .

This refactoring seems orthogonal to what we're doing here. I like it, but it seems orthogonal.

alessandrod:
> Use Arc<OwnedFd> inside ProgramData. This means we keep the same interface as we do now where the BTF file descriptor is a hidden detail inside ProgramData. The con is that we now have the "cost" of using Arc. I don't think this is a huge deal because Arc is pretty cheap to store and the creation cost is just at load time of Bpf where it'd be minimal compared to the rest. Not really a con nor a plus but it is worth noting that in this approach the mental model of the btf file descriptor is that it isn't owned by anyone but rather just shared by all the programs and maps.

I think this is what we should do. The cost of this `Arc` would be insignificant - we only allocate one per object file, the allocation is tiny and is going to fit in the smallest size class provided by the allocator so it's not actually going to allocate anything. 

The `WithBtf` approach doesn't seem viable to me for the same reason that `load(..., btf_fd)` isn't: the problem is the same, except now you've moved this weird BTF thing in return position, and I need to spell it out anytime I call `bpf.program_mut()` - from rustdocs:

```
    /// let mut program: aya::WithBtfFd<UProbe> = bpf.program_mut("SSL_read").unwrap().try_into()?;
```

tamird:
I agree that the WithBtf type should be invisible in user code. I think if we can make that work, this approach remains viable.

alessandrod:
> I agree that the WithBtf type should be invisible in user code. I think if we can make that work, this approach remains viable.

But how? program and program_mut want you to specify the return type

tamird:
> > I agree that the WithBtf type should be invisible in user code. I think if we can make that work, this approach remains viable.
> 
> But how? program and program_mut want you to specify the return type

If you could ascribe using only the program type, then you wouldn't need to care that you're getting a `WithBtf<T>` rather than a `T`.

nrxus:
> > > I agree that the WithBtf type should be invisible in user code. I think if we can make that work, this approach remains viable.
> > 
> > 
> > But how? program and program_mut want you to specify the return type
> 
> If you could ascribe using only the program type, then you wouldn't need to care that you're getting a `WithBtf<T>` rather than a `T`.

The only way I can think of ascribing only the `T` is by changing `program_mut` to something like:

```rust
fn mut_program<P>(&mut self, name: &str) -> Result<WithBtfFd<'_, P>, ProgramError> {
   let program = self.programs.get_mut(name).ok_or(ProgramError::NotFound)?;
   program.try_into()
}
```

But then we are requiring users to know statically what program type they are going to want when they ask for one whilst currently that is not a requirement. For example, in the way I've been using `aya` I have a list of program names that I need to load/attach, some are kprobes, some are tracepoints so my code looks something like:

```rust
program_names.iter().try_for_each(|p| {
    let Some(program) = bpf.program_mut(p).ok_or_else(|| Err(/*idk some error*/))?;
    match program {
        Program::KProbe(p) => p.load(),
        Program::Tracepoint(p) => p.load(),
        _ => return Err(/* unsupported program */)
    }
})
```

If we changed it to where `program_mut` needs to know the type statically then the above strategy is no longer doable. In my case I could just have a list of kprobe names and a list of tracepoint names since I know what they each are since I wrote the programs but that may not be the case across the board so I am concerned about cutting off potential workflows for users of `aya`.

tamird:
Agreed. Seems the best we can do is a shared FD then. Does it need to be Arc, or can it be Rc?

nrxus:
> Agreed. Seems the best we can do is a shared FD then. Does it need to be Arc, or can it be Rc?

As far as I can tell `ProgramData` currently implements `Send +Sync` so I'd lean towards `Arc` so as to not remove functionality. I could totally see someone loading a Bpf instance in one thread and then sending it to another thread to manage its lifecycle after the programs are all attached/running.

tamird:
Makes sense

tamird (approved):
> Reviewed 45 of 45 files at r12, 7 of 7 files at r13, 32 of 32 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod and @nrxus)

___
*[`aya/src/sys/bpf.rs` line 574 at r11](https://reviewable.io/reviews/aya-rs/aya/662#-NaLvUmA5fyAhSbFhv6h:-NaROJtKE6NOUsppzhlT:boxgwk2) ([raw file](https://github.com/aya-rs/aya/blob/b2b52aac9eb63c73371c52805ed382b8d9ff1f31/aya/src/sys/bpf.rs#L574)):*

can you avoid the `as` cast here? let's use .try_into (there's an example of this already in the file)
</blockquote></details>

Let's address this one please.

aya/src/bpf.rs
```diff
@@ -543,6 +547,7 @@ impl<'a> BpfLoader<'a> {
                 let section = prog_obj.section.clone();
                 let obj = (prog_obj, function_obj);
 
+                let btf_fd = btf_fd.clone();
```
alessandrod, line 550: Here and everywhere else, use Arc::clone instead of fd.clone()

alessandrod (commented):
See clone comment, then it's ready to go!

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: 49 of 52 files reviewed, 2 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`aya/src/bpf.rs` line 550 at r14](https://reviewable.io/reviews/aya-rs/aya/662#gh-1277628723:-NaSfB1K4VKdgV8IeeWY:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/fcabe5fe586ed974c68873cbc3aa301f2e0fa293/aya/src/bpf.rs#L550)):*
<details><summary><i>Previously, alessandrod (Alessandro Decina) wrote…</i></summary><blockquote>

Here and everywhere else, use Arc::clone(&fd) instead of fd.clone()
</blockquote></details>

Done.
___
*[`aya/src/sys/bpf.rs` line 574 at r11](https://reviewable.io/reviews/aya-rs/aya/662#-NaLvUmA5fyAhSbFhv6h:-NaSfIuJ7PW4-AuYx1FA:brodsdh) ([raw file](https://github.com/aya-rs/aya/blob/b2b52aac9eb63c73371c52805ed382b8d9ff1f31/aya/src/sys/bpf.rs#L574)):*

Let's address this one please.
</blockquote></details>

Done.

Although I personally think it's a bit unnecessary, the error here is impossible unless there's a kernel/rust bug right, it's not really something the user could handle.

tamird (commented):
> Reviewed 3 of 3 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod and @nrxus)

___
*[`aya/src/sys/bpf.rs` line 574 at r11](https://reviewable.io/reviews/aya-rs/aya/662#-NaLvUmA5fyAhSbFhv6h:-NaSkNnWCmDPr8fAL-Z8:b99a8cx) ([raw file](https://github.com/aya-rs/aya/blob/b2b52aac9eb63c73371c52805ed382b8d9ff1f31/aya/src/sys/bpf.rs#L574)):*

Done.

Although I personally think it's a bit unnecessary, the error here is impossible unless there's a kernel/rust bug right, it's not really something the user could handle. 
</blockquote></details>

Sure, feel free to panic if you think that's better - I'm just asking that we not cast the sign away.
___
*[`aya/src/sys/bpf.rs` line 467 at r15](https://reviewable.io/reviews/aya-rs/aya/662#-NaSklffCWJN9Xzl0sgC:-NaSklfgEloY08X12rIs:b-hq1lso) ([raw file](https://github.com/aya-rs/aya/blob/6fa7fdb08f9b7892e743dd4631d05ff5726d9a18/aya/src/sys/bpf.rs#L467)):*
> ```XML
> 
>     attr.__bindgen_anon_6.__bindgen_anon_1.prog_id = prog_id;
>     // SAFETY: BPF_PROG_GET_FD_BY_ID returns a new file descriptor
> ```

can we please end comments with periods? here and 3 other places (let's match the style in the pre-existing comment please)

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: 51 of 52 files reviewed, 2 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`aya/src/sys/bpf.rs` line 467 at r15](https://reviewable.io/reviews/aya-rs/aya/662#-NaSklffCWJN9Xzl0sgC:-NaSuptR1L0jXomnyZmr:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/6fa7fdb08f9b7892e743dd4631d05ff5726d9a18/aya/src/sys/bpf.rs#L467)):*

can we please end comments with periods? here and 3 other places (let's match the style in the pre-existing comment please)
</blockquote></details>

Done.

tamird (approved):
> Reviewed 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod)

tamird:
Looks like the integration tests are failing, are we dropping a file descriptor somewhere?

nrxus:
> Looks like the integration tests are failing, are we dropping a file descriptor somewhere?

Small goof in my refactor to extract the `fd_sys_bpf` function. Fixed now.

tamird (approved):
> Reviewed 1 of 1 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/662)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod)

tamird:
Thanks @nrxus!

PR #663, opened 2023-07-17
No Submodules

dave-tucker:
Submodules are notoriously annoying to manage.
 
We don't actually care about the libbpf version at all since it's only
ever used as a shortcut to:
    
    - Generate struct bindings from kernel structs
    - Provide the bpf/* headers to C-BPF files so they can be compiled

The UX for people running the integration test is not nice either - as I found out - since if you update via `git pull` to aya@main then integration tests will refuse to run with no clear error message.

Since the libbpf version used isn't special and doesn't need tracking (outside of git commit message for bindings updates) this PR reverts the use of submodules.
It keeps the assumption that libbpf is in the root of the checkout though, but if it is not found a nice error message is produced telling you how to acquire it.
Fully re-instating `--libbpf-dir` would be an ideal option, yet this isn't possible without also reverting the move to `build.rs` for compiling the C-eBPF probes.

### DIFF
--- a/.github/workflows/gen.yml
+++ b/.github/workflows/gen.yml
@@ -8,6 +8,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        with:
+          repository: libbpf/libbpf
+          path: libbpf
 
       - name: libbpf-version
         working-directory: libbpf
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,8 +18,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/checkout@v3
         with:
-          submodules: recursive
+          repository: libbpf/libbpf
+          path: libbpf
 
       - name: Install Pre-requisites
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ target/
 !.vscode/settings.json
 site/
 .idea/
+# Checkout of libbpf, used for codegen + compiling C test code
+libbpf/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "libbpf"]
-	path = libbpf
-	url = https://github.com/libbpf/libbpf
--- a/test/integration-test/build.rs
+++ b/test/integration-test/build.rs
@@ -71,6 +71,15 @@ fn main() {
             .unwrap()
             .join("libbpf");
 
+        if !libbpf_dir.exists() {
+            panic!(
+                "Directory does not exist at {:?}\n\
+                Please perform a shallow clone of libbpf:\n\
+                git clone https://github.com/libbpf/libbpf --depth 1\n",
+                libbpf_dir
+            );
+        }
+
         let libbpf_headers_dir = out_dir.join("libbpf_headers");
 
         let mut includedir = OsString::new();


tamird:
This doesn't make sense to me. Why not just have the xtask init the submodule?

dave-tucker:
> This doesn't make sense to me. Why not just have the xtask init the submodule?

Because performing operations that mutate git in xtask isn't something I would advocate for - nor am I willing to take responsibility for any unexpected side effects.

Even if we accepted that risk - it doesn't fix the myriad usability issues around submodules.

tamird:
Sorry, what exactly are the usability issues?

How exactly does `git submodule init` mutate git?

Even if there's a good reason not to automatically do this on behalf of the user (and I don't think there is, x.py in rust does exactly this), you already admit to the usability regression in this PR by emitting an error with instructions from the xtask. That error could just as easily say "please init the submodule".

ajwerner:
@dave-tucker https://github.com/aya-rs/aya/pull/664 is an alternative proposal. I too am struggling to understand how a submodule, which lets your precisely control the version of the depended upon code as opposed to a loosely coupled expectation that the user check out some code into a specific place.

dave-tucker:
@ajwerner  @tamird 

Let me re-iterate: **The libbpf version used isn't special and doesn't need tracking**.
Arbitrary checkouts of libbpf are just fine.

May I also refer you to the aya README:

> It does not rely on [libbpf](https://github.com/libbpf/libbpf) nor [bcc](https://github.com/iovisor/bcc) - it's built from the ground up purely in Rust, using only the [libc](https://docs.rs/libc) crate to execute syscalls.

Having an explicit dependency on `libbpf` at the root of the repository is a source of confusion.

---

Flipping this around, with the recent changes my workflow is now broken.
Can you justify to me why you're insisting that something - that I've told you isn't special - is so special that operations like `git pull`, `git clone` and `git checkout` should now no longer work correctly because of submodules?

Yes submodules aren't updated on pull, don't get automatically pulled in on clone and don't change when switching branches.

tamird:
> @ajwerner @tamird
> 
> Let me re-iterate: **The libbpf version used isn't special and doesn't need tracking**. Arbitrary checkouts of libbpf are just fine.

Having a submodule doesn't *only* solve for the version; it also codifies where it's expected to be located and generally shifts knowledge out of people's heads and into code and automation. This seems good.

> 
> May I also refer you to the aya README:
> 
> > It does not rely on [libbpf](https://github.com/libbpf/libbpf) nor [bcc](https://github.com/iovisor/bcc) - it's built from the ground up purely in Rust, using only the [libc](https://docs.rs/libc) crate to execute syscalls.
> 
> Having an explicit dependency on `libbpf` at the root of the repository is a source of confusion.

Confusion for whom? Can you cite someone who was confused by this?

> 
> Flipping this around, with the recent changes my workflow is now broken. Can you justify to me why you're insisting that something - that I've told you isn't special - is so special that operations like `git pull`, `git clone` and `git checkout` should now no longer work correctly because of submodules?

Can you please spell out how these operations no longer work for you?

> 
> Yes submodules aren't updated on pull, don't get automatically pulled in on clone and don't change when switching branches.

All this can be done in the xtask in #664.

tamird:
Sorry, to address the final point in more detail: an arbitrary libbpf directory doesn't get updated on pull, isn't automatically pulled in on clone, and doesn't change when switching branches.

dave-tucker:
> The points about stale modules I do not understand at all. Having a manually maintained directory in the repository or outside has all the same problems and then some (your tools can't tell you it's stale).

The previous workflow was this:
_Just show up with any old checkout of libbpf at any location and things will just work_

It is now at the following extreme:

_We REALLY care about the location and version of libbpf to the point it's codified in the git repo and there are extra hoops you have to jump through at checkout etc.. for this to work properly_

Notice the difference?
I'm asserting that version and location are totally unimportant.
You're asserting that version and location are so important we need to codify this + add extra tooling to make this work.

If you want to hide the complexity, while making sure your reasoning for using submodules holds, then #664 needs to cover initializing submodules, checking they are at the correct revision for the current branch and forcing recompile after update if they are not. 

> Regarding the perception issue we can rename the libbpf directory to anything you like, including something like "libbpf-for-codegen-and-integration-tests-only". Would that assuage your concerns in this area?

No. because long path names are impractical, but hiding it in `test/libbpf` or similar might be bearable.

> I agree we should fix the instructions, my apologies for having failed to update them previously.

The other documentation gap is how/when to update the submodule.
Previously the `codegen` workflow would bump libbpf and re-generate bindings at the same time.
It's unclear now what the new workflow is - especially since all codegen has to be done in CI since we've had issues with incorrect bindings being generated in the past.

tamird:
> The previous workflow was this: _Just show up with any old checkout of libbpf at any location and things will just work_
> 
> It is now at the following extreme:
> 
> _We REALLY care about the location and version of libbpf to the point it's codified in the git repo and there are extra hoops you have to jump through at checkout etc.. for this to work properly_
> 
> Notice the difference? I'm asserting that version and location are totally unimportant. You're asserting that version and location are so important we need to codify this + add extra tooling to make this work.

I'd restate this: previously the workflow was: please internalize this undeclared dependency and figure out how to acquire it. We will not help you. The new workflow is that the tooling takes care of it and that's it.

> 
> If you want to hide the complexity, while making sure your reasoning for using submodules holds, then #664 needs to cover initializing submodules, checking they are at the correct revision for the current branch and forcing recompile after update if they are not.

Agreed.

> No. because long path names are impractical, but hiding it in `test/libbpf` or similar might be bearable.

I'd have suggested that path, but we also use it for codegen. I don't feel strongly about the name or path.

> The other documentation gap is how/when to update the submodule. Previously the `codegen` workflow would bump libbpf and re-generate bindings at the same time. It's unclear now what the new workflow is - especially since all codegen has to be done in CI since we've had issues with incorrect bindings being generated in the past.

The workflow is; `(cd libbpf && git fetch origin && git checkout origin/HEAD) && git add libbpf` followed by the existing machinery. We will do this in #664.

astoycos (approved):

tamird:
Yeah, we're making this automatic #664. Please don't throw out the baby with the bathwater.

PR #668, opened 2023-07-17
integration-test: invalidate on bpf-linker

tamird:
It turns out that cargo ignores rerun-if-changed directives that point
into $CARGO_HOME; use a symlink to trick cargo into respecting my
authoritah.


ajwerner (approved):
LGTM

Issue #670, opened 2023-07-18
integration test documentation requires update


dave-tucker:
The old documented workflow in `test/README.md` was:

`cargo xtask integration-test --libbpf-dir ../../libbpf`

Since then:
- `--libbpf-dir` has been deprecated
- A checkout of libbpf is no longer required
- There appear to have been changes to the way the command handles rust toolchains since `cargo xtask integration test` fails

```
--- stderr
  error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
  See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
  thread 'main' panicked at '"cargo" "build" "-p" "integration-ebpf" "-Z" "build-std=core" "--release" "--message-format=json" "--target" "bpfel-unknown-none" "--target-dir" "/home/dave/dev/aya-rs/aya/target/debug/build/integration-test-3d3491ef010ffd35/out/integration-ebpf" exited with status code 101:
  ', test/integration-test/build.rs:215:25
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: error while building userspace application
```

The only working command I've found is:
```console
cargo +nightly xtask integration-test
```

Which may or may not be the intended workflow.

ajwerner:
> --libbpf-dir has been deprecated

This is no longer mentioned.

> A checkout of libbpf is no longer required

https://github.com/aya-rs/aya/pull/673

> There appear to have been changes to the way the command handles rust toolchains since cargo xtask integration test fails

https://github.com/aya-rs/aya/pull/674

ajwerner:
Fixed by https://github.com/aya-rs/aya/pull/673 and https://github.com/aya-rs/aya/pull/674.

PR #674, opened 2023-07-18
test/integration-test: use +nightly for ebpf

ajwerner:
The ebpf probes require a nightly compiler. Before this change, if you ran `cargo xtask integration-test` with a stable compiler toolchain as default, or you ran `cargo +stable xtask integration-test`, you would have seen an error like the one below. This is now fixed.

```
  --- stderr
  /home/ajwerner/src/github.com/aya-rs/aya/test/integration-test/bpf/ring_buf_sched_tracepoint.bpf.c:18:21: warning: declaration of 'struct switch_args' will not be visible outside of this function [-Wvisibility]
  int bpf_prog(struct switch_args* ctx)
                      ^
  1 warning generated.
  error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
  See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
  thread 'main' panicked at '"cargo" "build" "-p" "integration-ebpf" "-Z" "build-std=core" "--release" "--message-format=json" "--target" "bpfel-unknown-none" "--target-dir" "/home/ajwerner/src/github.com/aya-rs/aya/target/debug/build/integration-test-9bbcb3db5e9f8f57/out/integration-ebpf" exited
with status code 101:
  ', test/integration-test/build.rs:219:25
Error: error while building userspace application

Caused by:
    Child { stdin: None, stdout: None, stderr: None, .. } exited with status code 101:
```

Relates to #670.

### DIFF
--- a/test/integration-ebpf/rust-toolchain.toml
+++ b/test/integration-ebpf/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "nightly"
--- a/test/integration-test/build.rs
+++ b/test/integration-test/build.rs
@@ -163,8 +163,6 @@ fn main() {
         let mut cmd = Command::new("cargo");
         cmd.args([
             "build",
-            "-p",
-            "integration-ebpf",
             "-Z",
             "build-std=core",
             "--release",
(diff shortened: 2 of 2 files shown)


ajwerner:
> Can we put a toolchain file in the integration-ebpf directory instead? Does the tooling respect it when not at the workspace root?

It does if you clear `RUSTUP_TOOLCHAIN`.

PR #637, opened 2023-07-05
Add more helpful methods to `ProgramInfo`

astoycos:
This PR adds some helpful methods to the `ProgramInfo` aya structure.  More specifically we start by adding helpers to extract the following information in usable format from `bpf_prog_info`

```
"Name",
"Id",
"Type",
"Name",
"Loaded-At",
"Tag",
"Gpl-compatible",
"Map-ids",
"Btf-id",
"Bytes_xlated",
"Jited",
"Bytes_jited",
"Bytes_memlock",
"Verified_insns",
``` 

This is in line with the information extracted by the popular [bpftool](https://github.com/libbpf/bpftool) binary 

bpfd is using these commits downstream in 

https://github.com/bpfd-dev/bpfd/pull/570

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/637)

PR #675, opened 2023-07-18
Add helpers for querying info from existing bpf programs

preuss-adam:
This change adds additional getter functions to Aya's BPF objects to extract the map ids of a loaded program and extract the map name from a loaded map. It also supports loading a map object by fd, similar to what already exists for program objects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -73,7 +73,7 @@ use crate::{
         bpf_map_get_info_by_fd, bpf_map_get_next_key, bpf_map_update_elem_ptr, bpf_pin_object,
         iter_map_ids, SyscallError,
     },
-    util::{nr_cpus, KernelVersion},
+    util::{bytes_of_bpf_name, nr_cpus, KernelVersion},
     PinningType, Pod,
 };
 
@@ -927,17 +927,7 @@ impl MapInfo {
 
     /// The name of the map, limited to 16 bytes.
     pub fn name(&self) -> &[u8] {
-        let length = self
-            .0
-            .name
-            .iter()
-            .rposition(|ch| *ch != 0)
-            .map(|pos| pos + 1)
-            .unwrap_or(0);
-
-        // The name field is defined as [std::os::raw::c_char; 16]. c_char may be signed or
-        // unsigned depending on the platform; that's why we're using from_raw_parts here.
-        unsafe { std::slice::from_raw_parts(self.0.name.as_ptr() as *const _, length) }
+        bytes_of_bpf_name(&self.0.name)
     }
 
     /// The name of the map as a &str. If the name is not valid unicode, None is returned.
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -65,12 +65,13 @@ use obj::maps::InvalidMapTypeError;
 use thiserror::Error;
 
 use crate::{
+    generated::bpf_map_info,
     obj::{self, parse_map_info, BpfSectionKind},
     pin::PinError,
     sys::{
         bpf_create_map, bpf_get_object, bpf_map_freeze, bpf_map_get_fd_by_id,
         bpf_map_get_info_by_fd, bpf_map_get_next_key, bpf_map_update_elem_ptr, bpf_pin_object,
-        SyscallError,
+        iter_map_ids, SyscallError,
     },
     util::{nr_cpus, KernelVersion},
     PinningType, Pod,
@@ -639,21 +640,14 @@ impl MapData {
             call: "BPF_OBJ_GET",
             io_error,
         })?;
-        let fd = MapFd(fd);
 
-        let info = bpf_map_get_info_by_fd(fd.as_fd())?;
-
-        Ok(Self {
-            obj: parse_map_info(info, PinningType::ByName),
-            fd,
-        })
+        Self::from_fd(fd)
     }
 
     /// Loads a map from a map id.
     pub fn from_id(id: u32) -> Result<Self, MapError> {
-        bpf_map_get_fd_by_id(id)
-            .map_err(MapError::from)
-            .and_then(Self::from_fd)
+        let fd = bpf_map_get_fd_by_id(id)?;
+        Self::from_fd(fd)
     }
 
     /// Loads a map from a file descriptor.
@@ -662,12 +656,10 @@ impl MapData {
     /// This API is intended for cases where you have received a valid BPF FD from some other means.
     /// For example, you received an FD over Unix Domain Socket.
     pub fn from_fd(fd: OwnedFd) -> Result<Self, MapError> {
-        let info = bpf_map_get_info_by_fd(fd.as_fd())?;
-
-        let fd = MapFd(fd);
+        let MapInfo(info) = MapInfo::new_from_fd(fd.as_fd())?;
         Ok(Self {
             obj: parse_map_info(info, PinningType::None),
-            fd,
+            fd: MapFd(fd),
         })
     }
 
@@ -722,6 +714,11 @@ impl MapData {
     pub(crate) fn obj(&self) -> &obj::Map {
         let Self { obj, fd: _ } = self;
         obj
+    }
+
+    /// Returns the kernel's information about the loaded map.
+    pub fn info(&self) -> Result<MapInfo, MapError> {
+        MapInfo::new_from_fd(self.fd.as_fd())
     }
 }
 
@@ -911,6 +908,129 @@ impl<T: Pod> Deref for PerCpuValues<T> {
     }
 }
 
+/// Provides information about a loaded map, like name, id and size.
+#[derive(Debug)]
+pub struct MapInfo(bpf_map_info);
+
+impl MapInfo {
+    fn new_from_fd(fd: BorrowedFd<'_>) -> Result<Self, MapError> {
+        let info = bpf_map_get_info_by_fd(fd.as_fd())?;
+        Ok(Self(info))
+    }
+
+    /// Loads map info from a map id.
+    pub fn from_id(id: u32) -> Result<Self, MapError> {
+        bpf_map_get_fd_by_id(id)
+            .map_err(MapError::from)
+            .and_then(|fd| Self::new_from_fd(fd.as_fd()))
+    }
+
+    /// The name of the map, limited to 16 bytes.
+    pub fn name(&self) -> &[u8] {
+        let length = self
+            .0
+            .name
+            .iter()
+            .rposition(|ch| *ch != 0)
+            .map(|pos| pos + 1)
+            .unwrap_or(0);
+
+        // The name field is defined as [std::os::raw::c_char; 16]. c_char may be signed or
+        // unsigned depending on the platform; that's why we're using from_raw_parts here.
+        unsafe { std::slice::from_raw_parts(self.0.name.as_ptr() as *const _, length) }
+    }
+
+    /// The name of the map as a &str. If the name is not valid unicode, None is returned.
+    pub fn name_as_str(&self) -> Option<&str> {
+        std::str::from_utf8(self.name()).ok()
+    }
+
+    /// The id for this map. Each map has a unique id.
+    pub fn id(&self) -> u32 {
+        self.0.id
+    }
+
+    /// The map type as defined by the linux kernel enum
+    /// [`bpf_map_type`](https://elixir.bootlin.com/linux/v6.4.4/source/include/uapi/linux/bpf.h#L905).
+    pub fn map_type(&self) -> u32 {
+        self.0.type_
+    }
+
+    /// The key size for this map.
+    pub fn key_size(&self) -> u32 {
+        self.0.key_size
+    }
+
+    /// The value size for this map.
+    pub fn value_size(&self) -> u32 {
+        self.0.value_size
+    }
+
+    /// The maximum number of entries in this map.
+    pub fn max_entries(&self) -> u32 {
+        self.0.max_entries
+    }
+
+    /// The flags for this map.
+    pub fn map_flags(&self) -> u32 {
+        self.0.map_flags
+    }
+
+    /// Returns a file descriptor referencing the map.
+    ///
+    /// The returned file descriptor can be closed at any time and doing so does
+    /// not influence the life cycle of the map.
+    pub fn fd(&self) -> Result<MapFd, MapError> {
+        let Self(info) = self;
+        let fd = bpf_map_get_fd_by_id(info.id)?;
+        Ok(MapFd(fd))
+    }
+
+    /// Loads a map from a pinned path in bpffs.
+    pub fn from_pin<P: AsRef<Path>>(path: P) -> Result<Self, MapError> {
+        use std::os::unix::ffi::OsStrExt as _;
+
+        // TODO: avoid this unwrap by adding a new error variant.
+        let path_string = CString::new(path.as_ref().as_os_str().as_bytes()).unwrap();
+        let fd = bpf_get_object(&path_string).map_err(|(_, io_error)| SyscallError {
+            call: "BPF_OBJ_GET",
+            io_error,
+        })?;
+
+        Self::new_from_fd(fd.as_fd())
+    }
+}
+
+/// Returns an iterator over all loaded bpf maps.
+///
+/// This differs from [`crate::Bpf::maps`] since it will return all maps
+/// listed on the host system and not only maps for a specific [`crate::Bpf`] instance.
+///
+/// # Example
+/// ```
+/// # use aya::maps::loaded_maps;
+///
+/// for m in loaded_maps() {
+///     match m {
+///         Ok(map) => println!("{:?}", map.name_as_str()),
+///         Err(e) => println!("Error iterating maps: {:?}", e),
+///     }
+/// }
+/// ```
+///
+/// # Errors
+///
+/// Returns [`MapError::SyscallError`] if any of the syscalls required to either get
+/// next map id, get the map fd, or the [`MapInfo`] fail. In cases where
+/// iteration can't be performed, for example the caller does not have the necessary privileges,
+/// a single item will be yielded containing the error that occurred.
+pub fn loaded_maps() -> impl Iterator<Item = Result<MapInfo, MapError>> {
+    iter_map_ids().map(|id| {
+        let id = id?;
+        MapInfo::from_id(id)
+    })
+}
+
 #[cfg(test)]
 mod tests {
     use std::os::fd::AsRawFd as _;
@@ -992,6 +1112,96 @@ mod tests {
                 fd,
             }) => assert_eq!(fd.as_fd().as_raw_fd(), 42)
         );
+    }
+
+    #[test]
+    // Syscall overrides are performing integer-to-pointer conversions, which
+    // should be done with `ptr::from_exposed_addr` in Rust nightly, but we have
+    // to support stable as well.
+    #[cfg_attr(miri, ignore)]
+    fn test_name() {
+        use crate::generated::bpf_map_info;
+
+        const TEST_NAME: &str = "foo";
+
+        override_syscall(|call| match call {
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_MAP_CREATE,
+                ..
+            } => Ok(42),
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_OBJ_GET_INFO_BY_FD,
+                attr,
+            } => {
+                assert_eq!(
+                    unsafe { attr.info.info_len },
+                    mem::size_of::<bpf_map_info>() as u32
+                );
+                let map_info = unsafe { &mut *(attr.info.info as *mut bpf_map_info) };
+                map_info.name[..TEST_NAME.len()]
+                    .copy_from_slice(unsafe { std::mem::transmute(TEST_NAME) });
+                Ok(0)
+            }
+            _ => Err((-1, io::Error::from_raw_os_error(EFAULT))),
+        });
+
+        let map_data = MapData::create(new_obj_map(), TEST_NAME, None).unwrap();
+        assert_eq!(TEST_NAME, map_data.info().unwrap().name_as_str().unwrap());
+    }
+
+    #[test]
+    // Syscall overrides are performing integer-to-pointer conversions, which
+    // should be done with `ptr::from_exposed_addr` in Rust nightly, but we have
+    // to support stable as well.
+    #[cfg_attr(miri, ignore)]
+    fn test_loaded_maps() {
+        use crate::generated::bpf_map_info;
+
+        override_syscall(|call| match call {
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_MAP_GET_NEXT_ID,
+                attr,
+            } => unsafe {
+                let id = attr.__bindgen_anon_6.__bindgen_anon_1.start_id;
+                if id < 5 {
+                    attr.__bindgen_anon_6.next_id = id + 1;
+                    Ok(0)
+                } else {
+                    Err((-1, io::Error::from_raw_os_error(libc::ENOENT)))
+                }
+            },
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_MAP_GET_FD_BY_ID,
+                attr,
+            } => Ok((1000 + unsafe { attr.__bindgen_anon_6.__bindgen_anon_1.map_id }) as c_long),
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_OBJ_GET_INFO_BY_FD,
+                attr,
+            } => {
+                let map_info = unsafe { &mut *(attr.info.info as *mut bpf_map_info) };
+                map_info.id = unsafe { attr.info.bpf_fd } - 1000;
+                map_info.key_size = 32;
+                map_info.value_size = 64;
+                map_info.map_flags = 1234;
+                map_info.max_entries = 99;
+                Ok(0)
+            }
+            _ => Err((-1, io::Error::from_raw_os_error(EFAULT))),
+        });
+
+        let loaded_maps: Vec<_> = loaded_maps().collect();
+        assert_eq!(loaded_maps.len(), 5);
+
+        for (i, map_info) in loaded_maps.into_iter().enumerate() {
+            let i = i + 1;
+            let map_info = map_info.unwrap();
+            assert_eq!(map_info.id(), i as u32);
+            assert_eq!(map_info.key_size(), 32);
+            assert_eq!(map_info.value_size(), 64);
+            assert_eq!(map_info.map_flags(), 1234);
+            assert_eq!(map_info.max_entries(), 99);
+            assert_eq!(map_info.fd().unwrap().as_fd().as_raw_fd(), 1000 + i as i32);
+        }
     }
 
     #[test]
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -68,8 +68,9 @@ use crate::{
     obj::{self, parse_map_info, BpfSectionKind},
     pin::PinError,
     sys::{
-        bpf_create_map, bpf_get_object, bpf_map_freeze, bpf_map_get_info_by_fd,
-        bpf_map_get_next_key, bpf_map_update_elem_ptr, bpf_pin_object, SyscallError,
+        bpf_create_map, bpf_get_object, bpf_map_freeze, bpf_map_get_fd_by_id,
+        bpf_map_get_info_by_fd, bpf_map_get_next_key, bpf_map_update_elem_ptr, bpf_pin_object,
+        SyscallError,
     },
     util::{nr_cpus, KernelVersion},
     PinningType, Pod,
@@ -648,6 +649,13 @@ impl MapData {
         })
     }
 
+    /// Loads a map from a map id.
+    pub fn from_id(id: u32) -> Result<Self, MapError> {
+        bpf_map_get_fd_by_id(id)
+            .map_err(MapError::from)
+            .and_then(Self::from_fd)
+    }
+
     /// Loads a map from a file descriptor.
     ///
     /// If loading from a BPF Filesystem (bpffs) you should use [`Map::from_pin`](crate::maps::MapData::from_pin).
@@ -933,6 +941,38 @@ mod tests {
             symbol_index: Some(0),
             data: Vec::new(),
         })
+    }
+
+    #[test]
+    fn test_from_map_id() {
+        override_syscall(|call| match call {
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_MAP_GET_FD_BY_ID,
+                attr,
+            } => {
+                assert_eq!(
+                    unsafe { attr.__bindgen_anon_6.__bindgen_anon_1.map_id },
+                    1234
+                );
+                Ok(42)
+            }
+            Syscall::Bpf {
+                cmd: bpf_cmd::BPF_OBJ_GET_INFO_BY_FD,
+                attr,
+            } => {
+                assert_eq!(unsafe { attr.info.bpf_fd }, 42);
+                Ok(0)
+            }
+            _ => Err((-1, io::Error::from_raw_os_error(EFAULT))),
+        });
+
+        assert_matches!(
+            MapData::from_id(1234),
+            Ok(MapData {
+                obj: _,
+                fd,
+            }) => assert_eq!(fd.as_fd().as_raw_fd(), 42)
+        );
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


astoycos:
@preuss-adam https://github.com/aya-rs/aya/pull/637 _should_ have most of what you need

preuss-adam:
@astoycos 
Thanks. I’ll rebase my other two commits once that PR goes in. For the `loaded_maps` were you thinking something like just adding a `pub struct MapInfo(bpf_map_info);` and the associated iterator?

astoycos:
> For the loaded_maps were you thinking something like just adding a pub struct MapInfo(bpf_map_info); and the associated iterator?

Yeah at least for the first go that's what I was thinking! I'd also like to do the same for bpf_links as well

preuss-adam:
@astoycos please see latest changes

astoycos:
Will take a look here, sorry for the delay @preuss-adam !

tamird (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 8 unresolved discussions (waiting on @astoycos and @preuss-adam)

___
*[`aya/src/maps/mod.rs` line 553 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbD-qK-Udv1hNQ1_CC:-NhbD-qK-Udv1hNQ1_CD:brnpxri) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L553)):*
> ```XML
>     }
> 
>     /// Loads a map from a map id
> ```

missing period.
___
*[`aya/src/maps/mod.rs` line 554 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDqEqD6-S3r_YSWxe:-NhbDqEqD6-S3r_YSWxf:bo2qq4v) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L554)):*
> ```XML
> 
>     /// Loads a map from a map id
>     pub fn from_id(id: u32) -> Result<Self, MapError> {
> ```

this is public API, you'll need to run `cargo +nightly xtask public-api --bless` to update the API fixtures.
___
*[`aya/src/maps/mod.rs` line 555 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbBsKj3uu73ghXh--G:-NhbBsKkDU9LrphK3m3l:bf2lf6b) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L555)):*
> ```XML
>     /// Loads a map from a map id
>     pub fn from_id(id: u32) -> Result<Self, MapError> {
>         Self::from_fd(bpf_map_get_fd_by_id(id)?)
> ```

nit: consider `bpf_map_get_fd_by_id(id).and_then(Self::from_fd)`; the `?` inside an expresion can be hard to read.
___
*[`aya/src/maps/mod.rs` line 621 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbCzXRAT9LH6AkmWPP:-NhbCzXRAT9LH6AkmWPQ:brnpxri) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L621)):*
> ```XML
>     }
> 
>     /// Returns the name of the map
> ```

missing period.
___
*[`aya/src/maps/mod.rs` line 622 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDoO04HsUs5wZhrvT:-NhbDoO04HsUs5wZhrvU:b-yms38j) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L622)):*
> ```XML
> 
>     /// Returns the name of the map
>     pub fn name(&self) -> Result<String, MapError> {
> ```

this needs a test.
___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-NhbC5ec866_8h6qL16h:bj0wkol) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*
> ```XML
>         let info = bpf_map_get_info_by_fd(self.fd.as_fd())?;
>         Ok(
>             unsafe { std::str::from_utf8(std::mem::transmute(info.name.as_slice())).unwrap() }
> ```

is there a reason we can't use [`CStr::from_from_bytes_until_nul`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_bytes_until_nul) here? or even [`CStr::from_ptr`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_ptr) which would avoid the neeed for the transmute.
___
*[`aya/src/maps/mod.rs` line 849 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDlsIEZBANYB1jhtG:-NhbDlsIEZBANYB1jhtH:b-yms38j) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L849)):*
> ```XML
> /// iteration can't be performed, for example the caller does not have the necessary privileges,
> /// a single item will be yielded containing the error that occurred.
> pub fn loaded_maps() -> impl Iterator<Item = Result<MapData, MapError>> {
> ```

this needs a test.
___
*[`aya/src/maps/mod.rs` line 852 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDE0N9gYpKFEX3xrp:-NhbDE0N9gYpKFEX3xrq:bgulcyz) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L852)):*
> ```XML
>     iter_map_ids()
>         .map(|id| {
>             let id = id?;
> ```

nit: might be clearer as `id.and_then(bpf_map_get_fd_by_id).and_then(MapData::from_fd)`

aya/src/maps/mod.rs
```diff
@@ -612,6 +618,16 @@ impl MapData {
         fd
     }
 
+    /// Returns the name of the map
+    pub fn name(&self) -> Result<String, MapError> {
```
astoycos, line 622: again let's use the same method as `ProgramInfo.name()` here https://github.com/aya-rs/aya/blob/main/aya/src/programs/mod.rs#L1004

aya/src/sys/bpf.rs
```diff
@@ -550,6 +550,21 @@ pub(crate) fn bpf_prog_get_info_by_fd(
     })
 }
 
+pub(crate) fn bpf_map_get_fd_by_id(map_id: u32) -> Result<OwnedFd, SyscallError> {
+    let mut attr = unsafe { mem::zeroed::<bpf_attr>() };
+
+    attr.__bindgen_anon_6.__bindgen_anon_1.prog_id = map_id;
```
astoycos, line 556:
```suggestion
    attr.__bindgen_anon_6.__bindgen_anon_1.map_id = map_id;
```

astoycos (changes requested):
Thanks for the PR @preuss-adam!

So some general design things as mentioned in the review...specifically

There's kind of two separate APIs here 
1. A wrapper around the kernel's `bpf_map_info` struct 
2. Aya's `Map` object

I think it would be cool to have a `from_map_info` API as well as the `from_id` one you've provided here.

Also we probably should add more methods on `MapInfo` for the relevant/useful fields from https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L6358

I'm thinking at the very least for the following:

```
struct bpf_map_info {
	__u32 type;
	__u32 id;
	__u32 key_size;
	__u32 value_size;
	__u32 max_entries;
	__u32 map_flags;
	char  name[BPF_OBJ_NAME_LEN];
```

lastly we'll need some testing coverage for all of this functionality. Which ultimately should include

1. Loading an aya::Map from a map ID
2. Listing all `aya::map::MapInfo`s
3. Calling all methods on `aya::map::MapInfo`s
4. Converting a `MapInfo` into a `Map`

If you want to break this work into multiple PRs as well that's completely fine.

preuss-adam:
Thank you both for the review comments. For now, I've pulled out the last commit with `loaded_maps` as that will require a bit more reworking. The other two commits have been updated.

tamird:
> Thank you both for the review comments. For now, I've pulled out the last commit with `loaded_maps` as that will require a bit more reworking. The other two commits have been updated.

Could you respond to the comments (in reviewable, where applicable, please)?

aya/src/maps/mod.rs
```diff
@@ -804,6 +820,41 @@ impl<T: Pod> Deref for PerCpuValues<T> {
...
+/// }
+/// ```
+///
+/// # Errors
+///
+/// Returns [`MapError::SyscallError`] if any of the syscalls required to either get
+/// next map id, get the map fd, or the [`MapData`] fail. In cases where
+/// iteration can't be performed, for example the caller does not have the necessary privileges,
+/// a single item will be yielded containing the error that occurred.
+pub fn loaded_maps() -> impl Iterator<Item = Result<MapData, MapError>> {
```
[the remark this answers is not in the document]
preuss-adam: Sounds good. Removed this for now and will do this in follow up PR.

aya/src/sys/bpf.rs
preuss-adam: Oops. Fixed

aya/src/maps/mod.rs
preuss-adam: Can't quite make it like this yet, as we don't have the struct to reference. Will do as part of follow up PR once we have `MapInfo(bpf_map_info);`.

preuss-adam (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: 0 of 3 files reviewed, 9 unresolved discussions (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/maps/mod.rs` line 553 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbD-qK-Udv1hNQ1_CC:-Nhi8nZP0at9ptgRriPm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L553)):*

missing period.
</blockquote></details>

Done.
___
*[`aya/src/maps/mod.rs` line 554 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDqEqD6-S3r_YSWxe:-Nhi8pw2B_iyOkS3cOXh:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L554)):*

this is public API, you'll need to run `cargo +nightly xtask public-api --bless` to update the API fixtures.
</blockquote></details>

Done.
___
*[`aya/src/maps/mod.rs` line 621 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbCzXRAT9LH6AkmWPP:-Nhi8xNfDMk9brwcLZ_v:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L621)):*

missing period.
</blockquote></details>

Done.
___
*[`aya/src/maps/mod.rs` line 622 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDoO04HsUs5wZhrvT:-Nhi8y0L6SjuqZXoO6uW:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L622)):*

this needs a test.
</blockquote></details>

Done.
___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-Nhi8zQh4b24EqRUZlUu:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*

is there a reason we can't use [`CStr::from_from_bytes_until_nul`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_bytes_until_nul) here? or even [`CStr::from_ptr`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_ptr) which would avoid the neeed for the transmute.
</blockquote></details>

Done.
___
*[`aya/src/maps/mod.rs` line 849 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDlsIEZBANYB1jhtG:-Nhi9-ig9q8dC0HiStl7:btkogy7) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L849)):*

this needs a test.
</blockquote></details>

This will come in as part of the follow up PR. Commit removed for now. 
___
*[`aya/src/maps/mod.rs` line 852 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbDE0N9gYpKFEX3xrp:-Nhi9GhY5zhn14PDlGJX:bb6un1j) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L852)):*

nit: might be clearer as `id.and_then(bpf_map_get_fd_by_id).and_then(MapData::from_fd)`
</blockquote></details>

Done

tamird (changes requested):
> Reviewed 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 4 unresolved discussions (waiting on @alessandrod and @astoycos)

___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-NhiImvDDhok5aFbifpi:bvwqunt) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I think we should reconcile this with https://github.com/aya-rs/aya/blob/main/aya/src/programs/mod.rs#L1004 as @astoycos suggested. It's weird that we have different implementations for these things.

preuss-adam (commented):
Should be all addressed now.

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675#-:-Ni6Gtx17CbtIZReIqHX:b8zy09l)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/maps/mod.rs` line 622 at r1](https://reviewable.io/reviews/aya-rs/aya/675#gh-1371941779:-Ni6G_fyAKgYkTssVyuC:bo5n8uz) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L622)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Can't quite make it like this yet, as we don't have the struct to reference. Will do as part of follow up PR once we have `MapInfo(bpf_map_info);`.
</blockquote></details>

Done. New commit on this PR.
___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-Ni6GVd00c7BOQ4uR6M4:bo5n8uz) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*

I think we should reconcile this with https://github.com/aya-rs/aya/blob/main/aya/src/programs/mod.rs#L1004 as @astoycos suggested. It's weird that we have different implementations for these things.
</blockquote></details>

Done. New commit on this PR.
___
*[`aya/src/maps/mod.rs` line 849 at r1](https://reviewable.io/reviews/aya-rs/aya/675#gh-1371900006:-Ni6GpcG-Cb2_SOUCdkR:bo5n8uz) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L849)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Sounds good. Removed this for now and will do this in follow up PR. 
</blockquote></details>

Done. New commit on this PR.

tamird (changes requested):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 7 unresolved discussions (waiting on @alessandrod, @astoycos, and @preuss-adam)

___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-Ni6PfCr0vQQl118Gjzp:b-or2ca3) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Done. New commit on this PR.
</blockquote></details>

I think I don't follow. The implementation still seems to be duplicated rather than shared.
___
*[`aya/src/maps/mod.rs` line 661 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6OyNN8sVnxKF8SIuc:-Ni6OyNN8sVnxKF8SIud:b-s1kei5) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L661)):*
> ```XML
>         let info = MapInfo::new_from_fd(fd.as_fd())?;
>         Ok(Self {
>             obj: parse_map_info(info.0, PinningType::None),
> ```

consider writing `let MapInfo(info) = MapInfo::new_from_fd(...)?` to avoid this `.0`.
___
*[`aya/src/maps/mod.rs` line 719 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6P8X85-8bN0M_1yVi:-Ni6P8X99ncdnOUmbP4x:b4422lx) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L719)):*
> ```XML
>     }
> 
>     /// Returns information about a loaded map with the [`MapInfo`] structure.
> ```

consider avoiding repeating the signature in the commit. also "a map" is weird, it's _this_ map, isn't it?
___
*[`aya/src/maps/mod.rs` line 943 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6PNrECh6j0I6E3apq:-Ni6PNrF59GVP8wFbuIZ:b-ixm2ua) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L943)):*
> ```XML
>     }
> 
>     /// The name of the map as a &str. If the name was not valid unicode, None is returned.
> ```

s/was/is/

preuss-adam (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 7 unresolved discussions (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-Ni6TmXR3PRiHT6ZJR7L:b-ho1rt3) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*

I think I don't follow. The implementation still seems to be duplicated rather than shared.
</blockquote></details>

Depends how you interpret "same method" :)  In an earlier version, it did it completely differently. I could put the common code into util if you think that would be cleaner?
___
*[`aya/src/maps/mod.rs` line 661 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6OyNN8sVnxKF8SIuc:-Ni6TjPj1ADtrAPwD4N7:b-7ql1td) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L661)):*

consider writing `let MapInfo(info) = MapInfo::new_from_fd(...)?` to avoid this `.0`.
</blockquote></details>

fixed
___
*[`aya/src/maps/mod.rs` line 719 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6P8X85-8bN0M_1yVi:-Ni6VADLAYdfx86u37TO:bhv3vxs) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L719)):*

consider avoiding repeating the signature in the commit. also "a map" is weird, it's _this_ map, isn't it?
</blockquote></details>

yes, fixed.
___
*[`aya/src/maps/mod.rs` line 943 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6PNrECh6j0I6E3apq:-Ni6VKx_3jaOHC1BYXyM:b-7ql1td) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L943)):*

s/was/is/
</blockquote></details>

fixed

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 5 unresolved discussions (waiting on @alessandrod, @astoycos, and @preuss-adam)

___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-Ni6Vk4pDvfffGiVnDdy:b-amnm0o) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Depends how you interpret "same method" :)  In an earlier version, it did it completely differently. I could put the common code into util if you think that would be cleaner?
</blockquote></details>

Yeah, I think it would be preferable to avoid duplication.
___
*[`aya/src/maps/mod.rs` line 943 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6PNrECh6j0I6E3apq:-Ni6VtZn2gVPeiUspynI:b-drotnv) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L943)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

fixed
</blockquote></details>

on second thought, do we need this method? the caller just as easily do this, right?

preuss-adam (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 5 unresolved discussions (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/maps/mod.rs` line 943 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6PNrECh6j0I6E3apq:-Ni6WsL81dOhS9OikN0t:b5chcb0) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L943)):*

on second thought, do we need this method? the caller just as easily do this, right?
</blockquote></details>

Not really. Only reason is `ProgramInfo` has it.

preuss-adam (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 5 unresolved discussions (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/maps/mod.rs` line 625 at r1](https://reviewable.io/reviews/aya-rs/aya/675#-NhbC5ec866_8h6qL16g:-Ni6Z2bd2r14zzgNmQL0:bb7f2mf) ([raw file](https://github.com/aya-rs/aya/blob/39c3ca50cd9dbec53dd50100d388defa51f154dd/aya/src/maps/mod.rs#L625)):*

Yeah, I think it would be preferable to avoid duplication.
</blockquote></details>

done

tamird (approved):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 5 unresolved discussions (waiting on @alessandrod, @astoycos, and @preuss-adam)

___
*[`aya/src/util.rs` line 374 at r6](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6ZCx_B3v-sPHNq9uA:-Ni6ZCx_B3v-sPHNq9uB:b-44u2uq) ([raw file](https://github.com/aya-rs/aya/blob/2c89d73d64575e4b6a8b939578f98d4a68fd2141/aya/src/util.rs#L374)):*
> ```XML
>         .unwrap_or(0);
> 
>     // The name field is defined as [std::os::raw::c_char; 16]. c_char may be signed or
> ```

this comment is talking about something at quite some distance. maybe the signature of this function should be `bpf_name: &[std::os::raw::c_char; 16]` so that this contract is obvious? then you can just talk about the cast and why it's done.
___
*[`aya/src/maps/mod.rs` line 943 at r4](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6PNrECh6j0I6E3apq:-Ni6ZTXTCTq749IfDty1:b-yi0il5) ([raw file](https://github.com/aya-rs/aya/blob/62291186d8e6be39f3f1944e5b32484340263653/aya/src/maps/mod.rs#L943)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Not really. Only reason is `ProgramInfo` has it.
</blockquote></details>

will let @astoycos weigh in on whether either method makes sense.

preuss-adam:
@astoycos  Any other thoughts on this PR? What are the next steps to be able to submit? Thanks.

aya/src/maps/mod.rs
```diff
@@ -903,6 +908,119 @@ impl<T: Pod> Deref for PerCpuValues<T> {
...
+    }
+
+    /// Loads map info from a map id.
+    pub fn from_id(id: u32) -> Result<Self, MapError> {
+        bpf_map_get_fd_by_id(id)
+            .map_err(MapError::from)
+            .and_then(|fd| Self::new_from_fd(fd.as_fd()))
+    }
+
+    /// The name of the map.
```
astoycos, line 928:
```suggestion
    /// The name of the map, limited to 16 bytes.
```

aya/src/util.rs
```diff
@@ -364,6 +364,18 @@ pub(crate) fn bytes_of_slice<T: Pod>(val: &[T]) -> &[u8] {
     unsafe { slice::from_raw_parts(val.as_ptr().cast(), size) }
 }
 
+pub(crate) fn bytes_of_bpf_name(bpf_name: &[i8]) -> &[u8] {
```
astoycos, line 367: This won't be i8 for all architectures unfortunately (I've run into this before)

I think you'd either need to make use of generics OR just duplicate the logic for now unless @tamird Knows of a rusty way around this :)

astoycos (changes requested):
Sorry for the delay I was at Kubcon NA last week 😢 , This is looking really good to me just a few small things, thanks for the hard work especially with the unit tests 🥳  (I'll have to go back and add the same for the `loaded_programs` api) 

One last thing can you also add a smoke test like https://github.com/aya-rs/aya/blob/main/test/integration-test/src/tests/smoke.rs#L74 It should be super simple

> will let @astoycos weigh in on whether either method makes sense.

Providing the map name as a `&str` just makes it easier for the user, I like keeping it consistent with the `ProgramInfo` API 👍 

Lastly looks like we need a `cargo xtask public-api --bless` 

Thanks!

aya/src/util.rs
astoycos: Or maybe use a ffi c type?

preuss-adam: Done. Works well with the ffi c type.

aya/src/maps/mod.rs
preuss-adam: Done

astoycos (approved):
LGTM 

Thanks for all the hard work @preuss-adam I'm looking forward to this

Just need @alessandrod To checkout the API changes

tamird (changes requested):
> Reviewed 5 of 5 files at r7, 4 of 4 files at r8, 3 of 3 files at r9, all commit messages.
> Dismissed @astoycos from 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod, @astoycos, and @preuss-adam)

___
*[`aya/src/util.rs` line 367 at r6](https://reviewable.io/reviews/aya-rs/aya/675#gh-1394431823:-NjOyLVM3w0sozNInBjS:bh3t3bg) ([raw file](https://github.com/aya-rs/aya/blob/2c89d73d64575e4b6a8b939578f98d4a68fd2141/aya/src/util.rs#L367)):*
<details><summary><i>Previously, preuss-adam wrote…</i></summary><blockquote>

Done. Works well with the ffi c type.
</blockquote></details>

Do we need the leading `::` on `::core`?
___
*[`aya/src/util.rs` line 374 at r6](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6ZCx_B3v-sPHNq9uA:-NjOyGzy4goPf6mTXBUq:b-c41jnc) ([raw file](https://github.com/aya-rs/aya/blob/2c89d73d64575e4b6a8b939578f98d4a68fd2141/aya/src/util.rs#L374)):*

this comment is talking about something at quite some distance. maybe the signature of this function should be `bpf_name: &[std::os::raw::c_char; 16]` so that this contract is obvious? then you can just talk about the cast and why it's done.
</blockquote></details>

Ping this is still unresolved.

tamird (changes requested):
> Reviewed 1 of 1 files at r10, all commit messages.
> Dismissed @astoycos from a discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod, @astoycos, and @preuss-adam)

___
*[`aya/src/util.rs` line 374 at r6](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6ZCx_B3v-sPHNq9uA:-NjP1k981Z85CUdToKw6:b-cxpbul) ([raw file](https://github.com/aya-rs/aya/blob/2c89d73d64575e4b6a8b939578f98d4a68fd2141/aya/src/util.rs#L374)):*

Ping this is still unresolved.
</blockquote></details>

I still think the comment shouldn't repeat the API, and there's no "name field" in sight. I think you just want to say that `c_char` can be signed or unsigned - though I still don't understand why that leads to us using `from_raw_parts`...as opposed to what?

preuss-adam (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/util.rs` line 374 at r6](https://reviewable.io/reviews/aya-rs/aya/675#-Ni6ZCx_B3v-sPHNq9uA:-NjP3cCW0GI-e4A12W2C:bo4s7iu) ([raw file](https://github.com/aya-rs/aya/blob/2c89d73d64575e4b6a8b939578f98d4a68fd2141/aya/src/util.rs#L374)):*

I still think the comment shouldn't repeat the API, and there's no "name field" in sight. I think you just want to say that `c_char` can be signed or unsigned - though I still don't understand why that leads to us using `from_raw_parts`...as opposed to what?
</blockquote></details>

Yeah, I think the comment doesn't really add anything here anymore. I've just removed it.

preuss-adam (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @alessandrod, @astoycos, and @tamird)

___
*[`aya/src/util.rs` line 367 at r6](https://reviewable.io/reviews/aya-rs/aya/675#gh-1394431823:-NjP4-Mc9uDHgiBVdXav:b-pgib0l) ([raw file](https://github.com/aya-rs/aya/blob/2c89d73d64575e4b6a8b939578f98d4a68fd2141/aya/src/util.rs#L367)):*

Do we need the leading `::` on `::core`?
</blockquote></details>

nope. removed.

tamird (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/675)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alessandrod and @astoycos)

astoycos (approved):

alessandrod (approved):
Looks great, thanks everyone!

PR #676, opened 2023-07-18
integration-ebpf: artifact dependency bpf-linker

tamird:
Replace the manual dependency machinery in integration-test/build.rs
with:

- integration-ebpf -[artifact-dependency]-> bpf-linker.
- integration-test -[build-dependency]-> integration-ebpf.
- integration-ebpf: build.rs that makes bpf-linker available to rustc.
- integration-ebpf: lib.rs (empty) creates a lib target, allows dep from
  integration-test.
- xtask: filter for test binaries now that the build also produces the
  bpf-linker binary.
- github: remove `cargo install bpf-linker` and commentary about cache
  invalidation.
- integration-test: remove bpf-linker installation instructions.


tamird:
Drat. This breaks the workflows on stable because cargo refuses to read `test/integration-ebpf/Cargo.toml` which now contains `artifact = ...` even though we `--ignore integration-ebpf` on the command line.

I'm thinking about how to solve this.

PR #681, opened 2023-07-20
test: document cargo build scripts

tamird:


Cargo.toml
```diff
@@ -29,7 +29,8 @@ default-members = [
     "aya-log-parser",
     "aya-obj",
     "aya-tool",
-    "test/integration-test",
+    # test/integration-test is omitted; including it in this list causes `cargo test` to run its
+    # tests, and that doesn't work unless they've been build with `cargo xtask`.
```
dave-tucker, line 33:
```suggestion
    # tests, and that doesn't work outside of `cargo xtask integration-test`.
```

because it's not just the stub files that are the problem, it's permissions to make `bpf()` syscalls.

test/integration-test/build.rs
```diff
@@ -13,6 +13,29 @@ use cargo_metadata::{
...
+/// crate's undeclared dependency on artifacts from `integration-ebpf` would cause build (and `cargo check`,
+/// and `cargo clippy`) failures until the user ran certain other commands in the workspace. Conversely,
+/// those same tools (e.g. cargo test --no-run) would produce stale results if run naively because
+/// they'd make use of artifacts from a previous build of `integration-ebpf`.
+///
+/// Note that this solution is imperfect: in particular it has to balance correctness with
+/// performance; an environment variable is used to replace true builds of `integration-ebpf` with
+/// stubs to preserve the property that code generation and linking (in `integration-ebpf`) do not
+/// occur on metadata-only actions such as `cargo check` or `cargo clippy` of this crate. This means
+/// that naively attempting to `cargo test` this crate will fail at runtime because the stubs are
```
dave-tucker, line 34: You may wish to point out that naively running `cargo test` here would fail anyway, due to permissions.
So the net result is no worse that before.

ajwerner (commented):
This is very helpful. Thanks for doing it.

Cargo.toml
tamird: That's not quite true; you could use `cargo xtask build-integration-test` which will print the names of the files on stdout, and then copy them somewhere and run them with the necessary permissions.

test/integration-test/build.rs
tamird: Updated to indicate that this applies to `cargo test --no-run`.

dave-tucker (approved):

PR #682, opened 2023-07-20
xtask: Add cargo-public-api

dave-tucker:
Adds cargo-public-api. This allows for public API changes to get caught
in CI, requiring new changes to be "blessed" by using:

  cargo xtask public-api --bless

When this file is changed for aya, Alessandro will need to review the PR.


tamird (changes requested):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @dave-tucker)

___
*[`.cargo/config.toml` line 14 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o0W-rCuQ2PpuS6H2_:-N_o0W-rCuQ2PpuS6H2a:b344yvi) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/.cargo/config.toml#L14)):*
> ```TOML
> 
> [env]
> CARGO_WORKSPACE_DIR = { value = "", relative = true }
> ```

why is this needed? can you add a comment?
___
*[`aya/Cargo.toml` line 32 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o0YAU6VQnjNcpQztD:-N_o0YAU6VQnjNcpQztE:b70par) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/Cargo.toml#L32)):*
> ```TOML
> [dev-dependencies]
> assert_matches = { workspace = true }
> expect-test = "1.4.1"
> ```

can we put all these in the workspace? if ever we decide to enforce the public API of other crates, we'll want these same versions. I think the idea is that all version strings go in the root cargo.toml
___
*[`aya/src/lib.rs` line 65 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o0iGW1LHUASx7pbGx:-N_o0iGW1LHUASx7pbGy:b1cwmha) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L65)):*
> ```XML
> mod tests {
>     #[test]
>     #[cfg_attr(miri, ignore)]
> ```

can you use `ignore = "string that explains why this can't work with miri"`?
___
*[`aya/src/lib.rs` line 67 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o0qhw-Hvp2jsRfevI:-N_o0qhw-Hvp2jsRfevJ:bm58fbt) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L67)):*
> ```XML
>     #[cfg_attr(miri, ignore)]
>     fn public_api() {
>         // Install a compatible nightly toolchain if it is missing
> ```

is this a reasonable thing to do to a user's machine if they run `cargo test`?
___
*[`aya/src/lib.rs` line 70 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o12ytD5kYWnJl5r06:-N_o12ytD5kYWnJl5r07:b9mqarc) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L70)):*
> ```XML
>         rustup_toolchain::install(public_api::MINIMUM_NIGHTLY_RUST_VERSION).unwrap();
> 
>         // Build rustdoc JSON
> ```

nit: i think this comment and the ones below are all restating the code; i suggest just removing them.

conversly we **should** have a comment explaining why a nightly toolchain is required.
___
*[`aya/src/lib.rs` line 82 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o2-XL4_rIpjcKGz7G:-N_o2-XL4_rIpjcKGz7H:b5xofuv) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L82)):*
> ```XML
> 
>         // Assert that the public API looks correct
>         expect_test::expect_file!["./public-api.txt"].assert_eq(&public_api.to_string());
> ```

do you think we'd get prettier output from [`PublicApiDiff::between`](https://docs.rs/public-api/latest/public_api/diff/struct.PublicApiDiff.html#method.between)? It'd save a dependency as well.

We'd probably have to write the UPDATE_EXPECT code here ourselves, but it doesn't seem that onerous.

```rust
let update_expect = env::var("UPDATE_EXPECT")
    .as_deref()
    .map(str::parse)
    .map(Result::unwrap)
    .unwrap_or_default();
let Metadata { workspace_root, .. } = MetadataCommand::new()
        .no_deps()
        .exec()
        .unwrap();
let api_file = workspace_root.join("public-api.json");
if update_expect {
    std::fs::rename(rustc_json, api_file).unwrap();
} else {
    let before = public_api::Builder::from_rustdoc_json(rustdoc_json).build().unwrap();
    let after = public_api::Builder::from_rustdoc_json(api_file).build().unwrap();
    let diff = public_api::diff::PublicApiDiff::from(before, after);
    assert!(diff.is_empty(), "{:#?}", diff);
}
```
implicit in this is that we'd store the json on disk rather than text.
___
*[`aya/src/maps/mod.rs` line 227 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o1W6b7fTf9vBL2JIc:-N_o1OC8CwBAxNtfUYFS:blgdfz8) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/maps/mod.rs#L227)):*
> ```XML
> 
>         let limit: RlimitSize = RlimitSize(limit.rlim_cur.try_into().unwrap());
>         if limit.0 == TryInto::<usize>::try_into(RLIM_INFINITY).unwrap() {
> ```

how come this was necessary? it'd be terser to write `usize::try_from(RLIM_INFINITY)`

dave-tucker:
thanks @tamird - moving this to draft as there are still some UI issues to work through:

- Rust toolchain needs to meet the minimum required for `cargo_public_api` - out nightly version is _always latest_ so I think the right option would be to fail the test if the nightly doesn't meet the minimum requirement.

- There is some weird interaction between `cargo-public-api`, `expect_test` and potentially `cargo-hack` which is preventing the public API file from being found.

On your comment:

> _[`aya/src/lib.rs` line 82 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o2-XL4_rIpjcKGz7G:-N_o2-XL4_rIpjcKGz7H:b5xofuv) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L82)):_
> 
> > ```
> >         // Assert that the public API looks correct
> >         expect_test::expect_file!["./public-api.txt"].assert_eq(&public_api.to_string());
> > ```
> 
> do you think we'd get prettier output from [`PublicApiDiff::between`](https://docs.rs/public-api/latest/public_api/diff/struct.PublicApiDiff.html#method.between)? It'd save a dependency as well.
> 
> We'd probably have to write the UPDATE_EXPECT code here ourselves, but it doesn't seem that onerous.
> 
> ```rust
> let update_expect = env::var("UPDATE_EXPECT")
>     .as_deref()
>     .map(str::parse)
>     .map(Result::unwrap)
>     .unwrap_or_default();
> let Metadata { workspace_root, .. } = MetadataCommand::new()
>         .no_deps()
>         .exec()
>         .unwrap();
> let api_file = workspace_root.join("public-api.json");
> if update_expect {
>     std::fs::rename(rustc_json, api_file).unwrap();
> } else {
>     let before = public_api::Builder::from_rustdoc_json(rustdoc_json).build().unwrap();
>     let after = public_api::Builder::from_rustdoc_json(api_file).build().unwrap();
>     let diff = public_api::diff::PublicApiDiff::from(before, after);
>     assert!(diff.is_empty(), "{:#?}", diff);
> }
> ```
> 
> implicit in this is that we'd store the json on disk rather than text.

The rationale for storing the public-api in text format on disk is that it should render a decent human-readable diff when the public API changes. And yes, as I'm digging into it, we might just throw out update_expect as I believe this is where the root of the CI failures are coming from 🤔 

> _[`aya/src/maps/mod.rs` line 227 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o1W6b7fTf9vBL2JIc:-N_o1OC8CwBAxNtfUYFS:blgdfz8) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/maps/mod.rs#L227)):_
> 
> > ```
> >         let limit: RlimitSize = RlimitSize(limit.rlim_cur.try_into().unwrap());
> >         if limit.0 == TryInto::<usize>::try_into(RLIM_INFINITY).unwrap() {
> > ```
> 
> how come this was necessary? it'd be terser to write `usize::try_from(RLIM_INFINITY)`

🤷 the moment I added the test RA got upset by this line. will look into it.

tamird (commented):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)* status: all files reviewed, 7 unresolved discussions (waiting on @dave-tucker)

___
*[`aya/src/lib.rs` line 82 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o2-XL4_rIpjcKGz7G:-N_tR1DIE3286iE8x4m0:b-j182cg) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L82)):*

The rationale for storing the public-api in text format on disk is that it should render a decent human-readable diff when the public API changes. And yes, as I'm digging into it, we might just throw out update_expect as I believe this is where the root of the CI failures are coming from 🤔 
</blockquote></details>

I still think we should print a fancy diff - that's much better in e.g. CI where a reviewer would otherwise get no information.
___
*[`aya/src/lib.rs` line 68 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tRBDg9xnRjjGaApcp:-N_tRBDg9xnRjjGaApcq:btqvuef) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L68)):*
> ```XML
>         env,
>         fs::{read_to_string, File},
>         io::Write,
> ```

nit: prefer to import traits that aren't referenced `as _`.
___
*[`aya/src/lib.rs` line 75 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tlmAa5vyamBxLvDsA:-N_tlmAa5vyamBxLvDsB:b-j76pht) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L75)):*
> ```XML
>     fn public_api() {
>         let rustdoc_json = rustdoc_json::Builder::default()
>             .toolchain("nightly")
> ```

looks like the test fails when nightly isn't installed, which is expected. should we use [`rustup_toolchain::is_installed`](https://docs.rs/rustup-toolchain/latest/rustup_toolchain/fn.is_installed.html) here?
___
*[`aya/src/lib.rs` line 88 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tQN-e8cuM4zQalvp_:-N_tQN-e8cuM4zQalvpa:bw33ftr) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L88)):*
> ```XML
>             .no_deps()
>             .exec()
>             .expect("failed to run cargo metadata");
> ```

i think you can just unwrap here like you do everything else.
___
*[`aya/src/lib.rs` line 93 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tQR3S3f5dYm3HMFko:-N_tQR3S3f5dYm3HMFkp:b-qz62l3) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L93)):*
> ```XML
>         let path = workspace_root.join("aya/src/public-api.txt");
> 
>         if env::var("UPDATE_EXPECT").is_ok() {
> ```

if you're just checking presence, use env::var_os() which returns an option, avoiding the appearance of discarding an error here
___
*[`aya/src/lib.rs` line 94 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tQY6bDEKIPeZw7Z9s:-N_tQY6bDEKIPeZw7Z9t:bi5fg5p) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L94)):*
> ```XML
> 
>         if env::var("UPDATE_EXPECT").is_ok() {
>             let mut f = File::create(&path).expect("failed to create aya/src/public-api.txt");
> ```

```
std::fs::write(&path, public_api).unwrap_or_else(|err| panic!("failed to write {path}: {err}"))
```

dave-tucker:
> > Reviewed 6 of 6 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)_ status: all files reviewed, 7 unresolved discussions (waiting on @dave-tucker)
> 
> _[`aya/src/lib.rs` line 82 at r1](https://reviewable.io/reviews/aya-rs/aya/682#-N_o2-XL4_rIpjcKGz7G:-N_tR1DIE3286iE8x4m0:b-j182cg) ([raw file](https://github.com/aya-rs/aya/blob/ca2c360e69c9dde93bad920320d5528f30d4b4c8/aya/src/lib.rs#L82)):_
> 
> Previously, dave-tucker (Dave Tucker) wrote…
> I still think we should print a fancy diff - that's much better in e.g. CI where a reviewer would otherwise get no information.

Done. Semi-fancy diff implemented.

```console
---- tests::public_api stdout ----
thread 'tests::public_api' panicked at 'public api has changed.
please bless by re-running tests with UPDATE_EXPECT=1

+pub mod aya::maps2
-pub mod aya::maps
', aya/src/lib.rs:103:9
```

> _[`aya/src/lib.rs` line 68 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tRBDg9xnRjjGaApcp:-N_tRBDg9xnRjjGaApcq:btqvuef) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L68)):_
> 
> > ```
> >         env,
> >         fs::{read_to_string, File},
> >         io::Write,
> > ```
> 
> nit: prefer to import traits that aren't referenced `as _`.

Done

> _[`aya/src/lib.rs` line 75 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tlmAa5vyamBxLvDsA:-N_tlmAa5vyamBxLvDsB:b-j76pht) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L75)):_
> 
> > ```
> >     fn public_api() {
> >         let rustdoc_json = rustdoc_json::Builder::default()
> >             .toolchain("nightly")
> > ```
> 
> looks like the test fails when nightly isn't installed, which is expected. should we use [`rustup_toolchain::is_installed`](https://docs.rs/rustup-toolchain/latest/rustup_toolchain/fn.is_installed.html) here?

I backed out that dependency - it's already a requirement for Aya developers to have both stable and nightly available so I'm fine panicking here. Just need to get both nightly and stable in the `build-aya` CI job 🤞 

> _[`aya/src/lib.rs` line 88 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tQN-e8cuM4zQalvp_:-N_tQN-e8cuM4zQalvpa:bw33ftr) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L88)):_
> 
> > ```
> >             .no_deps()
> >             .exec()
> >             .expect("failed to run cargo metadata");
> > ```
> 
> i think you can just unwrap here like you do everything else.

Done.

> _[`aya/src/lib.rs` line 93 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tQR3S3f5dYm3HMFko:-N_tQR3S3f5dYm3HMFkp:b-qz62l3) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L93)):_
> 
> > ```
> >         let path = workspace_root.join("aya/src/public-api.txt");
> > 
> >         if env::var("UPDATE_EXPECT").is_ok() {
> > ```
> 
> if you're just checking presence, use env::var_os() which returns an option, avoiding the appearance of discarding an error here

Also Done.

> _[`aya/src/lib.rs` line 94 at r2](https://reviewable.io/reviews/aya-rs/aya/682#-N_tQY6bDEKIPeZw7Z9s:-N_tQY6bDEKIPeZw7Z9t:bi5fg5p) ([raw file](https://github.com/aya-rs/aya/blob/c8200edd6ae2fce3fe96832bc6497e61a433f611/aya/src/lib.rs#L94)):_
> 
> > ```
> >         if env::var("UPDATE_EXPECT").is_ok() {
> >             let mut f = File::create(&path).expect("failed to create aya/src/public-api.txt");
> > ```
> 
> ```
> std::fs::write(&path, public_api).unwrap_or_else(|err| panic!("failed to write {path}: {err}"))
> ```

And done again!

Thanks for the review. Once this passes CI I'll remove it from draft.

tamird (commented):
> Reviewed 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)* status: all files reviewed, 9 unresolved discussions (waiting on @dave-tucker)

___
*[`xtask/src/public_api.rs` line 33 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRu9s63gPfRv5ghCE:-N_xRu9s63gPfRv5ghCF:b-yhg4xh) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L33)):*
> ```XML
>     let Metadata { workspace_root, .. } = &metadata;
> 
>     let packages = ["aya", "aya-bpf"];
> ```

should we do this for all crates that we publish? https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Package.html#structfield.publish i think we all want all packages where `publish != Some(vec![])`.
___
*[`xtask/src/public_api.rs` line 34 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRjkI2o9VvjIVH4V3:-N_xRjkI2o9VvjIVH4V4:b4bzv68) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L34)):*
> ```XML
> 
>     let packages = ["aya", "aya-bpf"];
>     packages.iter().try_for_each(|package| {
> ```

i think this will bail on the first error - so if two packages changed we'll only get the first one. should let them all run?
___
*[`xtask/src/public_api.rs` line 38 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xREjT5z0iBauaAL88:-N_xREjT5z0iBauaAL89:bkrw4eb) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L38)):*
> ```XML
>             package,
>             toolchain,
>             options.bless,
> ```

nit: destructure the options at the top of the function? it's a nice assertion that we use them all.
___
*[`xtask/src/public_api.rs` line 41 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPGUc9UhnoKnra2JB:-N_xPGUc9UhnoKnra2JC:bikjhdn) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L41)):*
> ```XML
>             workspace_root.as_std_path(),
>         )
>     })?;
> ```

i think you can drop the `?; Ok(())`
___
*[`xtask/src/public_api.rs` line 57 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRLLA-q86W7zacYsJ:-N_xRLLA-q86W7zacYsK:bcjof4l) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L57)):*
> ```XML
>         .toolchain(toolchain)
>         .package(package)
>         .all_features(true)
> ```

do we need to do the power set of features like we discussed? what about API that is `#[cfg(not(some_feature))]`?
___
*[`xtask/src/public_api.rs` line 63 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPpqI8S5jaijamYXM:-N_xPpqI8S5jaijamYXN:bbrosib) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L63)):*
> ```XML
> 
>     if bless {
>         std::fs::write(&path, public_api.to_string().as_bytes())?;
> ```

somewhat strange this one is fully qualified but not read_to_string
___
*[`xtask/src/public_api.rs` line 66 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPk2F13wW0yPWdP-F:-N_xPk2F13wW0yPWdP-G:b7e618c) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L66)):*
> ```XML
>     }
>     let current_api =
>         read_to_string(&path).context(format!("{} does not exist", &path.display()))?;
> ```

"does not exist" is not correct. this can fail for various reasons
___
*[`xtask/src/public_api.rs` line 68 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xQPGD89VYkp1oJUEg:-N_xQPGD89VYkp1oJUEh:b-dolk5g) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L68)):*
> ```XML
>         read_to_string(&path).context(format!("{} does not exist", &path.display()))?;
> 
>     let delta = lines(&current_api, &public_api.to_string())
> ```

we seem to be pulling in a lot of dependencies for this diff. why not use PublicApi::Diff? We'd need to store json instead of txt on disk, is that a problem?
___
*[`xtask/src/public_api.rs` line 83 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPb6z7gJm_5R3Ipf2:-N_xPb6z7gJm_5R3Ipf3:bro7ool) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L83)):*
> ```XML
>             println!("{}", line);
>         });
>         println!("\n");
> ```

did you want two newlines here? if you just want one that's `println!()`

dave-tucker:
> > Reviewed 10 of 10 files at r3, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)_ status: all files reviewed, 9 unresolved discussions (waiting on @dave-tucker)
> 
> _[`xtask/src/public_api.rs` line 33 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRu9s63gPfRv5ghCE:-N_xRu9s63gPfRv5ghCF:b-yhg4xh) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L33)):_
> 
> > ```
> >     let Metadata { workspace_root, .. } = &metadata;
> > 
> >     let packages = ["aya", "aya-bpf"];
> > ```
> 
> should we do this for all crates that we publish? https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Package.html#structfield.publish i think we all want all packages where `publish != Some(vec![])`.

That should be easy enough, I'll take a look at this later.

> _[`xtask/src/public_api.rs` line 34 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRjkI2o9VvjIVH4V3:-N_xRjkI2o9VvjIVH4V4:b4bzv68) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L34)):_
> 
> > ```
> >     let packages = ["aya", "aya-bpf"];
> >     packages.iter().try_for_each(|package| {
> > ```
> 
> i think this will bail on the first error - so if two packages changed we'll only get the first one. should let them all run?

I can look at that too.

> _[`xtask/src/public_api.rs` line 38 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xREjT5z0iBauaAL88:-N_xREjT5z0iBauaAL89:bkrw4eb) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L38)):_
> 
> > ```
> >             package,
> >             toolchain,
> >             options.bless,
> > ```
> 
> nit: destructure the options at the top of the function? it's a nice assertion that we use them all.

👍 will incorporate this in the next revision

> _[`xtask/src/public_api.rs` line 41 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPGUc9UhnoKnra2JB:-N_xPGUc9UhnoKnra2JC:bikjhdn) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L41)):_
> 
> > ```
> >             workspace_root.as_std_path(),
> >         )
> >     })?;
> > ```
> 
> i think you can drop the `?; Ok(())`

and this.

> _[`xtask/src/public_api.rs` line 57 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRLLA-q86W7zacYsJ:-N_xRLLA-q86W7zacYsK:bcjof4l) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L57)):_
> 
> > ```
> >         .toolchain(toolchain)
> >         .package(package)
> >         .all_features(true)
> > ```
> 
> do we need to do the power set of features like we discussed? what about API that is `#[cfg(not(some_feature))]`?

That's going to have to happen in a follow up PR - the most sustainable way would be for `cargo-hack` to expose a lib with it's `powerset` functionality. Building this from cargo metadata ourselves isn't straightforward and I'd rather not block this PR on it.

> _[`xtask/src/public_api.rs` line 63 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPpqI8S5jaijamYXM:-N_xPpqI8S5jaijamYXN:bbrosib) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L63)):_
> 
> > ```
> >     if bless {
> >         std::fs::write(&path, public_api.to_string().as_bytes())?;
> > ```
> 
> somewhat strange this one is fully qualified but not read_to_string

👍 

> _[`xtask/src/public_api.rs` line 66 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPk2F13wW0yPWdP-F:-N_xPk2F13wW0yPWdP-G:b7e618c) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L66)):_
> 
> > ```
> >     }
> >     let current_api =
> >         read_to_string(&path).context(format!("{} does not exist", &path.display()))?;
> > ```
> 
> "does not exist" is not correct. this can fail for various reasons

👍 

> _[`xtask/src/public_api.rs` line 68 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xQPGD89VYkp1oJUEg:-N_xQPGD89VYkp1oJUEh:b-dolk5g) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L68)):_
> 
> > ```
> >         read_to_string(&path).context(format!("{} does not exist", &path.display()))?;
> > 
> >     let delta = lines(&current_api, &public_api.to_string())
> > ```
> 
> we seem to be pulling in a lot of dependencies for this diff. why not use PublicApi::Diff? We'd need to store json instead of txt on disk, is that a problem?

As I said before, yes it's a problem. The JSON is a lot more verbose + doesn't produce a nice diff.
`PublicApi::Diff` uses `diff` anyway (you can check via cargo tree) so the net result doesn't require any more deps.

> _[`xtask/src/public_api.rs` line 83 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xPb6z7gJm_5R3Ipf2:-N_xPb6z7gJm_5R3Ipf3:bro7ool) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L83)):_
> 
> > ```
> >             println!("{}", line);
> >         });
> >         println!("\n");
> > ```
> 
> did you want two newlines here? if you just want one that's `println!()`

I did! It helps make the output more legible, but I'll add a comment to that effect.


Will issue another revision of this on (or before) Monday.

tamird (changes requested):
> Reviewed 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)* status: all files reviewed, 5 unresolved discussions (waiting on @dave-tucker)

___
*[`xtask/src/public_api.rs` line 57 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRLLA-q86W7zacYsJ:-Na77utw7JAaRyFDzUdD:by7gqfd) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L57)):*

That's going to have to happen in a follow up PR - the most sustainable way would be for `cargo-hack` to expose a lib with it's `powerset` functionality. Building this from cargo metadata ourselves isn't straightforward and I'd rather not block this PR on it.
</blockquote></details>

seems alright.
___
*[`xtask/src/public_api.rs` line 68 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xQPGD89VYkp1oJUEg:-Na7898pAJOMDFf1nVbS:b-11hobi) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L68)):*

As I said before, yes it's a problem. The JSON is a lot more verbose + doesn't produce a nice diff.
`PublicApi::Diff` uses `diff` anyway (you can check via cargo tree) so the net result doesn't require any more deps.

</blockquote></details>

Alright.
___
*[`xtask/src/public_api.rs` line 11 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na78INC7yiEGlNtiGKT:-Na78INC7yiEGlNtiGKU:b-7e3p74) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L11)):*
> ```XML
> use colored::*;
> use dialoguer::{theme::ColorfulTheme, Confirm};
> use diff::{lines, Result as DiffResult};
> ```

is DiffResult more readable than diff::Result?
___
*[`xtask/src/public_api.rs` line 44 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na766am6L5wcq4Gqw1r:-Na766am6L5wcq4Gqw1s:bya9epq) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L44)):*
> ```XML
>         .iter()
>         .filter_map(|package| {
>             if package.publish.is_none() {
> ```

i think this is not correct. the only case we want to discard is Some(<empty vector>).

also, we should emit all the errors we got because we can't be certain of their causes.

how about:
```rust
let errors = packages.into_iter().filter_map(|Package { name, publish, .. }| {
  if matches!(publish, Some(publish) if publish.is_empty()) {
    return None;
  }
  check_package_api(name, toolchain, bless, workspace_root.as_std_path()).err() // <-- note `.err`; we only want to keep the ones that errored
};
if !errors.is_empty() {
  bail(...);
}
Ok(())
```
___
*[`xtask/src/public_api.rs` line 68 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na78aix6LDDpS5DkRl3:-Na78V8DAsgEjULPBDAB:b-6un0z1) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L68)):*
> ```XML
>     workspace_root: &Path,
> ) -> anyhow::Result<()> {
>     let path = workspace_root.join(format!("xtask/public-api/{}.txt", package));
> ```

```
workspace_root.join("xtask").join("public-api").join(package).with_extension("txt");
```
___
*[`xtask/src/public_api.rs` line 82 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na77nx7CxF_lJRD4Uyt:-Na77nx7CxF_lJRD4Uyu:b-x5pimz) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L82)):*
> ```XML
>     }
>     let current_api =
>         read_to_string(&path).context(format!("error reading {}", &path.display()))?;
> ```

can this be `with_context`?
___
*[`xtask/src/public_api.rs` line 94 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na77_iz9DRf041R4Jgc:-Na77_j-DWd64zcLUst-:b-napeip) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L94)):*
> ```XML
> 
>     if !delta.is_empty() {
>         println!("public api for {} changed", package);
> ```

instead of mixing printed strings and emitted errors, can we put all the information into the error?

dave-tucker:
> > Reviewed 12 of 12 files at r4, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)_ status: all files reviewed, 5 unresolved discussions (waiting on @dave-tucker)
> 
> _[`xtask/src/public_api.rs` line 57 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xRLLA-q86W7zacYsJ:-Na77utw7JAaRyFDzUdD:by7gqfd) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L57)):_
> 
> Previously, dave-tucker (Dave Tucker) wrote…
> seems alright.
> 
> _[`xtask/src/public_api.rs` line 68 at r3](https://reviewable.io/reviews/aya-rs/aya/682#-N_xQPGD89VYkp1oJUEg:-Na7898pAJOMDFf1nVbS:b-11hobi) ([raw file](https://github.com/aya-rs/aya/blob/6bee9740d7f3d4e02a865c88d21b8e70a3e61274/xtask/src/public_api.rs#L68)):_
> 
> Previously, dave-tucker (Dave Tucker) wrote…
> Alright.
> 
> _[`xtask/src/public_api.rs` line 11 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na78INC7yiEGlNtiGKT:-Na78INC7yiEGlNtiGKU:b-7e3p74) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L11)):_
> 
> > ```
> > use colored::*;
> > use dialoguer::{theme::ColorfulTheme, Confirm};
> > use diff::{lines, Result as DiffResult};
> > ```
> 
> is DiffResult more readable than diff::Result?

I feel weird about anything calling itself `Result` that isn't a `Result` type.
I could change this to `Result as Diff` which would yield `Diff::Both`, `Diff::Left`, Diff::Right` which seems more readable to me.

> _[`xtask/src/public_api.rs` line 44 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na766am6L5wcq4Gqw1r:-Na766am6L5wcq4Gqw1s:bya9epq) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L44)):_
> 
> > ```
> >         .iter()
> >         .filter_map(|package| {
> >             if package.publish.is_none() {
> > ```
> 
> i think this is not correct. the only case we want to discard is Some().
> 
> also, we should emit all the errors we got because we can't be certain of their causes.
> 
> how about:
> 
> ```rust
> let errors = packages.into_iter().filter_map(|Package { name, publish, .. }| {
>   if matches!(publish, Some(publish) if publish.is_empty()) {
>     return None;
>   }
>   check_package_api(name, toolchain, bless, workspace_root.as_std_path()).err() // <-- note `.err`; we only want to keep the ones that errored
> };
> if !errors.is_empty() {
>   bail(...);
> }
> Ok(())
> ```

:+1:

> _[`xtask/src/public_api.rs` line 68 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na78aix6LDDpS5DkRl3:-Na78V8DAsgEjULPBDAB:b-6un0z1) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L68)):_
> 
> > ```
> >     workspace_root: &Path,
> > ) -> anyhow::Result<()> {
> >     let path = workspace_root.join(format!("xtask/public-api/{}.txt", package));
> > ```
> 
> ```
> workspace_root.join("xtask").join("public-api").join(package).with_extension("txt");
> ```
> 
> _[`xtask/src/public_api.rs` line 82 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na77nx7CxF_lJRD4Uyt:-Na77nx7CxF_lJRD4Uyu:b-x5pimz) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L82)):_
> 
> > ```
> >     }
> >     let current_api =
> >         read_to_string(&path).context(format!("error reading {}", &path.display()))?;
> > ```
> 
> can this be `with_context`?

It can.

> _[`xtask/src/public_api.rs` line 94 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na77_iz9DRf041R4Jgc:-Na77_j-DWd64zcLUst-:b-napeip) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L94)):_
> 
> > ```
> >     if !delta.is_empty() {
> >         println!("public api for {} changed", package);
> > ```
> 
> instead of mixing printed strings and emitted errors, can we put all the information into the error?

Encapsulated this into an error type which carries the package name + diff.
Printing that error to console is handled in the main function of this subcommand.

tamird (approved):
:lgtm:

> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682#-:-Na8EQxVCD4FGQDD9kjN:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @dave-tucker)

___
*[`xtask/src/public_api.rs` line 11 at r4](https://reviewable.io/reviews/aya-rs/aya/682#-Na78INC7yiEGlNtiGKT:-Na8D7I31zLENKOFYOJR:b76f16w) ([raw file](https://github.com/aya-rs/aya/blob/7b9755bbf31cf9544b371b15beeb381553259b12/xtask/src/public_api.rs#L11)):*

I feel weird about anything calling itself `Result` that isn't a `Result` type.
I could change this to `Result as Diff` which would yield `Diff::Both`, `Diff::Left`, Diff::Right` which seems more readable to me.

</blockquote></details>

Seems fine.
___
*[`xtask/src/public_api.rs` line 64 at r5](https://reviewable.io/reviews/aya-rs/aya/682#-Na8DHzK1YbjLN-1zu2V:-Na8DHzK1YbjLN-1zu2W:b-yagoet) ([raw file](https://github.com/aya-rs/aya/blob/15ede52330044c7fe98a65eb1f17361165ba7073/xtask/src/public_api.rs#L64)):*
> ```XML
> 
>     if !errors.is_empty() {
>         errors.iter().for_each(|e| eprintln!("{}", e));
> ```

can we please not print? why not build up a string containing all the errors (instead of collecting into a vector above, even) and return it?
___
*[`xtask/src/public_api.rs` line 121 at r5](https://reviewable.io/reviews/aya-rs/aya/682#-Na8Dgkw9WkMyW7SdxtH:-Na8Dgkw9WkMyW7SdxtI:bbzcgvf) ([raw file](https://github.com/aya-rs/aya/blob/15ede52330044c7fe98a65eb1f17361165ba7073/xtask/src/public_api.rs#L121)):*
> ```XML
>             Diff::Left(line) => Some(format!("+{}", line).green()),
>         })
>         .collect::<Vec<_>>();
> ```

there's no reason for this vector any more, right? you can just do the thing you do with strings below; if you end up with an empty string, there was no diff.

consider using [`writeln`](https://doc.rust-lang.org/stable/std/macro.writeln.html) to write directly into the string without intermediate `format!`s.

tamird (commented):
Merging with unresolved review comments is not ideal.

> Reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682#-:-NaJ1BsK6PiCrQXcob9-:blevvkj)* status: 20 of 21 files reviewed, 4 unresolved discussions

___
*[`xtask/src/public_api.rs` line 1 at r6](https://reviewable.io/reviews/aya-rs/aya/682#-NaImTKRAs9G3Uq-xzkU:-NaImTKS6cavMfiFI8XQ:bo8rs6e) ([raw file](https://github.com/aya-rs/aya/blob/38969df748de66cada24be7ca959f587a5820333/xtask/src/public_api.rs#L1)):*
> ```XML
> use std::fmt::Write as _;
> ```

group it with the existing group please
___
*[`xtask/src/public_api.rs` line 74 at r6](https://reviewable.io/reviews/aya-rs/aya/682#-NaImzAIDWbbYcPmsfjd:-NaImzAJA2W-dKEeynnQ:b-kf3r0b) ([raw file](https://github.com/aya-rs/aya/blob/38969df748de66cada24be7ca959f587a5820333/xtask/src/public_api.rs#L74)):*
> ```XML
>     bless: bool,
>     workspace_root: &Path,
> ) -> Result<(), PublicApiError> {
> ```

wouldn't this be simpler as anyhow::Result<String>? you can trivially `.context(package-name)` in the caller rather than at every point of error here.

this would obviate the need for PublicApiError.

tamird (commented):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/682)* status: all files reviewed, 3 unresolved discussions

PR #683, opened 2023-07-21
Refactor log macro for readability

tamird:
This happens to fix the miscompilation that occurs when bpf-linker is
moved to LLVM's new pass manager. A later commit will avoid the
miscompilation more convincingly.

### DIFF
--- a/aya-log-ebpf-macros/src/expand.rs
+++ b/aya-log-ebpf-macros/src/expand.rs
@@ -144,34 +144,27 @@ pub(crate) fn log(args: LogArgs, level: Option<TokenStream>) -> Result<TokenStre
     let values_iter = values.iter();
 
     Ok(quote! {
-        {
-            if let Some(buf_ptr) = unsafe { ::aya_log_ebpf::AYA_LOG_BUF.get_ptr_mut(0) } {
-                let buf = unsafe { &mut *buf_ptr };
-                if let Ok(header_len) = ::aya_log_ebpf::write_record_header(
-                    &mut buf.buf,
-                    #target,
-                    #lvl,
-                    module_path!(),
-                    file!(),
-                    line!(),
-                    #num_args,
-                ) {
-                    let record_len = header_len;
-
-                    if let Ok(record_len) = {
-                        Ok::<_, ()>(record_len) #( .and_then(|record_len| {
-                            if record_len >= buf.buf.len() {
-                                return Err(());
-                            }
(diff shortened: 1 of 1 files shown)


ajwerner (approved):
LGTM

alessandrod:
Do you know what the miscompilation looks like and why this fixes it?

tamird:
The hypothesis is lining thresholds change under the new PM, and this rearrangement gets this inlined with the new PM.

The root cause is that the logging functions should take self by reference to avoid passing lots of parameters and triggering miscompilation

aya-log-ebpf-macros/src/expand.rs
```diff
@@ -144,34 +144,27 @@ pub(crate) fn log(args: LogArgs, level: Option<TokenStream>) -> Result<TokenStre
...
-                        }) )*
-                    } {
-                        unsafe { ::aya_log_ebpf::AYA_LOGS.output(
-                            #ctx,
-                            &buf.buf[..record_len], 0
-                        )}
-                    }
-                }
+        match unsafe { &mut ::aya_log_ebpf::AYA_LOG_BUF }.get_ptr_mut(0).and_then(|ptr| unsafe { ptr.as_mut() }) {
+            None => {},
```
alessandrod, line 148: isn't this if let? Why doesn't clippy complain about this?

alessandrod:
If you're going to rework the WriteToBuf impls or whtever it's called, it would be good to actually remove the inlining from aya-log. It's common to have several log calls and aya-log currently consumes a big part of the insn budget because of all the inlining.

tamird:
Are you saying we should `#[inline(never)]`? The only inlining annotations I see are [here](https://github.com/aya-rs/aya/blob/aed74d5ab31c7ab86848d24709d1117b3c32240b/aya-log-common/src/lib.rs#L193) and [here](https://github.com/aya-rs/aya/blob/aed74d5ab31c7ab86848d24709d1117b3c32240b/aya-log-common/src/lib.rs#L269); the first one doesn't seem significant, the second has to do with the header (not the fields, which is what we're discussing the inlining of here).

alessandrod:
And also the inline on writing the record header needs to go too

tamird:
Yeah, I agree. I started pulling on this thread and it's getting a bit involved; I'm going to merge this as-is now and revisit when we have working debuginfo which will make figuring out the verifier issues I'm chasing now easier.

PR #687, opened 2023-07-25
xtask: allow arbitrary flags to cargo build

ajwerner:
Extracted from https://github.com/aya-rs/aya/pull/661. Useful for https://github.com/aya-rs/bpf-linker/pull/72.


tamird (approved):

tamird:
Merging so I can use this in aya-rs/bpf-linker#72.

PR #688, opened 2023-07-25
sys: bpf_prog_get_fd_by_id returns OwnedFd

tamird:


astoycos (approved):
/LGTM 

I added this to #637 so either merge that and close this or vice versa

aya/src/programs/lirc_mode2.rs
```diff
@@ -101,12 +101,12 @@ impl LircMode2 {
...
             })?;
 
-            prog_fds.push(fd as RawFd);
+            prog_fds.push(fd);
         }
 
         Ok(prog_fds
             .into_iter()
-            .map(|prog_fd| LircLink::new(prog_fd, target_fd.as_raw_fd()))
+            .map(|prog_fd| LircLink::new(prog_fd.into_raw_fd(), target_fd.as_raw_fd()))
```
nrxus, line 109: @tamird 

I know this is already merged but the `OwnedFd` is dropped here since there are no other usages of `OwnedFd` and thus the file descriptor is closed and yet we are creating a `LircLink` with this closed file descriptor. `LircLink` has a method `info` that uses this file descriptor to get its program information but that file descriptor is now closed. I don't really use lirc probes so I am not how how to test/prove that but I figured it's worth calling out that this PR most likely broke getting the program information from a queried list of lirc programs.

aya/src/programs/mod.rs
```diff
@@ -955,7 +955,7 @@ impl ProgramInfo {
                 call: "bpf_prog_get_fd_by_id",
                 io_error,
             })?;
-        Ok(fd as RawFd)
+        Ok(fd.into_raw_fd())
```
nrxus, line 958: Here is another place that this PR probably broke things. The `fd` returned here is already closed because the `OwnedFd` is dropped so anyone calling for this `fd` is getting a bad file descriptor. Additionally, the comment of this method no longer makes sense:

```
/// The returned fd must be closed when no longer needed.
```

I think this would be a quick fix though, we just gotta make this function return an `OwnedFd` instead and remove the comment since `OwnedFd`s are closed on drop.

aya/src/programs/lirc_mode2.rs
nrxus: Derp nvm, I thought we were calling for `as_raw_fd`. Ignore

aya/src/programs/mod.rs
nrxus: Derp nvm, I thought we were calling for `as_raw_fd`. Ignore

PR #694, opened 2023-07-27
remove programfd in favor of borrowedfd

astoycos:
Remove the Aya specific Programfd type in favor of the built in BorrowedFd type. The semantics are essentially the same since `ProgramFd` is essentially just a wrapper around `RawFd` which doesn't implement Drop.

  The "Complete" fix for this will be to turn the[ `ProgramData.fd` ](https://github.com/aya-rs/aya/blob/main/aya/src/programs/mod.rs#L410) into an `OwnedFd` however I'm waiting to see what the outcome of https://github.com/aya-rs/aya/pull/662 is/ looks like. 

In the meantime this seems like a fairly isolated change. cc @nrxus 

The only thing I'm not 100% sure on is if the unsafe call to transform a `RawFd` into a `BorrowedFd`.

specifically it states for `BorrowedFd::borrow_raw(fd)`

```
The resource pointed to by fd must remain open for the duration of the returned BorrowedFd, and it must not have the value -1.
```

Which is something we were assuming anyways with `ProgramFd` which is why I figured it'd be alright

### DIFF
--- a/aya/src/maps/array/program_array.rs
+++ b/aya/src/maps/array/program_array.rs
@@ -2,12 +2,11 @@
 
 use std::{
     borrow::{Borrow, BorrowMut},
-    os::fd::{AsRawFd, RawFd},
+    os::fd::{AsRawFd, BorrowedFd, RawFd},
 };
 
 use crate::{
     maps::{check_bounds, check_kv_size, MapData, MapError, MapKeys},
-    programs::ProgramFd,
     sys::{bpf_map_delete_elem, bpf_map_update_elem},
 };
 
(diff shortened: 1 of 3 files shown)


dave-tucker (changes requested):
Nope. We need to figure out a way of giving some type-safety to these APIs.
ProgramFd is used in places where this is a PROGRAM.
With these changes you pass in any BorrowedFd - i.e that of a regular file, which is a footgun that would be nice to avoid.

tamird:
You can just make the inner type *held* in ProgramFd a BorrowedFd.

PR #695, opened 2023-07-27
ci(Mergify): configuration update

dave-tucker:
This change has been made by @dave-tucker from Mergify config editor.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,99 @@
+pull_request_rules:
+  - name: automatic merge for Dependabot pull request that pass CI
+    conditions:
+      - author=dependabot[bot]
+    actions:
+      comment:
+        message: "@dependabot merge"
+
+  # REVIEW MANAGEMENT
+
+  - name: ask alessandrod to review public API changes
+    conditions:
+      - files=xtask/public-api/aya.txt
+      - -closed
+    actions:
+      comment:
+        message: "Hey @alessandrod, this pull request changes the Aya Public API and requires your review."
+      request_reviews:
+        users:
+          - alessandrod
+      label:
+        add:
+          - api/needs-review
+
+  # LABEL MANAGEMENT
+
+  - name: warn on conflicts
+    conditions:
+      - conflict
+    actions:
+      comment:
+        message: "@{{author}}, this pull request is now in conflict and requires a rebase."
+      label:
+        add:
+          - needs-rebase
+
+  - name: remove conflict label if not needed
+    conditions:
+      - -conflict
+    actions:
+      label:
+        remove:
+          - needs-rebase
+
+  - name: add labels for aya PRs
+    conditions:
+      - files~=^aya/
+    actions:
+      label:
+        add:
+          - aya
+
+  - name: add labels for aya-bpf PRs
+    conditions:
+      - files~=^(aya-bpf-macros|bpf/aya-bpf(-(cty|bindings))?)/
+    actions:
+      label:
+        add:
+          - aya-bpf
+
+  - name: add labels for aya-log PRs
+    conditions:
+      - files~=^aya-log(-common)?/
+    actions:
+      label:
+        add:
+          - aya-log
+
+  - name: add labels for aya-log-ebpf PRs
+    conditions:
+      - files~=^(aya-log-common|bpf/aya-log-ebpf)/
+    actions:
+      label:
+        add:
+          - aya-log
+
+  - name: add labels for aya-obj PRs
+    conditions:
+      - files~=^aya-obj/
+    actions:
+      label:
+        add:
+          - aya-obj
+
+  - name: add labels for aya-tool PRs
+    conditions:
+      - files~=^aya-tool/
+    actions:
+      label:
+        add:
+          - aya-tool
+
+  - name: add labels for CI/Test and Automation PRs
+    conditions:
+      - files~=^(.github|test|xtask)/
+    actions:
+      label:
+        add:
+          - test


dave-tucker:
This does 3 things:

1. Automatically merges Dependabot PRs that pass CI
2. Labels all PRs 🙏 which was previously done manually for the release note generation
3. Assigns Alessandro to review any PR for which `xtask/public-api/aya.txt` has changed

tamird:
> 

Can you put this in commit message please?

.mergify.yml
```diff
@@ -0,0 +1,101 @@
+pull_request_rules:
+  - name: automatic merge for Dependabot pull request that pass CI
+    conditions:
+      - author=dependabot[bot]
+      - check-success=build-workflow-complete
+    actions:
+      merge:
```
tamird, line 7: can we have it leave DP a comment instead?

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  # REVIEW MANAGEMENT
+
+  - name: ask alessandrod to review public API changes
+    conditions:
+      - files=xtask/public-api/aya.txt
+      - -closed
+    actions:
+      comment:
+        message: "Hey @alessandrod, this pull request changes the Aya API and requires your review."
```
tamird, line 18: nit: "changes the public API"

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+      - files~=^aya/
+      - -closed
+    actions:
+      label:
+        add:
+          - aya
+
+  - name: add labels for aya-bpf PRs
+    conditions:
+      - files~=^(bpf/aya-bpf/|bpf/aya-bpf-cty/|bpf/aya-bpf-bindings/|aya-bpf-macros/)
```
tamird, line 57: should this be `(^bpf/|aya-bpf-(bindings|macros)/)`?

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+    actions:
+      label:
+        add:
+          - aya-bpf
+
+  - name: add labels for aya-log PRs
+    conditions:
+      - files~=^(aya-log/|aya-log-common/)
```
tamird, line 65: ditto, can simplify to `^aya-log(-common)?/`

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  - name: add labels for aya-tool PRs
+    conditions:
+      - files~=^aya-tool/
+    actions:
+      label:
+        add:
+          - aya-tool
+
+  - name: add labels for CI/Test PRs
```
tamird, line 95: should this group include xtask and be named appropriately?

tamird (commented):
This is neat!

.mergify.yml
dave-tucker: DP comment?

tamird: I don't know how smart mergify is, but DP is pretty smart. You can omit the check success thing if you leave a comment, and DP will also rebase the PR (and perhaps regenerate it completely in complex cases e.g. Cargo.lock conflicts).

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+      - files~=^aya/
+      - -closed
+    actions:
+      label:
+        add:
+          - aya
+
+  - name: add labels for aya-bpf PRs
+    conditions:
+      - files~=^(bpf/aya-bpf(-cty|bindings|macros)?/)
```
tamird, line 57: this looks wrong? aya-bpf-macros is at the top level, not in bpf/

Please take a look at my original suggestion.

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  # REVIEW MANAGEMENT
+
+  - name: ask alessandrod to review public API changes
+    conditions:
+      - files=xtask/public-api/aya.txt
+      - -closed
+    actions:
+      comment:
+        message: "Hey @alessandrod, this pull request changes the Aya Public API and requires your review."
```
tamird, line 18: my point here was to drop "Aya", but I see now that this is only applied to aya, and not any of the other crates, so maybe it's fine.

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+    conditions:
+      - files~=^aya-tool/
+    actions:
+      label:
+        add:
+          - aya-tool
+
+  - name: add labels for CI/Test and Automation PRs
+    conditions:
+      - files~=^(.github/|test/|xtask/)
```
tamird, line 97: slash can be on the outside

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+    conditions:
+      - files~=^aya-log(-common)?/
+    actions:
+      label:
+        add:
+          - aya-log
+
+  - name: add labels for aya-log-ebpf PRs
+    conditions:
+      - files~=^bpf/aya-log-(ebpf|common)/
```
tamird, line 73: this is wrong; aya-log-common is at the top level

.mergify.yml
```diff
@@ -0,0 +1,101 @@
...
+      - -conflict
+    actions:
+      label:
+        remove:
+          - needs-rebase
+
+  - name: add labels for aya PRs
+    conditions:
+      - files~=^aya/
+      - -closed
```
tamird, line 49: how come only aya gets this filter?

.mergify.yml
dave-tucker: This was your suggestion:

> should this be `(^bpf/|aya-bpf-(bindings|macros)/)`?

That would catch `bpf/aya-log-ebpf` also + aya-bpf-bindings are in the bpf directory, not top-level.
I've now gone with something that should catch `aya-bpf-macros` + only the aya-bpf related dirs in `bpf`.

.mergify.yml
```diff
@@ -0,0 +1,100 @@
+pull_request_rules:
+  - name: automatic merge for Dependabot pull request that pass CI
+    conditions:
+      - author=dependabot[bot]
+      - check-success=build-workflow-complete
```
tamird, line 5: this can be removed; DP won't merge unless all checks pass

.mergify.yml
tamird: Thanks, my suggestion was incorrect.

tamird (approved):

.mergify.yml
dave-tucker: 👍 removed.

dave-tucker:
Will merge this on green (unless someone else gets to it first) ✅

PR #696, opened 2023-07-27
test: add the possibility to run a test inside a network namespace

Tuetuopay:
For tests that do networking operations, this allows to have a clean-state network namespace and interfaces for each test. Mainly, this avoids "device or resource busy" errors when reusing the loopback interface across tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ integration-ebpf = { path = "test/integration-ebpf", default-features = false }
 lazy_static = { version = "1", default-features = false }
 libc = { version = "0.2.105", default-features = false }
 log = { version = "0.4", default-features = false }
+netns-rs = { version = "0.1", default-features = false }
 num_enum = { version = "0.6", default-features = false }
 object = { version = "0.31", default-features = false }
 parking_lot = { version = "0.12.0", default-features = false }
--- a/aya/src/sys/netlink.rs
+++ b/aya/src/sys/netlink.rs
@@ -3,9 +3,9 @@ use thiserror::Error;
 
 use libc::{
     close, getsockname, nlattr, nlmsgerr, nlmsghdr, recv, send, setsockopt, sockaddr_nl, socket,
-    AF_NETLINK, AF_UNSPEC, ETH_P_ALL, IFLA_XDP, NETLINK_EXT_ACK, NETLINK_ROUTE, NLA_ALIGNTO,
-    NLA_F_NESTED, NLA_TYPE_MASK, NLMSG_DONE, NLMSG_ERROR, NLM_F_ACK, NLM_F_CREATE, NLM_F_DUMP,
-    NLM_F_ECHO, NLM_F_EXCL, NLM_F_MULTI, NLM_F_REQUEST, RTM_DELTFILTER, RTM_GETTFILTER,
+    AF_NETLINK, AF_UNSPEC, ETH_P_ALL, IFF_UP, IFLA_XDP, NETLINK_EXT_ACK, NETLINK_ROUTE,
+    NLA_ALIGNTO, NLA_F_NESTED, NLA_TYPE_MASK, NLMSG_DONE, NLMSG_ERROR, NLM_F_ACK, NLM_F_CREATE,
+    NLM_F_DUMP, NLM_F_ECHO, NLM_F_EXCL, NLM_F_MULTI, NLM_F_REQUEST, RTM_DELTFILTER, RTM_GETTFILTER,
     RTM_NEWQDISC, RTM_NEWTFILTER, RTM_SETLINK, SOCK_RAW, SOL_NETLINK,
 };
 
@@ -238,6 +238,32 @@ pub(crate) unsafe fn netlink_find_filter_with_name(
     }
 
     Ok(filter_info)
+}
+
+#[doc(hidden)]
+pub unsafe fn netlink_set_link_up(if_index: i32) -> Result<(), io::Error> {
+    let sock = NetlinkSocket::open()?;
+
+    // Safety: Request is POD so this is safe
+    let mut req = mem::zeroed::<Request>();
+
+    let nlmsg_len = mem::size_of::<nlmsghdr>() + mem::size_of::<ifinfomsg>();
+    req.header = nlmsghdr {
+        nlmsg_len: nlmsg_len as u32,
+        nlmsg_flags: (NLM_F_REQUEST | NLM_F_ACK) as u16,
+        nlmsg_type: RTM_SETLINK,
+        nlmsg_pid: 0,
+        nlmsg_seq: 1,
+    };
+    req.if_info.ifi_family = AF_UNSPEC as u8;
+    req.if_info.ifi_index = if_index;
+    req.if_info.ifi_flags = IFF_UP as u32;
+    req.if_info.ifi_change = IFF_UP as u32;
+
+    sock.send(&bytes_of(&req)[..req.header.nlmsg_len as usize])?;
+    sock.recv()?;
+
+    Ok(())
 }
 
 #[repr(C)]
--- a/test/integration-test/Cargo.toml
+++ b/test/integration-test/Cargo.toml
@@ -12,6 +12,7 @@ aya-log = { workspace = true }
 aya-obj = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
+netns-rs = { workspace = true }
 object = { workspace = true }
 rbpf = { workspace = true }
 tokio = { workspace = true, default-features = false, features = [
--- a/test/integration-test/src/tests/smoke.rs
+++ b/test/integration-test/src/tests/smoke.rs
@@ -4,8 +4,12 @@ use aya::{
     Bpf, BpfLoader,
 };
 
+use crate::utils::NetNsGuard;
+
 #[test]
 fn xdp() {
+    let _netns = NetNsGuard::new();
+
     let kernel_version = KernelVersion::current().unwrap();
     if kernel_version < KernelVersion::new(5, 18, 0) {
         eprintln!("skipping test on kernel {kernel_version:?}, support for BPF_F_XDP_HAS_FRAGS was added in 5.18.0; see https://github.com/torvalds/linux/commit/c2f2cdb");
@@ -20,6 +24,8 @@ fn xdp() {
 
 #[test]
 fn extension() {
+    let _netns = NetNsGuard::new();
+
     let kernel_version = KernelVersion::current().unwrap();
     if kernel_version < KernelVersion::new(5, 9, 0) {
         eprintln!("skipping test on kernel {kernel_version:?}, XDP uses netlink");
--- a/test/integration-test/src/utils.rs
+++ b/test/integration-test/src/utils.rs
@@ -1,0 +1,72 @@
+//! Utilities to run tests
+
+use std::{
+    ffi::CString,
+    io, process,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+use aya::netlink_set_link_up;
+use libc::if_nametoindex;
+use netns_rs::{get_from_current_thread, NetNs};
+
+pub struct NetNsGuard {
+    name: String,
+    old_ns: NetNs,
+    ns: Option<NetNs>,
+}
+
+impl NetNsGuard {
+    pub fn new() -> Self {
+        let old_ns = get_from_current_thread().expect("Failed to get current netns");
+
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let pid = process::id();
+        let name = format!("aya-test-{pid}-{}", COUNTER.fetch_add(1, Ordering::Relaxed));
+
+        // Create and enter netns
+        let ns = NetNs::new(&name).unwrap_or_else(|e| panic!("Failed to create netns {name}: {e}"));
+        let netns = Self {
+            old_ns,
+            ns: Some(ns),
+            name,
+        };
+
+        let ns = netns.ns.as_ref().unwrap();
+        ns.enter()
+            .unwrap_or_else(|e| panic!("Failed to enter network namespace {}: {e}", netns.name));
+        println!("Entered network namespace {}", netns.name);
+
+        // By default, the loopback in a new netns is down. Set it up.
+        let lo = CString::new("lo").unwrap();
+        unsafe {
+            let idx = if_nametoindex(lo.as_ptr());
+            if idx == 0 {
+                panic!(
+                    "Interface `lo` not found in netns {}: {}",
+                    netns.name,
+                    io::Error::last_os_error()
+                );
+            }
+            netlink_set_link_up(idx as i32)
+                .unwrap_or_else(|e| panic!("Failed to set `lo` up in netns {}: {e}", netns.name));
+        }
+
+        netns
+    }
+}
+
+impl Drop for NetNsGuard {
+    fn drop(&mut self) {
+        // Avoid panic in panic
+        if let Err(e) = self.old_ns.enter() {
+            eprintln!("Failed to return to original netns: {e}");
+        }
+        if let Some(ns) = self.ns.take() {
+            if let Err(e) = ns.remove() {
+                eprintln!("Failed to remove netns {}: {e}", self.name);
+            }
+        }
+        println!("Exited network namespace {}", self.name);
+    }
+}
--- a/aya/src/lib.rs
+++ b/aya/src/lib.rs
@@ -58,3 +58,5 @@ pub mod util;
 pub use bpf::*;
 pub use obj::btf::{Btf, BtfError};
 pub use object::Endianness;
+#[doc(hidden)]
+pub use sys::netlink_set_link_up;
--- a/aya/src/sys/mod.rs
+++ b/aya/src/sys/mod.rs
@@ -12,6 +12,8 @@ use libc::{c_int, c_long, pid_t, SYS_bpf, SYS_perf_event_open};
 pub(crate) use bpf::*;
 #[cfg(test)]
 pub(crate) use fake::*;
+#[doc(hidden)]
+pub use netlink::netlink_set_link_up;
 pub(crate) use netlink::*;
 pub(crate) use perf_event::*;
 
--- a/test/integration-test/src/lib.rs
+++ b/test/integration-test/src/lib.rs
@@ -18,3 +18,5 @@ pub const BPF_PROBE_READ: &[u8] =
 
 #[cfg(test)]
 mod tests;
+#[cfg(test)]
+mod utils;


test/integration-test/src/tests/smoke.rs
```diff
@@ -4,8 +4,14 @@ use aya::{
...
 };
 
+use crate::utils::Netns;
+
 #[test]
 fn xdp() {
+    Netns::new().exec(do_xdp)
+}
+
+fn do_xdp() {
```
dave-tucker, line 14:
```suggestion
    Netns::new().exec(|| {
```
You can use a closure to avoid creating a new function.

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,76 @@
+//! Utilities to run tests
+
+use std::{
+    fs::File,
+    io::Write,
```
dave-tucker, line 5:
```suggestion
    io::Write as _,
```

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+    os::fd::AsRawFd,
```
dave-tucker, line 6:
```suggestion
    os::fd::AsRawFd as _,
```

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+        .arg("-e")
+        .stdin(Stdio::piped())
+        .spawn()
+        .unwrap();
+
+    // Send script
+    let mut stdin = bash.stdin.take().unwrap();
+    stdin.write_all(script.as_bytes()).unwrap();
+    stdin.write_all(b"\n").unwrap();
+    drop(stdin);
```
dave-tucker, line 26: Why the explicit drop of stdin?

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl Netns {
+    pub fn new() -> Self {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let name = format!("aya-test-{}", COUNTER.fetch_add(1, Ordering::Relaxed));
+        script(&format!(
+            "
+            ip netns add {name}
+            ip -netns {name} link set lo up
+        "
```
dave-tucker, line 47: Formatting here looks off

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+pub struct Netns {
+    pub name: String,
+    pub fd: File,
+}
+
+impl Netns {
+    pub fn new() -> Self {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
```
dave-tucker, line 41: Would it make sense to move `COUNTER` outside of Netns?
Wondering if we ever run in parallel whether 2 tests could pick netns 0 at the same time.

dave-tucker (changes requested):
Looking good! A couple of nits, but excited to get this merged.

test/integration-test/src/utils.rs
Tuetuopay: Otherwise stdin does not get closed, and the bash we execute does not close terminate due to a missing EOF, deadlocking in the process.

test/integration-test/src/utils.rs
Tuetuopay: Done in 66c6ed7

test/integration-test/src/utils.rs
Tuetuopay: Done in 66c6ed7

test/integration-test/src/utils.rs
Tuetuopay: I don't follow you -- this is a static atomic, so as long as it's the same process, no two test will pick the same number. And since those integration tests are unit tests in the cargo sense, that's what happens.

Moving it outside of netns (out of `new` for that matter) would only make it accessible to foreign code (respecting visibility ofc)

test/integration-test/src/utils.rs
Tuetuopay: sadly rustfmt won't let me do the following. fixed in 199d5ca

```rust
        script(&format!("
            ip netns add {name}
            ip -netns {name} link set lo up
        "));
```

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+            })
+            .join()
+            .unwrap()
+        })
+    }
+}
+
+impl Drop for Netns {
+    fn drop(&mut self) {
+        script(&format!("ip netns del {}", self.name))
```
tamird, line 72: should the signature of script be

```
fn script<T: AsRef<[u8]>, I: IntoIterator<Item=T>>(script: I)
```
?

test/integration-test/src/tests/smoke.rs
Tuetuopay: Done in a956e8f

test/integration-test/src/utils.rs
tamird: we might want to include the pid of the current process in the netns name to avoid concurrent processes from picking the same one.

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+    process::{Command, Stdio},
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+use libc::{setns, CLONE_NEWNET};
+
+fn script(script: &str) {
+    // We pipe stdout/err to capture it and print it ourselves, so that it gets captured by the
+    // test harness
+    let mut bash = Command::new("/bin/bash")
```
tamird, line 16: why do we need to shell out to bash rather than use `Command::new("ip", ...)` multiple times?

test/integration-test/src/utils.rs
Tuetuopay: Unless we want to pass binary data to the script (which I highly doubt), I'd prefer `T: AsRef<str>` instead to ensure _at least_ utf8. As for why a plain string and not `script: impl IntoIterator<Item=T>` where `T: AsRef<str>`, it all comes down to convenience. (see your further comment).

test/integration-test/src/utils.rs
Tuetuopay: Pure convenience, and let's be honest, laziness.

This allows to write setup scripts without repeating the `Command::new` boilerplate code every time. Since this is made to do some setup, it allows to make readable setup scripts. Here only to create the network namespace, but it could be used by a test to create veths, install IPs, etc. And having 6 invocations of `Command::new(..).args(..).output().unwrap();` is just unwieldy.

This comes out of what we do at work (granted for actual applications and not libraries), where having an easy way to write test setup goes a long way for people to actually write useful tests.

Tuetuopay:
> we might want to include the pid of the current process in the netns name to avoid concurrent processes from picking the same one.

great idea. done in 3a44804

test/integration-test/src/utils.rs
tamird: This laziness comes at a cost -- for example your current netns usage that executes two commands (one to create the network ns and another to `up` it) can fail to run the second half, and the Drop impl won't run in that case. I think we want to be more scrupulous here, not less.

Tuetuopay: Indeed, good point. Though I still stand that for tests themselves this helper for running a script is useful.

Properly split with cleanups in f4c3968. It does so by creating the handle asap so that drop gets called if link up fails.

tamird: I buy that we want to make this convenient for callers, but I don't think we should do that by using bash. Do we really want to end up with word splitting issues in our tests because we decided bash was an appropriate language?

Better would be to make this function's signature:
```rust
fn script<T: AsRef<OsStr>, I: IntoIterator<Item=T>>(script: impl IntoIterator<Item = I>)
```
note that this is different from my earlier suggestion in two ways:
- [u8] -> OsStr to match the bounds on various methods on `Command`
- `script` is now an iterator of iterators, so that this can be called as:
```rust
script([
  &["echo", "hello world"], // Look ma, no word splitting!
  &["rm", "-rf", "/"]
])
```

stepping back even further: we really really want to discourage shelling out in our tests because we want to drive toward having virtualized tests that exercise different kernel versions, and we don't want the userland of those VMs to become load bearing. We want to build the integration tests once, and then run them on a bunch of different VMs which ideally just have the kernel and little else.

Could we do this NetNS stuff using [the netns_rs crate](https://docs.rs/netns-rs/latest/netns_rs/) which doesn't depend on any userland at all?

tamird (changes requested):

Tuetuopay:
@tamird I'll remove that script function. Something more ad-hoc will be added in the future when needed (even in my xdp maps pr I don't need it).

Good idea for the netns crate. Especially since it handles the rollback to the original netns, it'll remove the thread we spawn.

As for userspace, I think that `iproute2` is ubiquitous enough that it's not a big constraint to require it to run tests. Especially since we use it to up a link, that even `busybox`'s implementation of the `ip` command can handle. Furthermore, I really don't want to bring the complexity of manual netlink in the codebase for something as simple as upping `lo`. I'll have a look at the raw netlink code aya already has, and see if it can be added without too much code.

test/integration-test/src/utils.rs
Tuetuopay: Resolving conversation as this code has been dropped.

test/integration-test/src/utils.rs
Tuetuopay: Resolving as the script function has been dropped.

test/integration-test/src/utils.rs
Tuetuopay: As requested:

- dropped the script function (well, not requested per se, but strongly implied)
- dropped the usage of iproute2, both for netns creation and for link up

Cargo.toml
```diff
@@ -71,6 +71,7 @@ integration-ebpf = { path = "test/integration-ebpf", default-features = false }
 lazy_static = { version = "1", default-features = false }
 libc = { version = "0.2.105", default-features = false }
 log = { version = "0.4", default-features = false }
+netns-rs = "0.1"
```
tamird, line 74: could you do the usual { version = .., default-features = false } please?

test/integration-test/Cargo.toml
```diff
@@ -12,6 +12,7 @@ aya-log = { workspace = true }
 aya-obj = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
+netns-rs.workspace = true
```
tamird, line 15: please use the prevailing style

aya/src/sys/netlink.rs
```diff
@@ -240,6 +240,32 @@ pub(crate) unsafe fn netlink_find_filter_with_name(
...
+        nlmsg_pid: 0,
+        nlmsg_seq: 1,
+    };
+    req.if_info.ifi_family = AF_UNSPEC as u8;
+    req.if_info.ifi_index = if_index;
+    req.if_info.ifi_flags = IFF_UP as u32;
+    req.if_info.ifi_change = IFF_UP as u32;
+
+    sock.send(&bytes_of(&req)[..req.header.nlmsg_len as usize])?;
+    sock.recv()?;
```
tamird, line 264: do we need to check what we got back?

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+    os::fd::RawFd,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+use aya::netlink_set_link_up;
+use libc::if_nametoindex;
+use netns_rs::{get_from_current_thread, NetNs};
+
+pub struct Netns {
+    pub name: String,
```
tamird, line 14: do we need everything to be pub? seems like maybe none of it needs to be?

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+    ffi::CString,
+    os::fd::RawFd,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+use aya::netlink_set_link_up;
+use libc::if_nametoindex;
+use netns_rs::{get_from_current_thread, NetNs};
+
+pub struct Netns {
```
tamird, line 13: let's call this `NetNsGuard` to signal that it has important drop semantics?

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Netns {
+    pub fn new() -> Self {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let pid = std::process::id();
+        let name = format!("aya-test-{pid}-{}", COUNTER.fetch_add(1, Ordering::Relaxed));
+
+        // Create and enter netns
+        let ns = NetNs::new(&name).unwrap_or_else(|e| panic!("Failed to create netns {name}: {e}"));
+        ns.enter().expect("Failed to enter network namespace");
```
tamird, line 27: nit: include the name in all the messages?

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+        let netns = Self {
+            old_ns: get_from_current_thread().expect("Failed to get current netns"),
```
tamird, line 29: if this fails we leak the created namespace; consider doing this before creating the new one

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+            ns: Some(ns),
+            name,
+        };
+
+        // By default, the loopback in a new netns is down. Set it up.
+        let lo = CString::new("lo").unwrap();
+        unsafe {
+            let idx = if_nametoindex(lo.as_ptr());
+            if idx == 0 {
+                panic!("Interface `lo` not found in netns {}", netns.name);
```
tamird, line 39: > On success, if_nametoindex() returns the index number of the
       network interface; on error, 0 is returned and [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is set to
       indicate the error.

please emit the value of [`std::io::Error::last_os_error`](https://doc.rust-lang.org/stable/std/io/struct.Error.html#method.last_os_error)

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+            }
+            netlink_set_link_up(idx as RawFd)
```
tamird, line 41: RawFd is wrong here

Tuetuopay: Why so? I copied this `as RawFd` from other uses of `if_nametoindex` in aya: see [`Xdp::attach`](https://github.com/aya-rs/aya/blob/main/aya/src/programs/xdp.rs#L99), that calls [`Xdp::attach_to_ifindex`](https://github.com/aya-rs/aya/blob/main/aya/src/programs/xdp.rs#L120) (uses `as RawFd`) which then calls [`aya::sys::netlink_set_xdp_fd`](https://github.com/aya-rs/aya/blob/main/aya/src/sys/netlink.rs#L26).

So if `RawFd` is wrong here, it's also wrong in the place I mentioned.

Cargo.toml
Tuetuopay: Sure. Done in 8ec9f73

Tuetuopay: Sure. Done in 8ec9f73

test/integration-test/Cargo.toml
Tuetuopay: Sure. Done in 8ec9f73

test/integration-test/src/utils.rs
Tuetuopay: Done in 8ec9f73

test/integration-test/src/utils.rs
Tuetuopay: Done in 8ec9f73

test/integration-test/src/utils.rs
Tuetuopay: Yup, this one slipped through. Done in 8ec9f73

test/integration-test/src/utils.rs
Tuetuopay: Sure. Done in 8ec9f73

test/integration-test/src/utils.rs
Tuetuopay: Indeed it does not. Habit of logging everything. Done in 8ec9f73

aya/src/sys/netlink.rs
Tuetuopay: AFAIK no. For an up request, we get back a simple ACK message with an error message, containing an error value of zero if successful, all of which is already checked in `NetlinkSocket::recv`.

test/integration-test/src/utils.rs
```diff
@@ -10,23 +11,26 @@ use aya::netlink_set_link_up;
...
+    ns: Option<NetNs>,
 }
 
-impl Netns {
+impl NetNsGuard {
     pub fn new() -> Self {
         static COUNTER: AtomicU64 = AtomicU64::new(0);
         let pid = std::process::id();
         let name = format!("aya-test-{pid}-{}", COUNTER.fetch_add(1, Ordering::Relaxed));
+        let old_ns = get_from_current_thread().expect("Failed to get current netns");
```
tamird, line 25: sorry to harp on this more: make this the first thing in the function?the preceding two lines are infallible.

test/integration-test/src/utils.rs
```diff
@@ -10,23 +11,26 @@ use aya::netlink_set_link_up;
...
+    name: String,
+    old_ns: NetNs,
+    ns: Option<NetNs>,
 }
 
-impl Netns {
+impl NetNsGuard {
     pub fn new() -> Self {
         static COUNTER: AtomicU64 = AtomicU64::new(0);
         let pid = std::process::id();
```
tamird, line 23: `use std::process` for consistency

test/integration-test/src/utils.rs
tamird: I don't know about those other cites, but I think we can agree that an interface index is not a file descriptor?

tamird (approved):

test/integration-test/src/utils.rs
Tuetuopay: > I don't know about those other cites, but I think we can agree that an interface index is not a file descriptor?

Huh the sun hit hard, I didn't even realize this. Indeed that other use is wrong. Changing to `as i32`.

test/integration-test/src/utils.rs
Tuetuopay: Done

Tuetuopay:
@tamird last comments taken into account. I squashed all commits in a single one (it ends up pretty short, to not pollute the log with useless commits) and rebased on `main`.

Thanks for the swift reviews!

test/integration-test/src/utils.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn new() -> Self {
+        let old_ns = get_from_current_thread().expect("Failed to get current netns");
+
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let pid = process::id();
+        let name = format!("aya-test-{pid}-{}", COUNTER.fetch_add(1, Ordering::Relaxed));
+
+        // Create and enter netns
+        let ns = NetNs::new(&name).unwrap_or_else(|e| panic!("Failed to create netns {name}: {e}"));
+        ns.enter()
```
tamird, line 29: We need to create the NetNsGuard before this to ensure cleanup.

Tuetuopay: Indeed. Fixed.

tamird:
@alessandrod PTAL since this technically affects public API.

tamird:
Per discord discussion we don't need API review here.

PR #698, opened 2023-07-27
bpf: Remove builtin global functions

tamird:
This commit removes memset and memcpy, relying instead on
implementations provided by std/compiler-builtins.

This commit adds `#![no_builtins]` to all the BPF programs written in
Rust, and the same should be propagated to aya-template and all examples
in the book and elsewhere before this commit is merged.

It turns out that without the `#![no_builtins]` annotation rustc
generates LLVM IR that calls LLVM intrinsics rather than libcalls. These
may end up as libcalls after lowering, but not before emitting errors in
BPF lowering[0].

This works thanks to https://github.com/rust-lang/rust/pull/113716 which
causes `#![no_builtins]` to behave similarly to `-fno-builtin` in clang,
which was added in https://reviews.llvm.org/D68028 with similar
motivation.

This commit implies that we now require rustc nightly >= 2023-07-20.

[0] https://github.com/llvm/llvm-project/blob/7b2745b/llvm/lib/Target/BPF/BPFISelLowering.cpp#L472-L474

### DIFF
--- a/bpf/aya-bpf/src/lib.rs
+++ b/bpf/aya-bpf/src/lib.rs
@@ -31,7 +31,7 @@ use core::ffi::c_void;
 
 pub use aya_bpf_cty as cty;
 pub use aya_bpf_macros as macros;
-use cty::{c_int, c_long};
+use cty::c_long;
 use helpers::{bpf_get_current_comm, bpf_get_current_pid_tgid, bpf_get_current_uid_gid};
 
 pub const TASK_COMM_LEN: usize = 16;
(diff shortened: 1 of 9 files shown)


alessandrod:
The compiler-builtins fallbacks don't pass the verifier, which is why we provide our own

tamird:
I don't  think that's correct; if it were, the integration tests would fail here.

alessandrod:
We've discussed this on discord, the tests are passing because of inlining. Can we close?

tamird:
I'd still like to write better tests, marking as draft if that's ok.

PR #699, opened 2023-07-28
Misc testing improvements

tamird:
- integration-test: rerun-if-changed=integration-ebpf
- Do not escape newlines on Err(LoadError).unwrap()


ajwerner (approved):
LGTM

PR #701, opened 2023-07-31
aya: Return `OwnedFd` for `perf_event_open`.

nrxus:
This fixes a file descriptor leak when creating a link of BPF_PERF_EVENT attach type.

### DIFF
--- a/aya/src/maps/perf/perf_buffer.rs
+++ b/aya/src/maps/perf/perf_buffer.rs
@@ -1,13 +1,13 @@
 use std::{
     ffi::c_void,
     io, mem,
-    os::fd::{AsRawFd, RawFd},
+    os::fd::{AsFd, AsRawFd, BorrowedFd, OwnedFd, RawFd},
     ptr, slice,
     sync::atomic::{self, AtomicPtr, Ordering},
 };
 
 use bytes::BytesMut;
-use libc::{c_int, close, munmap, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE};
+use libc::{c_int, munmap, MAP_FAILED, MAP_SHARED, PROT_READ, PROT_WRITE};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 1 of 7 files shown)


aya/src/programs/perf_attach.rs
```diff
@@ -55,12 +54,12 @@ impl Link for PerfLink {
...
     fn id(&self) -> Self::Id {
-        PerfLinkId(self.perf_fd)
+        PerfLinkId(self.perf_fd.as_raw_fd())
     }
 
     fn detach(mut self) -> Result<(), ProgramError> {
-        let _ = perf_event_ioctl(self.perf_fd, PERF_EVENT_IOC_DISABLE, 0);
-        unsafe { close(self.perf_fd) };
+        let _ = perf_event_ioctl(self.perf_fd.as_fd(), PERF_EVENT_IOC_DISABLE, 0);
+        std::mem::drop(self.perf_fd);
```
nrxus, line 62: I am not sure if this is necessary, or if we should just let it drop on it's own when the scope ends. I just added the manual `drop` to match the exact same behavior as before where we closed the file descriptor immediately after the `PERF_EVENT_IOC_DISABLE`

aya/src/programs/perf_attach.rs
```diff
@@ -72,10 +71,10 @@ impl Link for PerfLink {
     }
 }
 
-pub(crate) fn perf_attach(prog_fd: RawFd, fd: RawFd) -> Result<PerfLinkInner, ProgramError> {
+pub(crate) fn perf_attach(prog_fd: RawFd, fd: OwnedFd) -> Result<PerfLinkInner, ProgramError> {
     if FEATURES.bpf_perf_link() {
         let link_fd =
-            bpf_link_create(prog_fd, fd, BPF_PERF_EVENT, None, 0).map_err(|(_, io_error)| {
+            bpf_link_create(prog_fd, fd.as_raw_fd(), BPF_PERF_EVENT, None, 0).map_err(|(_, io_error)| {
```
nrxus, line 77: Prior to this PR this would leak the file descriptor as we would drop it at the end of this scope without closing it and no one else had a handle to it. Now at the end of this scope the `drop` will close the file descriptor for us.

tamird (approved):

aya/src/maps/perf/perf_buffer.rs
```diff
@@ -1,13 +1,13 @@
 use std::{
     ffi::c_void,
     io, mem,
-    os::fd::{AsRawFd, RawFd},
+    os::fd::{AsFd, AsRawFd, BorrowedFd, OwnedFd, RawFd},
```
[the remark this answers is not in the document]
rc-andres: Both `AsFd` and `AsRawFd` are explicitly needed in this file because we implement `AsFd` and `AsRawFd` manually for `PerfBuffer`.

tamird (approved):

PR #702, opened 2023-07-31
aya: Don't store btf_fd in MapData

dave-tucker:
This is only used in create and therefore can be passed
as a parameter.

### DIFF
--- a/aya/src/maps/bloom_filter.rs
+++ b/aya/src/maps/bloom_filter.rs
@@ -116,7 +116,6 @@ mod tests {
             obj: new_obj_map(),
             fd: None,
             pinned: false,
-            btf_fd: None,
         };
         assert_matches!(
             BloomFilter::<_, u16>::new(&map),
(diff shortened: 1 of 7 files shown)


aya/src/maps/bloom_filter.rs
```diff
@@ -118,7 +118,7 @@ mod tests {
             obj: new_obj_map(),
             fd: None,
             pinned: false,
-            btf_fd: None,
+            //
```
tamird, line 121: leftover?

aya/src/maps/lpm_trie.rs
```diff
@@ -282,7 +280,7 @@ mod tests {
                 data: Vec::new(),
             }),
             fd: None,
-            btf_fd: None,
+
```
tamird, line 283: spurious?

tamird (approved):
LGTM. Would be preferable to squash these two commits, imo.

dave-tucker:
Related to the api changes: #703

alessandrod (approved):

PR #706, opened 2023-07-31
integration-test: Remove runtime toolchain deps

tamird:
Move the use of clang and llvm-objcopy from run-time to build-time. This
allows the integration tests to run on VMs with simpler userlands.

Create a new CI job to build the integration tests separately from
running them. Ship them from that job to the runner job using github
actions artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
 
       - name: Run miri
         run: |
+          set -euxo pipefail
           cargo hack miri test --all-targets --feature-powerset \
             --exclude aya-bpf \
             --exclude aya-bpf-bindings \
@@ -79,6 +80,7 @@ jobs:
 
       - name: Build
         run: |
+          set -euxo pipefail
           cargo hack build --all-targets --feature-powerset \
             --exclude aya-bpf \
             --exclude aya-bpf-bindings \
@@ -90,6 +92,7 @@ jobs:
         env:
           RUST_BACKTRACE: full
         run: |
+          set -euxo pipefail
           cargo hack test --all-targets --feature-powerset \
             --exclude aya-bpf \
             --exclude aya-bpf-bindings \
@@ -122,7 +125,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: Prereqs
+      - name: bpf-linker
         run: cargo install bpf-linker --git https://github.com/aya-rs/bpf-linker.git
 
       - uses: taiki-e/install-action@cargo-hack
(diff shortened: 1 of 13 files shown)


tamird:
This happens to reduce CI time as well because of better caching; previously the integration tests weren't enjoying the benefits of [`Swatinem/rust-cache`](https://github.com/swatinem/rust-cache). On a hydrated cache this takes total CI from 14m22s to 8m31s.

ajwerner (approved):
Works for me. Let's see if others have opinions on the testing layout.

tamird:
Turns out the tests weren't running (because github artifacts strip the execute bit). They're running now, adding about 90 seconds.

ajwerner:
> They're running now, adding about 90 seconds.

Could parallelize the building and running in the two profiles to regain some speed I reckon.

.github/workflows/ci.yml
```diff
@@ -43,6 +43,7 @@ jobs:
 
       - name: Run miri
         run: |
+          set -euxo pipefail
```
dave-tucker, line 46: I'm assuming this was for debugging?

.github/workflows/ci.yml
```diff
@@ -122,34 +125,73 @@ jobs:
...
+          toolchain: nightly
+          components: rust-src
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: bpf-linker
+        run: cargo install bpf-linker --git https://github.com/aya-rs/bpf-linker.git
+
+      - name: Install dependencies
+        # ubuntu-20.04 comes with clang prior to
```
dave-tucker, line 160: We're on 22.04 now, does this comment still hold?

dave-tucker (commented):
Some nits/questions but overall LGTM

.github/workflows/ci.yml
tamird: Not even, I just added it as a proactive measure. Otherwise you can't see in the logs when each line is run.

.github/workflows/ci.yml
tamird: Let's find out! I copied this from a similar stanza in bpf-linker where we're on 20.04. Removed installing clang, let's see what happens.

.github/workflows/ci.yml
```diff
@@ -122,34 +125,73 @@ jobs:
...
+          cargo xtask build-integration-test --cargo-arg=--profile=release | xargs -I % cp % integration-test-binaries/release
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: integration-test-binaries
+          path: integration-test-binaries
+
+  run-integration-test:
+    runs-on: macos-latest
+    needs: build-integration-test
```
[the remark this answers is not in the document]
tamird: :vulcan_salute:

test/integration-test/bpf/reloc.bpf.c
```diff
@@ -0,0 +1,106 @@
+// clang-format off
```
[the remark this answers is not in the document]
tamird: There isn't, but I was locally formatting and getting annoyed that it would break compilation each time I did so.

test/integration-test/bpf/reloc.bpf.c
```diff
@@ -0,0 +1,106 @@
...
+  __u8 b;
+  __u8 c;
+};
+
+__attribute__((noinline)) int field_global() {
+  struct relocated_struct_with_scalars s = {1, 2, 3};
+  return set_output(__builtin_preserve_access_index(s.b));
+}
+
+SEC("uprobe/field") int field(void *ctx) {
```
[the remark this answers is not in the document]
tamird: Yeah, it doesn't work, but I'm not sure the reason is that aya is doing anything wrong. If I use `SEC("uprobe")` for all of these then the verifier rejects the programs:
```
thread 'tests::btf_relocations::relocation_tests::_struct_flavors_true_none_1_expects' panicked at 'called `Result::unwrap()` on an `Err` value: LoadError { io_error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, verifier_log: number of funcs in func_info doesn't match number of subprogs
verification time 503 usec
stack depth 0+0
 }', test/integration-test/src/tests/btf_relocations.rs:41:20
```

If I only change *one* of them to `SEC("uprobe")` then it all works, so my guess is that the name doesn't matter, we're just not allowed to have collisions.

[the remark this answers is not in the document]
dave-tucker: Ok, happy for this to stay as it is then and I'll investigate this when I start looking at fixing up ProgramSection etc... to be complaint with libbpfs strict section naming rules.

.github/workflows/ci.yml
tamird: Seems to work.

tamird: Just kidding, we do need a newer clang still. Added commentary.

Issue #610, opened 2023-05-23
tc load report 'no space left'


Sherlock-Holo:
when this error message reported, what is the real reason?
I can't attach this https://github.com/Sherlock-Holo/mahiro/commit/fe85104164ade7fd7fec944cfb2eae2f7efbb73e#diff-b47e6e512f3e362859296f1fca2c484e0699ca24db91fcf4813a04af127701a4R20 bpf program, however, I can attach https://github.com/Sherlock-Holo/mahiro/commit/fe85104164ade7fd7fec944cfb2eae2f7efbb73e#diff-b47e6e512f3e362859296f1fca2c484e0699ca24db91fcf4813a04af127701a4R16

Sherlock-Holo:
oh... I change `#[classifier(name = "dnat_ingress_with_redirect_route")]` to `#[classifier(name = "dnat_ingress_with_redirect")]` and it works
that makes me more confuse...

dave-tucker:
I would guess the name is too long. Would be interesting where the ENOSPC comes from though.
Either running with `strace` or Aya's debug logging enabled would be helpful.

Sherlock-Holo:
it seems the error reported by this https://github.com/aya-rs/aya/blob/58f1ecbf0089194d729327692adca6391fc24932/aya/src/sys/netlink.rs#L491

dave-tucker:
Ah yeah here's the issue. Name should probably be truncated here:
https://github.com/aya-rs/aya/blob/58f1ecbf0089194d729327692adca6391fc24932/aya/src/sys/netlink.rs#L134

pooladkhay:
Hi, I'd like to work on this issue.
As I understand the length of the name should not exceed 32 bytes (31 chars + null byte).
Is it ok to just truncate anything beyond this size ? also maybe show a warning ?

pooladkhay:
According to the kernel code, the name can be up to 256 bytes long:
https://github.com/torvalds/linux/blob/master/net/sched/cls_bpf.c#L28
```c
#define CLS_BPF_NAME_LEN	256
```
And realised you have set the total length of attributes to 64 bytes:
https://github.com/aya-rs/aya/blob/main/aya/src/sys/netlink.rs#L254
```rust
#[repr(C)]
struct TcRequest {
    header: nlmsghdr,
    tc_info: tcmsg,
    attrs: [u8; 64],
}
```

[...]

PR #708, opened 2023-07-31
aya: fix tc name limit

pooladkhay:
By increasing the `attrs` length to a number that can accommodate 256-bytes-long tc names enforced by the kernel, aya no longer limits this attribute.

292 is the smallest possible size which is divisible by 4 (alignment requirement).

Fixes: #610

### DIFF
--- a/aya-obj/src/btf/types.rs
+++ b/aya-obj/src/btf/types.rs
@@ -76,7 +76,7 @@ impl Const {
 
     pub(crate) fn new(btf_type: u32) -> Self {
         let info = (BtfKind::Const as u32) << 24;
-        Const {
+        Self {
             name_offset: 0,
             info,
             btf_type,
@@ -150,7 +150,7 @@ impl Ptr {
 
     pub fn new(name_offset: u32, btf_type: u32) -> Self {
         let info = (BtfKind::Ptr as u32) << 24;
-        Ptr {
+        Self {
             name_offset,
             info,
             btf_type,
@@ -181,7 +181,7 @@ impl Typedef {
 
     pub(crate) fn new(name_offset: u32, btf_type: u32) -> Self {
         let info = (BtfKind::Typedef as u32) << 24;
-        Typedef {
+        Self {
             name_offset,
             info,
             btf_type,
@@ -211,7 +211,7 @@ impl Float {
 
     pub fn new(name_offset: u32, size: u32) -> Self {
         let info = (BtfKind::Float as u32) << 24;
-        Float {
+        Self {
             name_offset,
             info,
             size,
@@ -262,7 +262,7 @@ impl Func {
     pub fn new(name_offset: u32, proto: u32, linkage: FuncLinkage) -> Self {
         let mut info = (BtfKind::Func as u32) << 24;
         info |= (linkage as u32) & 0xFFFF;
-        Func {
+        Self {
             name_offset,
             info,
             btf_type: proto,
@@ -301,7 +301,7 @@ impl TypeTag {
 
     pub fn new(name_offset: u32, btf_type: u32) -> Self {
         let info = (BtfKind::TypeTag as u32) << 24;
-        TypeTag {
+        Self {
             name_offset,
             info,
             btf_type,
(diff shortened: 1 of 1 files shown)


aya/src/sys/netlink.rs
```diff
@@ -277,7 +277,7 @@ struct Request {
 struct TcRequest {
     header: nlmsghdr,
     tc_info: tcmsg,
-    attrs: [u8; 64],
+    attrs: [u8; 292],
```
tamird, line 280: Please add a comment explaining this value.

tamird (changes requested):
Please rebase out the merge commit.

aya/src/sys/netlink.rs
```diff
@@ -277,7 +277,7 @@ struct Request {
 struct TcRequest {
     header: nlmsghdr,
     tc_info: tcmsg,
-    attrs: [u8; 64],
+    attrs: [u8; 292], // 292 is the smallest size that can accommodate 256-bytes-long tc names allowed by the kernel and is divisible by 4 (alignment requirement).
```
tamird, line 280: please:
- put the comment above the field, and wrap at 80 columns
- provide a citation for the claim that the kernel allows 256 bytes, preferably by linking to kernel source code. Please use 7 characters for the kernel SHA e.g. https://github.com/torvalds/linux/blob/343e537/kernel/bpf/verifier.c#L9
- provide a better explanation for the value selected; 256 is divisible by 4, so this doesn't make sense to me.

tamird (changes requested):

aya/src/sys/netlink.rs
pooladkhay: I described it based on what was already discussed in the issue.
Valid points, I'll update the comment accordingly.

tamird: Yes, github issues - particular closed ones - are not an acceptable repository of knowledge.

dave-tucker: This comment is incorrect.
Netlink messages can contain any number of attributes.
`attrs: [u8; 64]` is being used as a fixed-size buffer (see: `request_attributes()`)  that these attributes get written into.

Side note: This code could do with some love.

To move this forwards I would suggest you bump this to `[u8; 512]`, add a comment explaining that the buffer should be sized to hold at least `CLS_BPF_NAME_LEN = 256` bytes, and finally add an an integration test for a TC program with a long name.

pooladkhay: @dave-tucker  `request_attributes()` is being called from three places (i.e. `netlink_qdisc_add_clsact`, `netlink_qdisc_attach` and `netlink_find_filter_with_name`) where `req` is of type `TcRequest`.

There is only one place (inside `netlink_find_filter_with_name`) that an attribute of arbitrary size is being added to `attrs` buffer: `options.write_attr_bytes(TCA_BPF_NAME as u16, prog_name.to_bytes_with_nul())?;` which is the tc name.

Since all other attributes have a fixed sizes, the required size would be `all other attributes + size of name` and we know the maximum size of name allowed by the kernel is 256 bytes.
Based on what I observed `33 bytes` is what all other attributes require so 289 is the total required space for the `attrs` buffer. I added 3 more bytes for alignment requirement (which is not really necessary because `request_attributes` does that anyway).

With all that, Can you please explain why `[u8; 512]` ?

dave-tucker: > There is only one place (inside netlink_find_filter_with_name) that an attribute of arbitrary size is being added to attrs buffer: options.write_attr_bytes(TCA_BPF_NAME as u16, prog_name.to_bytes_with_nul())?; which is the tc name.

> Since all other attributes have a fixed sizes, the required size would be all other attributes + size of name

IIRC The netlink protocol allows for varying sized attributes. We've just been lucky that the ones we had fit into the size of buffer - IMO this was probably luck, not judgement - and you're unlucky to have hit the issue with the TC name being longer than the buffer we allocated.

The issue with sizing to EXACTLY 289 bytes is that this is going to break again the moment we add another attribute here alongside the name.

`512` isn't outrageously large, but at least gives us some headroom in the interim.

The right fix, which is significantly more work, is to not use a fized size buffer int the first place.

pooladkhay: I get that now, Thank you for the explanation.
I'll send the patch with a test soon.

PR #709, opened 2023-08-01
aya: Use OwnedFd in FdLink.

nrxus:
I had to add a few `into_raw_fd()` to avoid ballooning up this PR with changes to `MapData` or to `ProgramData` (because the syscall to `BPF_OBJ_GET` now returns an `OwnedFd`) Those are more involved so they should each be their own PR. 

As far as I could tell there weren't any leaked file descriptors being fixed by this, this is just future proofing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/709)

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -42,7 +42,7 @@ use std::{
     marker::PhantomData,
     mem,
     ops::Deref,
-    os::fd::{AsFd, AsRawFd, OwnedFd, RawFd},
+    os::fd::{AsFd as _, AsRawFd, IntoRawFd as _, OwnedFd, RawFd},
     path::Path,
     ptr,
     sync::Arc,
@@ -516,11 +516,9 @@ impl MapData {
                 code,
                 io_error,
             }
-        })? as RawFd;
+        })?;
 
-        self.fd = Some(fd);
-
-        Ok(fd)
+        Ok(*self.fd.insert(fd as RawFd))
     }
 
     pub(crate) fn open_pinned<P: AsRef<Path>>(
@@ -536,11 +534,9 @@ impl MapData {
         let fd = bpf_get_object(&path_string).map_err(|(_, io_error)| SyscallError {
             call: "BPF_OBJ_GET",
             io_error,
-        })? as RawFd;
+        })?;
 
-        self.fd = Some(fd);
-
-        Ok(fd)
+        Ok(*self.fd.insert(fd.into_raw_fd()))
     }
 
     /// Loads a map from a pinned path in bpffs.
@@ -558,29 +554,29 @@ impl MapData {
         let fd = bpf_get_object(&path_string).map_err(|(_, io_error)| SyscallError {
             call: "BPF_OBJ_GET",
             io_error,
-        })? as RawFd;
+        })?;
 
-        let info = bpf_map_get_info_by_fd(fd)?;
+        let info = bpf_map_get_info_by_fd(fd.as_fd())?;
 
         Ok(MapData {
             obj: parse_map_info(info, PinningType::ByName),
-            fd: Some(fd),
+            fd: Some(fd.into_raw_fd()),
             btf_fd: None,
             pinned: true,
         })
     }
 
-    /// Loads a map from a [`RawFd`].
+    /// Loads a map from a file descriptor.
     ///
     /// If loading from a BPF Filesystem (bpffs) you should use [`Map::from_pin`](crate::maps::MapData::from_pin).
     /// This API is intended for cases where you have received a valid BPF FD from some other means.
     /// For example, you received an FD over Unix Domain Socket.
-    pub fn from_fd(fd: RawFd) -> Result<MapData, MapError> {
-        let info = bpf_map_get_info_by_fd(fd)?;
+    pub fn from_fd(fd: OwnedFd) -> Result<MapData, MapError> {
+        let info = bpf_map_get_info_by_fd(fd.as_fd())?;
 
         Ok(MapData {
             obj: parse_map_info(info, PinningType::None),
-            fd: Some(fd),
+            fd: Some(fd.into_raw_fd()),
             btf_fd: None,
             pinned: false,
         })
--- a/aya/src/programs/cgroup_device.rs
+++ b/aya/src/programs/cgroup_device.rs
@@ -1,7 +1,7 @@
 //! Cgroup device programs.
 
 use crate::util::KernelVersion;
-use std::os::fd::{AsRawFd, RawFd};
+use std::os::fd::AsRawFd;
 
 use crate::{
     generated::{bpf_attach_type::BPF_CGROUP_DEVICE, bpf_prog_type::BPF_PROG_TYPE_CGROUP_DEVICE},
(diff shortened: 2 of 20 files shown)


tamird (approved):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/709)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod and @nrxus)

*[a discussion](https://reviewable.io/reviews/aya-rs/aya/709#gh-1280304891:-NakPKj-2ct-1sKvJFzl:b-rgighg) (no related file):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

technically a breaking API change but the old API was unsafe as it relied on the user knowing not to close the `RawFd` after the fact. This makes it more explicit that this function will take ownership of the file descriptor. 
</blockquote></details>

Looks like a win to me.

___
___
*[`aya/src/maps/mod.rs` line 584 at r1](https://reviewable.io/reviews/aya-rs/aya/709#-NakPBFKED9p8gcRr59G:-NakPBFKED9p8gcRr59H:bil20ng) ([raw file](https://github.com/aya-rs/aya/blob/0c9b6222cc477c311bc43881acf69b22e3eb0617/aya/src/maps/mod.rs#L584)):*
> ```XML
>     }
> 
>     /// Loads a map from an [`OwnedFd`].
> ```

I think it'd be better avoid repeating the type here.



_Code quote:_
```XML
OwnedFd
```

nrxus (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/709)* status: 19 of 20 files reviewed, 1 unresolved discussion (waiting on @alessandrod and @tamird)

___
*[`aya/src/maps/mod.rs` line 584 at r1](https://reviewable.io/reviews/aya-rs/aya/709#-NakPBFKED9p8gcRr59G:-NakRYAuDRJ9lq2qgdRd:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/0c9b6222cc477c311bc43881acf69b22e3eb0617/aya/src/maps/mod.rs#L584)):*

I think it'd be better avoid repeating the type here.
</blockquote></details>

Done.

tamird (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/709)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

aya/src/maps/mod.rs
```diff
@@ -563,35 +565,37 @@ impl MapData {
...
             obj: parse_map_info(info, PinningType::ByName),
-            fd: Some(fd),
+            fd: Some(fd.into_raw_fd()),
             btf_fd: None,
             pinned: true,
         })
     }
 
-    /// Loads a map from a [`RawFd`].
+    /// Loads a map from a file descriptor.
```
dave-tucker, line 584:
```suggestion
    /// Loads a map from an `[OwnedFd]`
```

aya/src/programs/cgroup_device.rs
```diff
@@ -1,7 +1,7 @@
 //! Cgroup device programs.
 
 use crate::util::KernelVersion;
-use std::os::fd::{AsRawFd, RawFd};
+use std::os::fd::AsRawFd;
```
dave-tucker, line 4:
```suggestion
use std::os::fd::AsRawFd as _;
```

aya/src/programs/cgroup_skb.rs
```diff
@@ -1,11 +1,7 @@
 //! Cgroup skb programs.
 
 use crate::util::KernelVersion;
-use std::{
-    hash::Hash,
-    os::fd::{AsRawFd, RawFd},
-    path::Path,
-};
+use std::{hash::Hash, os::fd::AsRawFd, path::Path};
```
dave-tucker, line 4:
```suggestion
use std::{hash::Hash, os::fd::AsRawFd as _, path::Path};
```

aya/src/maps/mod.rs
```diff
@@ -42,7 +42,7 @@ use std::{
     marker::PhantomData,
     mem,
     ops::Deref,
-    os::fd::{AsFd, AsRawFd, OwnedFd, RawFd},
+    os::fd::{AsFd as _, AsRawFd, IntoRawFd as _, OwnedFd, RawFd},
```
dave-tucker, line 45:
```suggestion
    os::fd::{AsFd as _, AsRawFd as _, IntoRawFd as _, OwnedFd, RawFd},
```

dave-tucker (changes requested):
A couple of nits but other than that LGTM

aya/src/maps/mod.rs
nrxus: This goes against the earlier PR feedback from @tamird . I don't have a strong preference either way but I would lean towards not having the type in the comment since that is already obvious from the signature but I am fine with whatever you all device.

aya/src/programs/cgroup_device.rs
nrxus: This trait is explicitely used in the signature for `CgroupDevice::attach` so it cannot be aliased to `_`.

aya/src/maps/mod.rs
nrxus: This trait is explicitly used in this module for `impl AsRawFd for MapFd`.

aya/src/maps/mod.rs
dave-tucker: That's fine, I didn't see the earlier review comment. Feel free to leave it as is.

dave-tucker (approved):

aya/src/maps/mod.rs
```diff
@@ -563,35 +565,37 @@ impl MapData {
...
     ///
     /// If loading from a BPF Filesystem (bpffs) you should use [`Map::from_pin`](crate::maps::MapData::from_pin).
     /// This API is intended for cases where you have received a valid BPF FD from some other means.
     /// For example, you received an FD over Unix Domain Socket.
-    pub fn from_fd(fd: RawFd) -> Result<MapData, MapError> {
-        let info = bpf_map_get_info_by_fd(fd).map_err(|io_error| MapError::SyscallError {
-            call: "BPF_OBJ_GET",
-            io_error,
-        })?;
+    pub fn from_fd(fd: OwnedFd) -> Result<MapData, MapError> {
```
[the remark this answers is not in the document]
alessandrod: I think we should do this in a separate PR? You're only changing the signature
but otherwise everything else stays the same. Why not do the signature change
when we actually change the impl?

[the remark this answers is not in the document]
tamird: That would mean the `bpf_map_info_by_fd` call on line 591 would require an `unsafe` call to [`BorrowedFd::borrow_raw`](https://doc.rust-lang.org/stable/std/os/fd/struct.BorrowedFd.html#method.borrow_raw).

I'm OK with either approach.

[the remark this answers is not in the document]
alessandrod: Yeah the bpf_map_info_by_fd change is also not related to links (it's for maps), so I think it could have gone in a separate PR too (the PR does state this is about links).

`MapData::from_fd` is effectively the only public API change in this PR - without this change I wouldn't have had to look at the PR at all :P

That said no need to add more churn! Thanks a lot @nrxus, looks good I'm approving

alessandrod (approved):

tamird:
@nrxus apologies for the conflict. Let's rebase and get this landed.

tamird:
I took care of the conflict.

PR #718, opened 2023-08-02
uprobe: refactor target resolution 

ajwerner:
This attempts to do fewer lossy conversions and to avoid some
allocations.

### DIFF
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -2,6 +2,7 @@
 use libc::pid_t;
 use object::{Object, ObjectSection, ObjectSymbol};
 use std::{
+    borrow::Cow,
     error::Error,
     ffi::CStr,
     fs,
@@ -82,7 +83,7 @@ impl UProbe {
         target: T,
         pid: Option<pid_t>,
     ) -> Result<UProbeLinkId, ProgramError> {
-        let path = resolve_attach_path(target, pid)?;
+        let path = resolve_attach_path(&target, pid)?;
 
         let sym_offset = if let Some(fn_name) = fn_name {
             resolve_symbol(&path, fn_name).map_err(|error| UProbeError::SymbolError {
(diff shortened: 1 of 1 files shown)


aya/src/programs/uprobe.rs
```diff
@@ -127,38 +127,44 @@ fn resolve_attach_path<T: AsRef<Path>>(
...
+            find_lib_in_proc_maps(pid, target_str)
+                .map_err(|io_error| UProbeError::FileError {
+                    filename: format!("/proc/{pid}/maps"),
+                    io_error,
+                })
+                .transpose()
+        })
+        .transpose()?;
+    path.or_else(|| {
+        Some(target)
```
tamird, line 148: This expression (`Some().filter().map()`) would be clearer as `target.is_absolute().then(|| target_str.to_owned())`

aya/src/programs/uprobe.rs
```diff
@@ -127,38 +127,44 @@ fn resolve_attach_path<T: AsRef<Path>>(
     target: T,
     pid: Option<pid_t>,
 ) -> Result<String, UProbeError> {
```
tamird, line 129: can we return &str here? the only caller immediately takes a reference.

aya/src/programs/uprobe.rs
```diff
@@ -281,16 +287,16 @@ fn find_lib_in_proc_maps(pid: pid_t, lib: &str) -> Result<Option<String>, io::Er
...
     } else {
-        let lib = lib.to_string();
-        let lib1 = lib.clone() + ".so";
-        let lib2 = lib + "-";
-        libs.iter()
-            .find(|(k, _)| k.starts_with(&lib1) || k.starts_with(&lib2))
+        libs.into_iter().find(|(k, _)| {
+            k.strip_prefix(lib)
+                .map(|k| k.starts_with(".so") || k.starts_with('-'))
+                .unwrap_or_default()
```
tamird, line 295: `.map(...).unwrap_or_default` may be clearer as `map_or(false, ...)`

tamird (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/718#-:-NaqkXCnEnNwGoSQGqwu:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @ajwerner)

PR #719, opened 2023-08-02
aya: refactor handling of /proc/$pid/maps

dave-tucker:
Since there are eyes on the uprobe code right now, I pulled this one out from #329.

This commit refactors the handling of /proc/$pid/maps since the collection previously assumed that all entries here could be represented in a HashMap. Those with a path component are stored in a HashMap for fast lookup via library name. All other entries are cached in a Vec to allow for filtering based on offsets, required for supporting USDT probes.

### DIFF
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -2,6 +2,7 @@
 use libc::pid_t;
 use object::{Object, ObjectSection, ObjectSymbol};
 use std::{
+    collections::HashMap,
     error::Error,
     ffi::CStr,
     fs,
@@ -86,10 +87,14 @@ impl UProbe {
         let target_str = &*target.as_os_str().to_string_lossy();
 
         let mut path = if let Some(pid) = pid {
-            find_lib_in_proc_maps(pid, target_str).map_err(|io_error| UProbeError::FileError {
-                filename: format!("/proc/{pid}/maps"),
-                io_error,
-            })?
+            let proc_map_libs =
+                ProcMap::new(pid).map_err(|e| UProbeError::ProcMapError { pid, source: e })?;
+            proc_map_libs
+                .find_by_name(target_str)
+                .map_err(|io_error| UProbeError::FileError {
+                    filename: format!("/proc/{}/maps", pid),
+                    io_error,
+                })?
         } else {
             None
         };
@@ -227,43 +232,17 @@ pub enum UProbeError {
         #[source]
         io_error: io::Error,
     },
+
+    /// There was en error resolving a path
+    #[error("error fetching libs for {pid}")]
+    ProcMapError {
+        /// The pid
+        pid: i32,
+        /// The [`ProcMapError`] that caused the error
+        #[source]
+        source: ProcMapError,
+    },
 }
-
-fn proc_maps_libs(pid: pid_t) -> Result<Vec<(String, String)>, io::Error> {
-    let maps_file = format!("/proc/{pid}/maps");
-    let data = fs::read_to_string(maps_file)?;
-
-    Ok(data
-        .lines()
-        .filter_map(|line| {
-            let line = line.split_whitespace().last()?;
-            if line.starts_with('/') {
-                let path = PathBuf::from(line);
-                let key = path.file_name().unwrap().to_string_lossy().into_owned();
-                Some((key, path.to_string_lossy().to_string()))
-            } else {
-                None
-            }
-        })
-        .collect())
-}
-
-fn find_lib_in_proc_maps(pid: pid_t, lib: &str) -> Result<Option<String>, io::Error> {
-    let libs = proc_maps_libs(pid)?;
-
-    let ret = if lib.contains(".so") {
-        libs.iter().find(|(k, _)| k.as_str().starts_with(lib))
-    } else {
-        let lib = lib.to_string();
-        let lib1 = lib.clone() + ".so";
-        let lib2 = lib + "-";
-        libs.iter()
-            .find(|(k, _)| k.starts_with(&lib1) || k.starts_with(&lib2))
-    };
-
-    Ok(ret.map(|(_, v)| v.clone()))
-}
-
 #[derive(Debug)]
 pub(crate) struct CacheEntry {
     key: String,
@@ -428,5 +407,154 @@ fn resolve_symbol(path: &str, symbol: &str) -> Result<u64, ResolveSymbolError> {
             )
         })?;
         Ok(sym.address() - section.address() + offset)
+    }
+}
+
+/// Error reading from /proc/pid/maps
+#[derive(Debug, Error)]
+pub enum ProcMapError {
+    /// An [`io::Error`]
+    #[error(transparent)]
+    IoError(io::Error),
+
+    /// Error parsing a line of /proc/pid/maps
+    #[error("proc map entry parse error")]
+    ParseError,
+}
+
+pub(crate) struct ProcMap {
+    _entries: Vec<ProcMapEntry>,
+    paths: HashMap<String, String>,
+}
+
+impl ProcMap {
+    fn new(pid: pid_t) -> Result<Self, ProcMapError> {
+        let maps_file = format!("/proc/{}/maps", pid);
+        let data = fs::read_to_string(maps_file).map_err(ProcMapError::IoError)?;
+        let mut entries = vec![];
+        let mut paths = HashMap::new();
+        for line in data.lines() {
+            let entry = ProcMapEntry::parse(line)?;
+            if let Some(path) = &entry.path {
+                let p = PathBuf::from(path);
+                let key = p.file_name().unwrap().to_string_lossy().into_owned();
+                let value = p.to_string_lossy().to_string();
+                paths.insert(key, value);
+            }
+            entries.push(entry);
+        }
+        Ok(ProcMap {
+            _entries: entries,
+            paths,
+        })
+    }
+
+    fn find_by_name(&self, lib: &str) -> Result<Option<String>, io::Error> {
+        let ret = if lib.contains(".so") {
+            self.paths.iter().find(|(k, _)| k.as_str().starts_with(lib))
+        } else {
+            let lib = lib.to_string();
+            let lib1 = lib.clone() + ".so";
+            let lib2 = lib + "-";
+            self.paths
+                .iter()
+                .find(|(k, _)| k.starts_with(&lib1) || k.starts_with(&lib2))
+        };
+
+        Ok(ret.map(|(_, v)| v.clone()))
+    }
+}
+
+pub(crate) struct ProcMapEntry {
+    _address: u64,
+    _address_end: u64,
+    _perms: String,
+    _offset: u64,
+    _dev: String,
+    _inode: u32,
+    path: Option<String>,
+}
+
+impl ProcMapEntry {
+    fn parse(line: &str) -> Result<Self, ProcMapError> {
+        let parts: Vec<&str> = line.split_whitespace().collect();
+        if parts.len() < 5 {
+            return Err(ProcMapError::ParseError);
+        }
+        let addr_parts: Vec<&str> = parts[0].split('-').collect();
+        let address =
+            u64::from_str_radix(addr_parts[0], 16).map_err(|_| ProcMapError::ParseError)?;
+        let address_end =
+            u64::from_str_radix(addr_parts[1], 16).map_err(|_| ProcMapError::ParseError)?;
+        let perms = parts[1];
+        let offset = u64::from_str_radix(parts[2], 16).map_err(|_| ProcMapError::ParseError)?;
+        let dev = parts[3];
+        let inode = parts[4].parse().map_err(|_| ProcMapError::ParseError)?;
+        let path = if parts.len() == 6 {
+            if parts[5].starts_with('/') {
+                Some(parts[5].to_string())
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
+        Ok(ProcMapEntry {
+            _address: address,
+            _address_end: address_end,
+            _perms: perms.to_string(),
+            _offset: offset,
+            _dev: dev.to_string(),
+            _inode: inode,
+            path,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_parse_proc_map_entry_shared_lib() {
+        let s = "7ffd6fbea000-7ffd6fbec000 r-xp 00000000 00:00 0                          [vdso]";
+        let proc_map = ProcMapEntry::parse(s).unwrap();
+        assert_eq!(proc_map._address, 0x7ffd6fbea000);
+        assert_eq!(proc_map._address_end, 0x7ffd6fbec000);
+        assert_eq!(proc_map._perms, "r-xp");
+        assert_eq!(proc_map._offset, 0x0);
+        assert_eq!(proc_map._dev, "00:00");
+        assert_eq!(proc_map._inode, 0);
+        assert_eq!(proc_map.path, None);
+    }
+
+    #[test]
+    fn test_parse_proc_map_entry_absolute_path() {
+        let s = "7f1bca83a000-7f1bca83c000 rw-p 00036000 fd:01 2895508                    /usr/lib64/ld-linux-x86-64.so.2";
+        let proc_map = ProcMapEntry::parse(s).unwrap();
+        assert_eq!(proc_map._address, 0x7f1bca83a000);
+        assert_eq!(proc_map._address_end, 0x7f1bca83c000);
+        assert_eq!(proc_map._perms, "rw-p");
+        assert_eq!(proc_map._offset, 0x00036000);
+        assert_eq!(proc_map._dev, "fd:01");
+        assert_eq!(proc_map._inode, 2895508);
+        assert_eq!(
+            proc_map.path,
+            Some("/usr/lib64/ld-linux-x86-64.so.2".to_string())
+        );
+    }
+
+    #[test]
+    fn test_parse_proc_map_entry_all_zeros() {
+        let s = "7f1bca5f9000-7f1bca601000 rw-p 00000000 00:00 0";
+        let proc_map = ProcMapEntry::parse(s).unwrap();
+        assert_eq!(proc_map._address, 0x7f1bca5f9000);
+        assert_eq!(proc_map._address_end, 0x7f1bca601000);
+        assert_eq!(proc_map._perms, "rw-p");
+        assert_eq!(proc_map._offset, 0x0);
+        assert_eq!(proc_map._dev, "00:00");
+        assert_eq!(proc_map._inode, 0);
+        assert_eq!(proc_map.path, None);
     }
 }


addisoncrump:
Don't suppose we could get an iterator over the entries + getters for fields in this PR?

dave-tucker:
> Don't suppose we could get an iterator over the entries + getters for fields in this PR?

Sure. I'll address the comments here tomorrow and will see about adding an iterator and some getters here too.

dave-tucker:
@addisoncrump Can you check the API here and let me know if this is what you had in mind?

dave-tucker (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`aya/src/programs/uprobe.rs` line 95 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NaqGoIG3IUtNStHKyT3:-NawIFpl438Tih03EI3Z:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L95)):*

why change this? what's wrong with {pid}?
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 236 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NapqVdo-0oN9Bo91bkx:-NawIHhn3UnEr8BbubvU:b-9rmk7y) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L236)):*

nit: not pub, so doc comments are dead code. Can we remove them?
</blockquote></details>

Is pub now
___
*[`aya/src/programs/uprobe.rs` line 415 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NapqKfq5uxnzCbyvCFL:-NawISUqC1wKR7BaTWro:b-swkl4v) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L415)):*

should we make this non-pub and have the relevant functions return dyn box errors? we're long talked about hiding errors from the public API.
</blockquote></details>

Yes eventually, although I don't have a good picture in my mind what our Box<dyn Error> future will look like exactly...
I'm hesitant to try it here since I'll need to redo all of UprobeError handling, which in turn links back into ProgramError and so on - it could very quickly snowball.
___
*[`aya/src/programs/uprobe.rs` line 425 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NaqFdxR32c-zbstyqK4:-NawJqD-1Fjd0s4opw4u:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L425)):*

why pub(crate)? only used in this file AFAICT.
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 479 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NaqGYVv3O-Ryrlz4Z5f:-NawJqeW9i_UNJUMRDpo:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L479)):*

can we implement FromStr instead of this bespoke method?
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 480 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NaqGOvZ7luxpQBBMJrk:-NawJrAi6MT7yzkU2WZP:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L480)):*

can we avoid the vector?

```
let mut parts = line.split_whitespace();
let next = || parts.next().ok_or(ProcMapError::ParseError);
let addr_parts = next()?.split(...);
let perms = next()?;
...
```
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 484 at r1](https://reviewable.io/reviews/aya-rs/aya/719#-NaqGQFM2c6nNfd2Vawz:-NawJrj3BPSz_qEt11qm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/7972b5d25d750c3c3104a99c301bfee9bafdd68c/aya/src/programs/uprobe.rs#L484)):*

ditto, avoid the vector. split_once?
</blockquote></details>

Done.

addisoncrump:
> @addisoncrump Can you check the API here and let me know if this is what you had in mind?

Looks good. If you've got some time, an `async` version of `new` would also be useful for e.g. AsyncPerfEventArray, where this will likely often be used.

tamird:
> > @addisoncrump Can you check the API here and let me know if this is what you had in mind?
> 
> Looks good. If you've got some time, an `async` version of `new` would also be useful for e.g. AsyncPerfEventArray, where this will likely often be used.

Async reading of `/proc/{pid}/maps`? Why would such code ever yield?

addisoncrump:
> Why would such code ever yield?

Fair point, the async is probably not worth it. Disregard.

addisoncrump:
Is this ready to merge? :slightly_smiling_face:

tamird (commented):
Looks like comments still need to be addressed.

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719#-:-NbL50edEbSq8RnD9ond:b-3pxz0v)* status: all files reviewed, 11 unresolved discussions (waiting on @alessandrod and @dave-tucker)

dave-tucker (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719)* status: all files reviewed, 10 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`aya/src/programs/uprobe.rs` line 438 at r3](https://reviewable.io/reviews/aya-rs/aya/719#-NawKhYe0jWlAtLOpBUw:-NbPGcB6-FVozCGF2bJr:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/649074f7b54d874569dedbf84da036cf9198c162/aya/src/programs/uprobe.rs#L438)):*

repeating the type - maybe this variant should be called "ProcfsMapsError"?
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 450 at r3](https://reviewable.io/reviews/aya-rs/aya/719#-NawK76t1z_SPqM9Cftj:-NbPGcm35zA_qkRHC0ay:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/649074f7b54d874569dedbf84da036cf9198c162/aya/src/programs/uprobe.rs#L450)):*

missing period - also, if you intend to separate these sentences into two paragraphs, there should be an empty comment line between them.
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 476 at r3](https://reviewable.io/reviews/aya-rs/aya/719#-NawKFC3ANu0JVGfR-Tq:-NbPGdH8-DKmerLOlSvC:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/649074f7b54d874569dedbf84da036cf9198c162/aya/src/programs/uprobe.rs#L476)):*

ditto empty comment line
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 518 at r3](https://reviewable.io/reviews/aya-rs/aya/719#-NawKRHR2fH5fFWxbcpG:-NbPGfDS70uRTkoJcjJ_:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/649074f7b54d874569dedbf84da036cf9198c162/aya/src/programs/uprobe.rs#L518)):*

all comments missing periods
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 259 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPQdm8EaWnwDs6Svsy:-NbPpk4J3HJGxr9px5dg:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L259)):*

what does this mean?
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 261 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPQg6D5npwYAyZgjwm:-NbPpkSv6c5ELWz3N_b8:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L261)):*

maybe drop "Error" here too? see comment below 
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 441 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPPsQe98lLN-MUEDTt:-NbPpn1Z3HY75GUO3x-N:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L441)):*

can we remove the Error suffix from all the variants? https://rust-lang.github.io/rust-clippy/master/index.html#/enum_variant_names

Clippy isn't flagging this because this enum is `pub`, but since we're introducing it here, let's follow the advice.
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 473 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPStTh17khmVDcP4PK:-NbPqpna9SqnAYBhLMYh:b-xl3ok4) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L473)):*

why lossy here and below? but also should this just be a map<PathBuf, PathBuf> to avoid the fallible conversions?
</blockquote></details>

I don't think we can be Map<PathBuf, PathBuf>
I need to map a filename (which is a library name) to it's full path...
Assuming I keep the PathBuf, the filename could then be.
```rust
let filename = p
                    .file_name()
                    .and_then(|s| s.to_str())
                    .map(|s| s.to_string())
                    .expect("library path should have a filename");
```
___
*[`aya/src/programs/uprobe.rs` line 475 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPT2ZdD0qhPmISocVf:-NbPqfc13O5nOCTcUrCa:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L475)):*

why `.entry.or_insert`? do we expect repetitions? are we testing that case?
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 486 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPRDfz1PvBgvZ2dlSv:-NbPqgb39dn3TCER5IjN:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L486)):*

why not return Option<&str> and let the caller clone if necessary?
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 576 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPRV03CMwFCgX-Hiq8:-NbPqjTZBMFVA185J69o:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L576)):*

should we check the errors right here rather than parse further? even more succinct would be:

let (address, adddress_end) = next()?.split_once('-').ok_or(ProcMapError::ParseError).and_then(|(start, end)| {
  let start = u64::from_str_radix(start, 16)?;
  let end = u64::from_str_radix(end, 16)?;
  Ok((start, end))
})?;
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 578 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPRc4s0_KaKVuKowJh:-NbPqkfR0HfayFBqHejD:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L578)):*

map_err here can go away since you have a `#[from]`; the try operator will do the right thing.
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 582 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPSK8E7P91sCmBQw-L:-NbPqlatEuWoM67vpYy4:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L582)):*

s.starts_with('/').then(|| s.to_string())
</blockquote></details>

Done.

tamird (changes requested):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719)* status: all files reviewed, 3 unresolved discussions (waiting on @alessandrod and @dave-tucker)

___
*[`aya/src/programs/uprobe.rs` line 473 at r4](https://reviewable.io/reviews/aya-rs/aya/719#-NbPStTh17khmVDcP4PK:-NbPtp-w9vSAAi-7ns41:b-r66gvb) ([raw file](https://github.com/aya-rs/aya/blob/c66518556a78dfbd285dd2bc82f6800901c6bd28/aya/src/programs/uprobe.rs#L473)):*

I don't think we can be Map<PathBuf, PathBuf>
I need to map a filename (which is a library name) to it's full path...
Assuming I keep the PathBuf, the filename could then be.
```rust
let filename = p
                    .file_name()
                    .and_then(|s| s.to_str())
                    .map(|s| s.to_string())
                    .expect("library path should have a filename");
```
</blockquote></details>

Actually, why even have a map at all? `find_library_path_by_name` is never able to use anything as a key for a precise lookup. If you want de-duplication, this may as well be a HashSet<PathBuf>.
___
*[`aya/src/programs/uprobe.rs` line 588 at r5](https://reviewable.io/reviews/aya-rs/aya/719#-NbPtF4z0v-7gdJ4SU_B:-NbPtF5-5Edr-wMPpMru:b-5mpd9p) ([raw file](https://github.com/aya-rs/aya/blob/4b3bd50e3ff2b6341afa5705dd4a829f9fbeab60/aya/src/programs/uprobe.rs#L588)):*
> ```XML
>         let offset = u64::from_str_radix(next()?, 16)?;
>         let dev = next()?;
>         let inode = next()?.parse().map_err(ProcMapError::ParseInt)?;
> ```

map_err can go away
___
*[`aya/src/programs/uprobe.rs` line 735 at r5](https://reviewable.io/reviews/aya-rs/aya/719#-NbPtOvn89791hSsV17j:-NbPtOvn89791hSsV17k:b34w2yd) ([raw file](https://github.com/aya-rs/aya/blob/4b3bd50e3ff2b6341afa5705dd4a829f9fbeab60/aya/src/programs/uprobe.rs#L735)):*
> ```XML
>                 .find_library_path_by_name("ld-linux-x86-64.so.2")
>                 .unwrap(),
>             Some("/usr/lib64/ld-linux-x86-64.so.2")
> ```

what's being tested here? does it pick the first one? the last one? they're all the same.

aya/src/programs/uprobe.rs
```diff
@@ -450,3 +432,307 @@ fn resolve_symbol(path: &str, symbol: &str) -> Result<u64, ResolveSymbolError> {
...
+    #[error("proc map entry parse error")]
+    Parse,
+}
+
+/// The memory maps of a process.
+///
+/// This is read from /proc/`pid`/maps.
+///
+/// The information here may be used to resolve addresses to paths.
+pub struct ProcMap {
```
alessandrod, line 457: Couldn't this be private for now? I understand that USDT might need it public at
some point, but can we flip it public when we merge that?

I don't think we have anything to lose from having it private for now. Instead
if we make it public now and by the time we merge USDT we need to change API for
whatever reason, then we're going to introduce yet another breaking change.

dave-tucker: It's pub because @addisoncrump asked for it 😉 
I'm happy to flip it to priv and let them convince you why it should be pub

alessandrod: Then I'm going to have to review the API 😋

tamird (changes requested):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719)* status: all files reviewed, 4 unresolved discussions (waiting on @alessandrod and @dave-tucker)

___
*[`aya/src/programs/uprobe.rs` line 481 at r6](https://reviewable.io/reviews/aya-rs/aya/719#-NbQ9f7d5OkosORsmOXo:-NbQ9f7d5OkosORsmOXp:b-m2d8zq) ([raw file](https://github.com/aya-rs/aya/blob/127860d1238d37f9d2b8b83cb2ca3870aaab7811/aya/src/programs/uprobe.rs#L481)):*
> ```XML
>                 p.file_name()
>                     .and_then(|s| s.to_str())
>                     .map(|s| s.starts_with(lib))
> ```

can this and the below be map_or(false, ...) to avoid unwrap_or_default? nit, but is more explicit that way
___
*[`aya/src/programs/uprobe.rs` line 488 at r6](https://reviewable.io/reviews/aya-rs/aya/719#-NbQ9LKA8SqcIs3zrhXS:-NbQ9LKA8SqcIs3zrhXT:bkftqju) ([raw file](https://github.com/aya-rs/aya/blob/127860d1238d37f9d2b8b83cb2ca3870aaab7811/aya/src/programs/uprobe.rs#L488)):*
> ```XML
>                 p.file_name()
>                     .and_then(|s| s.to_str())
>                     .map(|s| {
> ```

make this `and_then`, will remove one of the unwrap_or_default calls

aya/src/programs/uprobe.rs
```diff
@@ -450,3 +432,277 @@ fn resolve_symbol(path: &str, symbol: &str) -> Result<u64, ResolveSymbolError> {
...
+/// for the process, ranging from address to address_end.
+#[derive(Debug)]
+struct ProcMapEntry {
+    _address: u64,
+    _address_end: u64,
+    _perms: String,
+    _offset: u64,
+    _dev: String,
+    _inode: u32,
+    path: Option<String>,
```
addisoncrump, line 538: The path of the library is not guaranteed to be UTF-8.

aya/src/programs/uprobe.rs
```diff
@@ -450,3 +432,277 @@ fn resolve_symbol(path: &str, symbol: &str) -> Result<u64, ResolveSymbolError> {
...
+
+        Ok(ret.map(|p| {
+            p.as_os_str()
+                .to_str()
+                .expect("library path should be valid UTF-8")
+        }))
+    }
+}
+
+impl FromStr for ProcMap {
```
addisoncrump, line 505: It is not a valid assumption that the /proc/$pid/map is UTF-8, since the library name can be non-UTF-8 (though this is very unlikely, it still is a potential use case).

addisoncrump (changes requested):

aya/src/programs/uprobe.rs
dave-tucker: It's priv now. Once this is in someone can open a PR and we can all bikeshed on the API 😆 
My priority is to get the USDT work rebased and reviewable.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719)* status: all files reviewed, 6 unresolved discussions (waiting on @addisoncrump, @alessandrod, and @dave-tucker)

___
*[`aya/src/programs/uprobe.rs` line 457 at r5](https://reviewable.io/reviews/aya-rs/aya/719#gh-1288798174:-NbUpAhP2vpJ_ZWv0_cR:b-k3og84) ([raw file](https://github.com/aya-rs/aya/blob/4b3bd50e3ff2b6341afa5705dd4a829f9fbeab60/aya/src/programs/uprobe.rs#L457)):*

It's priv now. Once this is in someone can open a PR and we can all bikeshed on the API 😆 
My priority is to get the USDT work rebased and reviewable.
</blockquote></details>

ProcMapError is still pub. @alessandrod TAL?
___
*[`aya/src/programs/uprobe.rs` line 505 at r6](https://reviewable.io/reviews/aya-rs/aya/719#gh-1290006785:-NbV7lLN7eHAf_fL3tq9:b-5ysxyz) ([raw file](https://github.com/aya-rs/aya/blob/127860d1238d37f9d2b8b83cb2ca3870aaab7811/aya/src/programs/uprobe.rs#L505)):*
<details><summary><i>Previously, addisoncrump (Addison Crump) wrote…</i></summary><blockquote>

It is not a valid assumption that the /proc/$pid/map is UTF-8 (and therefore representable by &str), since the library name can be non-UTF-8 (though this is very unlikely, it still is a potential use case).
</blockquote></details>

Sniped me into looking into this, I took a shot at boiling away all the UTF-8 assumptions in https://github.com/aya-rs/aya/pull/742.

tamird:
Do we still want to do this?

tamird:
I rebased this and moved some code around to reduce the diff. I think there are still questions about Unicode assumptions.

dave-tucker:
> Do we still want to do this?

Yes, it was extracted from #329 to make that change easier to review in smaller chunks. I'd like to get back to #329 at some point as we've had multiple people ask for this in the past.

tamird:
@dave-tucker can you help me understand what the important bits of this change are? The description talks about hash map assumptions, but the base shape of this code no longer contains a hash map.

tamird (approved):
> Reviewed 2 of 2 files at r7, 2 of 2 files at r10, 1 of 1 files at r11, 2 of 2 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, all commit messages.
> Dismissed @addisoncrump and @alessandrod from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/719)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

___
*[`aya/src/programs/uprobe.rs` line 391 at r14](https://reviewable.io/reviews/aya-rs/aya/719#-OKb2uq263dtUFe1AHDD:-OKb2uq263dtUFe1AHDE:bqp70n5) ([raw file](https://github.com/aya-rs/aya/blob/b46d196ff5afeea305735325da0850fac3d6ba32/aya/src/programs/uprobe.rs#L391)):*
> ```XML
> ///
> /// The information here may be used to resolve addresses to paths.
> struct ProcMap<T> {
> ```

I made this generic so I could remove the `.to_vec` calls in the tests.

dave-tucker (commented):
Apparently I'm not able to review my own PR, even if it's been carried 😄 
But these changes do indeed LGTM - Thanks @tamird

tamird:
Great! Will merge when this is green.

PR #727, opened 2023-08-08
integration-test: avoid reliance on kernel headers

tamird:
This should allow us to build on any system.


ajwerner (approved):
LGTM

tamird:
@dave-tucker could you please have a look?

test/integration-test/build.rs
```diff
@@ -116,10 +116,15 @@ fn main() {
             target_arch.push(arch);
         };
 
+        let libbpf_vmlinux_dir = libbpf_dir.join(".github/actions/build-selftests");
```
dave-tucker, line 113: This could use a comment describing what is going on here. It's clear this is the path to a vmlinux.h, but what is not clear is:

1. How the file was generated?
2. Is this location stable? It looks like it's part of a GH action in libbpf, but from some simple grepping it doesn't appear to be used anywhere.

Is this something we want to do on all platforms?

test/integration-test/bpf/text_64_64_reloc.c
```diff
@@ -1,5 +1,5 @@
 // clang-format off
-#include <linux/bpf.h>
+#include <vmlinux.h>
```
dave-tucker, line 2: Should this be something like:
```c
#ifdef __linux__
#include <linux/bpf.h>
#else
#include <vmlinux.h>
#endif
```

You could even add our own `linux.h` file to DRY this up.
If possible it's probably better to let the preprocessor decide what to include than chosing `vmlinux` in all cases.

test/integration-test/bpf/reloc.bpf.c
```diff
@@ -46,20 +56,32 @@ __attribute__((noinline)) int pointer_global() {
...
-__attribute__((noinline)) int struct_flavors_global() {
+__noinline int struct_flavors_global() {
   struct relocated_struct_with_scalars s = {1, 2, 3};
+#ifndef TARGET
   if (bpf_core_field_exists(s.a)) {
     return set_output(__builtin_preserve_access_index(s.a));
+#else
+  if (bpf_core_field_exists(s.d)) {
+    return set_output(__builtin_preserve_access_index(s.d));
+#endif
```
dave-tucker, line 67: Shouldn't this be:
```c
if (bpf_core_field_exists(s.a)) {
   return set_output(__builtin_preserve_access_index(s.a));
} else if (bpf_core_field_exists(s.d)) {
  return set_output(__builtin_preserve_access_index(s.d));
} else {
   return 42
}
```

That would return 1 if a exists, 3 if d exists OR `42` if neither a or d exist.
I don't understand what adding the `ifndef TARGET` adds here.

dave-tucker (changes requested):
nit: Could you please include the crates touched in the first line of the commit message when practical please (i.e it's not all crates).

```test, xtask: Simplify ExitStatus```

It makes it easier to see what's happening on `git log --pretty=oneline`

test/integration-test/bpf/text_64_64_reloc.c
tamird: But why? We don't actually use any Linux data structures, we need this include to provide defines relied on by the libbpf headers.

test/integration-test/bpf/reloc.bpf.c
tamird: Any expression including `s.a` doesn't compile when TARGET is defined because `a` doesn't exist.

test/integration-test/build.rs
tamird: Done. Yes, we want to do it on all platforms.

dave-tucker: > // However in our tests we do not make use of kernel data structures, and so any vmlinux.h
> // which defines the constants we need (e.g. `__u8`, `__u64`, `BPF_MAP_TYPE_ARRAY`,
> // `BPF_ANY`, `XDP_PASS`, `XDP_DROP`, etc.) will suffice.

If it is just a handful of contants, why not just write a quick `linux.h`?
That would keep things self-contained.

dave-tucker (commented):
Generally LGTM. I'm still concerned about random vmlinux.h from libbpf's testing workflow - which we may find eventually disappears. I'm not concerned enough to block this merging though.

test/integration-test/bpf/text_64_64_reloc.c
tamird: Oh, you want the ones from libbpf? Those aren't self contained, see https://github.com/libbpf/libbpf/pull/41#issuecomment-1668143392.

There really aren't any viable options if we want things to be self-contained. We could write our own header, but then we own it for life. If this one disappears, we can do that then.

test/integration-test/build.rs
tamird: Yeah, we could. This is a shortcut, and we can take the long-cut if and when we need to. Does that seem reasonable?

dave-tucker (approved):
LGTM.
I'll open an issue to track the longer term item.

tamird:
Thanks for the review!

PR #690, opened 2023-07-26
aya: Return error messages from netlink

dave-tucker:
This returns error strings from netlink since they are more informative than the raw os error. For example:

"Device or Resource Busy" vs. "XDP program already attached".

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/690)

PR #728, opened 2023-08-08
aya: fix tc name limit

pooladkhay:
The buffer for attributes should be sized to hold at least 256 bytes, based on `CLS_BPF_NAME_LEN = 256` from the kernel:
https://github.com/torvalds/linux/blob/02aee814/net/sched/cls_bpf.c#L28

This pull request also includes integration tests with an eBPF program of type classifier with a 256-byte-long name.

Fixes: #610

### DIFF
--- a/aya/src/sys/netlink.rs
+++ b/aya/src/sys/netlink.rs
@@ -10,8 +10,8 @@ use libc::{
     getsockname, nlattr, nlmsgerr, nlmsghdr, recv, send, setsockopt, sockaddr_nl, socket,
     AF_NETLINK, AF_UNSPEC, ETH_P_ALL, IFF_UP, IFLA_XDP, NETLINK_EXT_ACK, NETLINK_ROUTE,
     NLA_ALIGNTO, NLA_F_NESTED, NLA_TYPE_MASK, NLMSG_DONE, NLMSG_ERROR, NLM_F_ACK, NLM_F_CREATE,
-    NLM_F_DUMP, NLM_F_ECHO, NLM_F_EXCL, NLM_F_MULTI, NLM_F_REQUEST, RTM_DELTFILTER, RTM_GETTFILTER,
-    RTM_NEWQDISC, RTM_NEWTFILTER, RTM_SETLINK, SOCK_RAW, SOL_NETLINK,
+    NLM_F_DUMP, NLM_F_ECHO, NLM_F_EXCL, NLM_F_MULTI, NLM_F_REQUEST, RTM_DELTFILTER, RTM_GETQDISC,
+    RTM_GETTFILTER, RTM_NEWQDISC, RTM_NEWTFILTER, RTM_SETLINK, SOCK_RAW, SOL_NETLINK,
 };
 use thiserror::Error;
 
@@ -78,6 +78,46 @@ pub(crate) unsafe fn netlink_set_xdp_fd(
     sock.recv()?;
 
     Ok(())
+}
+
+pub(crate) unsafe fn netlink_clsact_qdisc_exists(if_index: i32) -> Result<bool, io::Error> {
+    let sock = NetlinkSocket::open()?;
+
+    let mut req = mem::zeroed::<TcRequest>();
+
+    let nlmsg_len = mem::size_of::<nlmsghdr>() + mem::size_of::<tcmsg>();
+
+    req.header = nlmsghdr {
+        nlmsg_len: nlmsg_len as u32,
+        nlmsg_flags: (NLM_F_REQUEST | NLM_F_ACK | NLM_F_DUMP) as u16,
+        nlmsg_type: RTM_GETQDISC,
+        nlmsg_pid: 0,
+        nlmsg_seq: 1,
+    };
+    req.tc_info.tcm_family = AF_UNSPEC as u8;
+
+    sock.send(&bytes_of(&req)[..req.header.nlmsg_len as usize])?;
+
+    for msg in sock.recv()? {
+        if msg.header.nlmsg_type != RTM_NEWQDISC {
+            continue;
+        }
+
+        let tc_msg = ptr::read_unaligned(msg.data.as_ptr() as *const tcmsg);
+        if tc_msg.tcm_ifindex != if_index {
+            continue;
+        }
+
+        let attrs = parse_attrs(&msg.data[mem::size_of::<tcmsg>()..])?;
+
+        if let Some(opts) = attrs.get(&(TCA_KIND as u16)) {
+            if opts.data == b"clsact\0" {
+                return Ok(true);
+            }
+        }
+    }
+
+    Ok(false)
 }
 
 pub(crate) unsafe fn netlink_qdisc_add_clsact(if_index: i32) -> Result<(), io::Error> {
@@ -289,7 +329,13 @@ struct Request {
 struct TcRequest {
     header: nlmsghdr,
     tc_info: tcmsg,
-    attrs: [u8; 64],
+    // The buffer for attributes should be sized to hold at least 256 bytes,
+    // based on `CLS_BPF_NAME_LEN = 256` from the kernel:
+    // https://github.com/torvalds/linux/blob/02aee814/net/sched/cls_bpf.c#L28
+    // We currently use around ~30 bytes of attributes in addition to name.
+    // Rather than picking a "right sized buffer" for the payload (which is of
+    // varying length anyway) we use the next largest power of 2.
+    attrs: [u8; 512],
 }
 
 struct NetlinkSocket {
--- a/test/integration-ebpf/src/tc_name_limit.rs
+++ b/test/integration-ebpf/src/tc_name_limit.rs
@@ -1,0 +1,22 @@
+#![no_std]
+#![no_main]
+
+use aya_bpf::{macros::classifier, programs::TcContext};
+
+// A function with a 256-byte-long name (all 'a's) to be used as the name of
+// the ebpf program. This name must match the name passed to userspace side
+// of the program (i.e. test/integration-test/src/tests/load.rs).
+// 256 is the maximum length allowed by the kernel, so this test should pass.
+// https://github.com/torvalds/linux/blob/02aee814/net/sched/cls_bpf.c#L28
+#[classifier]
+pub fn aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(
+    _ctx: TcContext,
+) -> i32 {
+    0
+}
+
+#[cfg(not(test))]
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
--- a/test/integration-test/src/tests/load.rs
+++ b/test/integration-test/src/tests/load.rs
@@ -6,11 +6,15 @@ use std::{
     time::{Duration, SystemTime},
 };
 
+use assert_matches::assert_matches;
 use aya::{
     maps::Array,
     programs::{
         links::{FdLink, PinnedLink},
-        loaded_links, loaded_programs, KProbe, TracePoint, UProbe, Xdp, XdpFlags,
+        loaded_links, loaded_programs,
+        tc::{clsact_qdisc_exists, qdisc_add_clsact},
+        KProbe, ProgramError, SchedClassifier, TcAttachType, TcError, TracePoint, UProbe, Xdp,
+        XdpFlags,
     },
     util::KernelVersion,
     Bpf,
@@ -607,4 +611,52 @@ fn pin_lifecycle_uprobe() {
 
     // Make sure the function isn't optimized out.
     uprobe_function();
+}
+
+#[test]
+fn tc_name_limit() {
+    let clsact_exists = clsact_qdisc_exists("lo").unwrap();
+    if !clsact_exists {
+        qdisc_add_clsact("lo").unwrap();
+    }
+
+    let mut bpf = Bpf::load(crate::TC_NAME_LIMIT_TEST).unwrap();
+
+    let long_name = "a".repeat(256);
+
+    let program: &mut SchedClassifier = bpf
+        .program_mut(long_name.as_str())
+        .unwrap()
+        .try_into()
+        .unwrap();
+    program.load().unwrap();
+    program.attach("lo", TcAttachType::Ingress).unwrap();
+}
+
+#[test]
+fn tc_name_limit_exceeded() {
+    let clsact_exists = clsact_qdisc_exists("lo").unwrap();
+    if !clsact_exists {
+        qdisc_add_clsact("lo").unwrap();
+    }
+
+    let mut bpf = Bpf::load(crate::TC_NAME_LIMIT_EXCEEDED_TEST).unwrap();
+
+    let long_name = "a".repeat(257);
+
+    let program: &mut SchedClassifier = bpf
+        .program_mut(long_name.as_str())
+        .unwrap()
+        .try_into()
+        .unwrap();
+    program.load().unwrap();
+
+    assert_matches!(
+      program.attach("lo", TcAttachType::Ingress),
+      Err(ProgramError::TcError(TcError::NetlinkError { io_error })) => {
+        // An invalid argument error (EINVAL) with code 22 should occur.
+        // The invalid argument is the tc program name which is too long.
+        assert_eq!(io_error.raw_os_error(), Some(22))
+      }
+    );
 }
--- a/aya/src/programs/tc.rs
+++ b/aya/src/programs/tc.rs
@@ -14,8 +14,8 @@ use crate::{
     },
     programs::{define_link_wrapper, load_program, Link, ProgramData, ProgramError},
     sys::{
-        netlink_find_filter_with_name, netlink_qdisc_add_clsact, netlink_qdisc_attach,
-        netlink_qdisc_detach,
+        netlink_clsact_qdisc_exists, netlink_find_filter_with_name, netlink_qdisc_add_clsact,
+        netlink_qdisc_attach, netlink_qdisc_detach,
     },
     util::{ifindex_from_ifname, tc_handler_make},
     VerifierLogLevel,
@@ -357,4 +357,10 @@ fn qdisc_detach_program_fast(
     }
 
     Ok(())
+}
+
+/// Check if the `clasct` qdisc exists on the given interface.
+pub fn clsact_qdisc_exists(if_name: &str) -> Result<bool, io::Error> {
+    let if_index = ifindex_from_ifname(if_name)?;
+    unsafe { netlink_clsact_qdisc_exists(if_index as i32) }
 }
--- a/test/integration-ebpf/Cargo.toml
+++ b/test/integration-ebpf/Cargo.toml
@@ -55,3 +55,11 @@ path = "src/xdp_sec.rs"
 [[bin]]
 name = "ring_buf"
 path = "src/ring_buf.rs"
+
+[[bin]]
+name = "tc_name_limit"
+path = "src/tc_name_limit.rs"
+
+[[bin]]
+name = "tc_name_limit_exceeded"
+path = "src/tc_name_limit_exceeded.rs"
--- a/test/integration-ebpf/src/tc_name_limit_exceeded.rs
+++ b/test/integration-ebpf/src/tc_name_limit_exceeded.rs
@@ -1,0 +1,22 @@
+#![no_std]
+#![no_main]
+
+use aya_bpf::{macros::classifier, programs::TcContext};
+
+// A function with a 257-byte-long name (all 'a's) to be used as the name of
+// the ebpf program. This name must match the name passed to userspace side
+// of the program (i.e. test/integration-test/src/tests/load.rs).
+// 256 is the maximum length allowed by the kernel, so this test should fail.
+// https://github.com/torvalds/linux/blob/02aee814/net/sched/cls_bpf.c#L28
+#[classifier]
+pub fn aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(
+    _ctx: TcContext,
+) -> i32 {
+    0
+}
+
+#[cfg(not(test))]
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
--- a/test/integration-test/src/lib.rs
+++ b/test/integration-test/src/lib.rs
@@ -22,6 +22,10 @@ pub const BPF_PROBE_READ: &[u8] =
 pub const REDIRECT: &[u8] = include_bytes_aligned!(concat!(env!("OUT_DIR"), "/redirect"));
 pub const XDP_SEC: &[u8] = include_bytes_aligned!(concat!(env!("OUT_DIR"), "/xdp_sec"));
 pub const RING_BUF: &[u8] = include_bytes_aligned!(concat!(env!("OUT_DIR"), "/ring_buf"));
+pub const TC_NAME_LIMIT_TEST: &[u8] =
+    include_bytes_aligned!(concat!(env!("OUT_DIR"), "/tc_name_limit"));
+pub const TC_NAME_LIMIT_EXCEEDED_TEST: &[u8] =
+    include_bytes_aligned!(concat!(env!("OUT_DIR"), "/tc_name_limit_exceeded"));
 
 #[cfg(test)]
 mod tests;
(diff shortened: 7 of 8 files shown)


test/integration-ebpf/src/tc_name_limit.rs
```diff
@@ -0,0 +1,28 @@
+#![no_std]
+#![no_main]
+
+use aya_bpf::{macros::classifier, programs::TcContext};
+
+// This macro generates a function with arbitrary name
+macro_rules! generate_ebpf_function {
```
tamird, line 7: why is this macro needed? is this something forward-looking?

aya/src/sys/netlink.rs
```diff
@@ -277,7 +277,10 @@ struct Request {
 struct TcRequest {
     header: nlmsghdr,
     tc_info: tcmsg,
-    attrs: [u8; 64],
+    // The buffer for attributes should be sized to hold at least 256 bytes,
+    // based on `CLS_BPF_NAME_LEN = 256` from the kernel:
+    // https://github.com/torvalds/linux/blob/02aee814/net/sched/cls_bpf.c#L28
+    attrs: [u8; 512],
```
tamird, line 283: at least 256, but then why 512?

test/integration-test/src/tests/load.rs
```diff
@@ -13,6 +15,27 @@ use aya::{
 const MAX_RETRIES: usize = 100;
 const RETRY_DURATION: time::Duration = time::Duration::from_millis(10);
 
+#[test]
+fn tc_name_limit() {
+    let _ = qdisc_add_clsact("lo");
```
tamird, line 20: why is it ok to drop this value?

test/integration-ebpf/src/tc_name_limit.rs
```diff
@@ -0,0 +1,28 @@
...
+    };
+}
+
+/*
+Generating a function with a 256-byte-long name (all 'a's) to be used as
+the ebpf program. This name must match the name passed to userspace side.
+256 is the maximum length allowed by the kernel:
+https://github.com/torvalds/linux/blob/02aee814/net/sched/cls_bpf.c#L28
+*/
+generate_ebpf_function!(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa);
```
tamird, line 22: what happens if the program name is 257 bytes long?

tamird (changes requested):
Please squash these commits.

aya/src/sys/netlink.rs
dave-tucker: Because we currently use around ~30 bytes of attributes in addition to name.
Rather than picking a "right sized buffer" for the payload (which is of varying length anyway) we use the next largest power of 2.
I've got netlink changes planned soon, so will look at making the buffer size variable then.

dave-tucker:
I'm happy with the change to up the buffer to the next nearest power of 2.
Tamir's comments need addressing before merge.

test/integration-test/src/tests/load.rs
pooladkhay: We need the `clsact` qdisc attached to the interface before we can proceed, Actually a better approach would be a function that checks that so we can call `qdisc_add_clsact("lo")` only when that interface doesn't have that qdisc.
I can add that and update the code...

test/integration-ebpf/src/tc_name_limit.rs
pooladkhay: @tamird Sorry for the delay
This is the error you get when the name is more that 256 bytes:
```
Error: netlink error while attaching ebpf program to tc

Caused by:
```

test/integration-test/src/tests/load.rs
pooladkhay: @tamird I have added a function to check if `clsact` qdisc exists before trying to actually add it.
I'd really appreciate it if you review the code.

aya/src/sys/netlink.rs
```diff
@@ -1,12 +1,19 @@
-use std::{collections::HashMap, ffi::CStr, io, mem, os::fd::RawFd, ptr, slice};
+use std::{
+    collections::HashMap,
+    ffi::CStr,
+    io,
+    mem::{self},
```
tamird, line 5: {self} is redundant

aya/src/sys/netlink.rs
tamird: Can the comment say that please?

test/integration-ebpf/src/tc_name_limit.rs
tamird: But you still write it =/

test/integration-test/src/tests/load.rs
```diff
@@ -13,6 +15,30 @@ use aya::{
...
+#[test]
+fn tc_name_limit() {
+    let clsact_exists = clsact_qdisc_exists("lo").unwrap();
+    if !clsact_exists {
+        qdisc_add_clsact("lo").unwrap();
+    }
+
+    let mut bpf = Bpf::load(crate::TC_NAME_LIMIT_TEST).unwrap();
+
+    // This magic number (256) is derived from the fact that the kernel
```
tamird, line 27: this comment is somewhat misplaced; this 256 is just defined by the fact that this is the name used in the probe code.

test/integration-ebpf/src/tc_name_limit.rs
pooladkhay: I just found it cleaner to pass 256 'a's to a macro than writing a function with that name directly. If that's an issue, I can remove it.

test/integration-ebpf/src/tc_name_limit.rs
pooladkhay: That's a good idea, Macro is probably more useful now.

test/integration-test/src/tests/load.rs
pooladkhay: > Is there a different type of probe we can use that doesn't have this requirement?

By requirement, Do you mean the need to call `qdisc_add_clsact("lo")`?
If yes, I don't think so.
As far as I know we can either use `ingress` or `clsact` if attaching to ingress path and to attach an ebpf program to egress path, the only option is to use `clsact`:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=1f211a1b929c804100e138c5d3d656992cfd5622

aya/src/sys/netlink.rs
pooladkhay: Sure, will update the comment.

alessandrod:
@pooladkhay ping?

pooladkhay:
> @pooladkhay ping?

@alessandrod  pong! 😃 

Life hasn't been easy lately, I should have notified earlier. 
I'll get back to this asap, please let it be open...

alessandrod:
> > @pooladkhay ping?
> 
> @alessandrod pong! 😃
> 
> Life hasn't been easy lately, I should have notified earlier. I'll get back to this asap, please let it be open...

*hugs* ❤️ and sorry if I made you feel rushed it wasn't my intention

pooladkhay:
> > > @pooladkhay ping?
> > 
> > 
> > @alessandrod pong! 😃
> > Life hasn't been easy lately, I should have notified earlier. I'll get back to this asap, please let it be open...
> 
> _hugs_ ❤️ and sorry if I made you feel rushed it wasn't my intention

No worries at all 🙋‍♂️

test/integration-ebpf/src/tc_name_limit.rs
pooladkhay: Macro was replaced by just a function.

aya/src/sys/netlink.rs
pooladkhay: @tamird Updated the comment.

test/integration-ebpf/src/tc_name_limit.rs
pooladkhay: @tamird An integration test was added to make sure programs with a name longer than 256 bytes will fail to attach.

tamird (changes requested):
> Reviewed 1 of 5 files at r3, 1 of 2 files at r4, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)* status: all files reviewed, 7 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5SDXd2pHIUaJ-2D9N):*
can you please squash these commits? I don't understand why 12 commits are needed.
___
*[`aya/src/sys/netlink.rs` line 83 at r5](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5T_RhCOdhjLed1zta:-Nk5T_Ri6rr3HQAA0ocL:bkkapkl) ([raw file](https://github.com/aya-rs/aya/blob/df59ff37299c921a10aa75988b9d49cc3084cde6/aya/src/sys/netlink.rs#L83)):*
> ```XML
> }
> 
> pub(crate) unsafe fn netlink_clsact_qdisc_exists(if_index: i32) -> Result<bool, io::Error> {
> ```

i'm not familiar enough with netlink, @dave-tucker should review this.
___
*[`test/integration-ebpf/src/tc_name_limit.rs` line 6 at r5](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5TmrE6UJrsMZA3WhU:-Nk5TmrE6UJrsMZA3WhV:b-5n03kp) ([raw file](https://github.com/aya-rs/aya/blob/df59ff37299c921a10aa75988b9d49cc3084cde6/test/integration-ebpf/src/tc_name_limit.rs#L6)):*
> ```XML
> use aya_bpf::{macros::classifier, programs::TcContext};
> 
> /*
> ```

I don't think we use this comment style in this project. Can you please use `// ....` style instead?
___
*[`test/integration-ebpf/src/tc_name_limit_exceeded.rs` line 6 at r5](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5TwPQ3Ggkd-dBlurp:-Nk5TwPR42egCcJXGI0X:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/df59ff37299c921a10aa75988b9d49cc3084cde6/test/integration-ebpf/src/tc_name_limit_exceeded.rs#L6)):*
> ```XML
> use aya_bpf::{macros::classifier, programs::TcContext};
> 
> /*
> ```

ditto
___
*[`test/integration-test/src/tests/load.rs` line 27 at r2](https://reviewable.io/reviews/aya-rs/aya/728#gh-1303289865:-Nk5Uquo5LWJ1po0aFLu:b-iyjmnq) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L27)):*
<details><summary><i>Previously, pooladkhay (Mohammad Javad Pooladkhay) wrote…</i></summary><blockquote>

The idea is to ensure that Aya does not impose a limit and allows names up to the size allowed by the kernel.
I'll add this sentence as well to be more specific about the goal.
</blockquote></details>

I understand. I'm just saying that this repeats the comment that is present in the probe code. Here we just need to match the name of the probe. I think this comment can just be removed?
___
*[`test/integration-test/src/tests/load.rs` line 65 at r5](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5Uz8m42q1QEzJhy6T:-Nk5Uz8m42q1QEzJhy6U:b-sr9ot1) ([raw file](https://github.com/aya-rs/aya/blob/df59ff37299c921a10aa75988b9d49cc3084cde6/test/integration-test/src/tests/load.rs#L65)):*
> ```XML
>     let mut bpf = Bpf::load(crate::TC_NAME_LIMIT_EXCEEDED_TEST).unwrap();
> 
>     /*
> ```

ditto about this comment
___
*[`test/integration-test/src/tests/load.rs` line 87 at r5](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5U80HBO9Cjl16Nl5l:-Nk5U80HBO9Cjl16Nl5m:bh5nw8c) ([raw file](https://github.com/aya-rs/aya/blob/df59ff37299c921a10aa75988b9d49cc3084cde6/test/integration-test/src/tests/load.rs#L87)):*
> ```XML
>     // as a result, all arms except the desired Err() should be unreachable!().
>     match program.attach("lo", TcAttachType::Ingress) {
>         Ok(_) => unreachable!(),
> ```

unreachable? this is just a test failure, isn't it? can you instead write:

```rust
assert_matches!(
  program.attach("lo", TcAttachType::Ingress),
  Err(ProgramError::TcError(TcError::NetlinkError { io_error })) => {
    assert_eq!(io_error.raw_or_error(), Some(22))
  }
);
```

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)* status: all files reviewed, 8 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)

___
*[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-Nk5VSAG5JV1YxGCDNOX:bm8du96) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):*
> ```XML
> 
> #[test]
> fn tc_name_limit() {
> ```

Looks like these tests are failing:
```
failures:

---- tests::load::tc_name_limit stdout ----
thread 'tests::load::tc_name_limit' panicked at test/integration-test/src/tests/load.rs:31:32:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }

---- tests::load::tc_name_limit_exceeded stdout ----
thread 'tests::load::tc_name_limit_exceeded' panicked at test/integration-test/src/tests/load.rs:60:32:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

It's the `qdisc_add_clsact("lo").unwrap();` calls that seem to be failing.

pooladkhay:
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)_ status: all files reviewed, 8 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)
> 
> _[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-Nk5VSAG5JV1YxGCDNOX:bm8du96) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):_
> 
> > ```
> > #[test]
> > fn tc_name_limit() {
> > ```
> 
> Looks like these tests are failing:
> 
> ```
> failures:
> 
> ---- tests::load::tc_name_limit stdout ----
> thread 'tests::load::tc_name_limit' panicked at test/integration-test/src/tests/load.rs:31:32:
> called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> ---- tests::load::tc_name_limit_exceeded stdout ----
> thread 'tests::load::tc_name_limit_exceeded' panicked at test/integration-test/src/tests/load.rs:60:32:
> called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> ```
> 
> It's the `qdisc_add_clsact("lo").unwrap();` calls that seem to be failing.


That's true,
Two commands are being used to run the tests:
`cargo xtask integration-test local` which succeeds, and `cargo xtask integration-test vm <KERNEL_IMAGE>` which is the faulty one. Running it on my local machine fails too.

Is it possible that the loop back interface which I'm trying to attach an eBPF program to, is not available on that image?

tamird (changes requested):
> Reviewed 7 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)

___
*[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-Nkb3eon-VFFIX5_pNvr:brgkats) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):*
<details><summary><i>Previously, pooladkhay (Mohammad Javad Pooladkhay) wrote…</i></summary><blockquote>

That's true,
Two commands are being used to run the tests:
`cargo xtask integration-test local` which succeeds, and `cargo xtask integration-test vm <KERNEL_IMAGE>` which is the faulty one. Running it on my local machine fails too.

Is it possible that the loop back interface which I'm trying to attach an eBPF program to, is not available on that image? (hence the `Os { code: 2, kind: NotFound, message: "No such file or directory" }` error)

</blockquote></details>

Yep, seems likely that it is not available. Can we create a network namespace so that we aren't reliant on ambient resources?

pooladkhay:
> > Reviewed 7 of 7 files at r6, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)_ status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)
> 
> _[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-Nkb3eon-VFFIX5_pNvr:brgkats) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):_
> Previously, pooladkhay (Mohammad Javad Pooladkhay) wrote…
> 
> Yep, seems likely that it is not available. Can we create a network namespace so that we aren't reliant on ambient resources?

Could you please give me more hints, and maybe an example?

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)

___
*[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-Nl0BXrm5VzWznqJ46jD:bvnvg11) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):*
<details><summary><i>Previously, pooladkhay (Mohammad Javad Pooladkhay) wrote…</i></summary><blockquote>

Could you please give me more hints, and maybe an example?
</blockquote></details>

See https://github.com/aya-rs/aya/blob/b1769678f48f7abf6c987a1d686bbaffd5d1e664/test/integration-test/src/tests/xdp.rs#L137

pooladkhay:
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)_ status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)
> 
> _[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-Nl0BXrm5VzWznqJ46jD:bvnvg11) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):_
> Previously, pooladkhay (Mohammad Javad Pooladkhay) wrote…
> 
> > Could you please give me more hints, and maybe an example?
> 
> See
> 
> https://github.com/aya-rs/aya/blob/b1769678f48f7abf6c987a1d686bbaffd5d1e664/test/integration-test/src/tests/xdp.rs#L137

I tried that. Namespace is created but the test still fails with the same error:
```
test tests::load::tc_name_limit ... Entered network namespace aya-test-168-0
thread 'tests::load::tc_name_limit' panicked at test/integration-test/src/tests/load.rs:623:32:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
Exited network namespace aya-test-168-0
FAILED
```
I wonder what is the purpose of this step in testing? 
What is the main difference between `cargo xtask integration-test vm` and `cargo xtask integration-test local` if the tests are running on a linux machine?
Is this related to kernel versions? maybe being able to test on a specific version?

I took a look and the CI file and realised `runs-on: ubuntu-22.04` is set [at the top](https://github.com/aya-rs/aya/blob/41351ec56068da73c002486b60baf404ce8809a9/.github/workflows/ci.yml#L20) but there are also checks like [this one](https://github.com/aya-rs/aya/blob/41351ec56068da73c002486b60baf404ce8809a9/.github/workflows/ci.yml#L219C9-L219C33): `if: runner.os == 'macOS'`.
Considering those checks checking `runner.os`, maybe  `cargo xtask integration-test vm` is used in a scenario where the runner is not a linux image? 
If that's the case, I'm unable to make sense of it because of `runs-on: ubuntu-22.04`.

Or maybe I'm missing something here...

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/728)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @pooladkhay)

___
*[`test/integration-test/src/tests/load.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/728#-Nk5VSAG5JV1YxGCDNOW:-NlOqcRs0G81TiwR6hut:bv7h51z) ([raw file](https://github.com/aya-rs/aya/blob/5d608c5485f3cd3396deacf53c8a8f4062a03374/test/integration-test/src/tests/load.rs#L19)):*
<details><summary><i>Previously, pooladkhay (Mohammad Javad Pooladkhay) wrote…</i></summary><blockquote>

I tried that. Namespace is created but the test still fails with the same error:
```
test tests::load::tc_name_limit ... Entered network namespace aya-test-168-0
thread 'tests::load::tc_name_limit' panicked at test/integration-test/src/tests/load.rs:623:32:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
Exited network namespace aya-test-168-0
FAILED
```
I wonder what is the purpose of this step in testing? 
What is the main difference between `cargo xtask integration-test vm` and `cargo xtask integration-test local` if the tests are running on a linux machine?
Is this related to kernel versions? maybe being able to test on a specific version?

I took a look and the CI file and realised `runs-on: ubuntu-22.04` is set [at the top](https://github.com/aya-rs/aya/blob/41351ec56068da73c002486b60baf404ce8809a9/.github/workflows/ci.yml#L20) but there are also checks like [this one](https://github.com/aya-rs/aya/blob/41351ec56068da73c002486b60baf404ce8809a9/.github/workflows/ci.yml#L219C9-L219C33): `if: runner.os == 'macOS'`.
Considering those checks checking `runner.os`, maybe  `cargo xtask integration-test vm` is used in a scenario where the runner is not a linux image? 
If that's the case, I'm unable to make sense of it because of `runs-on: ubuntu-22.04`.

Or maybe I'm missing something here...

</blockquote></details>

Yes, this is both for testing on non-Linux and for future testing on different kernel versions.

As for the error you're seeing, I don't know how to help you. It seems we explicitly set up a loopback device in the network namespace: https://github.com/aya-rs/aya/blob/41351ec56068da73c002486b60baf404ce8809a9/test/integration-test/src/utils.rs#L41-L53

dave-tucker:
I'm going through old PRs - sorry for such a long delay on this one.
I've rebased it on #690 which gives more of a clue of why this fails:

```
---- tests::load::tc_name_limit stdout ----
thread 'tests::load::tc_name_limit' panicked at test/integration-test/src/tests/load.rs:587:32:
called `Result::unwrap()` on an `Err` value: NetlinkError(NetlinkError(Error { message: "Specified qdisc kind is unknown", s)
```

Root cause was tricky to track down...
In our kernel config, `CONFIG_NET_SCH_INGRESS=m` - you can follow along as it's the case on most distros.
If you do `lsmod | grep sch_ingress` you'll see it's not loaded.
Then you can do `sudo tc qdisc add dev lo clsact`, run `lsmod | grep sch_ingress` and see the module is magically loaded!

How you ask? Kernel module auto-loading is a thing (via calls to `request_module()` in the kernel code.
It seems like for this to work in our test VMs we'll have to bundle `/sbin/modprobe` (or something binary compatible) into our VM. I'll take a look at doing that and then we can revisit this PR

pooladkhay:
> I'm going through old PRs - sorry for such a long delay on this one. I've rebased it on #690 which gives more of a clue of why this fails:
> 
> ```
> ---- tests::load::tc_name_limit stdout ----
> thread 'tests::load::tc_name_limit' panicked at test/integration-test/src/tests/load.rs:587:32:
> called `Result::unwrap()` on an `Err` value: NetlinkError(NetlinkError(Error { message: "Specified qdisc kind is unknown", s)
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Root cause was tricky to track down... In our kernel config, `CONFIG_NET_SCH_INGRESS=m` - you can follow along as it's the case on most distros. If you do `lsmod | grep sch_ingress` you'll see it's not loaded. Then you can do `sudo tc qdisc add dev lo clsact`, run `lsmod | grep sch_ingress` and see the module is magically loaded!
> 
> How you ask? Kernel module auto-loading is a thing (via calls to `request_module()` in the kernel code. It seems like for this to work in our test VMs we'll have to bundle `/sbin/modprobe` (or something binary compatible) into our VM. I'll take a look at doing that and then we can revisit this PR

Apparently there is no longer a limit on the name of the classifier. I tried a name larger than 8 KB (8280 "a"s) which was successfully attached. I guess we can safely close this issue.

I still see that [CLS_BPF_NAME_LEN](https://github.com/torvalds/linux/blob/66701750d5565c574af42bef0b789ce0203e3071/net/sched/cls_bpf.c#L28) is 256. Curious to see how that's made possible. Any thoughts?

PR #729, opened 2023-08-08
log: annotate logging functions inlining

tamird:
Some of these functions fail to compile when not inlined, so we should
be explicit.

Before deciding on this approach I tried various ways of making all
these functions #[inline(never)] to save instructions but I ran into
blockers:
- These functions currently return Result, which is a structure. This is
  not permitted in BPF.
- I tried inventing a newtype that is a #[repr(transparent)] wrapper of
  u16, and having these functions return that; however it seems that
  even if the object code is legal, the verifier will reject such
  functions because the BTF (if present, and it was in my local
  experiments) would indicate that the return is a structure.
- I tried having these functions return a plain u16 where 0 means error,
  but the verifier still rejected the BTF because the receiver (even if
  made into &self) is considered a structure, and forbidden.

We can eventually overcome these problems by "lying" in our BTF once
support for it matures in the bpf-linker repo (e.g. Option<NonZeroU16>
should be perfectly legal as it is guaranteed to be word-sized), but we
aren't there yet, and this is the safest thing we can do for now.

### DIFF
--- a/aya-log-common/src/lib.rs
+++ b/aya-log-common/src/lib.rs
@@ -147,6 +147,10 @@ pub enum DisplayHint {
     UpperMac,
 }
 
+// Must be inlined, else the BPF backend emits:
+//
+// llvm: <unknown>:0:0: in function _ZN14aya_log_common5write17hc9ed05433e23a663E { i64, i64 } (i8, ptr, i64, ptr, i64): only integer returns supported
+#[inline(always)]
 pub(crate) fn write(tag: u8, value: &[u8], buf: &mut [u8]) -> Result<usize, ()> {
     let wire_len: LogValueLength = value
         .len()
@@ -175,6 +179,9 @@ pub trait WriteToBuf {
 macro_rules! impl_write_to_buf {
     ($type:ident, $arg_type:expr) => {
         impl WriteToBuf for $type {
+            // This need not be inlined because the return value is Result<N, ()> where N is
+            // mem::size_of<$type>, which is a compile-time constant.
+            #[inline(never)]
             fn write(self, buf: &mut [u8]) -> Result<usize, ()> {
                 write($arg_type.into(), &self.to_ne_bytes(), buf)
             }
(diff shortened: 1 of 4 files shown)


aya-log-common/src/lib.rs
```diff
@@ -147,53 +147,20 @@ pub enum DisplayHint {
...
-impl<T, V> TagLenValue<T, V> {
-    #[inline(always)]
-    pub(crate) fn new(tag: T, value: V) -> TagLenValue<T, V> {
-        TagLenValue { tag, value }
+pub(crate) fn write(tag: u8, value: &[u8], buf: &mut [u8]) -> Result<usize, ()> {
+    let wire_len: LogValueLength = value
+        .len()
+        .try_into()
+        .map_err(|num::TryFromIntError { .. }| ())?;
+    let values: &[&[u8]] = &[&[tag], &wire_len.to_ne_bytes(), value];
```
ajwerner, line 159: We shall pray the loop gets inlined and that the stack space requirements here in practice are less than the unoptimized code might imply. In practice, i suspect it does

ajwerner (commented):
I don't find myself understanding how some of these need to be inline(always) and some can be inline(never) despite having the same signature. It feels like either they should all be inline(always) or we should continue to throw up our hands and pray to the compiler gods like we do all over the place. 

The first commit is cleaner, but leaves more up to the compiler to inline and unroll, and if it doesn't you're in more trouble. There's probably an intermediate formulation that does the unrolling. I'm sorry to say that I find myself mostly feeling shrug about this whole endeavor. What is this PR achieving other than using more idiomatic rust? Also, it feels like the `inline(never)` additions may break somebody's working code that we just don't have tests that use similar behavior.

tamird:
> I don't find myself understanding how some of these need to be inline(always) and some can be inline(never) despite having the same signature. It feels like either they should all be inline(always) or we should continue to throw up our hands and pray to the compiler gods like we do all over the place.

I agree, they should all be inline(always); I need to beef up the tests to expose the conditions in which all these functions' return values are actually used.

> The first commit is cleaner, but leaves more up to the compiler to inline and unroll, and if it doesn't you're in more trouble. There's probably an intermediate formulation that does the unrolling.

Indeed. We can unroll it by hand if you prefer.

> I'm sorry to say that I find myself mostly feeling shrug about this whole endeavor. What is this PR achieving other than using more idiomatic rust? Also, it feels like the `inline(never)` additions may break somebody's working code that we just don't have tests that use similar behavior.

Recall that our initial attempts to use the new LLVM pass manager in bpf-linker exposed cases wherein *not* inlining some functions would produce miscompilation, leading us to discover that bpf-linker did not correctly handle diagnostics from LLVM. In other words: we can't leave these functions unannotated because changes in their inlining behavior can break compilation.

I agree that these functions should all end up inline(never), I just need to write the tests to prove it.

ajwerner (approved):
Thanks, this seems more reasonable. I still don't follow why have `inline(never)` for the cases that don't have to be inlined. It seems like it's still probably good if the compiler does inline them when it deems that it ought to. I feel like the comment about why it doesn't need `inline(always)` is good enough.

tamird:
> I still don't follow why have `inline(never)` for the cases that don't have to be inlined. It seems like it's still probably good if the compiler does inline them when it deems that it ought to. I feel like the comment about why it doesn't need `inline(always)` is good enough.

Yeah, it's a question. We can leave it up to the compiler -- the annotation is a stronger statement about the fact that these functions *can* be not-inline and still generate valid code, but it's not required.

Note that when BTF support arrives, these annotations will immediately expose the problems I described earlier in this PR and in the commit message, but if we remove the annotations then the problems won't become apparent until an unlucky user compiles a program where the compiler decided not to inline.

ajwerner (approved):
Ack, a possible approach here is to create a feature that the integration test uses that just enables that annotation, then I'd expect there to be no controversy. However, I'm cool with this either way. LGTM

PR #731, opened 2023-08-09
aya-obj: Mutate BTF in-place without clone

dave-tucker:
The BTF we're working on is Cow anyway so modifying in-place is fine. All we need to do is store some information before we start our mutable iteration to avoid concurrently borrowing types both mutably and immutably.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/731)

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -475,7 +475,7 @@ impl Btf {
     ) -> Result<(), BtfError> {
         let mut types = mem::take(&mut self.types);
         for i in 0..types.types.len() {
-            let t = &types.types[i];
+            let t = &mut types.types[i];
             let kind = t.kind();
             match t {
                 // Fixup PTR for Rust
@@ -483,9 +483,7 @@ impl Btf {
                 // While I figure out if this needs fixing in the Kernel or LLVM, we'll
                 // do a fixup here
                 BtfType::Ptr(ptr) => {
-                    let mut fixed_ty = ptr.clone();
-                    fixed_ty.name_offset = 0;
-                    types.types[i] = BtfType::Ptr(fixed_ty)
+                    ptr.name_offset = 0;
                 }
                 // Sanitize VAR if they are not supported
                 BtfType::Var(v) if !features.btf_datasec => {
@@ -496,7 +494,7 @@ impl Btf {
                     debug!("{}: not supported. replacing with STRUCT", kind);
 
                     // STRUCT aren't allowed to have "." in their name, fixup this if needed.
-                    let mut name_offset = t.name_offset();
+                    let mut name_offset = d.name_offset;
                     let name = self.string_at(name_offset)?;
 
                     // Handle any "." characters in struct names
@@ -506,18 +504,22 @@ impl Btf {
                         name_offset = self.add_string(&fixed_name);
                     }
 
-                    let mut members = vec![];
-                    for member in d.entries.iter() {
-                        let mt = types.type_by_id(member.btf_type).unwrap();
-                        members.push(BtfMember {
-                            name_offset: mt.name_offset(),
-                            btf_type: member.btf_type,
-                            offset: member.offset * 8,
+                    let entries = std::mem::take(&mut d.entries);
+
+                    let members = entries
+                        .iter()
+                        .map(|e| {
+                            let mt = types.type_by_id(e.btf_type).unwrap();
+                            BtfMember {
+                                name_offset: mt.name_offset(),
+                                btf_type: e.btf_type,
+                                offset: e.offset * 8,
+                            }
                         })
-                    }
+                        .collect();
 
                     types.types[i] =
-                        BtfType::Struct(Struct::new(name_offset, members, d.entries.len() as u32));
+                        BtfType::Struct(Struct::new(name_offset, members, entries.len() as u32));
                 }
                 // Fixup DATASEC
                 // DATASEC sizes aren't always set by LLVM
@@ -527,18 +529,16 @@ impl Btf {
                     let name = self.string_at(d.name_offset)?;
                     let name = name.into_owned();
 
-                    let mut fixed_ty = d.clone();
-
                     // Handle any "/" characters in section names
                     // Example: "maps/hashmap"
                     let fixed_name = name.replace('/', ".");
                     if fixed_name != name {
-                        fixed_ty.name_offset = self.add_string(&fixed_name);
+                        d.name_offset = self.add_string(&fixed_name);
                     }
 
                     // There are some cases when the compiler does indeed populate the
                     // size
-                    if t.size().unwrap() > 0 {
+                    if d.size > 0 {
                         debug!("{} {}: size fixup not required", kind, name);
                     } else {
                         // We need to get the size of the section from the ELF file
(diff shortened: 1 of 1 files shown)


tamird (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/731)* status: all files reviewed, 5 unresolved discussions (waiting on @dave-tucker)

___
*[`aya-obj/src/btf/btf.rs` line 531 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPKEc07Njdl2RYwcSs:-NbPKEc07Njdl2RYwcSt:b-l58qdj) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/btf.rs#L531)):*
> ```XML
> 
>                     let mut members = vec![];
>                     for member in d.entries.iter() {
> ```

so i think we can avoid the need for datasec_member_name_offset here if we realize that we're about to throw away `d`, so we can write:

let entries = std::mem::take(d.entries);
// now we've dropped the mutable borrow of types.types, i hope
//
// populate members as before
types.types[i] = ...
___
*[`aya-obj/src/btf/btf.rs` line 584 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPLhQ67-l8ILpU9l52:-NbPLhQ67-l8ILpU9l53:bapvjm9) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/btf.rs#L584)):*
> ```XML
>                         for e in &mut d.entries.iter_mut() {
>                             let (name_offset, linkage) =
>                                 var_types.get(&e.btf_type).ok_or(BtfError::InvalidDatasec)?;
> ```

i think we can do similar tricks here; hollow out `d` instead of cloning it, which drops the mutable reference, and then write a new one. that avoids `var_types` which is also allocates.
___
*[`aya-obj/src/btf/btf.rs` line 1256 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPJ0WB0APr5F48SZ3b:-NbPJ0WB0APr5F48SZ3c:b-4l5wky) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/btf.rs#L1256)):*
> ```XML
>         )));
> 
>         let name_offset = btf.add_string("foo");
> ```

should we extract local for this "foo", or perhaps assert that name_offset1 is equal to this name_offset below?
___
*[`aya-obj/src/btf/types.rs` line 868 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPIq_s8ciRy9bOfPEt:-NbPIq_s8ciRy9bOfPEu:bmdm10x) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/types.rs#L868)):*
> ```XML
> 
> #[repr(C)]
> #[derive(Clone, Debug)]
> ```

do we want to remove these clone impls?
___
*[`aya-obj/src/btf/types.rs` line 877 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPKsDlC4iyBqJIAS5Q:-NbPKsDlC4iyBqJIAS5R:b1rt77x) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/types.rs#L877)):*
> ```XML
> 
> impl DataSec {
>     pub fn name_offset(&self) -> u32 {
> ```

why are these getters needed? the fields are pub(crate).

dave-tucker (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/731)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @tamird)

___
*[`aya-obj/src/btf/btf.rs` line 531 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPKEc07Njdl2RYwcSs:-NbPXAgn6GvQO1Mm-mjP:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/btf.rs#L531)):*

so i think we can avoid the need for datasec_member_name_offset here if we realize that we're about to throw away `d`, so we can write:

let entries = std::mem::take(d.entries);
// now we've dropped the mutable borrow of types.types, i hope
//
// populate members as before
types.types[i] = ...
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 584 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPLhQ67-l8ILpU9l52:-NbPXBRsE7jchF2H34x-:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/btf.rs#L584)):*

i think we can do similar tricks here; hollow out `d` instead of cloning it, which drops the mutable reference, and then write a new one. that avoids `var_types` which is also allocates.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 1256 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPJ0WB0APr5F48SZ3b:-NbPXBsR91xRYpiBsqgE:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/btf.rs#L1256)):*

should we extract local for this "foo", or perhaps assert that name_offset1 is equal to this name_offset below?
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/types.rs` line 868 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPIq_s8ciRy9bOfPEt:-NbPXDXd6P7_tq8wdCtY:byx2p8j) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/types.rs#L868)):*

do we want to remove these clone impls?
</blockquote></details>

We can't because Btf is Clone on write. So all everything contained within also needs to be clone since we derive the clone impl.
___
*[`aya-obj/src/btf/types.rs` line 877 at r1](https://reviewable.io/reviews/aya-rs/aya/731#-NbPKsDlC4iyBqJIAS5Q:-NbPXD3lBvLxGbQg2MHg:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e67cd2bfaf55a9e85eeb4175f8e8f3e8e4405d14/aya-obj/src/btf/types.rs#L877)):*

why are these getters needed? the fields are pub(crate).
</blockquote></details>

Done.

tamird (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/731#-:-NbPYjh7EghBUtfGLy7h:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @dave-tucker)

___
*[`aya-obj/src/btf/btf.rs` line 511 at r2](https://reviewable.io/reviews/aya-rs/aya/731#-NbPZ0Ns89opoKvoqiZ4:-NbPZ0Ns89opoKvoqiZ5:b660ry1) ([raw file](https://github.com/aya-rs/aya/blob/098d4364bd0fb8551f0515cb84afda6aff23ed7f/aya-obj/src/btf/btf.rs#L511)):*
> ```XML
>                     let members = entries
>                         .iter()
>                         .map(|e| {
> ```

Can "e" be "member"? Reduces the diff.
___
*[`aya-obj/src/btf/btf.rs` line 562 at r2](https://reviewable.io/reviews/aya-rs/aya/731#-NbPZTKq2sW05O1wVkDW:-NbPZTKq2sW05O1wVkDX:b-sctmij) ([raw file](https://github.com/aya-rs/aya/blob/098d4364bd0fb8551f0515cb84afda6aff23ed7f/aya-obj/src/btf/btf.rs#L562)):*
> ```XML
>                         // we can query by name in symbol_offsets
>                         let mut entries = mem::take(&mut d.entries);
>                         let mut fixed_section = d.clone();
> ```

Is this needed because there's a private field?
___
*[`aya-obj/src/btf/btf.rs` line 564 at r2](https://reviewable.io/reviews/aya-rs/aya/731#-NbPZIHGAAtowImv-aSf:-NbPZIHGAAtowImv-aSg:b2lp6cw) ([raw file](https://github.com/aya-rs/aya/blob/098d4364bd0fb8551f0515cb84afda6aff23ed7f/aya-obj/src/btf/btf.rs#L564)):*
> ```XML
>                         let mut fixed_section = d.clone();
> 
>                         for e in entries.iter_mut() {
> ```

s/e/d/ to match old name and reduce diff

dave-tucker (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/731)* status: all files reviewed, 3 unresolved discussions (waiting on @tamird)

___
*[`aya-obj/src/btf/btf.rs` line 562 at r2](https://reviewable.io/reviews/aya-rs/aya/731#-NbPZTKq2sW05O1wVkDW:-NbPjM0D7QuWkFXpFDoL:b-kda0yw) ([raw file](https://github.com/aya-rs/aya/blob/098d4364bd0fb8551f0515cb84afda6aff23ed7f/aya-obj/src/btf/btf.rs#L562)):*

Is this needed because there's a private field?
</blockquote></details>

Yes. But at least we've pulled out the Vec now.

PR #733, opened 2023-08-09
integration-test: Implement running on VMs

tamird:
See individual commits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/733)


ajwerner (commented):
My biggest feedback about this PR at this point is that there's a lot of context in your brain about how these pieces fit together that aren't really apparent without unpacking the structure of the code.

In particular, a doc comment on the init process would be good. Also good would be a comment in `xtask::run` with a high-level overview of what's going on.

Thanks for making this all work on the mac! Very handy

xtask/src/run.rs
```diff
@@ -83,40 +102,383 @@ pub fn build(opts: BuildOptions) -> Result<Vec<(String, PathBuf)>> {
...
+                let mut qemu_child = qemu
+                    .stdout(Stdio::piped())
+                    .spawn()
+                    .with_context(|| format!("failed to spawn {qemu:?}"))?;
+                let Child { stdout, .. } = &mut qemu_child;
+                let stdout = stdout.take().unwrap();
+                let stdout = BufReader::new(stdout);
+
+                let mut outcome = None;
+                for line in stdout.lines() {
```
dave-tucker, line 458: Per previous comment re: stderr, I would hope that stderr would be empty in the success cases.

xtask/src/run.rs
```diff
@@ -83,40 +102,383 @@ pub fn build(opts: BuildOptions) -> Result<Vec<(String, PathBuf)>> {
...
+                    let line =
+                        line.with_context(|| format!("failed to read line from {qemu:?}"))?;
+                    println!("{}", line);
+                    if let Some(line) = line.strip_prefix("init: ") {
+                        let previous = match line {
```
dave-tucker, line 463: Some commenting around the expected outputs here might be helpful

xtask/src/run.rs
```diff
@@ -83,40 +102,383 @@ pub fn build(opts: BuildOptions) -> Result<Vec<(String, PathBuf)>> {
...
+                    .arg("-initrd")
+                    .arg(&initrd_image);
+                if guest_arch == "aarch64" {
+                    match OS {
+                        "linux" => {
+                            let mut cmd = Command::new("locate");
+                            let output = cmd
+                                .arg("QEMU_EFI.fd")
+                                .output()
+                                .with_context(|| format!("failed to run {cmd:?}"))?;
```
dave-tucker, line 417: Will the kernel not boot without UEFI?
It would be a lot simpler in both cases (x86/aarch64) to live without it.

xtask/src/run.rs
```diff
@@ -83,40 +102,383 @@ pub fn build(opts: BuildOptions) -> Result<Vec<(String, PathBuf)>> {
...
+                if status.code() != Some(0) {
+                    bail!("{cmd:?} failed: {output:?}")
+                }
+                let Output { stdout, .. } = output;
+
+                // Now parse the output of the file command, which looks something like
+                //
+                // - Linux kernel ARM64 boot executable Image, little-endian, 4K pages
+                //
+                // - Linux kernel x86 boot executable bzImage, version 6.1.0-10-cloud-amd64 [..]
```
dave-tucker, line 259: This works, but I find it hard to believe we have to call `file` to get this info.
I'm sure there's another way, but it's not hit me just yet

xtask/src/run.rs
```diff
@@ -83,40 +102,383 @@ pub fn build(opts: BuildOptions) -> Result<Vec<(String, PathBuf)>> {
...
-
-        println!("{name} running {cmd:?}");
-
-        let status = cmd
-            .status()
-            .with_context(|| format!("failed to run {cmd:?}"))?;
-        if status.code() != Some(0) {
-            writeln!(&mut failures, "{name} failed: {status:?}").context("String write failed")?
-        }
+    #[allow(clippy::type_complexity)]
```
dave-tucker, line 130: Some context on what this is suppressing would be useful.

xtask/src/run.rs
```diff
@@ -1,47 +1,66 @@
...
-#[derive(Debug, Parser)]
+#[derive(Parser)]
 pub struct Options {
-    #[command(flatten)]
-    pub build_options: BuildOptions,
-    /// The command used to wrap your application.
-    #[clap(short, long, default_value = "sudo -E")]
-    pub runner: String,
+    #[clap(subcommand)]
+    pub environment: Environment,
```
dave-tucker, line 46: Missing doc comment?

Cargo.toml
```diff
@@ -72,6 +74,7 @@ lazy_static = { version = "1", default-features = false }
 libc = { version = "0.2.105", default-features = false }
 log = { version = "0.4", default-features = false }
 netns-rs = { version = "0.1", default-features = false }
+nix = { version = "0.26.2", default-features = false }
```
dave-tucker, line 77: I for one am very pro `nix` and have 0 objections to using it in `xtask`.
ISTR in any we only used `libc` and I don't think we need to be as strict, but flagging it anyway.

xtask/src/run.rs
```diff
@@ -1,47 +1,66 @@
...
+        /// wget --accept-regex '.*/linux-image-[0-9\.-]+-cloud-.*-unsigned*' \
+        ///   --recursive ftp://ftp.us.debian.org/debian/pool/main/l/linux/
+        ///
+        /// You can then extract them with:
+        ///
+        /// find . -name '*.deb' -print0 \
+        /// | xargs -0 -I {} sh -c "dpkg --fsys-tarfile {} \
+        ///   | tar --wildcards --extract '*vmlinuz*' --file -"
+        #[clap(required = true)]
+        kernel_image: Vec<PathBuf>,
```
dave-tucker, line 39: Is it useful to pass > 1 kernel? In CI I'd assume we'd use a matrix job.
For local testing, I'm not sure...

tamird (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/733#-:-NbQIP6D9y8xNf-sB-zv:b-896fix)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @dave-tucker)

xtask/src/run.rs
dave-tucker: Thinking about this again, do we want/need to support a different guest arch to the host arch?

dave-tucker (changes requested):
This is looking great! Pure Rust 🦀 and few external deps.
Thanks for doing this - looks like it will be pretty easy to build upon.

A couple of nits and things that could use some commenting for coherence.

init/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+            flags: nix::mount::MsFlags::empty(),
+            data: None,
+            target_mode: None,
+        },
+    ] {
+        match target_mode {
+            None => {
+                // Must exist.
+                let nix::sys::stat::FileStat { st_mode, .. } = nix::sys::stat::stat(target)
+                    .with_context(|| format!("stat({target}) failed"))?;
```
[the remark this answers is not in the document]
tamird: Why? It's literally a function call.

init/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+        .split_whitespace()
+        .filter_map(|parameter| {
+            parameter
+                .strip_prefix("init.arg=")
+                .map(std::ffi::OsString::from)
+        })
+        .collect::<Vec<_>>();
+
+    // Iterate files in /bin.
+    let read_dir = std::fs::read_dir("/bin").context("read_dir(/bin) failed")?;
```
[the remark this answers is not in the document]
tamird: You get a summary at the end:
```
/bin/release-integration-test failed: ExitStatus(unix_wait_status(25856))
/bin/dev-integration-test failed: ExitStatus(unix_wait_status(25856))
```
So I don't think it matters?

init/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+fn main() {
+    match run() {
+        Ok(()) => {
+            println!("init: success");
+        }
+        Err(err) => {
+            println!("{err:?}");
+            println!("init: failure");
```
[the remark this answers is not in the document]
tamird: I don't think this buys us anything other than having to potentially deal with multiple streams from xtask.

init/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+    let _: std::convert::Infallible =
+        nix::sys::reboot::reboot(nix::sys::reboot::RebootMode::RB_POWER_OFF)
+            .expect("reboot failed");
```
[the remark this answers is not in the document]
tamird: Clarified.

ajwerner (approved):
Yes, the comments please. Thank you. 

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/733#-:-NbQToaV4DetafHKqFzA:b96hh08)* status: 0 of 11 files reviewed, 12 unresolved discussions (waiting on @dave-tucker and @tamird)

xtask/src/run.rs
tamird: Yes, it will not boot. Black screen and no progress.

xtask/src/run.rs
tamird: I don't think this is a good idea. I'd rather look for expected output in both cases than risk introducing false negatives because stderr came out empty.

tamird (commented):
Thanks for the review!

init/src/main.rs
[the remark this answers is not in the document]
dave-tucker: I missed the parens when reviewing. I see them now.

xtask/src/run.rs
dave-tucker: If it's useful we keep it.

xtask/src/run.rs
dave-tucker: It's worth calling out that `init: success` and `init: failure` are hardcoded outputs from `init` -  a crate under out control. If they change, they must change here too.

`if let Some(previous) = previous {` seems to be an odd condition to reach.
Under which conditions could we set outcome more than once?

tamird: Added callout. It's an unexpected condition we'd only reach if we had a bug.

Cargo.toml
tamird: Ack.

dave-tucker (approved):
LGTM on green ✅

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/733)* status: 0 of 11 files reviewed, 12 unresolved discussions (waiting on @dave-tucker)

*[a discussion](https://reviewable.io/reviews/aya-rs/aya/733#gh-1288879379:-NbQokJS5BhZN_OPGk3R:b-896fix) (no related file):*

I missed the parens when reviewing. I see them now.
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/733#gh-1288937130:-NbQolKq6RrNtGCY6OuC:b-896fix) (no related file):*

If it's useful we keep it.
</blockquote></details>

Done.

___

PR #735, opened 2023-08-09
aya-log: s/Result<usize, ()>/Option<NonZeroUsize>/

tamird:
`Option<NonZeroUsize>` is guaranteed to have the same size as `usize`, which
is not guaranteed for `Result`. This is a minor optimization, but also
results in simpler code.


ajwerner (approved):
LGTM

PR #737, opened 2023-08-09
github: run integration tests on local kernel

tamird:
This tests on 5.15.0-1041-azure.


tamird:
This passes now that GH runner are on 6.2.

PR #742, opened 2023-08-10
programs: avoid assumptions that paths are UTF-8

tamird:


tamird:
@addisoncrump care to review this as well?

addisoncrump:
Sure, I'll take a peek in a bit.

ajwerner (approved):
Seems reasonable to me.x

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742#-:-NbW6HOx5-C6cUKiTdWS:b-z6ojqn)* status: all files reviewed, 1 unresolved discussion (waiting on @dave-tucker and @tamird)

___
*[`aya/src/programs/uprobe.rs` line 295 at r1](https://reviewable.io/reviews/aya-rs/aya/742#-NbW4BZ6Chn_bSbZW_bc:-NbW4BZ6Chn_bSbZW_bd:b-rfnr8b) ([raw file](https://github.com/aya-rs/aya/blob/f061a97007b7c468b5db128401db3ab845e8a1cf/aya/src/programs/uprobe.rs#L295)):*
> ```XML
>         .filter_map(|mut line| {
>             loop {
>                 if let [stripped @ .., c] = line {
> ```

Could be the following, but as you noted, it'll have a bounds check. 
```rust
    let line = line
        .iter()
        .rev()
        .position(|&b| !b.is_ascii_whitespace())
        .and_then(|pos| line.get(..line.len() - pos))
        .unwrap_or(line);
```

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 3 of 9 files reviewed, all discussions resolved (waiting on @ajwerner and @dave-tucker)

___
*[`aya/src/programs/uprobe.rs` line 295 at r1](https://reviewable.io/reviews/aya-rs/aya/742#-NbW4BZ6Chn_bSbZW_bc:-NbW6dwCEo1XN-35A4z8:b-rrlw62) ([raw file](https://github.com/aya-rs/aya/blob/f061a97007b7c468b5db128401db3ab845e8a1cf/aya/src/programs/uprobe.rs#L295)):*

Could be the following, but as you noted, it'll have a bounds check. 
```rust
    let line = line
        .iter()
        .rev()
        .position(|&b| !b.is_ascii_whitespace())
        .and_then(|pos| line.get(..line.len() - pos))
        .unwrap_or(line);
```


</blockquote></details>

I think you'd want to protect yourself from bugs by writing

```rust
    let line = line
        .iter()
        .rev()
        .position(|&b| !b.is_ascii_whitespace())
        .map(|pos| line.get(..line.len() - pos).unwrap())
        .unwrap_or(line);
```

at which point this is looking rather ugly.

ajwerner (approved):

aya/src/programs/kprobe.rs
```diff
@@ -68,7 +68,7 @@ impl KProbe {
     ///
     /// The returned value can be used to detach from the given function, see [KProbe::detach].
     pub fn attach(&mut self, fn_name: &str, offset: u64) -> Result<KProbeLinkId, ProgramError> {
```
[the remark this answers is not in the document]
addisoncrump: As an example of this:
```
$ cat evil.c 
#include <stdio.h>

void thing() {
  puts("hello!");
}

int main(int argc, char **argv) {
  thing();
  return 0;
}
$ gcc -O0 -c evil.c 
$ objcopy --redefine-sym thing="$(printf '\x80')" evil.o 
$ objdump -t evil.o 

evil.o:     file format elf64-x86-64

SYMBOL TABLE:
0000000000000000 l    df *ABS*	0000000000000000 evil.c
0000000000000000 l    d  .text	0000000000000000 .text
0000000000000000 l    d  .rodata	0000000000000000 .rodata
0000000000000000 g     F .text	0000000000000011 �
0000000000000000         *UND*	0000000000000000 puts
0000000000000011 g     F .text	0000000000000020 main


$ gcc evil.o -o evil
$ ./evil 
hello!
```

addisoncrump (changes requested):

aya/src/programs/uprobe.rs
```diff
@@ -127,31 +160,30 @@ impl UProbe {
 fn resolve_attach_path<T: AsRef<Path>>(
     target: &T,
     pid: Option<pid_t>,
-) -> Result<Cow<'_, str>, UProbeError> {
+) -> Result<Cow<'_, Path>, UProbeError> {
```
dave-tucker, line 163: Do we still need Cow here? With the ldsocache changes to store PathBuf, 2/3  successful code paths yield a pathbuf, where the final one returns the input if the path is absolute.

aya/src/programs/uprobe.rs
```diff
@@ -249,45 +282,55 @@ pub enum UProbeError {
...
+        })
+    }))
 }
 
 #[derive(Debug)]
 pub(crate) struct CacheEntry {
-    key: String,
-    value: String,
+    key: OsString,
+    value: OsString,
```
dave-tucker, line 333: Value should be `PathBuf`

dave-tucker (changes requested):

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 3 of 9 files reviewed, 11 unresolved discussions (waiting on @addisoncrump, @ajwerner, and @dave-tucker)

___
*[`aya/src/programs/kprobe.rs` line 70 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291191539:-NbZoK_I99SBFq4s_RWP:bl4804p) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/kprobe.rs#L70)):*
<details><summary><i>Previously, addisoncrump (Addison Crump) wrote…</i></summary><blockquote>

As an example of this:
```
$ cat evil.c 
#include <stdio.h>

void thing() {
  puts("hello!");
}

int main(int argc, char **argv) {
  thing();
  return 0;
}
$ gcc -O0 -c evil.c 
$ objcopy --redefine-sym thing="$(printf '\x80')" evil.o 
$ objdump -t evil.o 

evil.o:     file format elf64-x86-64

SYMBOL TABLE:
0000000000000000 l    df *ABS*	0000000000000000 evil.c
0000000000000000 l    d  .text	0000000000000000 .text
0000000000000000 l    d  .rodata	0000000000000000 .rodata
0000000000000000 g     F .text	0000000000000011 �
0000000000000000         *UND*	0000000000000000 puts
0000000000000011 g     F .text	0000000000000020 main


$ gcc evil.o -o evil
$ ./evil 
hello!
```
</blockquote></details>

Sounds reasonable, but I'll leave that for a future change to avoid mixing API and non-API changes.
___
*[`aya/src/programs/kprobe.rs` line 71 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291163115:-NbZlqsvB0rX3TCU8oKC:b4tijtc) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/kprobe.rs#L71)):*

```rust
pub fn attach<T: AsRef<Path>(&mut self, fn_name: T, offset: u64) -> Result<KProbeLinkId, ProgramError> {
```

This won't break the API since both &str and &OsStr implement `AsRef<Path>`.
Not sure there's a function we'd kprobe that isn't UTF-8, but 🤷 we may as well be thorough.
</blockquote></details>

It might break type inference if someone's `.into()`ing the operand.

I think that's better, but I'd prefer to avoid API review here.
___
*[`aya/src/programs/probe.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291164862:-NbZmEZS4fNNdwWOrGmk:b9pa4jk) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/probe.rs#L170)):*

Why is it OK to do lossy conversion here but not in other places?
Ultimately it looks like this name is being written to a file as bytes so a lossy conversion is unnecessary.
</blockquote></details>

The very next line calls `str::replace`, so my feeling is that this need not match exactly. We can probably avoid the conversion but the code will become more complex. Do you know if we need to?
___
*[`aya/src/programs/uprobe.rs` line 53 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291139596:-NbZmvsv3U6KDlPkgbOI:b-90lccs) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L53)):*

What's a needle? I would have expected `pattern`
</blockquote></details>

It's not a pattern though. See the argument name here: https://doc.rust-lang.org/stable/std/primitive.slice.html#method.starts_with
___
*[`aya/src/programs/uprobe.rs` line 163 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291209103:-NbZnGOx2jO8xeSCNXkD:bx25ux0) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L163)):*

Do we still need Cow here? With the ldsocache changes to store PathBuf, 2/3  successful code paths yield a pathbuf, where the final one returns the input if the path is absolute.
</blockquote></details>

This is a fair point, but what would you have this return? `Result<Option<PathBuf>, UProbeError>`? Is that better?
___
*[`aya/src/programs/uprobe.rs` line 207 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291147207:-NbZnXOz8ey5Nm2K9Tjr:b-u1rqn0) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L207)):*

This could panic, because you're making another assumption about UTF-8 paths here.
The assertion below should probably be either that the `libc_path.file_name()` contains `libc`.
</blockquote></details>

This is a test.
___
*[`aya/src/programs/uprobe.rs` line 316 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291154752:-NbZncDd3XTNX4LX_vDm:bcjagq1) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L316)):*

lib should be an `&OsStr` - it's never used as a `&Path` here.
</blockquote></details>

I tried to keep OsStr out of the various function signatures. The caller has a Path, it doesn't care that we covert to OsStr for "reasons".
___
*[`aya/src/programs/uprobe.rs` line 333 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291210011:-NbZnqHWC3O_Nzjr5H5K:bb76mu4) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L333)):*

Value should be `PathBuf`
</blockquote></details>

Why?
___
*[`aya/src/programs/uprobe.rs` line 440 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291203552:-NbZnruJ3rzHVJa9ShsZ:bsxn7wb) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L440)):*

Again, I think `lib` should be `&OsStr` here too.
</blockquote></details>

See previous reply.
___
*[`aya/src/sys/netlink.rs` line 745 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291160397:-NbZnudf7KOfApJ5CTne:bi6mx54) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/sys/netlink.rs#L745)):*

Again, struggling to see how this needs to be part of this commit.
</blockquote></details>

It can be a separate commit. I was just hunting for unnecessary lossy string conversions.
___
*[`aya-obj/src/obj.rs` line 1650 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291135988:-NbZo3puEIat42uD6Bd9:bi6mx54) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya-obj/src/obj.rs#L1650)):*

I'm not sure how these changes are relevant to the commit?
</blockquote></details>

It can be a separate commit. I was just hunting for unnecessary lossy string conversions.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 3 of 9 files reviewed, 11 unresolved discussions (waiting on @addisoncrump, @ajwerner, and @dave-tucker)

___
*[`aya/src/sys/netlink.rs` line 745 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291160397:-NbZyJ66B-I5icYvD_-T:b-l7qfr9) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/sys/netlink.rs#L745)):*

It can be a separate commit. I was just hunting for unnecessary lossy string conversions.
</blockquote></details>

Extracted into a parent commit.
___
*[`aya-obj/src/obj.rs` line 1650 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291135988:-NbZyMLZBzPtlJiYQ-0D:b-l7qfr9) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya-obj/src/obj.rs#L1650)):*

It can be a separate commit. I was just hunting for unnecessary lossy string conversions.
</blockquote></details>

Extracted into a parent commit.

dave-tucker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @addisoncrump, @ajwerner, and @tamird)

___
*[`aya/src/programs/probe.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291164862:-Nb_79mp5dOexJTbVgRM:b-gn9o9y) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/probe.rs#L170)):*

The very next line calls `str::replace`, so my feeling is that this need not match exactly. We can probably avoid the conversion but the code will become more complex. Do you know if we need to?
</blockquote></details>

See:
https://docs.kernel.org/trace/kprobetrace.html
https://docs.kernel.org/trace/uprobetracer.html

Which describes the format of this file for kprobes/uprobes.

At`let probe`... later in this function .. we should be using the path verbatim - not after a lossy conversion. Assuming it follows standard kernel convention. Reference: https://lwn.net/Articles/71472/

The event alias component, where fixed_fn_name is used transforms `/bin/bash` into `bin_bash` where it's used to build a unique id - `aya_1234_bin_bash`. I think we also need to do further due diligence on whether the presence of https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html in this identifier would cause issues.
___
*[`aya/src/programs/uprobe.rs` line 163 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291209103:-Nb_B4Pv9zR4FzWlRvg_:b9elckm) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L163)):*

This is a fair point, but what would you have this return? `Result<Option<PathBuf>, UProbeError>`? Is that better?
</blockquote></details>

I feel `Result<Option<PathBuf>, UProbeError>` is better, but that's personal preference.
___
*[`aya/src/programs/uprobe.rs` line 333 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291210011:-Nb_BZKeBYNY4zVb2f2s:bsnnm8p) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L333)):*

Why?
</blockquote></details>

You can check the output of `ldconfig -p` the keys are OsString, the values are PathBuf.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @addisoncrump, @ajwerner, and @dave-tucker)

___
*[`aya/src/programs/probe.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291164862:-Nb_PpxV0XeCBGamNukO:b-veb8m6) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/probe.rs#L170)):*

See:
https://docs.kernel.org/trace/kprobetrace.html
https://docs.kernel.org/trace/uprobetracer.html

Which describes the format of this file for kprobes/uprobes.

At`let probe`... later in this function .. we should be using the path verbatim - not after a lossy conversion. Assuming it follows standard kernel convention. Reference: https://lwn.net/Articles/71472/

The event alias component, where fixed_fn_name is used transforms `/bin/bash` into `bin_bash` where it's used to build a unique id - `aya_1234_bin_bash`. I think we also need to do further due diligence on whether the presence of https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html in this identifier would cause issues.

</blockquote></details>

OK this was pretty involved, but done.
___
*[`aya/src/programs/uprobe.rs` line 163 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291209103:-Nb_HRus7qjG-feXGz5t:b-1j872f) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L163)):*

I feel `Result<Option<PathBuf>, UProbeError>` is better, but that's personal preference.
</blockquote></details>

Looking at this again: it's not correct. 1/3 code paths returns an owned PathBuf, the other 2 return Path. Even if ldsocache held PathBuf, it would return borrowed Path. Left as-is.
___
*[`aya/src/programs/uprobe.rs` line 333 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291210011:-Nb_GgJy80Jibx3DdsLY:bz23nex) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/uprobe.rs#L333)):*

You can check the output of `ldconfig -p` the keys are OsString, the values are PathBuf.
</blockquote></details>

OsString and PathBuf are the same type, we should choose the one that's easier to work with. In this case it's easier to work with OsString because all the usages (e.g. the various calls to strip_suffix and friends) want the OsString semantics, not the Path semantics.

ajwerner (approved):
> Reviewed 2 of 6 files at r2, 7 of 7 files at r5, 6 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 8 of 11 files reviewed, 7 unresolved discussions (waiting on @addisoncrump, @alessandrod, and @dave-tucker)

___
*[`aya/src/programs/probe.rs` line 51 at r6](https://reviewable.io/reviews/aya-rs/aya/742#-NbpAIM-Cfr4d8xRBHrM:-NbpAIM-Cfr4d8xRBHrN:by14ekj) ([raw file](https://github.com/aya-rs/aya/blob/2f3b8b050991a25abd5f062ae59001a3833e342b/aya/src/programs/probe.rs#L51)):*
> ```XML
> 
>     bytes.as_ref().split(|b| b == &b'\n').map(|mut line| {
>         loop {
> ```

nit: `while let` is a thing
```rust
    while let [stripped @ .., c] = line {
        if c.is_ascii_whitespace() {
            line = stripped;
            continue
        }
        break;
    }
```
___
*[`aya/src/programs/uprobe.rs` line 258 at r6](https://reviewable.io/reviews/aya-rs/aya/742#-NbpC0Rj65TG7YHce58G:-NbpC0Rk5uvPWLBwx9wq:b-lx786b) ([raw file](https://github.com/aya-rs/aya/blob/2f3b8b050991a25abd5f062ae59001a3833e342b/aya/src/programs/uprobe.rs#L258)):*
> ```XML
>         .split(|b| b == &b'\n')
>         .filter_map(|mut line| {
>             loop {
> ```

same nit from above: `while let` exists.

ajwerner (approved):
> Reviewed 3 of 9 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: all files reviewed, 7 unresolved discussions (waiting on @addisoncrump, @alessandrod, and @dave-tucker)

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: all files reviewed, 7 unresolved discussions (waiting on @addisoncrump, @alessandrod, and @dave-tucker)

___
*[`aya/src/programs/probe.rs` line 51 at r6](https://reviewable.io/reviews/aya-rs/aya/742#-NbpAIM-Cfr4d8xRBHrM:-NbpHqAlCjp8LsXqBpZz:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2f3b8b050991a25abd5f062ae59001a3833e342b/aya/src/programs/probe.rs#L51)):*

nit: `while let` is a thing
```rust
    while let [stripped @ .., c] = line {
        if c.is_ascii_whitespace() {
            line = stripped;
            continue
        }
        break;
    }
```
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 258 at r6](https://reviewable.io/reviews/aya-rs/aya/742#-NbpC0Rj65TG7YHce58G:-NbpHrS05V140EO6KaKm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2f3b8b050991a25abd5f062ae59001a3833e342b/aya/src/programs/uprobe.rs#L258)):*

same nit from above: `while let` exists.
</blockquote></details>

Done.

ajwerner (approved):
> Reviewed 8 of 9 files at r7, 1 of 9 files at r8.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 3 of 11 files reviewed, 7 unresolved discussions (waiting on @addisoncrump, @alessandrod, and @dave-tucker)

addisoncrump (approved):
LGTM, but we need to also apply similar changes for function names in the future.

tamird (commented):
> Dismissed @addisoncrump and @dave-tucker from 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/742)* status: 1 of 15 files reviewed, all discussions resolved

___
*[`aya/src/programs/kprobe.rs` line 70 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291191539:-Ncc76FS7ZJYQkE_i4KX:b-3czpvt) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/kprobe.rs#L70)):*

Sounds reasonable, but I'll leave that for a future change to avoid mixing API and non-API changes.
</blockquote></details>

https://github.com/aya-rs/aya/pull/765
___
*[`aya/src/programs/kprobe.rs` line 71 at r2](https://reviewable.io/reviews/aya-rs/aya/742#gh-1291163115:-Ncc772t3ftbJjyUJSu7:b-3czpvt) ([raw file](https://github.com/aya-rs/aya/blob/09bee10e79e03ee9960da19a8002db0d3aad6e7b/aya/src/programs/kprobe.rs#L71)):*

It might break type inference if someone's `.into()`ing the operand.

I think that's better, but I'd prefer to avoid API review here.
</blockquote></details>

https://github.com/aya-rs/aya/pull/765

PR #744, opened 2023-08-10
programs: `ProgramFd` is owned

tamird:
`ProgramData::fd` is now also owned. This means that `ProgramData` now
closes the file descriptor on drop. In the future we might consider
making `ProgramFd` hold a `BorrowedFd` but this requires API design work
due to overlapping borrows.

Since `ProgramFd` is no longer `Copy`, update methods to take it by
reference to allow callers to use it multiple times as they are
accustomed to doing.

This is an API breaking change.

Updates #612.

### DIFF
--- a/aya/src/programs/extension.rs
+++ b/aya/src/programs/extension.rs
@@ -1,5 +1,5 @@
 //! Extension programs.
-use std::os::fd::{AsFd as _, AsRawFd as _, OwnedFd};
+use std::os::fd::{AsFd as _, AsRawFd as _, BorrowedFd, OwnedFd};
 use thiserror::Error;
 
 use object::Endianness;
@@ -42,8 +42,9 @@ pub enum ExtensionError {
 /// prog.attach("eth0", XdpFlags::default())?;
 ///
 /// let prog_fd = prog.fd().unwrap();
+/// let prog_fd = prog_fd.try_clone().unwrap();
 /// let ext: &mut Extension = bpf.program_mut("extension").unwrap().try_into()?;
-/// ext.load(prog_fd, "function_to_replace")?;
+/// ext.load(&prog_fd, "function_to_replace")?;
 /// ext.attach()?;
 /// Ok::<(), aya::BpfError>(())
 /// ```
@@ -68,12 +69,12 @@ impl Extension {
     /// The extension code will be loaded but inactive until it's attached.
     /// There are no restrictions on what functions may be replaced, so you could replace
     /// the main entry point of your program with an extension.
-    pub fn load(&mut self, program: ProgramFd, func_name: &str) -> Result<(), ProgramError> {
-        let target_prog_fd = program.as_raw_fd();
+    pub fn load(&mut self, program: &ProgramFd, func_name: &str) -> Result<(), ProgramError> {
+        let target_prog_fd = program.as_fd();
         let (btf_fd, btf_id) = get_btf_info(target_prog_fd, func_name)?;
 
         self.data.attach_btf_obj_fd = Some(btf_fd);
-        self.data.attach_prog_fd = Some(target_prog_fd);
+        self.data.attach_prog_fd = Some(target_prog_fd.as_raw_fd());
         self.data.attach_btf_id = Some(btf_id);
         load_program(BPF_PROG_TYPE_EXT, &mut self.data)
     }
@@ -86,7 +87,9 @@ impl Extension {
     /// The returned value can be used to detach the extension and restore the
     /// original function, see [Extension::detach].
     pub fn attach(&mut self) -> Result<ExtensionLinkId, ProgramError> {
-        let prog_fd = self.data.fd_or_err()?;
+        let prog_fd = self.fd()?;
+        let prog_fd = prog_fd.as_fd();
+        let prog_fd = prog_fd.as_raw_fd();
         let target_fd = self.data.attach_prog_fd.ok_or(ProgramError::NotLoaded)?;
         let btf_id = self.data.attach_btf_id.ok_or(ProgramError::NotLoaded)?;
         // the attach type must be set as 0, which is bpf_attach_type::BPF_CGROUP_INET_INGRESS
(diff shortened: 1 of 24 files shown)


aya/src/programs/extension.rs
```diff
@@ -37,12 +37,22 @@ pub enum ExtensionError {
...
+/// let mut ext = None;
+/// for (name, program) in bpf.programs_mut() {
+///     match name {
+///         "main" => prog = Some(program),
+///         "extension" => ext = Some(program),
+///         _ => {},
+///     }
+/// }
+///
+/// let prog: &mut Xdp = prog.unwrap().try_into()?;
```
nrxus, line 50: Requiring to loop through the programs whenever we need to attach extensions doesn't seem ideal. I wonder what would the performance difference be of that versus having something like:

```rust
let extension = bpf.with_program_mut("extension", |bpf: &mut Bpf, extension: &mut Program| {
    let extensio: &mut Xdp = extension.try_into()?;
    let prog = bpf.program_mut("main").unwrap();
    prog.load()?;
    prog.attach("eth0", XdpFlags::default())?;
    extension.load(prog.fd().unwrap())?;
    Ok(extension)
}).unwrap()?;

extension.attach()?;
```

The way `with_program_mut` would be implemented would be that it'd essentially remove the given program from the program map, pass a mutable reference to the closure, and after the closure is finished put that program back in. The tricky thing I can see with this is that if the closure panics we aren't putting the program back in so it might get to an inconsistent state but I don't immediately see that as a big deal IMO.

tamird: I don't think this scales all that well - in particular what happens if you want to operate on N extensions at the same time? In the fullness of time we should use https://github.com/rust-lang/rust/issues/97601 (get_many_mut) but that's not available today. Though it is available via hashbrown.

Maybe @alessandrod has ideas?

tamird: Pulled in `hashbrown`. See how you like this.

dave-tucker: This is one of the reasons I said let's used OnwedFd and try_clone_to_owned in public APIs that expose FDs in #723 - as well as taking the same approach for Links.

I'm not opposed to optimizing later to use BorrowedFd, and even adding `get_many_mut`, but these changes probably require some design work upfront.
As it stands this PR will break users that are expecting the mutable borrow from `program_mut` to be dropped after getting a `ProgramFd` - aside from the example here this could also be an issue for users of `ProgramArray`.

We need to work up a design around how lifetimes in `BpfLoader` would work with all the myriad bpf lifecycle APIs we have.

dave-tucker (changes requested):

aya/src/programs/extension.rs
tamird: Agreed; deferring such design seems better. Done.

nrxus: I don't love the idea that the only way to get access to the file descriptor of a program by an user is by dup'ing it especially since it seems we are doing this to get around the awkwardness of passing a loaded program's fd to an extension which isn't the only reason one may need to get the file descriptor for. A use case I have for needing the file descriptor is for setting up the tail call table which doesn't require mutable references of multiple programs but with this change I now would have to pay the runtime cost created to work around extensions on a different workflow. 

Fwiw the example code I gave above would have worked for multiple extensions as well unless I am missing something:

```rust
// If you need to attach one program to many extensions
let extension = bpf.with_program_mut("main", |bpf: &mut Bpf, prog: &mut Program| {
   let prog: &mut Xdp = prog.try_into()?;
    prog.load()?;
    prog.attach("eth0", XdpFlags::default())?;
    let prog_fd = prog.fd().unwrap()?;
    
    let extension: &mut Extension = bpf.program_mut("extension1").unwrap().try_into()?;
    extension.load(prog_fd))?;
    extension.attach()?;
   
     // no longer need extension here so we can shadow the variable  and the borrow checker is happy
    let extension: &mut Extension = bpf.program_mut("extension2").unwrap().try_into()?;
    extension.load(prog_fd))?;
    extension.attach()?;
    
    Ok(())
}).unwrap()?;
```

The above *also* solves the issue of needing to borrow from the maps and programs at the same time fwiw which would be useful for the socketfd stuff.

aya/src/programs/mod.rs
```diff
@@ -718,8 +720,12 @@ macro_rules! impl_fd {
         $(
             impl $struct_name {
                 /// Returns the file descriptor of this Program.
-                pub fn fd(&self) -> Option<ProgramFd> {
-                    self.data.fd.map(|fd| ProgramFd(fd))
+                pub fn fd(&self) -> Result<ProgramFd, ProgramError> {
+                    self.data.fd
+                        .as_ref()
+                        .ok_or(ProgramError::NotLoaded)
+                        .and_then(|fd| fd.try_clone().map_err(Into::into))
```
nrxus, line 727: Having the only way to get a the loaded file descriptor of a loaded program be through duplicating the file descriptor feels heavy weight to me. If we *really* don't want to spend much time thinking of another design at this time for how to handle borrowing two programs at the same time (or a program and a map at the same time) then we should at least provide a cheap way to get a borrowed file descriptor for the cases where the user can handle the lifetime themselves. In my use case I use the loaded programs file descriptors to setup tail calls and I normally remove the map out of the `Bpf` struct to handle their lifecycle myself so I wouldn't run into borrow checker issues but yet would still be paying the runtime penalty here.

tamird: I've made `ProgramFd` return by reference. Does this appeal?

nrxus: Yeah that satisfies my concern enough. I am curious if we actually need the wrappers around the file descriptor at all but I guess that's tangential to this PR, they just feel kind of awkward and aren't providing that much value (preventing an user from trying to add a non-program fd to an extension/tail call table, which isn't unsafe it'd just return an error?).

tamird: Shrug. I'd rather not engage on that discussion here.

aya/src/programs/xdp.rs
```diff
@@ -174,7 +176,9 @@ impl Xdp {
     ///
     /// Ownership of the link will transfer to this program.
     pub fn attach_to_link(&mut self, link: XdpLink) -> Result<XdpLinkId, ProgramError> {
-        let prog_fd = self.data.fd_or_err()?;
+        let prog_fd = self.fd()?;
```
alessandrod, line 179: Here and everywhere else, why not do:

```
let prog_fd = self.fd()?.as_fd().as_raw_fd();
```

alessandrod (approved):
See nit, otherwise looks good

aya/src/programs/xdp.rs
tamird: Just personal preference - I prefer to avoid `?` in the middle of an expression. This will also make the diff cleaner when we get around to removing the need for `as_raw_fd`; it'll be a pure deletion.

tamird:
Will merge after @dave-tucker has a chance to approve.

PR #745, opened 2023-08-10
xtask: watch for kernel panic in stdout too

tamird:
Seems unclear whether kernel panics go to stdout or stderr, so do them
both.


ajwerner (approved):
Makes sense. Hope it works.

PR #747, opened 2023-08-10
sys: add map_ids to bpf_prog_get_info_by_fd

tamird:
Allows the caller to pass a slice which the kernel will populate with
map ids used by the program.

### DIFF
--- a/aya/src/sys/bpf.rs
+++ b/aya/src/sys/bpf.rs
@@ -477,10 +477,14 @@ pub(crate) fn bpf_prog_get_fd_by_id(prog_id: u32) -> Result<OwnedFd, SyscallErro
     })
 }
 
-fn bpf_obj_get_info_by_fd<T>(fd: BorrowedFd<'_>) -> Result<T, SyscallError> {
+fn bpf_obj_get_info_by_fd<T, F: FnOnce(&mut T)>(
+    fd: BorrowedFd<'_>,
+    init: F,
+) -> Result<T, SyscallError> {
     let mut attr = unsafe { mem::zeroed::<bpf_attr>() };
-    // info gets entirely populated by the kernel
-    let info = MaybeUninit::zeroed();
+    let mut info = unsafe { mem::zeroed() };
+
+    init(&mut info);
 
     attr.info.bpf_fd = fd.as_raw_fd() as u32;
     attr.info.info = &info as *const _ as u64;
@@ -489,7 +493,7 @@ fn bpf_obj_get_info_by_fd<T>(fd: BorrowedFd<'_>) -> Result<T, SyscallError> {
     match sys_bpf(bpf_cmd::BPF_OBJ_GET_INFO_BY_FD, &mut attr) {
         Ok(code) => {
             assert_eq!(code, 0);
-            Ok(unsafe { info.assume_init() })
+            Ok(info)
         }
         Err((code, io_error)) => {
             assert_eq!(code, -1);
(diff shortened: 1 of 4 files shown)


aya/src/sys/bpf.rs
```diff
@@ -501,13 +505,19 @@ fn bpf_obj_get_info_by_fd<T>(fd: BorrowedFd<'_>) -> Result<T, SyscallError> {
...
 
-pub(crate) fn bpf_prog_get_info_by_fd(fd: RawFd) -> Result<bpf_prog_info, SyscallError> {
+pub(crate) fn bpf_prog_get_info_by_fd(
+    fd: RawFd,
+    map_ids: &mut [u32],
+) -> Result<bpf_prog_info, SyscallError> {
     let fd = unsafe { BorrowedFd::borrow_raw(fd) };
-    bpf_obj_get_info_by_fd::<bpf_prog_info>(fd)
+    bpf_obj_get_info_by_fd(fd, |info: &mut bpf_prog_info| {
+        info.nr_map_ids = map_ids.len() as _;
```
astoycos, line 514: does this not need to be `as u64` ? Or does it already know that based on the memory layout?

aya/src/sys/bpf.rs
```diff
@@ -477,10 +477,14 @@ pub(crate) fn bpf_prog_get_fd_by_id(prog_id: u32) -> Result<OwnedFd, SyscallErro
...
 
-fn bpf_obj_get_info_by_fd<T>(fd: BorrowedFd<'_>) -> Result<T, SyscallError> {
+fn bpf_obj_get_info_by_fd<T, F: FnOnce(&mut T)>(
+    fd: BorrowedFd<'_>,
+    init: F,
+) -> Result<T, SyscallError> {
     let mut attr = unsafe { mem::zeroed::<bpf_attr>() };
-    // info gets entirely populated by the kernel
-    let info = MaybeUninit::zeroed();
+    let mut info = unsafe { mem::zeroed() };
```
astoycos, line 485: Can this stay as `MaybeUninit::zeroed();` ?

tamird: How's that better? We'll need to immediately `assume_init` so we can write to it.

aya/src/sys/bpf.rs
tamird: Yes, the type is inferred.

aya/src/sys/bpf.rs
astoycos: why not just call `as_mut_ptr()` on `info` in the init closure? Since we only need to assume_init in that one case

tamird: We'd still need to assume_init *later*, before returning. What is the benefit of using `MaybeUninit` here?

astoycos: Or just use as_mut_ptr() here and ONLY do unsafe code in that one case

tamird: This doesn't make sense - look at the diff. Before the change, we had unsafe code on the return path.

tamird: Writing through `MaybeUninit::as_mut_ptr` is nasty:

> Gets a mutable pointer to the contained value. Reading from this pointer or turning it into a reference is undefined behavior unless the MaybeUninit<T> is initialized.

https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.as_mut_ptr

That means we can't turn it into a mutable reference because that's UB. Again, how is this better?

astoycos: Fair enough I guess there is no way to protect against the unsafe code here 👍 In the docs it says 
```
This has the same effect as [MaybeUninit::zeroed().assume_init()](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.zeroed). It is useful for FFI sometimes, but should generally be avoided.
```
I guess this is the "usefult for FFI case" lol :)

tamird: Yup.

astoycos (approved):
/LGTM

PR #748, opened 2023-08-10
programs: Plug attach_btf_obj_fd leak

tamird:
`ProgramData::attach_btf_obj_fd` is now owned.  This means that
`ProgramData` now closes the file descriptor on drop.

Updates #612.

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -406,7 +406,7 @@ pub(crate) struct ProgramData<T: Link> {
     pub(crate) fd: Option<RawFd>,
     pub(crate) links: LinkMap<T>,
     pub(crate) expected_attach_type: Option<bpf_attach_type>,
-    pub(crate) attach_btf_obj_fd: Option<u32>,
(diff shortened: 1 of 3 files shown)


tamird:
cc @nrxus

aya/src/programs/mod.rs
```diff
@@ -604,7 +597,7 @@ fn load_program<T: Link>(
         kernel_version: target_kernel_version,
         expected_attach_type: *expected_attach_type,
         prog_btf_fd: btf_fd.as_ref().map(|f| f.as_fd()),
-        attach_btf_obj_fd: *attach_btf_obj_fd,
+        attach_btf_obj_fd: attach_btf_obj_fd.as_ref().map(|fd| fd.as_raw_fd() as _),
```
nrxus, line 600: why not make `attach_btf_obj_fd` be a `Option<BorrowedFd<'a>>`? We already do that for `prog_btf_fd`,

tamird: Good call. Done.

dave-tucker (approved):

PR #758, opened 2023-08-16
aya: MapData::fd is non-optional

tamird:
The primary driver of change here is that `MapData::create` is now a
factory function that returns `Result<Self, _>` rather than mutating
`&mut self`. The remaining changes are consequences of that change, the
most notable of which is the removal of several errors which are no
longer possible.

### DIFF
--- a/aya/src/maps/array/array.rs
+++ b/aya/src/maps/array/array.rs
@@ -39,8 +39,6 @@ impl<T: Borrow<MapData>, V: Pod> Array<T, V> {
         let data = map.borrow();
         check_kv_size::<u32, V>(data)?;
 
-        let _fd = data.fd_or_err()?;
-
         Ok(Array {
             inner: map,
             _v: PhantomData,
@@ -63,7 +61,7 @@ impl<T: Borrow<MapData>, V: Pod> Array<T, V> {
     pub fn get(&self, index: &u32, flags: u64) -> Result<V, MapError> {
         let data = self.inner.borrow();
         check_bounds(data, *index)?;
-        let fd = data.fd_or_err()?;
+        let fd = data.fd;
 
         let value =
             bpf_map_lookup_elem(fd, index, flags).map_err(|(_, io_error)| SyscallError {
@@ -90,7 +88,7 @@ impl<T: BorrowMut<MapData>, V: Pod> Array<T, V> {
     pub fn set(&mut self, index: u32, value: impl Borrow<V>, flags: u64) -> Result<(), MapError> {
         let data = self.inner.borrow_mut();
         check_bounds(data, index)?;
-        let fd = data.fd_or_err()?;
+        let fd = data.fd;
         bpf_map_update_elem(fd, Some(&index), value.borrow(), flags).map_err(|(_, io_error)| {
             SyscallError {
                 call: "bpf_map_update_elem",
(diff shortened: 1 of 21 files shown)


ajwerner (approved):
This seems reasonable, but is this change motivated as part of some other effort? I get the ways in which this is better. I suppose before it was possible to defer the actual creation of the map from the setup and it's not now, but I'm unsure as to why somebody might have wanted to do that.

tamird:
> This seems reasonable, but is this change motivated as part of some other effort?

I noticed the oddity during review of another PR and decided to fix it. I'm particular this makes life better when we later replace the RawFd with an OwnedFd (notice we are no longer passing a "42" file descriptor in tests)

> I get the ways in which this is better. I suppose before it was possible to defer the actual creation of the map from the setup and it's not now, but I'm unsure as to why somebody might have wanted to do that.

Indeed. Unclear why anyone would want such deferral.

aya/src/maps/array/array.rs
```diff
@@ -39,7 +39,7 @@ impl<T: Borrow<MapData>, V: Pod> Array<T, V> {
         let data = map.borrow();
         check_kv_size::<u32, V>(data)?;
 
-        let _fd = data.fd_or_err()?;
+        let _fd = data.fd;
```
alessandrod, line 42: This can go now? We used to get the fd to ensure that the map was created, but
now it's always created

aya/src/maps/array/per_cpu_array.rs
```diff
@@ -58,7 +58,7 @@ impl<T: Borrow<MapData>, V: Pod> PerCpuArray<T, V> {
         let data = map.borrow();
         check_kv_size::<u32, V>(data)?;
 
-        let _fd = data.fd_or_err()?;
+        let _fd = data.fd;
```
alessandrod, line 61: Same here (and everywhere else)

aya/src/maps/array/array.rs
tamird: Done. Thanks for catching that.

aya/src/maps/mod.rs
```diff
@@ -513,28 +498,41 @@ impl MapData {
...
-
-        Ok(*self.fd.insert(fd.into_raw_fd()))
+        }) {
+            Ok(fd) => Ok(Self {
+                obj,
+                fd: fd.into_raw_fd(),
+                pinned: false,
+            }),
+            Err(_) => {
+                let mut map = Self::create(obj, name, None)?;
```
alessandrod, line 528: Why is btf_fd hardcoded to None here?

tamird: :facepalm: 

Done.

alessandrod (commented):
Thanks again for doing this, see comments

tamird:
I've addressed all the comments, PTAL.

alessandrod:
> > I get the ways in which this is better. I suppose before it was possible to defer the actual creation of the map from the setup and it's not now, but I'm unsure as to why somebody might have wanted to do that.
> 
> Indeed. Unclear why anyone would want such deferral.

context for posterity https://discord.com/channels/855676609003651072/855676609003651075/1141557678955577445

alessandrod (approved):

PR #763, opened 2023-08-22
aya: deny various allow-by-default lints

tamird:
Notably:
- clippy::use_self: replaced many T with Self.
- single_use_lifetimes: removed some single use lifetimes.
- unreachable_pub: removed some unreachable pub items.
- unused_crate_dependencies: removed unused futures,parking_lot deps.
- unused_qualifications: found a potential `crate` vs `$crate` bug.
- let_underscore_drop: not enabled, seems to trigger false positives.
- missing_copy_implementations: not enabled, unclear if we want this.
- unsafe_op_in_unsafe_fn: not enabled, unclear if we want this.
- unused_results: not enabled, needs many fixes (but I think wanted).

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -261,22 +261,22 @@ impl Map {
     /// Returns the low level map type.
     fn map_type(&self) -> u32 {
         match self {
-            Map::Array(map) => map.obj.map_type(),
-            Map::PerCpuArray(map) => map.obj.map_type(),
-            Map::ProgramArray(map) => map.obj.map_type(),
-            Map::HashMap(map) => map.obj.map_type(),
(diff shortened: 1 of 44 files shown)


aya/src/maps/mod.rs
```diff
@@ -500,9 +500,11 @@ impl MapData {
                 }
             })?;
 
+        #[allow(trivial_numeric_casts)]
+        let fd = fd as RawFd;
```
dave-tucker, line 504: I assume this will go away with when `bpf_create_map` returns an OwnedFd?

dave-tucker (approved):

aya/src/maps/mod.rs
tamird: Yep, though it might end up relocating elsewhere.

PR #765, opened 2023-08-24
aya: support non-UTF8 probing

### LINKED ISSUE #751: Support non-UTF-8 symbol probing
See conversation: https://github.com/aya-rs/aya/pull/742#discussion_r1291191539

tamird:
Fixes #751.

### DIFF
--- a/aya/src/programs/kprobe.rs
+++ b/aya/src/programs/kprobe.rs
@@ -1,5 +1,6 @@
 //! Kernel space probes.
 use std::{
+    ffi::OsStr,
     io,
     os::fd::AsFd as _,
     path::{Path, PathBuf},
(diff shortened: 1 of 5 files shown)


aya/src/programs/kprobe.rs
```diff
@@ -71,8 +71,12 @@ impl KProbe {
     /// target function.
     ///
     /// The returned value can be used to detach from the given function, see [KProbe::detach].
-    pub fn attach(&mut self, fn_name: &str, offset: u64) -> Result<KProbeLinkId, ProgramError> {
-        attach(&mut self.data, self.kind, Path::new(fn_name), offset, None)
+    pub fn attach<T: AsRef<Path>>(
```
alessandrod, line 74: As per discord conversation, this is not a `Path`.

I think the most accurate and intuitive (for users) type would be `OsStr` and `AsRef<OsStr>` works for the common case `attach("foo")`. To work with bytes though you have to type `attach(OsStr::from_bytes(b"foo"))` since `OsStr` can only be created from bytes via `std::os::unix::ffi::OsStrExt`.

Therefore I'm leaning towards just making it `AsRef<[u8]>` for convenience.

alessandrod (approved):

aya/src/programs/kprobe.rs
tamird: Thanks, I dug around a bit and came to the same conclusion. It looks like the meaning of this argument differs based on the type of probe. We could probably do better with the type safety of `ProbeKind` to encode this in the type system, but it seems correct. I've added a comment.

dave-tucker (approved):

PR #770, opened 2023-08-29
maps: `MapFd` and `SockMapFd` are owned

tamird:
`MapData::fd` is now a `MapFd`. This means that `MapData` now closes the
file descriptor on drop. In the future we might consider making `MapFd`
hold a `BorrowedFd` but this requires API design work due to overlapping
borrows.

Since `SockMapFd` is no longer `Copy`, attach methods to take it by
reference to allow callers to use it multiple times as they are
accustomed to doing.

`SockMapFd` implements `try_clone`. `MapFd` and `SockMapFd` are now
returned by reference to allow callers to avoid file descriptor cloning
when desired.

This is an API breaking change.

Updates #612.

### DIFF
--- a/aya-obj/src/relocation.rs
+++ b/aya-obj/src/relocation.rs
@@ -105,7 +105,7 @@ pub(crate) struct Symbol {
 
 impl Object {
     /// Relocates the map references
-    pub fn relocate_maps<'a, I: Iterator<Item = (&'a str, i32, &'a Map)>>(
+    pub fn relocate_maps<'a, I: Iterator<Item = (&'a str, std::os::fd::RawFd, &'a Map)>>(
         &mut self,
         maps: I,
         text_sections: &HashSet<usize>,
@@ -178,8 +178,8 @@ impl Object {
 fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
     fun: &mut Function,
     relocations: I,
-    maps_by_section: &HashMap<usize, (&str, i32, &Map)>,
-    maps_by_symbol: &HashMap<usize, (&str, i32, &Map)>,
+    maps_by_section: &HashMap<usize, (&str, std::os::fd::RawFd, &Map)>,
+    maps_by_symbol: &HashMap<usize, (&str, std::os::fd::RawFd, &Map)>,
     symbol_table: &HashMap<usize, Symbol>,
     text_sections: &HashSet<usize>,
 ) -> Result<(), RelocationError> {
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -36,12 +36,11 @@ use crate::{
         SkMsg, SkSkb, SkSkbKind, SockOps, SocketFilter, TracePoint, UProbe, Xdp,
     },
     sys::{
-        bpf_load_btf, bpf_map_freeze, bpf_map_update_elem_ptr, is_bpf_cookie_supported,
-        is_bpf_global_data_supported, is_btf_datasec_supported, is_btf_decl_tag_supported,
-        is_btf_enum64_supported, is_btf_float_supported, is_btf_func_global_supported,
-        is_btf_func_supported, is_btf_supported, is_btf_type_tag_supported, is_perf_link_supported,
+        bpf_load_btf, is_bpf_cookie_supported, is_bpf_global_data_supported,
+        is_btf_datasec_supported, is_btf_decl_tag_supported, is_btf_enum64_supported,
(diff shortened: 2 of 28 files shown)


tamird:
cc @nrxus

aya/src/bpf.rs
```diff
@@ -510,7 +510,7 @@ impl<'a> BpfLoader<'a> {
 
         obj.relocate_maps(
             maps.iter()
-                .map(|(s, data)| (s.as_str(), data.fd, &data.obj)),
+                .map(|(s, data)| (s.as_str(), data.fd().as_fd().as_raw_fd(), &data.obj)),
```
nrxus, line 513: why not pass a `BorrowedFd<'_>` here instead? The less references we have to `RawFd`s the easier it'll be gauge that our file descriptors are all valid.

aya/src/maps/mod.rs
```diff
@@ -582,11 +597,13 @@ impl MapData {
     /// This API is intended for cases where you have received a valid BPF FD from some other means.
     /// For example, you received an FD over Unix Domain Socket.
     pub fn from_fd(fd: OwnedFd) -> Result<Self, MapError> {
+        // TODO(https://github.com/aya-rs/aya/issues/612): should this function take a MapFd?
+        let fd = MapFd(fd);
```
nrxus, line 601: nit: probably would read better from a domain POV to only wrap it in `MapFd` after confirming that it's a `MapFd` (i.e., after `bpf_map_get_info_by_fd` succeeds).

aya/src/maps/mod.rs
```diff
@@ -610,30 +627,13 @@ impl MapData {
...
-impl Drop for MapData {
-    fn drop(&mut self) {
-        // TODO: Replace this with an OwnedFd once that is stabilized.
-        //
-        // SAFETY: `drop` is only called once.
-        unsafe { libc::close(self.fd) };
-    }
-}
-
-impl Clone for MapData {
```
nrxus, line 629: should we implement a `try_clone()` for `MapData` so maps can still be cloned?

aya/src/bpf.rs
tamird: IIRC the function we're calling here is available in no_std, so `BorrowedFd` can't appear in its signature.

tamird: Made that function available only in `std`. Unclear to me how anyone would use such a thing without access to real file descriptors.

aya/src/maps/mod.rs
tamird: Done.

aya-obj/src/relocation.rs
```diff
@@ -105,7 +110,11 @@ pub(crate) struct Symbol {
 
 impl Object {
     /// Relocates the map references
-    pub fn relocate_maps<'a, I: Iterator<Item = (&'a str, i32, &'a Map)>>(
+    #[cfg(feature = "std")]
+    pub fn relocate_maps<
```
alessandrod, line 114: This makes aya-obj a lot less useful without std. Do you think it would be too
painful to make this work with both RawFd and BorrowedFd? (Doesn't seem like but
I haven't tried)

alessandrod (commented):
Other than the aya-obj std thing, lgtm

aya-obj/src/relocation.rs
tamird: The previous discussion on this topic is [here](https://github.com/aya-rs/aya/pull/770#discussion_r1310993790). How do you get your hands on an FD in no_std?

aya/src/bpf.rs
alessandrod: I think we should undo the aya-obj bit. Relocation puts file descriptors in the instruction stream where they must be cast as integers anyway, so lifetime tracking doesn't buy us much and it makes aya-obj essentially useless on no_std.

tamird: I can do that. But can you help me understand how someone would use this in no_std given you can't get your hands on an FD? Or is the idea that the user has an FD through some other means?

alessandrod: Yes, it's unfortunate that *Fd are defined in `std::os`, in an ideal world some of the `os` definitions should be moved to `core` like it happened for `net`. But in a `no_std` program it's easy to imagine that one could still create fds using lower level linux APIs, and it seems silly that we would preclude them from using aya-obj just for a type alias.

tamird: I don't think this is equivalent to the `core::net` types since those are just scalars like IP addresses. Anyway, done -- I just defined a local alias for RawFd in this crate in no_std.

dave-tucker (approved):

PR #775, opened 2023-08-31
async_perf_event_array: access inner through async

tamird:
See commit messages.


tamird:
Note that the API changes here aren't substantive because [`BorrowMut: Borrow`](https://doc.rust-lang.org/stable/std/borrow/trait.BorrowMut.html).

alessandrod (approved):

Issue #636, opened 2023-07-05
cant' use LruHashMap in userspace


Sherlock-Holo:
the userspace HashMap try_from impl

```rust
impl<'a, V: Pod, K: Pod> TryFrom<&'a mut Map> for HashMap<&'a mut MapData, V, K> {
    type Error = MapError;

    fn try_from(map: &'a mut Map) -> Result<HashMap<&'a mut MapData, V, K>, MapError> {
        match map {
            Map::HashMap(m) => {
                HashMap::new(m)
            }
            _ => Err(MapError::InvalidMapType { map_type: map.map_type() }),
        }
    }
}
impl<V: Pod, K: Pod> TryFrom<Map> for HashMap<MapData, V, K> {
    type Error = MapError;

    fn try_from(map: Map) -> Result<HashMap<MapData, V, K>, MapError> {
        match map {
            Map::HashMap(m) => HashMap::new(m),
            _ => Err(MapError::InvalidMapType { map_type: map.map_type() }),
        }
    }
}
```

the expanded codes show HashMap only can from bpf hashmap, but not lru hashmap, I remember HashMap can try_from a LruHashMap at previous commit

Sherlock-Holo:
https://docs.rs/aya/latest/src/aya/maps/hash_map/hash_map.rs.html#48

the crates.io version handled the lruhashmap
maybe we can add a new type LruHashMap to fix this problem

tamird:
Do you have a test? I don't understand this issue.

Sherlock-Holo:
sorry for the late reply

kernel space codes

main.rs

```rust
#[xdp(name = "lru_test")]
fn lru_test(ctx: XdpContext) -> u32 {
    kernel::lru::lru_test(ctx)
}

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    unsafe { core::hint::unreachable_unchecked() }
}
```

lru.rs

```rust
use aya_bpf::bindings::xdp_action::XDP_PASS;
use aya_bpf::macros::map;
use aya_bpf::maps::LruHashMap;
use aya_bpf::programs::XdpContext;

#[map]
static LRU_MAP: LruHashMap<u32, u32> = LruHashMap::with_max_entries(100, 0);

pub fn lru_test(_ctx: XdpContext) -> u32 {
    let key = 1;
    let value = 1;
    let _ = LRU_MAP.insert(&key, &value, 0);
    
    XDP_PASS
}
```

user space codes

[...]

PR #777, opened 2023-09-01
aya/maps: rework TryFrom macros

ajwerner:
The old macros were repetitive and inflexible. This unifies the various
macros used to generate TryFrom implementations for map implementations
from the relevant map enum variants.

Cleanup in anticipation of fixing https://github.com/aya-rs/aya/issues/636.

The API changes are just about renaming the return to Self and
Self::Error; they are not real changes.

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -281,155 +281,82 @@ impl Map {
     }
 }
 
+// Implements TryFrom<Map> for different map implementations. Different map implementations can be
+// constructed from different variants of the map enum. Also, the implementation may have type
+// parameters (which we assume all have the bound `Pod` and nothing else).
 macro_rules! impl_try_from_map {
-    ($($tx:ident from Map::$ty:ident),+ $(,)?) => {
-        $(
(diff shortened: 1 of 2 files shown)


aya/src/maps/mod.rs
```diff
@@ -281,155 +281,122 @@ impl Map {
...
-                        Map::$ty(m) => {
-                            $tx::new(m)
-                        },
-                        _ => Err(MapError::InvalidMapType{ map_type: map.map_type()}),
-                    }
-                }
+        }
+    };
+    // Wrap the type parameters in parentheses so that they can be treated as a single TokenTree
+    // (tt) because you cannot compose repetitions but you can expand a scalar in a repetition.
```
tamird, line 302: i think this comment could be clearer; the point is that we're not interested in repetition, we just want to paste all the type parameters in a few places.

I think everything here will be simpler if you push the parentheses out into the macro invocation and include empty parentheses in the case of no type parameters.

tamird (approved):

alessandrod (approved):

PR #782, opened 2023-09-08
aya: add program_info() api to `Program`

astoycos:
Add a new api to the outer level `Program` structure which allows users to get the program's kernel info before casting it to an explicit program variant.

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -410,6 +410,39 @@ impl Program {
             Self::CgroupDevice(p) => p.fd(),
         }
     }
+
+    /// Returns information about a loaded program with the [`ProgramInfo`] structure.
+    ///
+    /// This information is populated at load time by the kernel and can be used
+    /// to get kernel details for a given [`Program`].
+    pub fn info(&self) -> Result<ProgramInfo, ProgramError> {
+        match self {
+            Self::KProbe(p) => p.info(),
+            Self::UProbe(p) => p.info(),
+            Self::TracePoint(p) => p.info(),
+            Self::SocketFilter(p) => p.info(),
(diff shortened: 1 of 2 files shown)


aya/src/programs/mod.rs
```diff
@@ -410,6 +410,39 @@ impl Program {
             Self::CgroupDevice(p) => p.fd(),
         }
     }
+
+    /// Returns information about a loaded program with the [`ProgramInfo`] structure.
+    ///
+    /// This information is populated at load time by the kernel and can be used
+    /// to get kernel details for a given [`Program`].
+    pub fn program_info(&self) -> Result<ProgramInfo, ProgramError> {
```
alessandrod, line 418: Why `program_info` and not `info`?

aya/src/programs/mod.rs
```diff
@@ -410,6 +410,39 @@ impl Program {
...
+        match self {
+            Self::KProbe(p) => p.program_info(),
```
tamird, line 420: This pattern (match on every variant, pull out the MapData, call a function on the MapData) repeats at least 4 times in this file. Can you extract a private function that pulls out the MapData and use it in all these functions? Ideally, that would done in a parent comment.

alessandrod: This is not using MapData, wrong file :P The inner types are actually different, so I think you could only do it with a macro.

tamird: Ah, thanks for explaining.

aya/src/programs/mod.rs
astoycos: Done! Also fixed up the existing API so that we're consistent.

alessandrod (approved):

PR #783, opened 2023-09-08
aya: Implement dedicated map pinning

astoycos:
Align with libbpf's  LIBBPF_PIN_BY_NAME map flag behavior.  More specifically if `map_pin_path` is not provided but a program has a map defined with the flag, pin it in `/sys/fs/bpf` otherwise pin it in the path provided by `map_pin_path(P)`.

Additionally, allow users to manually manage map pinning much like we do already for `Programs`, see commit messages for more details.

TODO support the pinning api for PerfEventArray's and AsyncPerfEventArray's

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -492,8 +492,13 @@ impl<'a> BpfLoader<'a> {
             let mut map = match obj.pinning() {
                 PinningType::None => MapData::create(obj, &name, btf_fd)?,
                 PinningType::ByName => {
-                    let path = map_pin_path.as_ref().ok_or(BpfError::NoPinPath)?;
-                    MapData::create_pinned(path, obj, &name, btf_fd)?
+                    // pin maps in /sys/fs/bpf by default to align with libbpf
+                    // behavior https://github.com/libbpf/libbpf/blob/v1.2.2/src/libbpf.c#L2161.
+                    let path = map_pin_path
+                        .as_deref()
+                        .unwrap_or_else(|| Path::new("/sys/fs/bpf"));
+
+                    MapData::create_pinned_by_name(path, obj, &name, btf_fd)?
                 }
             };
             map.finalize()?;
@@ -950,10 +955,6 @@ pub enum BpfError {
         /// The original io::Error
         error: io::Error,
     },
-
-    /// Pinning requested but no path provided
-    #[error("pinning requested but no path provided")]
-    NoPinPath,
 
     /// Unexpected pinning type
     #[error("unexpected pinning type {name}")]
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -398,7 +398,7 @@ impl<'a> BpfLoader<'a> {
             if let Some(btf) = obj.fixup_and_sanitize_btf(features)? {
                 match load_btf(btf.to_bytes(), *verifier_log_level) {
                     Ok(btf_fd) => Some(Arc::new(btf_fd)),
-                    // Only report an error here if the BTF is truely needed, otherwise proceed without.
+                    // Only report an error here if the BTF is truly needed, otherwise proceed without.
                     Err(err) => {
                         for program in obj.programs.values() {
                             match program.section {
@@ -859,6 +859,29 @@ impl Bpf {
     /// ```
     pub fn maps(&self) -> impl Iterator<Item = (&str, &Map)> {
         self.maps.iter().map(|(name, map)| (name.as_str(), map))
+    }
+
+    /// A mutable iterator over all the maps.
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # use std::path::Path;
+    /// # #[derive(thiserror::Error, Debug)]
+    /// # enum Error {
+    /// #     #[error(transparent)]
+    /// #     Bpf(#[from] aya::BpfError),
+    /// #     #[error(transparent)]
+    /// #     Pin(#[from] aya::pin::PinError)
+    /// # }
+    /// # let mut bpf = aya::Bpf::load(&[])?;
+    /// # let pin_path = Path::new("/tmp/pin_path");
+    /// for (_, map) in bpf.maps_mut() {
+    ///     map.pin(pin_path)?;
+    /// }
+    /// # Ok::<(), Error>(())
+    /// ```
+    pub fn maps_mut(&mut self) -> impl Iterator<Item = (&str, &mut Map)> {
+        self.maps.iter_mut().map(|(name, map)| (name.as_str(), map))
     }
 
     /// Returns a reference to the program with the given name.
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -48,6 +48,7 @@
 //! versa. Because of that, all map values must be plain old data and therefore
 //! implement the [Pod] trait.
 use std::{
+    borrow::BorrowMut,
     ffi::CString,
     fmt, io,
     marker::PhantomData,
@@ -170,8 +171,8 @@ pub enum MapError {
     #[error(transparent)]
     SyscallError(#[from] SyscallError),
 
-    /// Could not pin map by name
-    #[error("map `{name:?}` requested pinning by name. pinning failed")]
+    /// Could not pin map
+    #[error("map `{name:?}` requested pinning. pinning failed")]
     PinError {
         /// The map name
         name: Option<String>,
@@ -308,7 +309,88 @@ impl Map {
             Self::Unsupported(map) => map.obj.map_type(),
         }
     }
+
+    /// Pins the map to a BPF filesystem.
+    ///
+    /// When a map is pinned it will remain loaded until the corresponding file
+    /// is deleted. All parent directories in the given `path` must already exist.
+    pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
+        match self {
+            Self::Array(map) => map.pin(path),
+            Self::PerCpuArray(map) => map.pin(path),
+            Self::ProgramArray(map) => map.pin(path),
+            Self::HashMap(map) => map.pin(path),
+            Self::LruHashMap(map) => map.pin(path),
+            Self::PerCpuHashMap(map) => map.pin(path),
+            Self::PerCpuLruHashMap(map) => map.pin(path),
+            Self::PerfEventArray(map) => map.pin(path),
+            Self::SockHash(map) => map.pin(path),
+            Self::SockMap(map) => map.pin(path),
+            Self::BloomFilter(map) => map.pin(path),
+            Self::LpmTrie(map) => map.pin(path),
+            Self::Stack(map) => map.pin(path),
+            Self::StackTraceMap(map) => map.pin(path),
+            Self::Queue(map) => map.pin(path),
+            Self::CpuMap(map) => map.pin(path),
+            Self::DevMap(map) => map.pin(path),
+            Self::DevMapHash(map) => map.pin(path),
+            Self::XskMap(map) => map.pin(path),
+            Self::Unsupported(map) => map.pin(path),
+        }
+    }
 }
+
+// Implements map pinning for different map implementations
+// TODO add support for PerfEventArrays and AsyncPerfEventArrays
+macro_rules! impl_map_pin {
+    ($ty_param:tt {
+        $($ty:ident),+ $(,)?
+    }) => {
+        $(impl_map_pin!(<$ty_param> $ty);)+
+    };
+    (
+      <($($ty_param:ident),*)>
+      $ty:ident
+    ) => {
+            impl<T: BorrowMut<MapData>, $($ty_param: Pod),*> $ty<T, $($ty_param),*>
+            {
+                    /// Pins the map to a BPF filesystem.
+                    ///
+                    /// When a map is pinned it will remain loaded until the corresponding file
+                    /// is deleted. All parent directories in the given `path` must already exist.
+                    pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
+                        let data = self.inner.borrow_mut();
+                        data.pin(path)
+                    }
+            }
+
+    };
+}
+
+impl_map_pin!(() {
+    ProgramArray,
+    SockMap,
+    StackTraceMap,
+    CpuMap,
+    DevMap,
+    DevMapHash,
+    XskMap,
+});
+
+impl_map_pin!((V) {
+    Array,
+    PerCpuArray,
+    SockHash,
+    BloomFilter,
+    Queue,
+    Stack,
+});
+
+impl_map_pin!((K, V) {
+    HashMap,
+    PerCpuHashMap,
+    LpmTrie,
+});
 
 // Implements TryFrom<Map> for different map implementations. Different map implementations can be
 // constructed from different variants of the map enum. Also, the implementation may have type
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -461,7 +461,7 @@ impl MapData {
         Ok(Self { obj, fd })
     }
 
-    pub(crate) fn create_pinned<P: AsRef<Path>>(
+    pub(crate) fn create_pinned_by_name<P: AsRef<Path>>(
         path: P,
         obj: obj::Map,
         name: &str,
@@ -490,7 +490,6 @@ impl MapData {
             }
             Err(_) => {
                 let mut map = Self::create(obj, name, btf_fd)?;
-                let path = path.join(name);
                 map.pin(&path).map_err(|error| MapError::PinError {
                     name: Some(name.into()),
                     error,
--- a/test/integration-test/src/tests/load.rs
+++ b/test/integration-test/src/tests/load.rs
@@ -1,5 +1,7 @@
 use std::{
     convert::TryInto as _,
+    fs::remove_file,
+    path::Path,
     thread,
     time::{Duration, SystemTime},
 };
@@ -40,6 +42,8 @@ fn multiple_btf_maps() {
 
     let map_1: Array<_, u64> = bpf.take_map("map_1").unwrap().try_into().unwrap();
     let map_2: Array<_, u64> = bpf.take_map("map_2").unwrap().try_into().unwrap();
+    let map_pin_by_name: Array<_, u64> =
+        bpf.take_map("map_pin_by_name").unwrap().try_into().unwrap();
 
     let prog: &mut UProbe = bpf.program_mut("bpf_prog").unwrap().try_into().unwrap();
     prog.load().unwrap();
@@ -51,9 +55,75 @@ fn multiple_btf_maps() {
     let key = 0;
     let val_1 = map_1.get(&key, 0).unwrap();
     let val_2 = map_2.get(&key, 0).unwrap();
+    let val_3 = map_pin_by_name.get(&key, 0).unwrap();
 
     assert_eq!(val_1, 24);
     assert_eq!(val_2, 42);
+    assert_eq!(val_3, 44);
+    let map_pin = Path::new("/sys/fs/bpf/map_pin_by_name");
+    assert!(&map_pin.exists());
+
+    remove_file(map_pin).unwrap();
+}
+
+#[test]
+fn pin_lifecycle_multiple_btf_maps() {
+    let mut bpf = Bpf::load(crate::MULTIMAP_BTF).unwrap();
+
+    // "map_pin_by_name" should already be pinned, unpin and pin again later
+    let map_pin_by_name_path = Path::new("/sys/fs/bpf/map_pin_by_name");
+
+    assert!(map_pin_by_name_path.exists());
+    remove_file(map_pin_by_name_path).unwrap();
+
+    // pin and unpin all maps before casting to explicit types
+    for (i, (name, map)) in bpf.maps_mut().enumerate() {
+        // Don't pin system maps or the map that's already pinned by name.
+        if name.contains(".rodata") || name.contains(".bss") {
+            continue;
+        }
+        let map_pin_path = &Path::new("/sys/fs/bpf/").join(i.to_string());
+
+        map.pin(map_pin_path).unwrap();
+
+        assert!(map_pin_path.exists());
+        remove_file(map_pin_path).unwrap();
+    }
+
+    let mut map_1: Array<_, u64> = bpf.take_map("map_1").unwrap().try_into().unwrap();
+    let mut map_2: Array<_, u64> = bpf.take_map("map_2").unwrap().try_into().unwrap();
+    let mut map_pin_by_name: Array<_, u64> =
+        bpf.take_map("map_pin_by_name").unwrap().try_into().unwrap();
+
+    let prog: &mut UProbe = bpf.program_mut("bpf_prog").unwrap().try_into().unwrap();
+    prog.load().unwrap();
+    prog.attach(Some("trigger_bpf_program"), 0, "/proc/self/exe", None)
+        .unwrap();
+
+    trigger_bpf_program();
+
+    let key = 0;
+    let val_1 = map_1.get(&key, 0).unwrap();
+    let val_2 = map_2.get(&key, 0).unwrap();
+    let val_3 = map_pin_by_name.get(&key, 0).unwrap();
+
+    assert_eq!(val_1, 24);
+    assert_eq!(val_2, 42);
+    assert_eq!(val_3, 44);
+
+    let map_1_pin_path = Path::new("/sys/fs/bpf/map_1");
+    let map_2_pin_path = Path::new("/sys/fs/bpf/map_2");
+
+    map_1.pin(map_1_pin_path).unwrap();
+    map_2.pin(map_2_pin_path).unwrap();
+    map_pin_by_name.pin(map_pin_by_name_path).unwrap();
+    assert!(map_1_pin_path.exists());
+    assert!(map_2_pin_path.exists());
+    assert!(map_pin_by_name_path.exists());
+
+    remove_file(map_1_pin_path).unwrap();
+    remove_file(map_2_pin_path).unwrap();
+    remove_file(map_pin_by_name_path).unwrap();
 }
 
 #[no_mangle]
--- a/test/integration-test/src/tests/rbpf.rs
+++ b/test/integration-test/src/tests/rbpf.rs
@@ -34,7 +34,7 @@ fn run_with_rbpf() {
     assert_eq!(vm.execute_program().unwrap(), XDP_PASS);
 }
 
-static mut MULTIMAP_MAPS: [*mut Vec<u64>; 2] = [null_mut(), null_mut()];
+static mut MULTIMAP_MAPS: [*mut Vec<u64>; 3] = [null_mut(); 3];
 
 #[test]
 fn use_map_with_rbpf() {
@@ -47,11 +47,11 @@ fn use_map_with_rbpf() {
     );
 
     // Initialize maps:
-    // - fd: 0xCAFE00 or 0xCAFE01 (the 0xCAFE00 part is used to distinguish fds from indices),
+    // - fd: Bitwise OR of the map_id with 0xCAFE00 (used to distinguish fds from indices),
     // - Note that rbpf does not convert fds into real pointers,
     //   so we keeps the pointers to our maps in MULTIMAP_MAPS, to be used in helpers.
     let mut maps = HashMap::new();
-    let mut map_instances = vec![vec![0u64], vec![0u64]];
+    let mut map_instances = vec![vec![0u64], vec![0u64], vec![0u64]];
     for (name, map) in object.maps.iter() {
         assert_eq!(map.key_size(), size_of::<u32>() as u32);
         assert_eq!(map.value_size(), size_of::<u64>() as u32);
@@ -60,8 +60,14 @@ fn use_map_with_rbpf() {
             aya_obj::generated::bpf_map_type::BPF_MAP_TYPE_ARRAY as u32
         );
 
-        let map_id = if name == "map_1" { 0 } else { 1 };
-        let fd = map_id as std::os::fd::RawFd | 0xCAFE00;
+        let map_id = match name.as_str() {
+            "map_1" => 0,
+            "map_2" => 1,
+            "map_pin_by_name" => 2,
+            n => panic!("Unexpected map: {n}"),
+        };
+
+        let fd = map_id as i32 | 0xCAFE00;
         maps.insert(name.to_owned(), (fd, map.clone()));
 
         unsafe {
@@ -102,18 +108,16 @@ fn use_map_with_rbpf() {
         .expect("Helper failed");
     assert_eq!(vm.execute_program().unwrap(), 0);
 
-    assert_eq!(map_instances[0][0], 24);
-    assert_eq!(map_instances[1][0], 42);
+    assert_eq!(map_instances, [[24], [42], [44]]);
 
     unsafe {
-        MULTIMAP_MAPS[0] = null_mut();
-        MULTIMAP_MAPS[1] = null_mut();
+        MULTIMAP_MAPS.iter_mut().for_each(|v| *v = null_mut());
     }
 }
 
 #[track_caller]
 fn bpf_map_update_elem_multimap(map: u64, key: u64, value: u64, _: u64, _: u64) -> u64 {
-    assert_matches!(map, 0xCAFE00 | 0xCAFE01);
+    assert_matches!(map, 0xCAFE00 | 0xCAFE01 | 0xCAFE02);
     let key = *unsafe { (key as usize as *const u32).as_ref().unwrap() };
     let value = *unsafe { (value as usize as *const u64).as_ref().unwrap() };
     assert_eq!(key, 0);
--- a/aya/src/maps/array/array.rs
+++ b/aya/src/maps/array/array.rs
@@ -31,7 +31,7 @@ use crate::{
 /// ```
 #[doc(alias = "BPF_MAP_TYPE_ARRAY")]
 pub struct Array<T, V: Pod> {
-    inner: T,
+    pub(crate) inner: T,
     _v: PhantomData<V>,
 }
 
--- a/aya/src/maps/array/per_cpu_array.rs
+++ b/aya/src/maps/array/per_cpu_array.rs
@@ -50,7 +50,7 @@ use crate::{
 /// ```
 #[doc(alias = "BPF_MAP_TYPE_PERCPU_ARRAY")]
 pub struct PerCpuArray<T, V: Pod> {
-    inner: T,
+    pub(crate) inner: T,
     _v: PhantomData<V>,
 }
 
--- a/aya/src/maps/array/program_array.rs
+++ b/aya/src/maps/array/program_array.rs
@@ -48,7 +48,7 @@ use crate::{
 /// ```
 #[doc(alias = "BPF_MAP_TYPE_PROG_ARRAY")]
 pub struct ProgramArray<T> {
-    inner: T,
+    pub(crate) inner: T,
 }
 
 impl<T: Borrow<MapData>> ProgramArray<T> {
--- a/aya/src/maps/bloom_filter.rs
+++ b/aya/src/maps/bloom_filter.rs
@@ -36,7 +36,7 @@ use crate::{
 #[doc(alias = "BPF_MAP_TYPE_BLOOM_FILTER")]
 #[derive(Debug)]
 pub struct BloomFilter<T, V: Pod> {
-    inner: T,
+    pub(crate) inner: T,
     _v: PhantomData<V>,
 }
 
--- a/aya/src/maps/hash_map/hash_map.rs
+++ b/aya/src/maps/hash_map/hash_map.rs
@@ -34,7 +34,7 @@ use crate::{
 #[doc(alias = "BPF_MAP_TYPE_LRU_HASH")]
 #[derive(Debug)]
 pub struct HashMap<T, K, V> {
-    inner: T,
+    pub(crate) inner: T,
     _k: PhantomData<K>,
     _v: PhantomData<V>,
 }
--- a/aya/src/maps/hash_map/per_cpu_hash_map.rs
+++ b/aya/src/maps/hash_map/per_cpu_hash_map.rs
@@ -43,7 +43,7 @@ use crate::{
 #[doc(alias = "BPF_MAP_TYPE_LRU_PERCPU_HASH")]
 #[doc(alias = "BPF_MAP_TYPE_PERCPU_HASH")]
 pub struct PerCpuHashMap<T, K: Pod, V: Pod> {
-    inner: T,
+    pub(crate) inner: T,
     _k: PhantomData<K>,
     _v: PhantomData<V>,
 }
--- a/aya/src/maps/lpm_trie.rs
+++ b/aya/src/maps/lpm_trie.rs
@@ -48,7 +48,7 @@ use crate::{
 #[doc(alias = "BPF_MAP_TYPE_LPM_TRIE")]
 #[derive(Debug)]
 pub struct LpmTrie<T, K, V> {
-    inner: T,
+    pub(crate) inner: T,
     _k: PhantomData<K>,
     _v: PhantomData<V>,
 }
--- a/aya/src/maps/queue.rs
+++ b/aya/src/maps/queue.rs
@@ -30,7 +30,7 @@ use crate::{
 /// ```
 #[doc(alias = "BPF_MAP_TYPE_QUEUE")]
 pub struct Queue<T, V: Pod> {
-    inner: T,
+    pub(crate) inner: T,
     _v: PhantomData<V>,
 }
 
--- a/aya/src/maps/sock/sock_hash.rs
+++ b/aya/src/maps/sock/sock_hash.rs
@@ -65,7 +65,7 @@ use crate::{
 /// ```
 #[doc(alias = "BPF_MAP_TYPE_SOCKHASH")]
 pub struct SockHash<T, K> {
-    inner: T,
+    pub(crate) inner: T,
     _k: PhantomData<K>,
 }
 
--- a/aya/src/maps/stack.rs
+++ b/aya/src/maps/stack.rs
@@ -30,7 +30,7 @@ use crate::{
 /// ```
 #[doc(alias = "BPF_MAP_TYPE_STACK")]
 pub struct Stack<T, V: Pod> {
-    inner: T,
+    pub(crate) inner: T,
     _v: PhantomData<V>,
 }
 
--- a/aya/src/maps/stack_trace.rs
+++ b/aya/src/maps/stack_trace.rs
@@ -67,7 +67,7 @@ use crate::{
 #[derive(Debug)]
 #[doc(alias = "BPF_MAP_TYPE_STACK_TRACE")]
 pub struct StackTraceMap<T> {
-    inner: T,
+    pub(crate) inner: T,
     max_stack_depth: usize,
 }
 
--- a/aya/src/maps/xdp/cpu_map.rs
+++ b/aya/src/maps/xdp/cpu_map.rs
@@ -50,7 +50,7 @@ use crate::{
 /// Kernel documentation: <https://docs.kernel.org/next/bpf/map_cpumap.html>
 #[doc(alias = "BPF_MAP_TYPE_CPUMAP")]
 pub struct CpuMap<T> {
-    inner: T,
+    pub(crate) inner: T,
 }
 
 impl<T: Borrow<MapData>> CpuMap<T> {
--- a/aya/src/maps/xdp/dev_map.rs
+++ b/aya/src/maps/xdp/dev_map.rs
@@ -42,7 +42,7 @@ use crate::{
 /// Kernel documentation: <https://docs.kernel.org/next/bpf/map_devmap.html>
 #[doc(alias = "BPF_MAP_TYPE_DEVMAP")]
 pub struct DevMap<T> {
-    inner: T,
+    pub(crate) inner: T,
 }
 
 impl<T: Borrow<MapData>> DevMap<T> {
--- a/aya/src/maps/xdp/dev_map_hash.rs
+++ b/aya/src/maps/xdp/dev_map_hash.rs
@@ -42,7 +42,7 @@ use crate::{
 /// Kernel documentation: <https://docs.kernel.org/next/bpf/map_devmap.html>
 #[doc(alias = "BPF_MAP_TYPE_DEVMAP_HASH")]
 pub struct DevMapHash<T> {
-    inner: T,
+    pub(crate) inner: T,
 }
 
 impl<T: Borrow<MapData>> DevMapHash<T> {
--- a/aya/src/maps/xdp/xsk_map.rs
+++ b/aya/src/maps/xdp/xsk_map.rs
@@ -36,7 +36,7 @@ use crate::{
 /// Kernel documentation: <https://docs.kernel.org/next/bpf/map_xskmap.html>
 #[doc(alias = "BPF_MAP_TYPE_XSKMAP")]
 pub struct XskMap<T> {
-    inner: T,
+    pub(crate) inner: T,
 }
 
 impl<T: Borrow<MapData>> XskMap<T> {
--- a/test/integration-test/bpf/multimap-btf.bpf.c
+++ b/test/integration-test/bpf/multimap-btf.bpf.c
@@ -17,13 +17,24 @@ struct {
   __uint(max_entries, 1);
 } map_2 SEC(".maps");
 
+struct {
+  __uint(type, BPF_MAP_TYPE_ARRAY);
+  __type(key, __u32);
+  __type(value, __u64);
+  __uint(max_entries, 1);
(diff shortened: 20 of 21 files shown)


test/integration-test/src/tests/rbpf.rs
```diff
@@ -48,7 +48,7 @@ fn use_map_with_rbpf() {
     // - Note that rbpf does not convert fds into real pointers,
```
tamird, line 48: the comment above here needs an update? maybe you can generalize it so it's not so tightly coupled to the code

aya/src/maps/mod.rs
```diff
@@ -511,28 +624,56 @@ impl MapData {
...
+
+    pub(crate) fn unpin(&mut self) -> Result<(), io::Error> {
+        let Self {
+            fd: _,
+            paths,
+            obj: _,
+            name: _,
+        } = self;
+        for path in paths.drain(..) {
+            std::fs::remove_file(path)?;
```
tamird, line 675: i think you probably want to try removing them all rather than bailing after the first failure?

aya/src/maps/mod.rs
```diff
@@ -511,28 +624,56 @@ impl MapData {
...
-        let path = path.as_ref().join(name);
-        let path_string = CString::new(path.as_os_str().as_bytes())
-            .map_err(|error| PinError::InvalidPinPath { path, error })?;
+        let path_string = CString::new(path.as_ref().as_os_str().as_bytes()).map_err(|error| {
+            PinError::InvalidPinPath {
+                path: path.as_ref().to_path_buf(),
+                error,
+            }
+        })?;
+        debug!("Attempting to pin map {name} at {path_string:?}");
```
tamird, line 658: why not log `path.display()`?

aya/src/maps/mod.rs
```diff
@@ -511,28 +624,56 @@ impl MapData {
     pub fn from_fd(fd: OwnedFd) -> Result<Self, MapError> {
         let info = bpf_map_get_info_by_fd(fd.as_fd())?;
 
+        let name_bytes = bytes_of_c_char(&info.name);
         Ok(Self {
             obj: parse_map_info(info, PinningType::None),
             fd: fd.into_raw_fd(),
-            pinned: false,
+            name: std::str::from_utf8(name_bytes)
+                .expect("unable to parse map name from bpf_map_info")
```
tamird, line 632: why is panicking here OK?

aya/src/maps/mod.rs
```diff
@@ -496,10 +604,15 @@ impl MapData {
...
         let info = bpf_map_get_info_by_fd(fd.as_fd())?;
 
+        let name_bytes = bytes_of_c_char(&info.name);
+
         Ok(Self {
             obj: parse_map_info(info, PinningType::ByName),
             fd: fd.into_raw_fd(),
-            pinned: true,
+            name: std::str::from_utf8(name_bytes)
+                .expect("unable to parse map name from bpf_map_info")
```
tamird, line 613: why is panicking here OK?

aya/src/programs/mod.rs
```diff
@@ -977,17 +977,7 @@ impl ProgramInfo {
 
     /// The name of the program as was provided when it was load. This is limited to 16 bytes
     pub fn name(&self) -> &[u8] {
```
dave-tucker, line 979: This is probably one for Alessandro, but why are we returning &[u8] instead of `std::ffi::Cstr`

tamird: I think that's because the string is not null terminated if it is exactly 16 bytes long.

test/integration-test/src/tests/rbpf.rs
```diff
@@ -101,6 +107,7 @@ fn use_map_with_rbpf() {
 
     assert_eq!(map_instances[0][0], 24);
```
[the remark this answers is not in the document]
astoycos: Hrm so rbpf is just a VM for running the program, the program (multimap-bpf.bpf.c) just updates it's maps with those values 

```
SEC("tracepoint")
int bpf_prog(void *ctx) {
  __u32 key = 0;
  __u64 twenty_four = 24;
  __u64 forty_two = 42;
  __u64 forty_four = 44;

  bpf_map_update_elem(&map_1, &key, &twenty_four, BPF_ANY);
  bpf_map_update_elem(&map_2, &key, &forty_two, BPF_ANY);
  bpf_map_update_elem(&map_pin_by_name, &key, &forty_four, BPF_ANY);
  return 0;
}
```

Here we're asserting that the program executed and successfully updated it's maps.

I don't think checking each map pointer here is a bad thing personally

test/integration-test/src/tests/rbpf.rs
```diff
@@ -101,6 +107,7 @@ fn use_map_with_rbpf() {
...
     assert_eq!(map_instances[1][0], 42);
+    assert_eq!(map_instances[2][0], 44);
 
     unsafe {
         MULTIMAP_MAPS[0] = null_mut();
```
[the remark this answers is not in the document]
astoycos: Yep, sounds good!

test/integration-test/src/tests/rbpf.rs
astoycos: Updated and gave generalizing a shot

test/integration-test/src/tests/load.rs
```diff
@@ -45,9 +47,80 @@ fn multiple_btf_maps() {
...
+    let mut map_1: Array<_, u64> = bpf.take_map("map_1").unwrap().try_into().unwrap();
+    let mut map_2: Array<_, u64> = bpf.take_map("map_2").unwrap().try_into().unwrap();
+    let mut map_pin_by_name: Array<_, u64> =
+        bpf.take_map("map_pin_by_name").unwrap().try_into().unwrap();
+
+    let prog: &mut TracePoint = bpf.program_mut("bpf_prog").unwrap().try_into().unwrap();
+    prog.load().unwrap();
+    prog.attach("sched", "sched_switch").unwrap();
+
+    thread::sleep(time::Duration::from_secs(3));
```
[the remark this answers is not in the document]
astoycos: Done converted to existing way of doing things.

aya/src/util.rs
```diff
@@ -337,6 +337,18 @@ pub(crate) fn bytes_of_slice<T: Pod>(val: &[T]) -> &[u8] {
     unsafe { slice::from_raw_parts(val.as_ptr().cast(), size) }
 }
 
+pub(crate) fn bytes_of_c_char(val: &[std::os::raw::c_char]) -> &[u8] {
+    let length = val
```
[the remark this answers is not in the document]
astoycos: I'm alright with that, makes more sense to me I'm also adding a comment

```
    // length is the end of the C-style string (nul byte).
```

the the only difference is with your suggestion is that we default to `val.len()` but the C char should ALWAYS end with a NULL pointer so I'm going to leave it as `unwrap_or(0)`

aya/src/maps/mod.rs
astoycos: True, sounds good.

aya/src/maps/mod.rs
```diff
@@ -511,28 +624,56 @@ impl MapData {
...
-            return Err(PinError::AlreadyPinned { name: name.into() });
+        let Self {
+            fd,
+            paths,
+            obj: _,
+            name,
+        } = self;
+        if paths.contains(&path.as_ref().to_path_buf()) {
+            return Err(PinError::AlreadyPinned {
+                name: name.to_string(),
```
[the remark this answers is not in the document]
astoycos: Nope 
```
the trait `From<&mut String>` is not implemented for `String`
```
 :/

aya/src/util.rs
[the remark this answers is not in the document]
tamird: I don't think that's true. See the other discussion with @dave-tucker above: in the case where the slice is backed by an array, the string is allowed to take up the whole array without a NUL terminator.

Perhaps this method should take an array reference rather than a slice so that this is clearer.

`fn bytes_of_c_char<const N: usize>(vale: &[std::os::raw::c_char; N]) -> &[u8] { ... }`

test/integration-test/src/tests/rbpf.rs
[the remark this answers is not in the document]
tamird: Yeah, I think you should check them all, I just mean can we do it all at once?

tamird (commented):
Did you forget to push?

aya/src/maps/mod.rs
```diff
@@ -511,28 +624,56 @@ impl MapData {
...
-        let Self { fd, pinned, obj: _ } = self;
-        if *pinned {
-            return Err(PinError::AlreadyPinned { name: name.into() });
+        let Self {
+            fd,
+            paths,
+            obj: _,
+            name,
+        } = self;
+        if paths.contains(&path.as_ref().to_path_buf()) {
```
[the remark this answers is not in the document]
astoycos: Sure changed to a hashSet

aya/src/maps/mod.rs
astoycos: I made an dedicated error type instead.

aya/src/programs/mod.rs
astoycos: Is that true? from what I've read the string is always null terminated

https://stackoverflow.com/questions/47288881/are-c-constant-character-strings-always-null-terminated

test/integration-test/src/tests/rbpf.rs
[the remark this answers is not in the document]
astoycos: Done!

aya/src/bpf.rs
```diff
@@ -849,6 +854,29 @@ impl Bpf {
         self.maps.iter().map(|(name, map)| (name.as_str(), map))
     }
 
+    /// An iterator mutably referencing all the maps.
```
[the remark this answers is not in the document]

astoycos:
Pushed in a new commit @tamird so it's easier to check diff, I'll squash everything whenever you approve it

astoycos:
Sorry usually I respond to comments as I fix things and then resolve them after I push the changes.

aya/src/util.rs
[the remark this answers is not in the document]
astoycos: I haven't done this yet but if we can get some confirmation on the all c strings end in null thing I can do this 👍

astoycos:
I'll bless the public API on squash 👍

aya/src/util.rs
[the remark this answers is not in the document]
tamird: I don't understand what you're asking for. C is the wild west - strings usually terminate in NULL, but not always. I grepped around for BPF_OBJ_NAME_LEN in the kernel, and I think it *is* NULL-terminated (https://github.com/search?q=repo%3Atorvalds%2Flinux%20BPF_OBJ_NAME_LEN&type=code), so dave is probably right and this should return a CStr.

But we should make that change in another PR.

tamird (changes requested):
> Sorry usually I respond to comments as I fix things and then resolve them / re-request review after I push the changes.

You can "start a review" instead of publishing the comments as you go. That would be friendlier to reviewers.

tamird (commented):
Moving my review to Reviewable because GitHub reviews are so painful :(

> Reviewed 13 of 19 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783#-:-NeFMuoS2UH5kyrrkL6M:b-nj6jmn)* status: all files reviewed, 29 unresolved discussions (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`aya/src/maps/mod.rs` line 519 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMWlS1-_OV3feHcrN:-NeFMWlS1-_OV3feHcrO:b-z58jh1) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L519)):*
> ```XML
>     pub fn create(
>         obj: obj::Map,
>         name: &str,
> ```

we should maybe take a String here so that the caller can control allocation.
___
*[`aya/src/maps/mod.rs` line 554 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMc9C9iX3BK_-s_JK:-NeFMc9C9iX3BK_-s_JL:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L554)):*
> ```XML
>         path: P,
>         obj: obj::Map,
>         name: &str,
> ```

ditto
___
*[`aya/src/maps/mod.rs` line 654 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMkQdDSvrXOKOF8dk:-NeFMkQdDSvrXOKOF8dl:bwxbxrx) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L654)):*
> ```XML
>         if paths.contains(&path.as_ref().to_path_buf()) {
>             return Err(PinError::AlreadyPinned {
>                 name: name.to_string(),
> ```

nit: .clone() because the thing is already a String
___
*[`aya/src/maps/mod.rs` line 693 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMnvM9Y6USCtn7PaX:-NeFMnvM9Y6USCtn7PaY:bhmalv8) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L693)):*
> ```XML
>             let (paths, errors) = errors.into_iter().unzip();
>             return Err(PinError::UnpinError {
>                 name: name.to_string(),
> ```

.clone

aya/src/maps/mod.rs
```diff
@@ -397,8 +508,8 @@ pub(crate) fn check_v_size<V>(map: &MapData) -> Result<(), MapError> {
 pub struct MapData {
     pub(crate) obj: obj::Map,
     pub(crate) fd: RawFd,
-    /// Indicates if this map has been pinned to bpffs
-    pub pinned: bool,
+    pub(crate) name: String,
+    pub(crate) paths: HashSet<PathBuf>,
```
dave-tucker, line 512: We shouldn't track paths here either. See #789.
It's fragile because "things that are not aya ™️ " can interact with the filesystem rendering the check pointless.
BPF_OBJ_PIN returns -EEXISTS in the case of pinning to a location that already exists... and there are already plenty of ways to check if a path exists before pinning.

aya/src/maps/mod.rs
```diff
@@ -279,8 +285,113 @@ impl Map {
...
+            Self::LpmTrie(map) => map.pin(path),
+            Self::Stack(map) => map.pin(path),
+            Self::StackTraceMap(map) => map.pin(path),
+            Self::Queue(map) => map.pin(path),
+            Self::Unsupported(map) => map.pin(path),
+        }
+    }
+
+    /// Removes the pinned map from a BPF filesystem.
+    pub fn unpin(&mut self) -> Result<(), PinError> {
```
dave-tucker, line 317: `unpin` should also take a `Path`, or just not be exposed since you can easily rm it from the filesystem yourself.

aya/src/maps/mod.rs
```diff
@@ -511,28 +629,73 @@ impl MapData {
...
         bpf_pin_object(*fd, &path_string).map_err(|(_, io_error)| SyscallError {
             call: "BPF_OBJ_PIN",
             io_error,
         })?;
-        *pinned = true;
+        paths.insert(path.as_ref().to_path_buf());
+        Ok(())
+    }
+
+    pub(crate) fn unpin(&mut self) -> Result<(), PinError> {
```
dave-tucker, line 675: Per comment above, this needs to take a `Path`.
We shouldn't assume that all pins we know about need unpinning.

dave-tucker (changes requested):

astoycos:
Waiting for https://github.com/aya-rs/aya/pull/790 to merge then I'll build back on top of it

test/integration-test/src/tests/load.rs
```diff
@@ -46,9 +48,80 @@ fn multiple_btf_maps() {
...
+    let mut map_1: Array<_, u64> = bpf.take_map("map_1").unwrap().try_into().unwrap();
+    let mut map_2: Array<_, u64> = bpf.take_map("map_2").unwrap().try_into().unwrap();
+    let mut map_pin_by_name: Array<_, u64> =
+        bpf.take_map("map_pin_by_name").unwrap().try_into().unwrap();
+
+    let prog: &mut UProbe = bpf.program_mut("bpf_prog").unwrap().try_into().unwrap();
+    prog.load().unwrap();
+    prog.attach(Some("trigger_bpf_program"), 0, "/proc/self/exe", None)
+        .unwrap();
+    trigger_bpf_program();
```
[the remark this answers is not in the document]
astoycos: Done

test/integration-test/src/tests/rbpf.rs
```diff
@@ -99,18 +105,16 @@ fn use_map_with_rbpf() {
         .expect("Helper failed");
     assert_eq!(vm.execute_program().unwrap(), 0);
 
-    assert_eq!(map_instances[0][0], 24);
-    assert_eq!(map_instances[1][0], 42);
+    assert_eq!(map_instances, vec![vec![24], vec![42], vec![44]]);
```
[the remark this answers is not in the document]
astoycos: Done

astoycos (commented):
Much of this is not relevant anymore with @dave-tucker's changes in #790 

I tried to leave any relevant comments un-resolved  @tamird 

Sorry for the slow feedback loop here!

tamird (changes requested):
> Reviewed 19 of 19 files at r3, all commit messages.
> Dismissed @dave-tucker from a discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: all files reviewed, 15 unresolved discussions (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`aya/src/bpf.rs` line 480 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexdE_R6NhKjqzjoeR9:-NexdE_R6NhKjqzjoeRA:bq3a6my) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/aya/src/bpf.rs#L480)):*
> ```XML
>                 PinningType::None => MapData::create(obj, &name, btf_fd)?,
>                 PinningType::ByName => {
>                     let path = match &map_pin_path {
> ```

this is allocating an owned object and then passing it to create_pinned_by_name which only knows that it is AsRef<Path>. How about:

```
let path = map_pin_path.as_ref().unwrap_or_else(|| Path::new("/sys/fs/bpf"));
```
___
*[`aya/src/bpf.rs` line 480 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexfKbK063CS-lFqgZM:-NexfKbK063CS-lFqgZN:b4ozhdt) ([raw file](https://github.com/aya-rs/aya/blob/42fd82e32b85715df78508ef77f79eb8c2edd890/aya/src/bpf.rs#L480)):*
> ```XML
>                 PinningType::None => MapData::create(obj, &name, btf_fd)?,
>                 PinningType::ByName => {
>                     let path = map_pin_path.as_ref().ok_or(BpfError::NoPinPath)?;
> ```

this error variant is no longer ever used
___
*[`aya/src/maps/mod.rs` line 472 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeB7LAXPTemmCttrJ:-NexeB7LAXPTemmCttrK:b-u3kodm) ([raw file](https://github.com/aya-rs/aya/blob/42fd82e32b85715df78508ef77f79eb8c2edd890/aya/src/maps/mod.rs#L472)):*
> ```XML
>             Err(_) => {
>                 let mut map = Self::create(obj, name, btf_fd)?;
>                 let path = path.join(name);
> ```

looks like you reverted most of the changes except this?
___
*[`aya/src/maps/mod.rs` line 694 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeI7IEatA9ehjVSKQ:-NexeI7IEatA9ehjVSKR:bghmbiz) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/aya/src/maps/mod.rs#L694)):*
> ```XML
> 
>     pub(crate) fn unpin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), io::Error> {
>         let Self { fd: _, obj: _ } = self;
> ```

this is surprising - we aren't consulting self at all. should we? is there a way to confirm that `path` is in fact a pin of `self`?
___
*[`test/integration-test/src/tests/load.rs` line 56 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexcQXX8nkoCIzwT0dx:-NexcQXX8nkoCIzwT0dy:b-85bd78) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L56)):*
> ```XML
>     assert_eq!(val_2, 42);
>     assert_eq!(val_3, 44);
>     let map_pin = PathBuf::from("/sys/fs/bpf/map_pin_by_name");
> ```

```
let map_pin = Path::new(".....")
```
no need for an owned thing here (and the clone below is not necessary)
___
*[`test/integration-test/src/tests/load.rs` line 68 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeeMK3h4gH2vKdma0:-NexeeMK3h4gH2vKdma1:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L68)):*
> ```XML
>     // "map_pin_by_name" should already be pinned, unpin and pin again later
>     let map_pin_by_name = bpf.map_mut("map_pin_by_name").unwrap();
>     let map_pin_by_name_path = &PathBuf::from("/sys/fs/bpf/map_pin_by_name");
> ```

ditto
___
*[`test/integration-test/src/tests/load.rs` line 80 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-Nexehnv9wta5RySj7sn:-Nexehnv9wta5RySj7so:b45n1ow) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L80)):*
> ```XML
>             continue;
>         }
>         let map_pin_path = &PathBuf::from("/sys/fs/bpf/").join(i.to_string());
> ```

ditto (join is a method on Path)f
___
*[`test/integration-test/src/tests/load.rs` line 110 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexelEu-Cl2atweBIVv:-NexelEu-Cl2atweBIVw:b-dkexcg) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L110)):*
> ```XML
>     assert_eq!(val_3, 44);
> 
>     let map_1_pin_path = &PathBuf::from("/sys/fs/bpf/map_1");
> ```

both of these as well
___
*[`test/integration-test/src/tests/rbpf.rs` line 108 at r2](https://reviewable.io/reviews/aya-rs/aya/783#gh-1325010542:-NexcZNs8pB9GSNHMOdv:bp1oj5x) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/test/integration-test/src/tests/rbpf.rs#L108)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Done
</blockquote></details>

odd trailing comma left behind?
___
*[`test/integration-test/src/tests/rbpf.rs` line 34 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeqmyDOZlhpQ3BcNU:-NexeqmyDOZlhpQ3BcNV:br7blmf) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/rbpf.rs#L34)):*
> ```XML
> }
> 
> static mut MULTIMAP_MAPS: [*mut Vec<u64>; 3] = [null_mut(), null_mut(), null_mut()];
> ```

can this be written `[null_mut(); 3]`?
___
*[`test/integration-test/src/tests/rbpf.rs` line 60 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-Nexf-Bx8_ZV7v-6fR6q:-Nexf-Bx8_ZV7v-6fR6r:b-yja1g3) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/rbpf.rs#L60)):*
> ```XML
>         );
> 
>         let map_id = if name == "map_1" {
> ```

nit: `match map_id` and blow up if the name is anything other than map_{1,2,3}?
___
*[`aya/src/pin.rs` line 42 at r2](https://reviewable.io/reviews/aya-rs/aya/783#gh-1325008447:-NexceBx9tO0GDmVl6PG:b7a1vaw) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/pin.rs#L42)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

I did, adds a bit more complexity and I'm sure you'll see some stuff I did wrong there...
</blockquote></details>

I think this gone now because paths are no longer tracked?

astoycos (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: all files reviewed, 15 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @tamird)

___
*[`aya/src/bpf.rs` line 480 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexdE_R6NhKjqzjoeR9:-NexmqMi24ySso-8Sxvd:bigaov3) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/aya/src/bpf.rs#L480)):*

this is allocating an owned object and then passing it to create_pinned_by_name which only knows that it is AsRef<Path>. How about:

```
let path = map_pin_path.as_ref().unwrap_or_else(|| Path::new("/sys/fs/bpf"));
```
</blockquote></details>

Good point, note it needed to be .as_deref() though or else 

```
error[E0308]: mismatched types
   --> aya/src/bpf.rs:482:72
    |
482 |                     let path = map_pin_path.as_ref().unwrap_or_else(|| Path::new("/sys/fs/bpf"));
    |                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^ expected `&PathBuf`, found `&Path`
    |
    = note: expected reference `&PathBuf`
               found reference `&Path`
```
___
*[`aya/src/bpf.rs` line 480 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexfKbK063CS-lFqgZM:-Nexn92m71h_Su53blb0:b1f4ww8) ([raw file](https://github.com/aya-rs/aya/blob/42fd82e32b85715df78508ef77f79eb8c2edd890/aya/src/bpf.rs#L480)):*

this error variant is no longer ever used
</blockquote></details>

Yep good catch!
___
*[`aya/src/maps/mod.rs` line 519 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMWlS1-_OV3feHcrN:-NexodUHB7FNnM8X_o1x:bb6un1j) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L519)):*

we should maybe take a String here so that the caller can control allocation.
</blockquote></details>

Done
___
*[`aya/src/maps/mod.rs` line 554 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMc9C9iX3BK_-s_JK:-Nexof53E9lxLEtPhiQ-:bb6un1j) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L554)):*

ditto
</blockquote></details>

Done
___
*[`aya/src/maps/mod.rs` line 472 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeB7LAXPTemmCttrJ:-NexpDyP86xhJYxz_dAQ:b76docf) ([raw file](https://github.com/aya-rs/aya/blob/42fd82e32b85715df78508ef77f79eb8c2edd890/aya/src/maps/mod.rs#L472)):*

looks like you reverted most of the changes except this?
</blockquote></details>

We already append `name` to the path earlier, no need to do it twice 
___
*[`aya/src/maps/mod.rs` line 694 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeI7IEatA9ehjVSKQ:-NexpZ7j3rfpuD83B9ti:b-suxw9m) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/aya/src/maps/mod.rs#L694)):*

this is surprising - we aren't consulting self at all. should we? is there a way to confirm that `path` is in fact a pin of `self`?
</blockquote></details>

Good point... I'm not sure what to do here, we could get rid of the unpin API completely I just like the wrapper for ease of use.

I don't think there's an easy way to confirm path is self without a bit more diff

i.e we could load `from_pin(<PATH>)` before cleaning up but I don't think that's worth the extra syscall 
___
*[`test/integration-test/src/tests/load.rs` line 56 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexcQXX8nkoCIzwT0dx:-NexrfwdALkuWCl-uUB3:b-qzvpwr) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L56)):*

```
let map_pin = Path::new(".....")
```
no need for an owned thing here (and the clone below is not necessary)
</blockquote></details>

Yep good catch.
___
*[`test/integration-test/src/tests/load.rs` line 68 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeeMK3h4gH2vKdma0:-NexriNOD5WkSnT-Cciv:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L68)):*

ditto
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/load.rs` line 80 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-Nexehnv9wta5RySj7sn:-NexrioVDiJX2e-tqBbj:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L80)):*

ditto (join is a method on Path)f
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/load.rs` line 110 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexelEu-Cl2atweBIVv:-NexrkiSEj4SI2DBTrZV:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L110)):*

both of these as well
</blockquote></details>

Done.

astoycos:
I gave in an used reviewable @tamird 🍺

tamird (changes requested):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: all files reviewed, 10 unresolved discussions (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`aya/src/maps/mod.rs` line 519 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMWlS1-_OV3feHcrN:-Ney8VZ_D20gyOka9FYE:b-q6z43o) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L519)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Done
</blockquote></details>

I think this comment made sense earlier when this function's API was changing. Now that it is not, we should leave it as-is to reduce breakage.
___
*[`aya/src/maps/mod.rs` line 554 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMc9C9iX3BK_-s_JK:-Ney8z908edTiO-qJbI6:bk5lxas) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L554)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Done
</blockquote></details>

I think revert this (i had retracted these comments)
___
*[`aya/src/maps/mod.rs` line 472 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeB7LAXPTemmCttrJ:-Ney8vsCC-1QbULGZ3Nt:b-s1lnx5) ([raw file](https://github.com/aya-rs/aya/blob/42fd82e32b85715df78508ef77f79eb8c2edd890/aya/src/maps/mod.rs#L472)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

We already append `name` to the path earlier, no need to do it twice
</blockquote></details>

are you saying this was a bug? do we have a test?
___
*[`aya/src/maps/mod.rs` line 694 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeI7IEatA9ehjVSKQ:-Ney959F0DahJFrGGSdl:bliybwj) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/aya/src/maps/mod.rs#L694)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Good point... I'm not sure what to do here, we could get rid of the unpin API completely I just like the wrapper for ease of use.

I don't think there's an easy way to confirm path is self without a bit more diff

i.e we could load `from_pin(<PATH>)` before cleaning up but I don't think that's worth the extra syscall
</blockquote></details>

I don't think we should call something unpin that just removes arbitrary paths from the filesystem. you could use a witness type (i.e. some newtype that can only be obtained from a call to `pin`), but this is not appropriate as written IMO.
___
*[`aya/src/maps/mod.rs` line 528 at r4](https://reviewable.io/reviews/aya-rs/aya/783#-Ney8bMq3ixj6eJTp2gu:-Ney8bMq3ixj6eJTp2gv:b-qswzdp) ([raw file](https://github.com/aya-rs/aya/blob/e90e73ace6e5aac8d95c3863335033b6ff8caef1/aya/src/maps/mod.rs#L528)):*
> ```XML
>     ) -> Result<Self, MapError> {
>         let c_name =
>             CString::new(name.clone()).map_err(|_| MapError::InvalidName { name: name.clone() })?;
> ```

these two clones are negating the point of changing this function to take String - but as noted above we shouldn't change the API
___
*[`test/integration-test/src/tests/load.rs` line 80 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-Nexehnv9wta5RySj7sn:-Ney9M0T8OyE_q6JiJma:b-7kihh7) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L80)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

s/&//

astoycos (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: all files reviewed, 9 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @tamird)

*[a discussion](https://reviewable.io/reviews/aya-rs/aya/783#gh-1321660415:-NeyLfaWE6Z8CKnZdJKZ:b-896fix) (no related file):*

I don't understand what you're asking for. C is the wild west - strings usually terminate in NULL, but not always. I grepped around for BPF_OBJ_NAME_LEN in the kernel, and I think it *is* NULL-terminated (https://github.com/search?q=repo%3Atorvalds%2Flinux%20BPF_OBJ_NAME_LEN&type=code), so dave is probably right and this should return a CStr.

But we should make that change in another PR.
</blockquote></details>

Done.

___
___
*[`aya/src/maps/mod.rs` line 519 at r2](https://reviewable.io/reviews/aya-rs/aya/783#-NeFMWlS1-_OV3feHcrN:-NeyAfMr6hlA9eS90d9E:b2ngthx) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L519)):*

I think this comment made sense earlier when this function's API was changing. Now that it is not, we should leave it as-is to reduce breakage.
</blockquote></details>

+1 
___
*[`aya/src/maps/mod.rs` line 675 at r2](https://reviewable.io/reviews/aya-rs/aya/783#gh-1330073478:-NeyLk_8AGGEViegj89f:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/aya/src/maps/mod.rs#L675)):*

Per comment above, this needs to take a `Path`.
We shouldn't assume that all pins we know about need unpinning.
</blockquote></details>

Done.
___
*[`aya/src/maps/mod.rs` line 472 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeB7LAXPTemmCttrJ:-NeyAqWf-Yi9wN1s3jOU:b-bmsurr) ([raw file](https://github.com/aya-rs/aya/blob/42fd82e32b85715df78508ef77f79eb8c2edd890/aya/src/maps/mod.rs#L472)):*

are you saying this was a bug? do we have a test?
</blockquote></details>

We didn't but we do now 

i.e https://github.com/aya-rs/aya/pull/783/files#diff-8bb6ea6c207f19d27fe77eeb4e24f08a975fb5a3bed2054a192c187d83d91f4fR62
___
*[`aya/src/maps/mod.rs` line 694 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-NexeI7IEatA9ehjVSKQ:-NeyGElxE_mIrwoEARBU:bts5ez2) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/aya/src/maps/mod.rs#L694)):*

I don't think we should call something unpin that just removes arbitrary paths from the filesystem. you could use a witness type (i.e. some newtype that can only be obtained from a call to `pin`), but this is not appropriate as written IMO.
</blockquote></details>

I will go ahead and remove it then 
___
*[`aya/src/maps/mod.rs` line 528 at r4](https://reviewable.io/reviews/aya-rs/aya/783#-Ney8bMq3ixj6eJTp2gu:-NeyAhreCchM4hepgTLC:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e90e73ace6e5aac8d95c3863335033b6ff8caef1/aya/src/maps/mod.rs#L528)):*

these two clones are negating the point of changing this function to take String - but as noted above we shouldn't change the API
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/load.rs` line 80 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-Nexehnv9wta5RySj7sn:-NeyHqEg2UXRQgJ4miAW:b-qxzegx) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L80)):*

s/&//
</blockquote></details>

Join is a method on path but it turns it into an owned PathBuf https://doc.rust-lang.org/std/path/struct.Path.html#method.join

Then we get a bunch of "borrow of moved value" errors

using an &Path which is coerced into an &PathBuf on .join fixes this
___
*[`test/integration-test/src/tests/rbpf.rs` line 108 at r2](https://reviewable.io/reviews/aya-rs/aya/783#gh-1325010542:-NeyJuhp8M6BaKEqLMDo:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/9b5fce92d9f6cfc7b1965f31f2afb6f027b6bbac/test/integration-test/src/tests/rbpf.rs#L108)):*

odd trailing comma left behind?
</blockquote></details>

Done.

tamird (approved):
> Reviewed 2 of 19 files at r5, 17 of 17 files at r6, all commit messages.
> Dismissed @dave-tucker from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod, @astoycos, and @dave-tucker)

___
*[`test/integration-test/src/tests/load.rs` line 80 at r3](https://reviewable.io/reviews/aya-rs/aya/783#-Nexehnv9wta5RySj7sn:-NezabKa392UmViMtALR:bg2r84f) ([raw file](https://github.com/aya-rs/aya/blob/e1ba78e9dfc9fd6f4c344f32b2ff08cfdb960db0/test/integration-test/src/tests/load.rs#L80)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Join is a method on path but it turns it into an owned PathBuf https://doc.rust-lang.org/std/path/struct.Path.html#method.join

Then we get a bunch of "borrow of moved value" errors

using an &Path which is coerced into an &PathBuf on .join fixes this
</blockquote></details>

Ah I missed the .join.
___
*[`test/integration-test/src/tests/rbpf.rs` line 64 at r6](https://reviewable.io/reviews/aya-rs/aya/783#-Nezaj-R7DltwUwe4ylP:-Nezaj-R7DltwUwe4ylQ:brzst1b) ([raw file](https://github.com/aya-rs/aya/blob/9d06a812b69bfbd78ee54c16b3360a7ce5c27694/test/integration-test/src/tests/rbpf.rs#L64)):*
> ```XML
>             "map_2" => 1,
>             "map_pin_by_name" => 2,
>             _ => panic!("Unexpected map name: {}", name),
> ```

Nit: s/_/name/ to clarify that it's not discarded

dave-tucker (approved):

astoycos (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: 17 of 26 files reviewed, all discussions resolved (waiting on @alessandrod and @tamird)

___
*[`test/integration-test/src/tests/rbpf.rs` line 64 at r6](https://reviewable.io/reviews/aya-rs/aya/783#-Nezaj-R7DltwUwe4ylP:-NfBLIXC7hq4yur3pe_E:bb6un1j) ([raw file](https://github.com/aya-rs/aya/blob/9d06a812b69bfbd78ee54c16b3360a7ce5c27694/test/integration-test/src/tests/rbpf.rs#L64)):*

Nit: s/_/name/ to clarify that it's not discarded
</blockquote></details>

Done

astoycos:
@alessandrod Just need a final pass on the API from you please 👼

tamird (approved):
> Reviewed 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

___
*[`test/integration-test/src/tests/rbpf.rs` line 64 at r6](https://reviewable.io/reviews/aya-rs/aya/783#-Nezaj-R7DltwUwe4ylP:-NfBc75D-nptAs5Ontru:b-crbkek) ([raw file](https://github.com/aya-rs/aya/blob/9d06a812b69bfbd78ee54c16b3360a7ce5c27694/test/integration-test/src/tests/rbpf.rs#L64)):*
<details><summary><i>Previously, astoycos (Andrew Stoycos) wrote…</i></summary><blockquote>

Done
</blockquote></details>

Sorry I meant replace the underscore with "name"

astoycos (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: 23 of 26 files reviewed, all discussions resolved (waiting on @alessandrod and @tamird)

___
*[`test/integration-test/src/tests/rbpf.rs` line 64 at r6](https://reviewable.io/reviews/aya-rs/aya/783#-Nezaj-R7DltwUwe4ylP:-NfBlBHf9Nh-Irhv9krU:b-oqzlp3) ([raw file](https://github.com/aya-rs/aya/blob/9d06a812b69bfbd78ee54c16b3360a7ce5c27694/test/integration-test/src/tests/rbpf.rs#L64)):*

Sorry I meant replace the underscore with "name"
</blockquote></details>

My bad fixed!

tamird (approved):
> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

astoycos:
Poke @alessandrod For a quick look, TIA!

aya/src/maps/mod.rs
```diff
@@ -309,8 +310,93 @@ impl Map {
             Self::Unsupported(map) => map.obj.map_type(),
         }
     }
+
+    /// Pins the map to a BPF filesystem.
+    ///
+    /// When a BPF map is pinned to a BPF filesystem it will remain loaded after
```
alessandrod, line 316:
```
Pins the map to a BPF filesystem.

When a map is pinned it will remain loaded until the corresponding file is deleted. All parent directories in the given `path` must already exist.  
```

aya/src/maps/mod.rs
```diff
@@ -309,8 +310,93 @@ impl Map {
...
+            Self::DevMap(map) => map.pin(path),
+            Self::DevMapHash(map) => map.pin(path),
+            Self::XskMap(map) => map.pin(path),
+            Self::Unsupported(map) => map.pin(path),
+        }
+    }
 }
 
+// Implements map pinning for different map implementations
+// TODO add support for PerfEventArrays and AsyncPerfEventArrays
```
alessandrod, line 347: Why this todo? What's missing to implement it?

aya/src/maps/mod.rs
```diff
@@ -309,8 +310,93 @@ impl Map {
...
+    }) => {
+        $(impl_map_pin!(<$ty_param> $ty);)+
+    };
+    (
+      <($($ty_param:ident),*)>
+      $ty:ident
+    ) => {
+            impl<T: BorrowMut<MapData>, $($ty_param: Pod),*> $ty<T, $($ty_param),*>
+            {
+                    /// Pins the map to a BPF filesystem.
```
alessandrod, line 360: Update this comment too pls

aya/src/bpf.rs
```diff
@@ -856,6 +861,29 @@ impl Bpf {
         self.maps.iter().map(|(name, map)| (name.as_str(), map))
     }
 
+    /// An iterator over all the maps that allows modification.
```
alessandrod, line 864: A mutable iterator over all the maps.

alessandrod (commented):
Just a few nits!

aya/src/maps/mod.rs
astoycos: Done!

aya/src/bpf.rs
astoycos: Done! I think this is what I had before all the reviews 😜

aya/src/maps/mod.rs
```diff
@@ -308,8 +309,89 @@ impl Map {
             Self::Unsupported(map) => map.obj.map_type(),
         }
     }
+
+    /// Pins the map to a BPF filesystem.
+
```
alessandrod, line 314: missing ///

tamird (approved):
> Reviewed 10 of 10 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @astoycos)

aya/src/maps/mod.rs
astoycos: Done

astoycos:
> > Dismissed @alessandrod from 3 discussions.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)_ status: all files reviewed, 3 unresolved discussions (waiting on @astoycos)
> 
> _[`aya/src/maps/mod.rs` line 358 at r10](https://reviewable.io/reviews/aya-rs/aya/783#-NgPK6O08sdN__HJ3wSb:-NgPK6O08sdN__HJ3wSc:bps1s89) ([raw file](https://github.com/aya-rs/aya/blob/a1a188cad548811e85b1ead1f292a273cdc5f693/aya/src/maps/mod.rs#L358)):_
> 
> > ```
> >             {
> >                     /// Pins the map to a BPF filesystem.
> > ```
> 
> here too

Done!

tamird (commented):
> Reviewed 21 of 21 files at r12, all commit messages.
> Dismissed @alessandrod from a discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/783)* status: all files reviewed, 1 unresolved discussion (waiting on @astoycos)

PR #785, opened 2023-09-11
aya/ci: fixup broken debian kernel image url

astoycos:


tamird (approved):

tamird:
These are kernel image URLs - not source.

PR #789, opened 2023-09-18
aya: Remove MapData::pinned

dave-tucker:
BPF objects can be pinned mutliple times, to multiple different places. Tracking whether or not a map is pinned in a bool is therefore not sufficient. We could track this in a HashSet<PathBuf>, but there is really no reason to track it at all.


astoycos:
Head's up @dave-tucker I've removed this in https://github.com/aya-rs/aya/pull/783

astoycos:
Just haven't gotten a chance to respond to Tamir's comments

dave-tucker:
@astoycos ack. I'll close this and comment on your PR.

tamird:
FWIW I do think it's better to land this change separately ahead of #783.

PR #791, opened 2023-09-22
aya: Fix program loading on kernels with a patch > 255

nrxus:
This fixes an issue that made aya not able to load eBPF programs on kernels with a patch number greater than 255. 

I noticed this when trying to load some eBPF programs on an AWS instance running a `4.14.304-226.531.amzn2.x86_64` kernel. The kernel code that it tried to load was for a 4.15 kernel instead which caused a verifier error. 

I considered doing this directly in the `KernelVersion::current` function but per the comments in: https://github.com/torvalds/linux/commit/9b82f13e7ef316cdc0a8858f1349f4defce3f9e0 it looks like the clamp should only be applied for the context of creating the `LINUX_VERSION_CODE` so I limited the fix to only the calculating of the code for bpf prog loading.

### DIFF
--- a/aya/src/util.rs
+++ b/aya/src/util.rs
@@ -49,23 +49,43 @@ impl KernelVersion {
 
     /// Returns the kernel version of the currently running kernel.
     pub fn current() -> Result<Self, impl Error> {
-        let kernel_version = Self::get_kernel_version();
+        Self::get_kernel_version()
+    }
 
-        // The kernel version is clamped to 4.19.255 on kernels 4.19.222 and above.
-        //
-        // See https://github.com/torvalds/linux/commit/a256aac.
-        const CLAMPED_KERNEL_MAJOR: u8 = 4;
-        const CLAMPED_KERNEL_MINOR: u8 = 19;
-        if let Ok(Self {
-            major: CLAMPED_KERNEL_MAJOR,
-            minor: CLAMPED_KERNEL_MINOR,
-            patch: 222..,
-        }) = kernel_version
-        {
-            return Ok(Self::new(CLAMPED_KERNEL_MAJOR, CLAMPED_KERNEL_MINOR, 255));
+    /// The equivalent of LINUX_VERSION_CODE.
+    pub(crate) fn code(self) -> u32 {
+        let Self {
+            major,
+            minor,
+            mut patch,
+        } = self;
+
+        // Certain LTS kernels went above the "max" 255 patch so
+        // backports were done to cap the patch version
+        let max_patch = match (major, minor) {
+            // On 4.4 + 4.9, any patch 257 or above was hardcoded to 255.
+            // See: https://github.com/torvalds/linux/commit/a15813a +
+            // https://github.com/torvalds/linux/commit/42efb098
+            (4, 4 | 9) => 257,
+            // On 4.14, any patch 252 or above was hardcoded to 255.
+            // See: https://github.com/torvalds/linux/commit/e131e0e
+            (4, 14) => 252,
+            // On 4.19, any patch 222 or above was hardcoded to 255.
+            // See: https://github.com/torvalds/linux/commit/a256aac
+            (4, 19) => 222,
+            // For other kernels (i.e., newer LTS kernels as other
+            // ones won't reach 255+ patches) clamp it to 255. See:
+            // https://github.com/torvalds/linux/commit/9b82f13e
+            _ => 255,
+        };
+
+        // anything greater or equal to `max_patch` is hardcoded to
+        // 255.
+        if patch >= max_patch {
+            patch = 255;
         }
 
-        kernel_version
+        (u32::from(major) << 16) + (u32::from(minor) << 8) + u32::from(patch)
     }
 
     // This is ported from https://github.com/torvalds/linux/blob/3f01e9f/tools/lib/bpf/libbpf_probes.c#L21-L101.
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -617,14 +617,8 @@ fn load_program<T: Link>(
         },
     ) = obj;
 
-    let target_kernel_version = kernel_version.unwrap_or_else(|| {
-        let KernelVersion {
-            major,
-            minor,
-            patch,
-        } = KernelVersion::current().unwrap();
-        (u32::from(major) << 16) + (u32::from(minor) << 8) + u32::from(patch)
-    });
+    let target_kernel_version =
+        kernel_version.unwrap_or_else(|| KernelVersion::current().unwrap().code());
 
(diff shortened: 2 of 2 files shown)


alessandrod:
Thanks! Could you also paste the verifier error? @marysaka has been using kernels with patch > 255 so I'm wondering how we missed this failure and where it shows up

marysaka:
I'm pretty sure I tested kernel with patch > 255 apart from Debian ones as I was testing all kernel of every major distro around (cc @Arignir could you confirm that?)

nrxus:
Could you have been using kernels that are above 5.0? Starting on 5.0 the checks for `kern_code` were skipped: https://github.com/torvalds/linux/commit/6c4fc209fcf9

The verifier doesn't have any output on the error, it just returns an `EINVAL`. The way we realize it was due to an incorrect `kern_code` was by `strace`ing the syscall and saw that it was emitting a code for a 4.15 kernel

tamird:
I added clamping logic in https://github.com/aya-rs/aya/commit/6e570f0f14e615ccd0eeb80dfb68f3674f6ee74a -- maybe that was overly specific to 4.19.x and can be relaxed rather than adding disjoint logic?

tamird:
The commit you linked first appeared in 5.x kernels - how does libbpf deal with this case on that aws 4.14 kernel?

nrxus:
> I added clamping logic in [6e570f0](https://github.com/aya-rs/aya/commit/6e570f0f14e615ccd0eeb80dfb68f3674f6ee74a) -- maybe that was overly specific to 4.19.x and can be relaxed rather than adding disjoint logic?

I saw that clamping code but I wasn't sure if that was meant for checking the kernel version in major/minor/patch form or if it was meant specifically to clamp the code during prog load. For this PR I specifically only want to target the kern code during prog load. 

> The commit you linked first appeared in 5.x kernels - how does libbpf deal with this case on that aws 4.14 kernel?

The helper in libbpf clamp the patch to 256 regardless of the major/minor version: 
https://github.com/torvalds/linux/blob/d90b0276af8f25a0b8ae081a30d1b2a61263393b/tools/lib/bpf/bpf_helpers.h#L74

tamird:
Even that macro appeared fairly late https://github.com/torvalds/linux/commit/9ae2c26e43248b722e79fe867be38062c9dd1e5f

nrxus:
It looks like libbpf when doing prog load relies on the kernel provided `KERNEL_VERSION_CODE` macro if I understand this thread correctly: https://lore.kernel.org/lkml/2023041729-luckiness-active-3ddb@gregkh/T/. Since `aya` couldn't use that then I think it should clamp to 256 which is what newer kernel versions do when calculating the kernel version code.

alessandrod:
My thoughts so far: I like the idea of clamping only when generating the kernel code, instead of always clamping in KernelVersion::current() - API wise a caller might want to get the actual patch level for other puposes not necessarily for generating a kernel code. 

Once we split code generation, we can clamp regardless of 4.x 5.x X.x?

nrxus:
Does this mean you'd like me to move the pre-existing clamping code into the crate public method I created?

alessandrod:
> Does this mean you'd like me to move the pre-existing clamping code into the crate public method I created?

Yes, I don't think clamping slightly differently in two difference places is a good idea

aya/src/util.rs
```diff
@@ -49,23 +49,40 @@ impl KernelVersion {
...
-        const CLAMPED_KERNEL_MAJOR: u8 = 4;
-        const CLAMPED_KERNEL_MINOR: u8 = 19;
-        if let Ok(Self {
-            major: CLAMPED_KERNEL_MAJOR,
-            minor: CLAMPED_KERNEL_MINOR,
-            patch: 222..,
-        }) = kernel_version
-        {
-            return Ok(Self::new(CLAMPED_KERNEL_MAJOR, CLAMPED_KERNEL_MINOR, 255));
+    /// The equivalent of LINUX_VERSION_CODE
```
tamird, line 55: periods on all the comments please

aya/src/util.rs
```diff
@@ -49,23 +49,40 @@ impl KernelVersion {
...
+            (4, 14) => 252,
+            // On 4.19, any patch 222 or above was hardcoded to 255
+            // See: https://github.com/torvalds/linux/commit/a256aac
+            (4, 19) => 222,
+            // Newer LTS versions simply clamp to 255
+            // See: https://github.com/torvalds/linux/commit/9b82f13e
+            _ => 255,
+        };
+
+        if patch >= max_patch {
```
tamird, line 81: nit: let patch = patch.min(max_patch) or even let patch = patch.min(match ...);

aya/src/util.rs
```diff
@@ -49,23 +49,40 @@ impl KernelVersion {
...
+            // See: https://github.com/torvalds/linux/commit/a15813a +
+            // https://github.com/torvalds/linux/commit/42efb098
+            (4, 4 | 9) => 257,
+            // On 4.14, any patch 252 or above was hardcoded to 255
+            // See: https://github.com/torvalds/linux/commit/e131e0e
+            (4, 14) => 252,
+            // On 4.19, any patch 222 or above was hardcoded to 255
+            // See: https://github.com/torvalds/linux/commit/a256aac
+            (4, 19) => 222,
+            // Newer LTS versions simply clamp to 255
```
tamird, line 76: nit: this will apply to all versions, not strictly newer ones. i'm not sure how strict we want to be, but maybe we can just cite libbpf?

tamird (commented):
Nice archaeology digging up all the commits.

aya/src/util.rs
nrxus: That wouldn't work unfortunately. We aren't capping at at `max_patch` but instead saying if we are >= to `max_patch` then we need to set it to 255 regardless of what `max_patch`is.

aya/src/util.rs
nrxus: I put a comment that this will apply to all other versions but it's there mainly for newer LTS releases. Other kernels that aren't LTS are unlikely to reach 255 patches and older LTS releases haven't reached 255 patches yet and are already EOL so we are really only guarding for newer LTS kernels.

tamird (approved):

nrxus:
The integration tests are failing on something unrelated, the image that it is trying to use is no longer available in the Debian ftp. I am not really sure what image would be more stable so you all don't have to keep fixing it every few weeks though

tamird:
I'm pretty confused about that - I took a look earlier and the most recent 6.4.0-x image is

```
linux-image-6.4.0-0.deb12.2-cloud-amd64-unsigned_6.4.4-3~bpo12+1_amd64.deb	2023-08-16 06:27	26M
```

I don't understand why it would have been reverted back to August 16.

PR #792, opened 2023-09-22
aya-bpf: Add read_kernel! helper macro

Tacklebox:


alessandrod:
Thanks for this! See failing CI. Also we would need some integration tests

tamird (changes requested):
Needs tests and a rebase.

PR #796, opened 2023-09-27
integration-test: fix when host != target

tamird:
- integration-test: distinguish between success and noop
- integration-test: replace macro with function
- integration-test: build eBPF for the proper arch


ajwerner (approved):
LGTM

PR #799, opened 2023-09-28
aya: use montonic SinceBoot for loaded_at

ajwerner:
This commit was written to deflake test_loaded_at. The flakiness is due
to the lack of monotonicity in SystemTime clock calculations. See
https://github.com/aya-rs/aya/actions/runs/6340369670/job/17221591723.

This PR addresses that problem by introducing a wrapper type to
encapsulate the monotonically increasing duration since boot,
`SinceBoot`. This type has a method `into_system_time()` to convert it
into a `SystemTime`.

### DIFF
--- a/aya/src/lib.rs
+++ b/aya/src/lib.rs
@@ -87,6 +87,7 @@ pub mod pin;
 pub mod programs;
 pub use programs::loaded_programs;
 mod sys;
+pub mod time;
 pub mod util;
 
 pub use bpf::*;
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -71,7 +71,6 @@ use std::{
     os::fd::{AsFd, AsRawFd, BorrowedFd, OwnedFd},
     path::{Path, PathBuf},
     sync::Arc,
-    time::{Duration, SystemTime},
 };
 
 pub use cgroup_device::CgroupDevice;
@@ -110,13 +109,14 @@ use crate::{
     maps::MapError,
     obj::{self, btf::BtfError, VerifierLog},
     pin::PinError,
-    programs::utils::{boot_time, get_fdinfo},
+    programs::utils::get_fdinfo,
     sys::{
         bpf_btf_get_fd_by_id, bpf_get_object, bpf_link_get_fd_by_id, bpf_link_get_info_by_fd,
         bpf_load_program, bpf_pin_object, bpf_prog_get_fd_by_id, bpf_prog_get_info_by_fd,
         bpf_prog_query, iter_link_ids, iter_prog_ids, retry_with_verifier_logs,
         BpfLoadProgramAttrs, SyscallError,
     },
+    time::SinceBoot,
     util::KernelVersion,
     VerifierLogLevel,
 };
@@ -1086,8 +1086,9 @@ impl ProgramInfo {
     }
 
     /// The time the program was loaded.
-    pub fn loaded_at(&self) -> SystemTime {
-        boot_time() + Duration::from_nanos(self.0.load_time)
+    pub fn loaded_at(&self) -> SinceBoot {
+        // See https://github.com/torvalds/linux/blob/633b47cb/include/linux/bpf.h#L1418.
+        SinceBoot::from_nanos(self.0.load_time)
     }
 
     /// Returns a file descriptor referencing the program.
--- a/aya/src/programs/utils.rs
+++ b/aya/src/programs/utils.rs
@@ -5,7 +5,6 @@ use std::{
     io::{self, BufRead, BufReader},
     os::fd::{AsFd as _, AsRawFd as _, BorrowedFd},
     path::Path,
-    time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use crate::{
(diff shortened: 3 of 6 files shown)


tamird (changes requested):
> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/799)* status: all files reviewed, 9 unresolved discussions (waiting on @ajwerner and @alessandrod)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/aya-rs/aya/799#-NfRU6T81LiJy52TH3Ja):*
"calculations" is wrong - this doesn't have to do with calculations, it has to do with reading wall time more than once.
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRUHjZ9YU_Pb7QKYZa):*
s/PR/commit/ or just drop the word.

consider doing the same in the previous paragraph - drop the noun
___
*[`aya/src/time.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRcjFg3_gx5ztEHfFv:-NfRcjFg3_gx5ztEHfFw:b5agn7a) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L7)):*
> ```XML
> /// A timestamp relative to the system boot time.
> #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
> pub struct SinceBoot(Duration);
> ```

should the nomenclature be something like Instant?
___
*[`aya/src/time.rs` line 10 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRUWfX5V1-On8jxMrQ:-NfRUWfX5V1-On8jxMrR:bk8hsgl) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L10)):*
> ```XML
> 
> impl SinceBoot {
>     /// The current SinceBoot.
> ```

mirror the comment on https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html#method.now?
___
*[`aya/src/time.rs` line 12 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRUd9J8S1VLS3c4QUc:-NfRUd9J8S1VLS3c4QUd:b35ak86) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L12)):*
> ```XML
>     /// The current SinceBoot.
>     ///
>     /// Note that these calls will be monotonic.
> ```

not appropriate to document CLOCK_BOOTTIME, and "these calls" is not the right subject? also, this is used in testing only -- can we perhaps move it there?
___
*[`aya/src/time.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRUmruDwI3RBGMTqIb:-NfRUmruDwI3RBGMTqIc:bk0wnmk) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L19)):*
> ```XML
>     /// Converts the timestamp to a `SystemTime`.
>     ///
>     /// Note that this will not be robust to changes in the system clock, and thus these
> ```

I don't think we should document the semantics of ST here. it's all discussed at length: https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html#
___
*[`aya/src/time.rs` line 26 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRd14T17O1hPSjvyLN:-NfRd14T17O1hPSjvyLO:b6jvr0j) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L26)):*
> ```XML
>     }
> 
>     pub(crate) fn from_nanos(nanos: u64) -> Self {
> ```

maybe just `new` and take a duration? no reason to make the reader read "nanos" 3 times.
___
*[`aya/src/time.rs` line 34 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRbuSjAiSQw0R2gNRk:-NfRbuSjAiSQw0R2gNRl:b21tab9) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L34)):*
> ```XML
>     let since_boot = get_time(libc::CLOCK_BOOTTIME);
>     let since_epoch = get_time(libc::CLOCK_REALTIME);
>     UNIX_EPOCH + since_epoch - since_boot
> ```

this has only one caller, and when you combine it with the caller this is just `SystemTime::now() + (self - Self::now())`, I think?
___
*[`aya/src/time.rs` line 55 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRea667m7kVNgpZhql:-NfRea667m7kVNgpZhqm:b-gs2u9t) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L55)):*
> ```XML
>     #[test]
>     fn test_since_boot_now_into_system_near_system_time() {
>         let since_boot = SinceBoot::now().into_system();
> ```

let's avoid reading the real clock more than once -- as that is the source of flakes

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/799)* status: all files reviewed, 9 unresolved discussions (waiting on @ajwerner and @alessandrod)

___
*[`aya/src/time.rs` line 34 at r2](https://reviewable.io/reviews/aya-rs/aya/799#-NfRbuSjAiSQw0R2gNRk:-NfRgSP04_PU5_tSxuv2:byznsc1) ([raw file](https://github.com/aya-rs/aya/blob/02272cce0539a62d2e4b4c0fc30f38a267636b90/aya/src/time.rs#L34)):*

this has only one caller, and when you combine it with the caller this is just `SystemTime::now() + (self - Self::now())`, I think?
</blockquote></details>

edit: i think this should be `SystemTime::now() - (Self::now() - self)` to avoid a negative duration.

PR #800, opened 2023-09-28
integration-test: deflake test_loaded_at

ajwerner:
The test could fail due to the lack of clock monotonicity. This PR deflakes the test by adding retries.

See
https://github.com/aya-rs/aya/actions/runs/6340369670/job/17221591723.

### DIFF
--- a/test/integration-test/src/tests/load.rs
+++ b/test/integration-test/src/tests/load.rs
@@ -154,22 +154,36 @@ fn unload_xdp() {
 fn test_loaded_at() {
     let mut bpf = Bpf::load(crate::TEST).unwrap();
     let prog: &mut Xdp = bpf.program_mut("pass").unwrap().try_into().unwrap();
-    let t1 = SystemTime::now();
-    prog.load().unwrap();
-    let t2 = SystemTime::now();
-    assert_loaded("pass");
 
-    let loaded_at = prog.info().unwrap().loaded_at();
-
-    let range = t1..t2;
+    // SystemTime is not monotonic, which can cause this test to flake. We don't expect the clock
+    // timestamp to continuously jump around, so we add some retries. If the test is ever correct,
+    // we know that the value returned by loaded_at() was reasonable relative to SystemTime::now().
+    let mut failures = Vec::new();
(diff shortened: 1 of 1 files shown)


test/integration-test/src/tests/load.rs
```diff
@@ -154,22 +154,39 @@ fn unload_xdp() {
...
     let prog: &mut Xdp = bpf.program_mut("pass").unwrap().try_into().unwrap();
-    let t1 = SystemTime::now();
-    prog.load().unwrap();
-    let t2 = SystemTime::now();
-    assert_loaded("pass");
-
-    let loaded_at = prog.info().unwrap().loaded_at();
 
-    let range = t1..t2;
+    // The SystemTime::now() calls are not monotonic, which can cause this test to flake. We don't
```
tamird, line 158: "SystemTime is not monotonic"

test/integration-test/src/tests/load.rs
```diff
@@ -154,22 +154,39 @@ fn unload_xdp() {
...
+    // SystemTime::now().
+    let mut failures = Vec::new();
+    for _ in 0..5 {
+        let t1 = SystemTime::now();
+        prog.load().unwrap();
+        let t2 = SystemTime::now();
+        assert_loaded("pass");
+        let loaded_at = prog.info().unwrap().loaded_at();
+        let range = t1..t2;
+        prog.unload().unwrap();
```
tamird, line 170: shouldn't the assert unloaded be here?

since you're moving things around maybe

let t1 = ...
load()
let t2 = ...
let loaded_at = 

maybe these assert loaded/unloaded calls can all just go away.

tamird (approved):

test/integration-test/src/tests/load.rs
ajwerner: Done.

test/integration-test/src/tests/load.rs
ajwerner: Done.

ajwerner (commented):
TFTR!

tamird (approved):

PR #801, opened 2023-09-28
xtask: terminate QEMU on "BUG: soft lockup"

tamird:
See
https://github.com/aya-rs/aya/actions/runs/6329828507/job/17190931752.

Unlike on kernel panic this message prints periodically so it is permitted to print twice before QEMU is killed.


ajwerner (approved):
LGTM

PR #807, opened 2023-10-04
PerfEvent: add option to set the inherit bit

akhramov:
The inherit bit allows extending perf scope to the new child processes.

(I am not sure about the API, though)

### DIFF
--- a/aya/src/programs/perf_event.rs
+++ b/aya/src/programs/perf_event.rs
@@ -66,6 +66,11 @@ pub enum PerfEventScope {
         /// process id
         pid: u32,
     },
+    /// one process, any cpu, including child processes
+    OneProcessIncludingChildren {
+        /// process id
+        pid: u32,
+    },
     /// one process, one cpu
     OneProcessOneCpu {
         /// cpu id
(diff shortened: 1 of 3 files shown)


alessandrod:
Thanks for the PR!

> (I am not sure about the API, though)

What does libbpf do? I'm inclined to say we should set inherit=1 by default?

akhramov:
> What does libbpf do

https://github.com/torvalds/linux/blob/82714078aee4ccbd6ee7579d5a21f8a72155d0fb/tools/lib/bpf/libbpf.c#L12441-L12448

It doesn't set the bit either.

>  I'm inclined to say we should set inherit=1 by default?

We can, but it can ruin it for users who don't expect events from the sub-processes.

alessandrod:
I see.

I think inherit applies to all variants tho right? So maybe it should be an extra bool argument to attach()?

alessandrod:
Ping? 😊

akhramov:
Hey, sorry for the delay. I found my way around this, so probably we should close the PR.

Issue #808, opened 2023-10-09
Bug in Aya log 


qjerome:
The verifier fails at loading BPF program because of (supposed) DisplayHint::write function. 
```
Error: the BPF_PROG_LOAD syscall failed. Verifier output: Validating write() func#1...
Global function write() doesn't return scalar. Only those are supported.
verification time 40 usec
stack depth 0+0
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```
This bug appears only under some conditions:
- bpf-linker branch `feature/fix-di`is used
- `debug-info = 2` is present in eBPF's `Cargo.toml` file

Supporting files:
* disassembled eBPF bytecode: [poc-bug-aya-log.disass.txt](https://github.com/aya-rs/aya/files/12843248/poc-bug-aya-log.disass.txt)
* pre-optimization modules dump: [pre-opt.ll.gz](https://github.com/aya-rs/aya/files/12843274/pre-opt.ll.gz)
* post optimization modules dump: [post-opt.ll.gz](https://github.com/aya-rs/aya/files/12843277/post-opt.ll.gz)

A repro project can be found here:  https://github.com/0xrawsec/poc-bug-aya-log
**IMPORTANT**: a compiled version of `bpf-linker` is provided with the project and is located under the eBPF's directory.

tamird:
Can you reproduce this with a test in this repo? I understand if bpf-linker needs to be built from the `feature/fix-di` branch.

qjerome:
It is **mandatory** to have a bpf-linker `feature/fix-di` build **and** to be built with `debug-info = 2`. I am not familiar with the Aya integration tests, and I don't know whether those conditions can be met in the current implementation. 
If it is possible, please let me know how to do it.

tamird:
debug-info = 2 is easy to do - just put it in the appropriate manifest or .cargo/config.toml. Having a custom bpf-linker is not easy to do, but just leave that part out. The tests use whatever bpf-linker is installed on the machine, if you have a repro that requires me to install bpf-linker from the DI branch, that is good enough.

qjerome:
Have you managed to reproduce the issue ?
I actually shipped a compiled version of `bpf-linker` (nothing else than latest build of `feature/fix-di` build commit=60ce107a15333946ba0116297af30c9fd27ae6f4 ) with the project to be in the exact same conditions I was when the bug occurred.

tamird:
[...]

PR #810, opened 2023-10-10
repro issue #808

qjerome:
Implemented test aiming to reproduce issue #808
There is a little quirk though !
It actually triggers a bug, but not the expected one ! So I hope, once the unexpected bug gets fixed it will trigger the expected one.

### DIFF
--- a/test/integration-ebpf/Cargo.toml
+++ b/test/integration-ebpf/Cargo.toml
@@ -55,3 +55,7 @@ path = "src/xdp_sec.rs"
 [[bin]]
 name = "ring_buf"
 path = "src/ring_buf.rs"
+
+[[bin]]
+name = "log_str_random"
+path = "src/repro/log_str_random.rs"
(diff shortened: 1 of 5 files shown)


alessandrod:
>  It actually triggers a bug, but not the expected one ! So I hope, once the unexpected bug gets fixed it will trigger the expected one.

RUSTFLAGS="-C debuginfo=2" 😂

And then you get:

```
running 1 test
test tests::load::log_str_random ... FAILED

failures:

---- tests::load::log_str_random stdout ----
thread 'tests::load::log_str_random' panicked at 'called `Result::unwrap()` on an `Err` value: LoadError { io_error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, verifier_log: Validating write() func#1...
Global function write() doesn't return scalar. Only those are supported.
verification time 18 usec
stack depth 0+0
 }', test/integration-test/src/tests/load.rs:624:17
```

alessandrod:
The issue is functions from external crates get compiled as global functions instead of static. Global functions can't return non-scalar types. We should probably make all functions except program functions static in bpf-linker.

alessandrod:
Fixed in https://github.com/aya-rs/bpf-linker/pull/173

PR #811, opened 2023-10-11
aya: import types from std::ffi rather than libc

tamird:


tamird:
@alessandrod FYI this now changes API but the types are identical.

PR #818, opened 2023-10-16
aya: don't parse labels as programs

alessandrod:
Fixes a bug introduced by https://github.com/aya-rs/aya/pull/413 where we were generating a bunch of spurious LBB* programs.

### DIFF
--- a/aya-obj/src/obj.rs
+++ b/aya-obj/src/obj.rs
@@ -599,12 +599,12 @@ impl Object {
                 .get(symbol_index)
                 .expect("all symbols in symbols_by_section are also in symbol_table");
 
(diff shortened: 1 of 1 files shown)


dave-tucker (approved):
LGTM

tamird:
Can we add a test?

alessandrod:
All programs test this code. I can add an arbitrary "ensure that there are no programs that start with LBB*" check somewhere but seemed a bit silly, since from rust you can't control which labels are created, so the test might pass because we skip the labels, or because there are no labels, who knows.

tamird:
Can you control which labels are created from C? We have tests against eBPF programs written in C.

PR #819, opened 2023-10-16
tests: add AF_XDP test using xdpilone

Tuetuopay:
This is a follow-up PR to https://github.com/aya-rs/aya/pull/527 where the AF_XDP integration test got dropped due to them depending on libbpf. We use the `xdpilone` crate there that is a pure-rust AF_XDP userspace library.

### DIFF
--- a/test/integration-test/src/tests/xdp.rs
+++ b/test/integration-test/src/tests/xdp.rs
@@ -1,6 +1,4 @@
-use std::{
-    ffi::CStr, mem::MaybeUninit, net::UdpSocket, num::NonZeroU32, ptr::NonNull, time::Duration,
-};
+use std::{ffi::CStr, mem::MaybeUninit, net::UdpSocket, num::NonZeroU32, time::Duration};
 
 use aya::{
     maps::{Array, CpuMap, XskMap},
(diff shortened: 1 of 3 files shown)


test/integration-test/src/tests/xdp.rs
```diff
@@ -1,15 +1,86 @@
...
+        .unwrap()
+        .try_into()
+        .unwrap();
+    xdp.load().unwrap();
+    xdp.attach("lo", XdpFlags::default()).unwrap();
+
+    #[repr(align(4096))]
+    struct PacketMap(MaybeUninit<[u8; 4096]>);
+
+    let mem = Box::new(PacketMap(MaybeUninit::uninit()));
```
alessandrod, line 34: nit: you're never freeing this

Tuetuopay: @alessandrod are there any chance the tests will ever run on apple silicon? pages on M1/M2 are 16k instead of 4k, so it may be different for the `PacketMap` newtype. Is it worth it to change the `repr` of the type?

Tuetuopay: @alessandrod I align'ed to 16384 and I'm calling it a day.

Tuetuopay:
Rebased on main + align on 16k for Apple Silicon.

cameronelliott:
Hey, I think `xdpilone` is under the EUPL license, which is very similar to the GPL.
Can it be used in `aya`?  I'm not sure it can.

I have also been studying this a little bit:
https://github.com/aws/s2n-quic/tree/main/tools/xdp/s2n-quic-xdp

Which may (or may not) have many of the pieces to do AF_XDP from Rust without libbpf or libxdp.

Tuetuopay:
> Hey, I think `xdpilone` is under the EUPL license, which is very similar to the GPL. Can it be used in `aya`? I'm not sure it can.

This has been discussed in my original XDP maps PR. IANAL but this is for tests only, this should not affect the library itself. Since artifacts based on xdpilone are not distributed it should not matter.

> I have also been studying this a little bit: https://github.com/aws/s2n-quic/tree/main/tools/xdp/s2n-quic-xdp
> 
> Which may (or may not) have many of the pieces to do AF_XDP from Rust without libbpf or libxdp. (And it is Apache licensed)

Oh that's a pretty nice one. Seems a bit "as-is" but that's promising, if that's possible to pry it out of the aws lib.

> Also, I think `xdpilone` only has the parts to do tx, and not rx, so maybe another path would be good from that perspective.

It definitely has the parts for RX since it's what's being tested in the integration tests. It lacks the XDP probe to send the packet to the socket map, but that's where Aya comes in. So it's not batteries included like libbpf is, but that does not matter.

alessandrod (approved):

PR #822, opened 2023-10-20
aya/maps: pin for (async)perf_event_array

astoycos:
Implement pinning for perf_event_array and async_perf_event_array. Additionally make the core MapData.pin method operate on a reference rather than a mutable reference.

TODO from #783

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -48,7 +48,7 @@
 //! versa. Because of that, all map values must be plain old data and therefore
 //! implement the [Pod] trait.
 use std::{
-    borrow::BorrowMut,
+    borrow::Borrow,
     ffi::{c_long, CString},
     fmt, io,
     marker::PhantomData,
@@ -332,7 +332,7 @@ impl Map {
     ///
     /// When a map is pinned it will remain loaded until the corresponding file
     /// is deleted. All parent directories in the given `path` must already exist.
-    pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
+    pub fn pin<P: AsRef<Path>>(&self, path: P) -> Result<(), PinError> {
         match self {
             Self::Array(map) => map.pin(path),
             Self::BloomFilter(map) => map.pin(path),
@@ -360,7 +360,6 @@ impl Map {
 }
 
 // Implements map pinning for different map implementations
-// TODO add support for PerfEventArrays and AsyncPerfEventArrays
 macro_rules! impl_map_pin {
     ($ty_param:tt {
         $($ty:ident),+ $(,)?
@@ -371,14 +370,14 @@ macro_rules! impl_map_pin {
       <($($ty_param:ident),*)>
       $ty:ident
     ) => {
-            impl<T: BorrowMut<MapData>, $($ty_param: Pod),*> $ty<T, $($ty_param),*>
+            impl<T: Borrow<MapData>, $($ty_param: Pod),*> $ty<T, $($ty_param),*>
             {
                     /// Pins the map to a BPF filesystem.
                     ///
                     /// When a map is pinned it will remain loaded until the corresponding file
                     /// is deleted. All parent directories in the given `path` must already exist.
-                    pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
-                        let data = self.inner.borrow_mut();
+                    pub fn pin<P: AsRef<Path>>(self, path: P) -> Result<(), PinError> {
+                        let data = self.inner.borrow();
                         data.pin(path)
                     }
             }
@@ -591,7 +590,7 @@ impl MapData {
                 Ok(Self { obj, fd })
             }
             Err(_) => {
-                let mut map = Self::create(obj, name, btf_fd)?;
+                let map = Self::create(obj, name, btf_fd)?;
                 map.pin(&path).map_err(|error| MapError::PinError {
                     name: Some(name.into()),
                     error,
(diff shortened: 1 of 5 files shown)


tamird (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/822)* status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod and @astoycos)

___
*[`aya/src/maps/perf/async_perf_event_array.rs` line 116 at r1](https://reviewable.io/reviews/aya-rs/aya/822#-NhbY3BNCbJSo4whjgE_:-NhbY3BNCbJSo4whjgEa:b-za1hen) ([raw file](https://github.com/aya-rs/aya/blob/58b9cbf8825872e7765ce81c3b0a2fbc0a14d544/aya/src/maps/perf/async_perf_event_array.rs#L116)):*
> ```XML
>     /// When a map is pinned it will remain loaded until the corresponding file
>     /// is deleted. All parent directories in the given `path` must already exist.
>     pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
> ```

shouldn't this take &self?
___
*[`aya/src/maps/perf/perf_event_array.rs` line 206 at r1](https://reviewable.io/reviews/aya-rs/aya/822#-NhbY4rp74ZZwelKhQVR:-NhbY4rp74ZZwelKhQVS:b-za1hen) ([raw file](https://github.com/aya-rs/aya/blob/58b9cbf8825872e7765ce81c3b0a2fbc0a14d544/aya/src/maps/perf/perf_event_array.rs#L206)):*
> ```XML
>     /// When a map is pinned it will remain loaded until the corresponding file
>     /// is deleted. All parent directories in the given `path` must already exist.
>     pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
> ```

shouldn't this take &self?

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/822)* status: all files reviewed, 3 unresolved discussions (waiting on @alessandrod and @astoycos)

___
*[`aya/src/maps/perf/perf_event_array.rs` line 7 at r1](https://reviewable.io/reviews/aya-rs/aya/822#-NhbYEkW5kE6ecjRRUr4:-NhbYEkW5kE6ecjRRUr5:b-g7j6cj) ([raw file](https://github.com/aya-rs/aya/blob/58b9cbf8825872e7765ce81c3b0a2fbc0a14d544/aya/src/maps/perf/perf_event_array.rs#L7)):*
> ```XML
>     borrow::{Borrow, BorrowMut},
>     ops::Deref,
>     os::fd::{AsFd as _, AsRawFd, BorrowedFd, RawFd},
> ```

this `as _` is incorrect conflict resolution

astoycos:
> > Reviewed 5 of 5 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/aya-rs/aya/822)_ status: all files reviewed, 2 unresolved discussions (waiting on @alessandrod and @astoycos)
> 
> _[`aya/src/maps/perf/async_perf_event_array.rs` line 116 at r1](https://reviewable.io/reviews/aya-rs/aya/822#-NhbY3BNCbJSo4whjgE_:-NhbY3BNCbJSo4whjgEa:b-za1hen) ([raw file](https://github.com/aya-rs/aya/blob/58b9cbf8825872e7765ce81c3b0a2fbc0a14d544/aya/src/maps/perf/async_perf_event_array.rs#L116)):_
> 
> > ```
> >     /// When a map is pinned it will remain loaded until the corresponding file
> >     /// is deleted. All parent directories in the given `path` must already exist.
> >     pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
> > ```
> 
> shouldn't this take &self?
> 
> _[`aya/src/maps/perf/perf_event_array.rs` line 206 at r1](https://reviewable.io/reviews/aya-rs/aya/822#-NhbY4rp74ZZwelKhQVR:-NhbY4rp74ZZwelKhQVS:b-za1hen) ([raw file](https://github.com/aya-rs/aya/blob/58b9cbf8825872e7765ce81c3b0a2fbc0a14d544/aya/src/maps/perf/perf_event_array.rs#L206)):_
> 
> > ```
> >     /// When a map is pinned it will remain loaded until the corresponding file
> >     /// is deleted. All parent directories in the given `path` must already exist.
> >     pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<(), PinError> {
> > ```
> 
> shouldn't this take &self?

Done, good catch thanks!

tamird (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/822#-:-Nhbgmbf7_ICXnn9jkeR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

aya/src/maps/mod.rs
```diff
@@ -369,14 +369,14 @@ macro_rules! impl_map_pin {
       <($($ty_param:ident),*)>
       $ty:ident
     ) => {
-            impl<T: BorrowMut<MapData>, $($ty_param: Pod),*> $ty<T, $($ty_param),*>
+            impl<T: Borrow<MapData>, $($ty_param: Pod),*> $ty<T, $($ty_param),*>
```
alessandrod, line 372: The TODO at the top of the macro can go now?

aya/src/maps/perf/perf_event_array.rs
```diff
@@ -175,7 +176,7 @@ impl<T: Borrow<MapData>> PerfEventArray<T> {
     }
 }
 
-impl<T: BorrowMut<MapData>> PerfEventArray<T> {
+impl<T: Borrow<MapData>> PerfEventArray<T> {
```
alessandrod, line 179: Why change this? I think we want to keep open() only for BorrowMut. I think
`pin` should go in the impl above?

astoycos: Good catch can't remember why I did this thanks!

aya/src/maps/mod.rs
astoycos: Yep!

alessandrod (approved):

PR #829, opened 2023-10-31
public-api: bless with new nightly

tamird:
See https://github.com/rust-lang/rust/commit/58a80c85b9323e59e1b7e744d6.


vadorovsky (approved):
Amazing, so much cleaner!

PR #834, opened 2023-11-09
xtask: attempt to open /dev/kvm

tamird:
We're seeing test failures where KVM is present but we aren't able to use it.

See https://github.com/actions/runner-images/issues/7670#issuecomment-1760184162.


tamird:
Sample failures:
- https://github.com/aya-rs/aya/actions/runs/6793719823/job/18469014146
- https://github.com/aya-rs/aya/actions/runs/6807368000/job/18510104410

vadorovsky (approved):

PR #839, opened 2023-11-18
aya-log-parser: add support of :p format

lyuts:
Extending log parser with the support for formatting pointers.


alessandrod (approved):
Thank you!

PR #843, opened 2023-11-25
maps/ringbuf: make RingBuf: Send + Sync

ajwerner:
There was no reason for them not to be -- the APIs all require mutable references and hold onto mutable references, so there cannot be internal concurrency. The !Send + !Sync came from the MMap, but not for any good reason.


alessandrod (approved):

ajwerner:
TFTR!

PR #846, opened 2023-12-01
Improve AF_XDP test

arctic-alpaca:
- Fix UB in AF_XDP test: [Calling `assume_init_mut()` on uninitialized memory causes UB](https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html#method.assume_init_mut). By using `zeroed()` the memory is guaranteed to be zero and with that is properly initialized for `u8`.
- Add queue ID matching to AF_XDP test: Improves clarity how AF_XDP works for people using the test as an example. Basically does the same as the [kernel docs](https://www.kernel.org/doc/html/latest/bpf/map_xskmap.html#kernel).

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/846)

### DIFF
--- a/test/integration-ebpf/src/redirect.rs
+++ b/test/integration-ebpf/src/redirect.rs
@@ -31,7 +31,9 @@ pub fn redirect_sock(ctx: XdpContext) -> u32 {
     // Check whether incoming packet's queue ID matches the queue ID of the socket in XSKMAP at index `queue_id`.
     if SOCKS.get(queue_id) == Some(queue_id) {
         // Queue ID matches, redirect to AF_XDP socket.
-        SOCKS.redirect(0, 0).unwrap_or(xdp_action::XDP_ABORTED)
+        SOCKS
+            .redirect(queue_id, 0)
+            .unwrap_or(xdp_action::XDP_ABORTED)
     } else {
         // Queue ID did not match, pass packet to kernel network stack.
         xdp_action::XDP_PASS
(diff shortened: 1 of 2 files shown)


test/integration-ebpf/src/redirect.rs
```diff
@@ -25,8 +25,17 @@ static CPUS: CpuMap = CpuMap::with_max_entries(1, 0);
...
 #[xdp]
-pub fn redirect_sock(_ctx: XdpContext) -> u32 {
-    SOCKS.redirect(0, 0).unwrap_or(xdp_action::XDP_ABORTED)
+pub fn redirect_sock(ctx: XdpContext) -> u32 {
+    // Retrieve queue ID of incoming packet.
+    let queue_id = unsafe { *ctx.ctx }.rx_queue_index;
+    // Check whether incoming packet's queue ID matches the queue ID of the socket in XSKMAP at index `queue_id`.
+    if SOCKS.get(queue_id) == Some(queue_id) {
+        // Queue ID matches, redirect to AF_XDP socket.
+        SOCKS.redirect(0, 0).unwrap_or(xdp_action::XDP_ABORTED)
```
Tuetuopay, line 34: shouldn't this be `SOCKS.redirect(queue_id, 0)`? The test will still work as veth only has one queue, but since we're after correctness, this feels like it's missing.

arctic-alpaca: Good point, I changed it accordingly.

esemeniuc:
bump on merging this

PR #848, opened 2023-12-06
aya: Add `XskMap::unset`

arctic-alpaca:
- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/848)

### DIFF
--- a/aya/src/maps/xdp/xsk_map.rs
+++ b/aya/src/maps/xdp/xsk_map.rs
@@ -2,12 +2,12 @@
 
 use std::{
     borrow::{Borrow, BorrowMut},
-    os::fd::{AsFd, AsRawFd, RawFd},
+    os::fd::{AsFd, AsRawFd, BorrowedFd, RawFd},
 };
 
 use crate::{
     maps::{MapData, MapError, check_bounds, check_kv_size},
-    sys::{SyscallError, bpf_map_update_elem},
+    sys::{SyscallError, bpf_map_delete_elem, bpf_map_update_elem},
 };
 
 /// An array of AF_XDP sockets.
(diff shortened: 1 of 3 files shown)


aya/src/maps/xdp/xsk_map.rs
```diff
@@ -78,4 +78,23 @@ impl<T: BorrowMut<MapData>> XskMap<T> {
         )?;
         Ok(())
     }
+
+    /// Removes the `AF_XDP` socket stored at `index` from the map.
+    ///
+    /// Returns [`MapError::OutOfBounds`] if `index` is out of bounds, [`MapError::SyscallError`]
+    /// if `bpf_map_update_elem` fails.
```
dave-tucker, line 85:
```suggestion
    /// if `bpf_map_delete_elem` fails.
```

aya/src/maps/xdp/xsk_map.rs
```diff
@@ -78,4 +78,23 @@ impl<T: BorrowMut<MapData>> XskMap<T> {
...
+    pub fn clear_index(&mut self, index: u32) -> Result<(), MapError> {
```
dave-tucker, line 86: This should likely be either `remove` or `delete`. Hold please while we hear from @alessandrod.

dave-tucker (changes requested):
One small nit and one API related query for @alessandrod:

For eBPF maps that have are array-like we expose `get` and `set`.
For eBPF maps that have are hashmap-like we expose `get`, `insert` and `remove`.

Since BPF_MAP_TYPE_ARRAY doesn't support bpf_map_delete_elem we never implemented it, so there is no precedent set for naming.

Would you go with, `delete`, `remove` or `clear`?

arctic-alpaca:
@alessandrod Could you take a look this please? Without the function, it's not possible (as far as I know) to clean up a single AF_XDP socket without detaching/unloading the whole program.

aya/src/maps/xdp/xsk_map.rs
alessandrod: My thinking:

XskMap is fixed side - if you call map.len() you get `max_entries`.  So calling it `remove(index)` would be weird because len() before and after remove() is unchanged - so it doesn't remove?

I think this is why originally I suggested clear_index. Could also be unset()? I'm not really sure.

dave-tucker (approved):

tamird:
I've changed this to `unset` to match `set`. Assuming this looks good to everyone, let's merge it?

PR #854, opened 2023-12-14
.github: avoid homebrew automatic update

tamird:
This is slow and prone to failure; see
https://github.com/aya-rs/aya/actions/runs/7204446013/job/19625908097.


tamird:
This takes about 6 minutes off the mac integration test run time. 

Before: https://github.com/aya-rs/aya/actions/runs/7190700238/job/19584252022
After: https://github.com/aya-rs/aya/actions/runs/7211289293/job/19646377933

Issue #820, opened 2023-10-17
cargo test Fails on Documentation Test in aya/src/maps/ring_buf.rs


tonsiasy:
Summary:

Running cargo test results in a failure specifically related to the documentation test in aya/src/maps/ring_buf.rs.

Steps to Reproduce:

Clone the repository
Navigate to the repository directory
Run cargo test
Expected Behavior:

All tests, including documentation tests, should pass successfully.

Actual Behavior:

The documentation test for aya/src/maps/ring_buf.rs fails with multiple errors, causing cargo test to fail.
```
Errors:
failures:

---- aya/src/maps/ring_buf.rs - maps::ring_buf::RingBuf (line 57) stdout ----
error: expected `;`, found keyword `while`
  --> aya/src/maps/ring_buf.rs:80:37
   |
26 |     let ring_buf = guard.inner_mut()
   |                                     ^ help: add `;` here
27 |     while let Some(item) = ring_buf.next() {
   |     ----- unexpected token

error[E0277]: the `?` operator can only be used on `Result`s, not `Option`s, in a function that returns `Result`
  --> aya/src/maps/ring_buf.rs:76:54
   |
4  | fn main() { #[allow(non_snake_case)] fn _doctest_main_aya_src_maps_ring_buf_rs_57_0() -> Result<(), impl core::fmt::Debug> {
   |                                      ------------------------------------------------------------------------------------- this function returns a `Result`
...
22 | let ring_buf = RingBuf::try_from(bpf.map_mut("ARRAY")?)?;
   |                                                      ^ use `.ok_or(...)?` to provide an error compatible with `Result<(), _>`
   |
   = help: the trait `FromResidual<Option<Infallible>>` is not implemented for `Result<(), _>`
   = help: the following other types implement trait `FromResidual<R>`:
             <Result<T, F> as FromResidual<Yeet<E>>>
             <Result<T, F> as FromResidual<Result<Infallible, E>>>

error[E0277]: the `?` operator can only be applied to values that implement `Try`
  --> aya/src/maps/ring_buf.rs:79:21
   |
25 |     let mut guard = poll.readable()?;
   |                     ^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `Guard<'_, aya::maps::RingBuf<&mut MapData>>`
   |
   = help: the trait `Try` is not implemented for `Guard<'_, aya::maps::RingBuf<&mut MapData>>`

error: aborting due to 3 previous errors

[...]

PR #858, opened 2023-12-19
Fix docs build

dave-tucker:
Doctests aren't running in CI!
This PR:

1. Adds doctests to CI
2. Fixes ringbuf docs in aya
3. Fixes XDP map docs in aya-bpf

Fixes https://github.com/aya-rs/aya/issues/820


ajwerner (approved):
LGTM, thanks for doing it!

PR #859, opened 2023-12-31
Add SchedClassifier::attach_to_link()

alessandrod:
Similar to XdpLink::attach_to_link(), can be used to replace/upgrade the program attached to a link.

### DIFF
--- a/aya/src/programs/tc.rs
+++ b/aya/src/programs/tc.rs
@@ -152,27 +152,54 @@ impl SchedClassifier {
         attach_type: TcAttachType,
         options: TcOptions,
     ) -> Result<SchedClassifierLinkId, ProgramError> {
-        let prog_fd = self.fd()?;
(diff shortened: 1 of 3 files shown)


vadorovsky (commented):
In general looks good, but I'm wondering if we could add an integration test? Or would it be to much hassle to get it working with network namespaces and loopback devices?

vadorovsky (approved):
My intention was to approve, but my previous question still remains :stuck_out_tongue:

alessandrod:
Thanks for the review!

I wrote a small aya-template generated program to test this, and @jhutchins tested with his project too. I did try to write an integration test, but integration tests seem to be broken on my mac at the moment. I'm going to merge this as is, I will add tests when I'm done with some other bpf-linker stuff I'm doing, in the meantime we can unblock @jhutchins.

PR #862, opened 2024-01-07
public-api: regenerate

tamird:
`Send + Sync` became `Sync + Send` for reasons I don't understand.


alessandrod:
Why is this needed? Locally I don't seem to get Sync + Send

tamird:
Did you try with today's nightly?

tamird:
Failure @ main: https://github.com/aya-rs/aya/actions/runs/7443162986/job/20247658356.

PR #877, opened 2024-01-29
aya: perf_event: add inherit argument to attach()

alessandrod:
When inherit=true, processes spawned by the sampled processes get automatically sampled too.


dave-tucker (approved):
LGTM

PR #880, opened 2024-02-07
fix(aya-sock-map): invalid transmute when calling fd

erebe:
Corrent an invalid transmutation for sock_map.
fd is already a ref of MapFd, so transmuting &fd to &SockMapFd is equivalent to transmuting &&SockMapFd into &SockMapFd which is buggy.

### Step to reproduce
This simple program terminates in error with `Os error 9, invalid file descriptor`
```
    let mut intercept_ingress: SockMap<_> = bpf.take_map("INTERCEPT_INGRESS").unwrap().try_into()?;
    let map_fd = intercept_ingress.fd().try_clone()?;
```

SockHash is doing it correctly BTW https://github.com/aya-rs/aya/blob/main/aya/src/maps/sock/sock_hash.rs#L113


alessandrod (approved):
Oops, nice catch and thanks for the PR!

erebe:
You welcome, thanks for aya :)

PR #884, opened 2024-02-18
Install libelf on macOS

tamird:
Seems it was removed in
https://github.com/actions/runner-images/releases/tag/macos-13%2F20240204.1
though it's not called out in the notes.


tamird:
Dupe of #883.

PR #892, opened 2024-02-28
docs(aya): Document more breaking changes

dave-tucker:


alessandrod:
PerfEvent::attach has changed too

dave-tucker:
> exceptionally for this release we only document the major changes, then once we have automation to track changes we can be more strict.

🙏 Thanks!
I've got most of them now I think, but I'll document the perf_event ones also before merging this one.

PR #894, opened 2024-03-01
Add `CgroupDevice::query`

nbdd0121:
This follows closely the existing `LircMode2::query`.


alessandrod (commented):
Thanks for this! Look good, it seems that clippy is complaining about something

alessandrod (approved):

alessandrod:
Thank you!

PR #900, opened 2024-03-06
aya-log: allow re-attach and read previously created logs

catalin-h:
This feature is useful if someone wants to view the log contents of a program that is already running. 
For e.g. a pinned program or an XDP program attached to a net interface.

With the current API there is no viable way to re-attach to the aya-log once the user app exists. 

The only workaround that I've found is to _replace_ the loaded bpf program and have some of
the maps containing the program state pinned to bpffs so the new program can used that state.
But this implies that you can (always) reproduce the issue after replacing the program. This is 
something that not all bpf programs can do.


alessandrod:
Hey! Sorry for the delay in reviewing!

Thanks this seems useful, I'll do a proper review in the next couple of days.

aya-log/src/lib.rs
```diff
@@ -108,12 +110,54 @@ impl BpfLogger {
...
-        let mut logs: AsyncPerfEventArray<_> = bpf
-            .take_map(MAP_NAME)
-            .ok_or(Error::MapNotFound)?
-            .try_into()?;
+        let map = bpf.take_map(MAP_NAME).ok_or(Error::MapNotFound)?;
+        Self::read_logs_async(map, logger)?;
+        Ok(BpfLogger {})
+    }
+
+    /// Attaches to an existing `aya-log-ebpf` log created by a running ebpf
```
alessandrod, line 118: nit:

```
/// Starts reading log records created by the eBPF program identified by `program_id` and logs them
/// with the default logger. See [log::logger].
///
/// It can be used to read the logs of a pinned program.
```

I wouldn't say anything about background threads.

aya-log/src/lib.rs
```diff
@@ -108,12 +110,54 @@ impl BpfLogger {
...
+    pub fn init_from_id(program_id: u32) -> Result<BpfLogger, Error> {
+        Self::init_from_id_with_logger(program_id, log::logger())
+    }
 
+    /// Attaches to an existing `aya-log-ebpf` log created by a running ebpf
+    /// program identified by `program_id` and starts a background thread that
+    /// writes the log records to the provided logger.
+    /// It can be used to review the log for a pinned program after the user
+    /// application exits.
+    pub fn init_from_id_with_logger<T: Log + 'static>(
```
alessandrod, line 132: nit:

```
/// Starts reading log records created by the eBPF program identified by `program_id` and logs them  
/// with the default logger. See [log::logger]. 
```

aya-log/src/lib.rs
```diff
@@ -370,6 +413,12 @@ pub enum Error {
 
     #[error("invalid /sys/devices/system/cpu/online format")]
     InvalidOnlineCpu(#[source] io::Error),
+
+    #[error("program not found")]
+    ProgramNotFound,
+
+    #[error("error getting in use program maps")]
+    GetProgramMapIds(#[source] ProgramError),
```
alessandrod, line 421: I think that this variant should be called ProgramError, and should be error(transparent)

alessandrod (changes requested):
Great work! See comments.

aya-log/src/lib.rs
catalin-h: Ok, I've reworked the comment(s) w/o mentioning the thread.

aya-log/src/lib.rs
catalin-h: Reworked this change also

aya-log/src/lib.rs
```diff
@@ -112,12 +114,54 @@ impl EbpfLogger {
...
-            .try_into()?;
+        let map = bpf.take_map(MAP_NAME).ok_or(Error::MapNotFound)?;
+        Self::read_logs_async(map, logger)?;
+        Ok(EbpfLogger {})
+    }
+
+    /// Attaches to an existing `aya-log-ebpf` log created by a running ebpf
+    /// program identified by `program_id`. The log records will be written
+    /// to the default logger. See [log::logger].
+    /// It can be used to review the log for a pinned program after the user
```
alessandrod, line 125: nit: please add an empty line before this one. Separate paragraphs must be split by empty lines.

Also it should say "It can be used to read logs generated by a pinned program."

Review is not the correct word here, and the application doesn't necessarily have to exit, you could read logs of an application that is still running but wants another process to process for whatever reason (centralized logs for example).

aya-log/src/lib.rs
```diff
@@ -112,12 +114,54 @@ impl EbpfLogger {
...
+    /// application exits.
+    pub fn init_from_id(program_id: u32) -> Result<EbpfLogger, Error> {
+        Self::init_from_id_with_logger(program_id, log::logger())
+    }
 
+    /// Attaches to an existing `aya-log-ebpf` log created by a running ebpf
+    /// program identified by `program_id`. The log records will be written
+    /// to the user provided logger.
+    /// It can be used to review the log for a pinned program after the user
```
alessandrod, line 134: Same here

alessandrod (commented):
Thank you! Just a little doc change then it's ready to go

aya-log/src/lib.rs
catalin-h: I've used _review_ in the sense that someone or some process would want to _examine_ the logs later. I changed it to _read_ to not confuse about the meaning.
You're right it's not necessarily that the first app to exit.

alessandrod (approved):
Thanks a lot!

PR #901, opened 2024-03-07
aya: include license in crate workspace

danielmellado:
This PR includes the licenses files in the crate workspace subdirectory.
Without this, they won't be showing on crates.io and would be giving out
errors on tooling such as rust2rpm.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

### DIFF
--- a/aya/LICENSE-APACHE
+++ b/aya/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
--- a/aya/LICENSE-MIT
+++ b/aya/LICENSE-MIT
@@ -1,0 +1,1 @@
+../LICENSE-MIT


alessandrod:
Thanks!

Do symlinks actually work? Are they resolved when the .crate file is created or do they end up being included as symlinks? Also, we'd need to do the same for all the other crates I guess? (aya-bpf, aya-obj etc)

danielmellado:
Hi! Yeah, symlinks do work! And I didn't realize but deff we should duplicate this PR to the other workspaces even if in my case I was mainly interested in the main aya ;) 

In case you'd like to test the symlink thing you can test with `cargo-package`

```
cargo package -p path+file:///home/dmellado/aya/aya#0.12.0
...
Compiling aya v0.12.0 (/home/dmellado/aya/target/package/aya-0.12.0)

$ ls
BREAKING-CHANGES.md
Cargo.toml
Cargo.toml.orig
CHANGELOG.md
LICENSE-APACHE
LICENSE-MIT
README.md
src
target
```

As I see different crates.io packages for the workspace targets (aya-log and so) I'd split this into different PR, wdyt?

alessandrod:
> As I see different crates.io packages for the workspace targets (aya-log and so) I'd split this into different PR, wdyt?

Up to you, I'm fine with one PR but feel free to split!

alessandrod (approved):

PR #916, opened 2024-03-29
Implement InodeStorage

addisoncrump:
Addresses #202. Still need to write tests but I needed to pause for the evening, wanted to get some initial feedback.

### DIFF
--- a/ebpf/aya-ebpf/src/args.rs
+++ b/ebpf/aya-ebpf/src/args.rs
@@ -33,6 +33,14 @@ unsafe impl<T> FromBtfArgument for *const T {
     }
 }
 
+unsafe impl<T> FromBtfArgument for *mut T {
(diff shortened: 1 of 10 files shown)


addisoncrump:
This is known broken -- getting invalid argument here when loading the program. Not sure how to debug. Any tips?

vineethrp:
> nges the Aya Public AP



> This is known broken -- getting invalid argument here when loading the program at the map. Not sure how to debug. Any tips?

I was trying to implement BPF_MAP_TYPE_TASK_STORAGE and faced the same issue - invalid arguement. I have figured out that these new maps are implemented using the newer btf format and requires key_type_id and value_type_id. Aya supports only LegacyMap. So BPF_MAP_CREATE returns -EINVAL

addisoncrump:
Closing as this is not going to work, maybe this can be implemented another way in the future.

Issue #914, opened 2024-03-29
Instructions on codegen


kevinji:
How do you get `cargo +nightly xtask codegen` to work? I'm running into a bunch of missing header files on aarch64 Ubuntu (as a VM on macOS) so I'm wondering if there's instructions to get codegen working.

tyrone-wu:
Hello! 👋
I just made a PR #917  that should fix the current issue for this. Also, these packages need to be installed as well for codegen to work. 
```bash
sudo apt-get install libc6-dev libc6-dev-{arm64,armel,riscv64}-cross
```
This should probably be documented somewhere since I only found this when looking through the codegen GHA workflow.

PR #917, opened 2024-04-01
fix(xtask): corrected bpf to ebpf for path to aya-ebpf-bindings in codegen

tyrone-wu:
Currently, when running `cargo +nightly xtask codegen` locally and in the codegen GHA workflow, an error occurs with only "Error: bindgen failed" displayed.

This was due to a path using "bpf/..." instead of "ebpf/...". It is now corrected to "ebpf", so the codegen command and GHA workflow shouldn't fail. Additionally, any bindgen fails should now display a more direct message on why it failed.

Fixes: #914


tamird (approved):
Thanks!

PR #927, opened 2024-04-13
Mark aya-tool publish=false

tamird:


alessandrod (approved):
aya-tool is not meant to be an internal tool. That said it's little more than a prototype at the moment so if it's causing maintenance overhead i'm ok with getting it out of the way.

PR #928, opened 2024-04-13
Expose io_error in SyscallError

seanyoung:
If a bpf syscall fails, it would be useful to know if this is due to not having enough permissions or that the kernel does not have BPF support.

Ubuntu kernels are compiled without `CONFIG_BPF_LIRC_MODE2` so this is important for https://github.com/seanyoung/cir/ : I would like to present a useful error saying your kernel does not support BPF IR decoders, go fix your kernel.

Also, when removing all BPF programs using `cir config --clear` then if there is no support for BPF, there is nothing to clear and the syscall error can be ignored.

### DIFF
--- a/aya/src/sys/mod.rs
+++ b/aya/src/sys/mod.rs
@@ -48,10 +48,10 @@ pub(crate) enum Syscall<'a> {
 #[error("`{call}` failed")]
 pub struct SyscallError {
     /// The name of the syscall which failed.
-    pub(crate) call: &'static str,
+    pub call: &'static str,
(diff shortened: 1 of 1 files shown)


alessandrod (approved):

seanyoung:
It would be great to have this merged, please let me know if there is anything I can do to make this more amenable

alessandrod:
Oops sorry forgot to merge when I approved two weeks ago!

seanyoung:
Wonderful thank you @alessandrod

Issue #936, opened 2024-04-21
How about supporting for searching the symbols file from the gnu.debuglink section?


swananan:
Hi, I recently used aya to write some eBPF programs for tracing user programs. In my scenario, the release version is stripped, and the symbol file's relative path is placed in the _gnu.debuglink_ section. Unfortunately, aya doesn't currently support finding symbols from the _gnu.debuglink_ section, in my opinion, which is a pretty common use case. I've made some changes locally to support this small feature in my test environment. I'm wondering if there have been any plans or developments for adding this feature upstream. If needed, I'm happy to raise a PR for this.

alessandrod:
Hi! Yes I would love to see support for debuglink, and ideally for build_id too :)

WindySha:
Is it possible to support symbol in .gnu_debugdata section ? For Android platform, many elf symbols are hidden in .gnu_debugdata section.

PR #938, opened 2024-04-23
aya: add symbol lookup in associated debug files

swananan:
This change enhances the logic for symbol lookup in uprobe or uretprobe. If the symbol is not found in the original binary, the search continues in the debug file associated through the debuglink section. Before searching the symbol table, it compares the build IDs of the two files. The symbol lookup will only be terminated if both build IDs exist and do not match. This modification does not affect the existing symbol lookup logic.

Refs: #936

### DIFF
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -648,7 +648,7 @@ mod tests {
         obj.write()
     }
 
-    fn aligned_slice<'a>(vec: &'a mut Vec<u8>) -> &'a mut [u8] {
+    fn aligned_slice(vec: &mut Vec<u8>) -> &mut [u8] {
         let alignment = 8;
 
         let original_size = vec.len();
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -6,13 +6,13 @@ use std::{
     fs,
     io::{self, BufRead, Cursor, Read},
     mem,
-    os::{fd::AsFd as _, raw::c_char},
+    os::{fd::AsFd as _, raw::c_char, unix::ffi::OsStrExt},
     path::{Path, PathBuf},
     sync::Arc,
 };
 
 use libc::pid_t;
-use object::{Object, ObjectSection, ObjectSymbol};
+use object::{Object, ObjectSection, ObjectSymbol, Symbol};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


alessandrod:
Thanks! This looks great. Could you add a test for the debuglink resolution?

alessandrod:
> Thanks! This looks great. Could you add a test for the debuglink resolution?

Actually thinking of it, given our current integration-test setup this is _probably_ harder than it should be. Please take a look but if it's too cumbersome I'm fine just smoke testing this locally for now.

swananan:
> > Thanks! This looks great. Could you add a test for the debuglink resolution?
> 
> Actually thinking of it, given our current integration-test setup this is _probably_ harder than it should be. Please take a look but if it's too cumbersome I'm fine just smoke testing this locally for now.

I was initially planning to add UT, haha. However, I realized I needed to hook the `std::fs::read` function. If I use the `object` crate to prepare the testing file in memory (since preparing a real file might be too burdensome), and I'm unsure whether introducing a file mocking crate like `fake-fs` would be a good idea, especially since I am a new beginner to Rust and Aya.

I took a look at the integration test, but for this particular change, I think UT might be a more suitable choice because it doesn't affect the ebpf code part. If I am mistaken, please correct me.

If it's okay, I will add some UTs using the `object` and `fake-fs` crates as soon as possible.

alessandrod:
> I was initially planning to add UT, haha. However, I realized I needed to hook the std::fs::read function

unit test works! But why do you need to hook std::fs::read?

alessandrod:
> Since std::fs::read is called by resolve_symbol, I think I need to mock this function. This is because I don't intend to provide a real file path for resolve_symbol to process.

You could introduce a new resolve_symbol_data(...) and call it from resolve_symbol. resolve_symbol would fs::read then call resolve_symbol_data with the content. Then you could unit test resolve_symbol_data and not need fs access?

addisoncrump:
Super cool. I work a lot with stripped binaries for which I have separate debug symbols, so this will be a real help.

swananan:
@alessandrod could you review this pr? thanks

alessandrod:
> @alessandrod could you review this pr? thanks

Yep sorry been travelling this week, will take a look soon

aya/src/programs/uprobe.rs
```diff
@@ -432,17 +432,108 @@ enum ResolveSymbolError {
...
+                    e,
+                )
+            })?;
+            let debug_obj = object::read::File::parse(&*debug_data)?;
+
+            verify_build_ids(&obj, &debug_obj, symbol)?;
+
+            debug_obj_keeper = Some(debug_obj);
+            debug_obj_keeper.as_ref().map_or(
+                Err(ResolveSymbolError::Unknown(symbol.to_string())),
```
alessandrod, line 528: nit: why not unwrap here?

alessandrod (commented):
quality PR, thanks so much! Just a nit and please fix the lint issue

aya/src/programs/uprobe.rs
swananan: oh, I totally agree.

swananan:
Hi @alessandrod this PR failed the public api check phase, but I am not sure if it's the root cause, could you help take a look? thanks!

addisoncrump:
I think you can do `cargo xtask public-api --bless` or something similar to fix this.

alessandrod:
> Hi @alessandrod this PR failed the public api check phase, but I am not sure if it's the root cause, could you help take a look? thanks!

I think adding the write feature to the object crate pulls in a new transitive dependency - the equivalent crate. That crate has a blanket impl (smh), which impacts some of our types that we use as keys in maps. I just pushed a commit which hopefully makes CI green.

alessandrod:
Can you take a look at the test failure under miri?

swananan:
> Can you take a look at the test failure under miri?

I looked into this failure and found an exception, **Invalid ELF header size or alignment**, originating from the object crate. I checked the main_bytes generated by _create_elf_with_build_id_, and the ELF header size is 64, which seems correct.

However, further investigation into the object crate showed that it has a strict address alignment requirement, leading to a failure in the object::File::parse API (see [this link](https://github.com/gimli-rs/object/blob/0643b61e1db0eb97e320c3ca80d07a9e82897a9a/src/pod.rs#L34)). My understanding is that in Rust, a Vec's memory comes from the heap and should be aligned to 8 bytes on x64 architectures. However, since I'm new to Rust, I'm not entirely certain about this. Obviously, this alignment is not guaranteed with the miri interpreter (this is my first time hearing about miri, haha).

Given the strict limitation in the object crate, I decided to adjust the Vec<u8> to ensure 8-byte alignment in the tests as a workaround for miri (I am not sure if it's a good idea = =||). I've added a new commit with these changes. Could you please review it and let me know your thoughts?

alessandrod:
Yep ELF files must be aligned. Thank you again!

PR #939, opened 2024-04-24
fix CI failures on macos

alessandrod:
This fixes the current CI failures we're seeing on macos https://github.com/aya-rs/aya/actions/runs/8810405556/job/24182725423

The fix is to make sure bpf-linker is always rebuilt/relinked against the latest llvm installed via brew.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,9 +256,10 @@ jobs:
 
       - name: bpf-linker
         if: runner.os == 'macOS'
-        # NB: rustc doesn't ship libLLVM.so on macOS, so disable proxying (default feature).
+        # NB: rustc doesn't ship libLLVM.so on macOS, so disable proxying (default feature). We also
+        # --force so that bpf-linker gets always relinked against the latest LLVM installed by brew.
         # Remove --rev when LLVM18 is released.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -258,7 +258,7 @@ jobs:
         if: runner.os == 'macOS'
         # NB: rustc doesn't ship libLLVM.so on macOS, so disable proxying (default feature).
         # Remove --rev when LLVM18 is released.
```
tamird, line 260: looks like llvm is 18 now: https://formulae.brew.sh/formula/llvm

.github/workflows/ci.yml
```diff
@@ -258,7 +258,7 @@ jobs:
...
-        run: cargo install bpf-linker --git https://github.com/aya-rs/bpf-linker.git --rev 821f92990074cb7e950e25129dcd55e20424cede --no-default-features
+        run: cargo install --force bpf-linker --git https://github.com/aya-rs/bpf-linker.git --rev 821f92990074cb7e950e25129dcd55e20424cede --no-default-features
```
tamird, line 261: can you add a comment to explain why we're doing this?

.github/workflows/ci.yml
alessandrod: removed

.github/workflows/ci.yml
alessandrod: done

tamird (approved):

PR #941, opened 2024-04-30
Leak an FD to /dev/null to appease miri

tamird:
- **Deduplicate test helpers**
- **Leak an FD to /dev/null to appease miri**


tamird:
@vadorovsky PTAL

aya/src/sys/mod.rs
```diff
@@ -86,11 +86,12 @@ impl std::fmt::Debug for Syscall<'_> {
     }
 }
 
-fn syscall(call: Syscall<'_>) -> SysResult<c_long> {
+fn syscall(call: Syscall<'_>) -> SysResult<i64> {
     #[cfg(test)]
     return TEST_SYSCALL.with(|test_impl| unsafe { test_impl.borrow()(call) });
 
     #[cfg_attr(test, allow(unreachable_code))]
+    #[allow(clippy::useless_conversion)]
```
alessandrod, line 94: where is the useless conversion? this probably can be moved to a better place?

alessandrod (approved):
Thanks for fixing this, I definitely appreciate all your work on this chore stuff. Just a question but otherwise lgtm

aya/src/sys/mod.rs
tamird: It's the `.into()` at the bottom. I tried assigning a local but then we need more `allow(unreachable)`s.

tamird: Alright I've narrowed this.

Issue #587, opened 2023-04-24
kprobe attach __x64_sys_open get empty file from arg 0


nichoalsliu:
//env 
aya = { version = ">=0.11", features=["async_tokio"] }
ubuntu 22.04

//user space code 

```rust
    let program: &mut KProbe = bpf.program_mut("probe_open").unwrap().try_into()?;
    program.load()?;
    program.attach("__x64_sys_open", 0)?;

    let mut perf_array = AsyncPerfEventArray::try_from(bpf.map_mut("EVENTS")?)?;
     for cpu_id in online_cpus()? {
            let mut buf = perf_array.open(cpu_id, None)?;
            task::spawn(async move{
                let mut buffers = (0..10)
                .map(|_| BytesMut::with_capacity(1024))
                .collect::<Vec<_>>();

            loop{
                let events = buf.read_events(&mut buffers).await.unwrap();
                for buf in buffers.iter_mut().take(events.read) {
                    let ptr = buf.as_ptr() as *const PacketLog;
                    let data = unsafe { ptr.read_unaligned() };
                    let proc_name = String::from_utf8(data.prog_name.to_vec()).unwrap();
                    let file_name = String::from_utf8_lossy(&data.fname.to_vec()).into_owned();
                    println!("LOG: proc_name:{}, filename:{},flag:{}", proc_name,file_name,data.flg);                 
                }
            }
            });
     }
```

//ebpf code 

```rust
    fn try_probe_open(ctx: ProbeContext) -> Result<u32, u32> {
        let ptr_fname: *const u8 = ctx.arg(0).ok_or(1u32).map_err(|_| {info!(&ctx,"get arg 1 err");1 as u32})?;
        let  mut fname = [0u8; 100];
        unsafe{helpers::bpf_probe_read_kernel_str_bytes(ptr_fname,&mut fname).map_err(|e| {info!(&ctx,"read kernel buff err:{}",e); e as u32 })? };
        let flg: i32= ctx.arg(1).ok_or(1u32).map_err(|_| {info!(&ctx,"get arg 2 err");1 as u32})?;
  
        let prog_name = helpers::bpf_get_current_comm().map_err(|e| e as u32)?;

        let log_entry = PacketLog {
            fname,flg,prog_name,};
  
        unsafe {
            EVENTS.output(&ctx, &log_entry, 0);
        }

        Ok(0)
    }
```

//result 
I get the result like this:
"LOG: proc_name:xxxx, filename:,flag:0"
[...]

PR #944, opened 2024-05-14
aya-ebpf: ProbeContext: rewrite the implement of FromPtRegs trait.

WindySha:
The original implementation of the  trait is unable to correctly retrieve the arguments of a function, necessitating a rewrite. This includes two interfaces: `from_argument` and `from_retval`.

To access a function's argument, it is necessary to accurately identify the register address where the argument is stored. Subsequently, we can use the  function to read
the contents at this register address.
This allows us to correctly retrieve the argument of the function.

Fixes: #587

### DIFF
--- a/ebpf/aya-ebpf/src/args.rs
+++ b/ebpf/aya-ebpf/src/args.rs
@@ -92,149 +92,203 @@ impl PtRegs {
 pub trait FromPtRegs: Sized {
     /// Coerces a `T` from the `n`th argument of a pt_regs context where `n` starts
     /// at 0 and increases by 1 for each successive argument.
-    fn from_argument(ctx: &pt_regs, n: usize) -> Option<Self>;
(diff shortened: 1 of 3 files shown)


WindySha:
Why did the CI flow failed? I need some help.

> ---- tests::ring_buf::ring_buf::write_half_the_capacity_items stdout ----
> thread 'tests::ring_buf::ring_buf::write_half_the_capacity_items' panicked at test/integration-test/src/tests/ring_buf.rs:83:21:
> called `Result::unwrap()` on an `Err` value: LoadError { io_error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, verifier_log: 0: R1=ctx(off=0,imm=0) R10=fp0

Ablu:
> This includes two interfaces:  and .

This reads like there should be the interfaces spelled out? :)

Issue #948, opened 2024-05-17
feature: More public APIs for `XdpContext` to grow/shrink packets


gabhijit:
Currently the APIs available for `XdpContext` support mainly accessing the `data` and `data_end`. It would be handy to have the APIs that allow to grow or shrink the packet as are provided by `bpf_xdp_xdp_adjust_{head|tail|meta}`. 

Such API will be useful during actual packet processing, whenever additional headers are to be pushed to the packet etc. 

Will submit a draft PR for this.

PR #949, opened 2024-05-17
aya-ebpf: `XdpContext` adjust_{head|tail|meta} API

gabhijit:
Added new public API for the `XdpContext`:
- `adjust_head` : To grow the packet buffer at the beginning
- `adjist_tail` : To grow the packet buffer at the end
- `adjust_meta` : To grow metadata area

These API are handy for packet processing when additional headers etc. are to be pushed to the packet.

Also added these APIs using `cargo xtask public-api --bless`

Refs: #948

### DIFF
--- a/ebpf/aya-ebpf/src/programs/xdp.rs
+++ b/ebpf/aya-ebpf/src/programs/xdp.rs
@@ -1,6 +1,10 @@
 use core::ffi::c_void;
 
-use crate::{bindings::xdp_md, EbpfContext};
+use crate::{
+    bindings::xdp_md,
+    helpers::{bpf_xdp_adjust_head, bpf_xdp_adjust_meta, bpf_xdp_adjust_tail},
+    EbpfContext,
+};
 
 pub struct XdpContext {
     pub ctx: *mut xdp_md,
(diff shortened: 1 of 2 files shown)


gabhijit:
> I noticed from the comments in the code that passing a negative value can reduce the packet size, which would be very convenient for me. However, passing a negative value to the adjust_head method seems to increase the packet size. Am I missing something, or could there be an issue? Thank you.

Yes, using the negative value adjust the `data` pointer to the lower value and thus can be used to 'push' a header say and the opposite to 'pop' a header.

ebpf/aya-ebpf/src/programs/xdp.rs
```diff
@@ -32,6 +36,42 @@ impl XdpContext {
     pub fn metadata_end(&self) -> usize {
         self.data()
     }
+
+    /// Adjusts the head of the Packet by given 'delta' (both positive and negative values are
+    /// possible.)
+    #[inline(always)]
+    pub fn adjust_head(&mut self, delta: crate::cty::c_int) -> Result<(), ()> {
```
alessandrod, line 43: here and in the other methods I think we should use i32, not c_int.

ebpf/aya-ebpf/src/programs/xdp.rs
```diff
@@ -32,6 +36,42 @@ impl XdpContext {
...
+        unsafe {
+            match bpf_xdp_adjust_head(self.ctx, delta) {
+                0 => Ok(()),
+                _ => Err(()),
```
alessandrod, line 47: why not return the error?

alessandrod (changes requested):
This looks ok -

kjetijor:
@gabhijit are you planning on moving this forward ? If not, I might take this over and try to get it merged.

gabhijit:
@kjetijor : Sorry I had completely forgotten about it. Please go ahead and get it merged!! Apologies for letting it fall through the crack.

Issue #957, opened 2024-05-28
How to Detect Support for Ringbuf in Rust eBPF Code?


swananan:
Hi, I encountered a problem with my eBPF program written in Rust. I’m using the ringbuf, but it fails to run on the lower version kernel, presumably due to a lack of ringbuf support in the eBPF implementation of those kernels.

I was considering using Rust compile-time macros to conditionally use ringbuf based on kernel support, but this approach would result in multiple binaries, which I want to avoid.

So I did some research, I found that eBPF CO-RE and the **bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_ringbuf_reserve)** function can help detect the existence of ringbuf support without complaints from the eBPF verifier. However, I noticed that this API isn't available in the aya helper library. I guess **bpf_core_enum_value_exists** relies on LLVM eBPF functions, which might explain its absence in aya. If I understand it incorrectly, please let me know. 

Is there a better solution for detecting ringbuf or other features support in Rust eBPF code other than using Rust compile-time macros? Any guidance or suggestions would be greatly appreciated.

swananan:
Update: Refer to this link [here](https://lore.kernel.org/bpf/CADxym3b5zGP_zX3cUurqmwfR-dPhDXKx=QoxVG+-5EVtiadnFA@mail.gmail.com/T/). We have two potential solutions. The first is bpf_core_enum_value_exists, which Aya does not yet support. The second solution involves checking for support in user space and using an eBPF global value to control eBPF code. This approach needs to allow the loader or kernel to remove dead code and unnecessary maps from the eBPF code. It looks like these PRs (https://github.com/aya-rs/aya/pull/970 and https://github.com/aya-rs/aya/pull/968) are working on the second solution, and I am very much looking forward to using it.

swananan:
Update: Refer to this link [here](https://lore.kernel.org/bpf/CADxym3b5zGP_zX3cUurqmwfR-dPhDXKx=QoxVG+-5EVtiadnFA@mail.gmail.com/T/). We have two potential solutions. The first is bpf_core_enum_value_exists, which Aya does not yet support. The second solution involves checking for support in user space and using an eBPF global value to control eBPF code. This approach needs to allow the loader or kernel to remove dead code and unnecessary maps from the eBPF code. It looks like these PRs (https://github.com/aya-rs/aya/pull/970 and https://github.com/aya-rs/aya/pull/968) are working on the second solution, and I am very much looking forward to using it.

[...]

PR #961, opened 2024-05-31
aya: tidy bpf syscalls and feature support functions into submodules

tyrone-wu:
This PR organizes the bpf syscall related functions and feature support functions in `aya/src/sys/bpf.rs` into submodules: 
```
aya/src/sys/
  ├─ btf.rs
  ├─ link.rs
  ├─ map.rs
  ├─ object.rs
  ├─ program.rs
  ├─ utils.rs
  ├─ feature_probe.rs
  └─ ...
```
This is to help make development slightly less daunting to parse and expand on. For example, #959 would fit in `utils.rs`, and additional feature detection stuff (like #957) would go in `feature_probe.rs`.

This also updates all the imports that involve these submodules.

Still kind of a working progress with organizing the feature probe stuff.


tyrone-wu:
Actually on second thought, I don't think this is necessary. There's some things im looking at in aya-obj anyway.

PR #968, opened 2024-06-13
Draft: aya: skip map creation and relocation before loading for maps that should be deactivated

martinsoees:
Allows programs containing unsupported maps for the current 
kernel to be loaded by skipping map creation and relocation before loading.

This is useful when you have a single program containing e.g. a `RingBuf`
and a `PerfEventArray` and you decide from user space which one to use 
before loading the bytecode by checking if the kernel you are running 
on has support for e.g. `RingBuf`.
That way you can use the same eBPF bytecode across multiple kernels 
regardless of the kernel has support for it or not.
Can be used with `.set_global()` to signal if the map is supported 
until support for `bpf_core_enum_value_exists`.

Example user space code:
```Rust
let ringbuf_supported = ...

let mut set = HashSet::new();
set.insert(bpf_map_type::BPF_MAP_TYPE_RINGBUF);

let mut ebpf = EbpfLoader::new()
    .deactivate_maps(set)
    .set_global("RINGBUF_SUPPORTED", &ringbuf_supported, true)
    .load(include_bytes_aligned!(
    "../../target/bpfel-unknown-none/debug/disable_maps"
))?;

// ...
if ringbuf_supported {
    ringbuf_events()?.await;
} else {
    perfbuf_events()?.await;
}
// ...
```

Example eBPF bytecode:
```Rust
#[no_mangle]
static RINGBUF_SUPPORTED: i32 = 0;

#[map]
static mut RINGBUF: RingBuf = RingBuf::with_byte_size(1024 * 10, 0);

#[map]
static mut PERFBUF: PerfEventArray<Buffer> = PerfEventArray::with_max_entries(1, 0);

#[xdp]
pub fn my_func(ctx: XdpContext) -> u32 {
    match try_my_func(ctx) {
        Ok(ret) => ret,
        Err(_) => xdp_action::XDP_ABORTED,
    }
}

fn try_my_func(ctx: XdpContext) -> Result<u32, u32> {
    info!(&ctx, "received a packet");
    let ringbuf = unsafe { core::ptr::read_volatile(&RINGBUF_SUPPORTED) };
    if ringbuf == 1 {
        info!(&ctx, "Ringbuf supported");
        // do ringbuf output
    } else {
        info!(&ctx, "Defaulting to perfevent");
        // do perfbuf output
    }
    Ok(xdp_action::XDP_PASS)
}
```

### DIFF
--- a/aya-obj/src/relocation.rs
+++ b/aya-obj/src/relocation.rs
@@ -2,6 +2,7 @@
 
 use alloc::{borrow::ToOwned, collections::BTreeMap, string::String};
 use core::mem;
+use std::println;
 
 use log::debug;
 use object::{SectionIndex, SymbolKind};
@@ -108,9 +109,20 @@ impl Object {
         &mut self,
         maps: I,
         text_sections: &HashSet<usize>,
+        ignored_maps: &HashMap<String, Map>,
     ) -> Result<(), EbpfRelocationError> {
         let mut maps_by_section = HashMap::new();
         let mut maps_by_symbol = HashMap::new();
+        let mut ignored_by_section = HashSet::new();
+        let mut ignored_by_symbol = HashSet::new();
+
+        for (name, map) in ignored_maps {
+            ignored_by_section.insert(map.section_index());
+            if let Some(index) = map.symbol_index() {
+                ignored_by_symbol.insert(index);
+            }
+        }
+
         for (name, fd, map) in maps {
             maps_by_section.insert(map.section_index(), (name, fd, map));
             if let Some(index) = map.symbol_index() {
@@ -127,6 +139,8 @@ impl Object {
                     &maps_by_symbol,
                     &self.symbol_table,
                     text_sections,
+                    &ignored_by_section,
+                    &ignored_by_symbol,
                 )
                 .map_err(|error| EbpfRelocationError {
                     function: function.name.clone(),
@@ -176,6 +190,7 @@ impl Object {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
     fun: &mut Function,
     relocations: I,
@@ -183,6 +198,8 @@ fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
     maps_by_symbol: &HashMap<usize, (&str, std::os::fd::RawFd, &Map)>,
     symbol_table: &HashMap<usize, Symbol>,
     text_sections: &HashSet<usize>,
+    ignored_by_section: &HashSet<usize>,
+    ignored_by_symbol: &HashSet<usize>,
 ) -> Result<(), RelocationError> {
     let section_offset = fun.section_offset;
     let instructions = &mut fun.instructions;
@@ -212,6 +229,7 @@ fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
                 index: rel.symbol_index,
             })?;
 
+
         let Some(section_index) = sym.section_index else {
             // this is not a map relocation
             continue;
@@ -222,7 +240,9 @@ fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
             continue;
         }
 
-        let (_name, fd, map) = if let Some(m) = maps_by_symbol.get(&rel.symbol_index) {
+        let (_name, fd, map) = if ignored_by_symbol.contains(&rel.symbol_index) {
+            continue
+        } else if let Some(m) = maps_by_symbol.get(&rel.symbol_index) {
             let map = &m.2;
             debug!(
                 "relocating map by symbol index {:?}, kind {:?} at insn {ins_index} in section {}",
@@ -232,6 +252,8 @@ fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
             );
             debug_assert_eq!(map.symbol_index().unwrap(), rel.symbol_index);
             m
+        } else if ignored_by_section.contains(&section_index) {
+            continue
         } else {
             let Some(m) = maps_by_section.get(&section_index) else {
                 debug!("failed relocating map by section index {}", section_index);
@@ -580,6 +602,8 @@ mod test {
             &maps_by_symbol,
             &symbol_table,
             &HashSet::new(),
+            &HashSet::new(),
+            &HashSet::new(),
         )
         .unwrap();
 
@@ -636,6 +660,8 @@ mod test {
             &maps_by_symbol,
             &symbol_table,
             &HashSet::new(),
+            &HashSet::new(),
+            &HashSet::new(),
         )
         .unwrap();
 
@@ -674,6 +700,8 @@ mod test {
             &maps_by_section,
             &maps_by_symbol,
             &symbol_table,
+            &HashSet::new(),
+            &HashSet::new(),
             &HashSet::new(),
         )
         .unwrap();
@@ -730,6 +758,8 @@ mod test {
             &maps_by_section,
             &maps_by_symbol,
             &symbol_table,
+            &HashSet::new(),
+            &HashSet::new(),
             &HashSet::new(),
         )
         .unwrap();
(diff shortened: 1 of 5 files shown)


martinsoees:
Perhaps this should be scoped to only kernel-userspace data streaming (`RingBuf`and `PerfEvent`).
Then the API could define a primary and a fallback data map, where `RingBuf` is primary and `PerfEvent` is a fallback map.
Would obviously require that the dev correctly handles this and reads from the correct map

alessandrod:
```
 fn try_my_func(ctx: XdpContext) -> Result<u32, u32> {
    info!(&ctx, "received a packet");
    let ringbuf = unsafe { core::ptr::read_volatile(&RINGBUF_SUPPORTED) };
    if ringbuf == 1 {
        info!(&ctx, "Ringbuf supported");
        // do ringbuf output
    } else {
        info!(&ctx, "Defaulting to perfevent");
        // do perfbuf output
    }
    Ok(xdp_action::XDP_PASS)
}
```

Does this actually work, if in the ringbuf==1 case you access the RINGBUF map, and then in the else you access the PERFBUF map? Because if you skip relocating it, I would have expected the verifier to complain that the map fd is not valid.

If this does work, can you add an integration test please?

alessandrod:
> Perhaps this should be scoped to only kernel-userspace data streaming (`RingBuf`and `PerfEvent`). Then the API could define a primary and a fallback data map, where `RingBuf` is primary and `PerfEvent` is a fallback map. Would obviously require that the dev correctly handles this and reads from the correct map

I tend to think the opposite: the current mechanism is too restrictive and it should be more generic. Right now you can ignore loading a map type, but what if I want to ignore loading only _an instance_ of a map type (because eg, a bit of functionality is disabled, and I want to save the memory?).

I'm thinking it's probably better to have something like `EbpfLoader::ignore_maps(&["RINGBUF"])` - do the ignore by name that is. Then an user can ignore for whatever reason, not only based on map type.

martinsoees:
> > Perhaps this should be scoped to only kernel-userspace data streaming (`RingBuf`and `PerfEvent`). Then the API could define a primary and a fallback data map, where `RingBuf` is primary and `PerfEvent` is a fallback map. Would obviously require that the dev correctly handles this and reads from the correct map
> 
> I tend to think the opposite: the current mechanism is too restrictive and it should be more generic. Right now you can ignore loading a map type, but what if I want to ignore loading only _an instance_ of a map type (because eg, a bit of functionality is disabled, and I want to save the memory?).
> 
> I'm thinking it's probably better to have something like `EbpfLoader::ignore_maps(&["RINGBUF"])` - do the ignore by name that is. Then an user can ignore for whatever reason, not only based on map type.

It actually started out like that, with the name rather than the type. I can revert it back to that or add a new function that filters by name and keep the one by id also? I see no reason why both can't exist

martinsoees:
This PR has a clippy allow for `too_many_arguments` on `relocate_maps`. Ideally this should be fixed, but would require a larger rewrite of the map relocation section, so I suggest that it should be a separate PR. Unless you want it in this PR?

aya/src/bpf.rs
```diff
@@ -224,6 +228,62 @@ impl<'a> EbpfLoader<'a> {
...
+    /// let mut set = HashSet::new();
+    /// set.insert(bpf_map_type::BPF_MAP_TYPE_RINGBUF);
+    /// let ebpf = EbpfLoader::new()
+    ///     .ignore_maps_by_type(set)
+    ///     .set_global("RINGBUF_SUPPORTED", &ringbuf_supported, true)
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::EbpfError>(())
+    /// ```
+    ///
+    pub fn ignore_maps_by_type(&mut self, set: HashSet<bpf_map_type>) -> &mut Self {
```
alessandrod, line 256: We shouldn't expose `bpf_map_type` here. It's a type that we don't directly
control so it'll break our API. I don't think we already have our own MapType,
maybe it's time to have it.

aya/src/bpf.rs
```diff
@@ -224,6 +228,62 @@ impl<'a> EbpfLoader<'a> {
         self
     }
 
+    /// Allows programs containing unsupported maps for the current kernel to be loaded
```
alessandrod, line 231: nit: I think this says too much about the implementation

I'd say:

```
Don't load maps of the given types.  
```

aya/src/bpf.rs
```diff
@@ -224,6 +228,62 @@ impl<'a> EbpfLoader<'a> {
...
+    /// by skipping map creation and relocation before loading.
+    ///
+    /// This is useful when you have a single ebpf program containing e.g. a `RingBuf`
+    /// and a `PerfEventArray` and you decide which one to use before loading the bytecode.
+    ///
+    /// Must be used with `.set_global()` to signal if the map is supported in the ebpf program.
```
alessandrod, line 237: Instead of `.set_global()` we should use whatever is the correct rustdoc syntax
to get a link to the method's doc.

aya/src/bpf.rs
```diff
@@ -224,6 +228,62 @@ impl<'a> EbpfLoader<'a> {
...
+    ///
+    /// let ringbuf_supported = 0;
+    /// let ebpf = EbpfLoader::new()
+    ///     .ignore_maps_by_name(&["RINGBUF"])
+    ///     .set_global("RINGBUF_SUPPORTED", &ringbuf_supported, true)
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::EbpfError>(())
+    /// ```
+    ///
+    pub fn ignore_maps_by_name(&mut self, name: &'a [&'a str]) -> &mut Self {
```
alessandrod, line 282: Why does this take a slice, and by_type takes a set?

aya/src/bpf.rs
```diff
@@ -224,6 +228,62 @@ impl<'a> EbpfLoader<'a> {
...
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::EbpfError>(())
+    /// ```
+    ///
+    pub fn ignore_maps_by_type(&mut self, set: HashSet<bpf_map_type>) -> &mut Self {
+        self.ignore_maps_by_type = set;
+        self
+    }
+
+    /// Allows programs containing unsupported maps for the current kernel to be loaded
```
alessandrod, line 261:
```
Don't load maps with the given names.

Similar to `ignore_maps_by_type()` but skips based on map name instead of type.
```

No need to repeat the same example. Or alternatively provide a different example
(for inspiration: something I plan to do with this is to avoid loading the MEMORY_EVENTS map of
my profiler when memory profiling is disabled)

alessandrod (changes requested):

alessandrod (changes requested):
Very nice work! See comments

aya/src/bpf.rs
martinsoees: Do you think the function should be removed until we have our own MapType, or do you want me to introduce it?

I can remove it for this PR and create a new PR that introduces MapType if you think that's best?

martinsoees:
> Very nice work! See comments

Thanks for the feedback. I left a comment to one of your review comments. The rest is straight forward. I'll see when I have time to fix it

aya/src/bpf.rs
alessandrod: Up to you I'm fine with either!

aya/src/bpf.rs
alessandrod: btw I think a slice is better, even if technically set is more correct, slice results in better ergonomics

aya/src/bpf.rs
martinsoees: Removed by_type, that can come in a later PR if someone needs it

aya-obj/src/relocation.rs
```diff
@@ -232,6 +250,8 @@ fn relocate_maps<'a, I: Iterator<Item = &'a Relocation>>(
             );
             debug_assert_eq!(map.symbol_index().unwrap(), rel.symbol_index);
             m
+        } else if ignored_by_section.contains(&section_index) {
```
alessandrod, line 253: I don't think this is correct, it's just not getting exercised by the test because we do have symbols. We switched to putting all maps in the same section (like libbpf does), so I think that this would skip all the maps?

PR #975, opened 2024-06-28
chore(xtask): Add s390x and powerpc64 to codegen

dave-tucker:
This commit adds the s390x and powerpc architectures to codegen.
This will enable an upcoming PR to add support for aya to support theses architectures in both aya and aya-ebpf.


dave-tucker:
Merging as this is a minor xtask change. Then I'll get the codegen workflow kicked off.

PR #976, opened 2024-06-28
ci: Fix codegen and prep for s390/ppc64el

dave-tucker:
Adds missing libelf headers that are now included
as part of libbpf-internal.h. Adds ppc64el and
s390x to the cross environment.


dave-tucker:
Another smol change to fix a CI workflow 🙏 Or at least get past the first error it's throwing at us.

PR #977, opened 2024-06-28
chore(xtask): Create bindings dir

dave-tucker:
If the bindings directory doesn't exist then create it.


dave-tucker:
Another smol xtask change. Sorry for the noise.

xtask/src/codegen/aya_ebpf_bindings.rs
```diff
@@ -112,6 +115,10 @@ pub fn codegen(opts: &SysrootOptions, libbpf_dir: &Path) -> Result<(), anyhow::E
         }
 
         let generated = dir.join("src").join(arch.to_string());
+        if !generated.exists() {
```
tamird, line 118: technically not necessary check

PR #979, opened 2024-06-29
aya,aya-obj: conversion from `u32` to prog, map, and link type

tyrone-wu:
I have a usecase that extracts program type information from from `bpf_prog_info`. 😺

This PR adds conversion from `u32` to `bpf_prog_type` so that the `type_` field in `bpf_prog_info` can map to the program type enum. This also implements Display for `bpf_prog_type` for condensed output format. 😺 

Also, `codegen` and `public-api` appears to be acting a little funky for me.


tyrone-wu:
So the scope of this PR expanded to include metadata info from `bpf_map_info` and `bpf_link_info`. I'm also planning to info for `bpf_link_info__bindgen_ty_1`.

tyrone-wu:
I almost accidently deleted this PR bc of the branch name I had. Since the scope of this PR expanded, this will go under a new branch since the one I'm working on locally is wacky.

PR #985, opened 2024-07-10
Add the option to support multiple and overrideable programs per cgroup

reyzell:
Add the option to support multiple and overrideable programs per cgroup
    
This change allows multiple BPF programs to attach to a cgroup (via the option
`CgroupAttachMode::AllowMultiple`), and allows a program to specify that it can be
overridden by one in a sub-cgroup (via the option `CgroupAttachMode::AllowOverride`).

### DIFF
--- a/aya/src/programs/cgroup_skb.rs
+++ b/aya/src/programs/cgroup_skb.rs
@@ -8,7 +8,8 @@ use crate::{
         bpf_prog_type::BPF_PROG_TYPE_CGROUP_SKB,
     },
     programs::{
-        define_link_wrapper, load_program, FdLink, Link, ProgAttachLink, ProgramData, ProgramError,
+        define_link_wrapper, load_program, CgroupAttachMode, FdLink, Link, ProgAttachLink,
+        ProgramData, ProgramError,
     },
     sys::{bpf_link_create, LinkTarget, SyscallError},
     util::KernelVersion,
@@ -43,12 +44,12 @@ use crate::{
 /// # }
 /// # let mut bpf = aya::Ebpf::load(&[])?;
 /// use std::fs::File;
-/// use aya::programs::{CgroupSkb, CgroupSkbAttachType};
+/// use aya::programs::{CgroupAttachMode, CgroupSkb, CgroupSkbAttachType};
 ///
 /// let file = File::open("/sys/fs/cgroup/unified")?;
 /// let egress: &mut CgroupSkb = bpf.program_mut("egress_filter").unwrap().try_into()?;
 /// egress.load()?;
-/// egress.attach(file, CgroupSkbAttachType::Egress)?;
+/// egress.attach(file, CgroupSkbAttachType::Egress, CgroupAttachMode::Single)?;
 /// # Ok::<(), Error>(())
 /// ```
 #[derive(Debug)]
@@ -87,6 +88,7 @@ impl CgroupSkb {
         &mut self,
         cgroup: T,
         attach_type: CgroupSkbAttachType,
+        mode: CgroupAttachMode,
     ) -> Result<CgroupSkbLinkId, ProgramError> {
         let prog_fd = self.fd()?;
         let prog_fd = prog_fd.as_fd();
@@ -97,18 +99,24 @@ impl CgroupSkb {
             CgroupSkbAttachType::Egress => BPF_CGROUP_INET_EGRESS,
         };
         if KernelVersion::current().unwrap() >= KernelVersion::new(5, 7, 0) {
-            let link_fd = bpf_link_create(prog_fd, LinkTarget::Fd(cgroup_fd), attach_type, None, 0)
-                .map_err(|(_, io_error)| SyscallError {
-                    call: "bpf_link_create",
-                    io_error,
-                })?;
+            let link_fd = bpf_link_create(
+                prog_fd,
+                LinkTarget::Fd(cgroup_fd),
+                attach_type,
+                None,
+                mode.into(),
+            )
+            .map_err(|(_, io_error)| SyscallError {
+                call: "bpf_link_create",
+                io_error,
+            })?;
             self.data
                 .links
                 .insert(CgroupSkbLink::new(CgroupSkbLinkInner::Fd(FdLink::new(
                     link_fd,
                 ))))
         } else {
-            let link = ProgAttachLink::attach(prog_fd, cgroup_fd, attach_type)?;
+            let link = ProgAttachLink::attach(prog_fd, cgroup_fd, attach_type, mode)?;
 
             self.data
                 .links
--- a/aya/src/programs/links.rs
+++ b/aya/src/programs/links.rs
@@ -10,7 +10,7 @@ use std::{
 use thiserror::Error;
 
 use crate::{
-    generated::bpf_attach_type,
+    generated::{bpf_attach_type, BPF_F_ALLOW_MULTI, BPF_F_ALLOW_OVERRIDE},
     pin::PinError,
     programs::{ProgramError, ProgramFd},
     sys::{bpf_get_object, bpf_pin_object, bpf_prog_attach, bpf_prog_detach, SyscallError},
@@ -26,6 +26,30 @@ pub trait Link: std::fmt::Debug + 'static {
 
     /// Detaches the LinkOwnedLink is gone... but this doesn't work :(
     fn detach(self) -> Result<(), ProgramError>;
+}
+
+/// Program attachment mode.
+#[derive(Clone, Copy, Debug, Default)]
+pub enum CgroupAttachMode {
+    /// Allows only one BPF program in the cgroup subtree.
+    #[default]
+    Single,
+
+    /// Allows the program to be overridden by one in a sub-cgroup.
+    AllowOverride,
+
+    /// Allows multiple programs to be run in the cgroup subtree.
+    AllowMultiple,
+}
+
+impl From<CgroupAttachMode> for u32 {
+    fn from(mode: CgroupAttachMode) -> Self {
+        match mode {
+            CgroupAttachMode::Single => 0,
+            CgroupAttachMode::AllowOverride => BPF_F_ALLOW_OVERRIDE,
+            CgroupAttachMode::AllowMultiple => BPF_F_ALLOW_MULTI,
+        }
+    }
 }
 
 #[derive(Debug)]
@@ -252,6 +276,7 @@ impl ProgAttachLink {
         prog_fd: BorrowedFd<'_>,
         target_fd: BorrowedFd<'_>,
         attach_type: bpf_attach_type,
+        mode: CgroupAttachMode,
     ) -> Result<Self, ProgramError> {
         // The link is going to own this new file descriptor so we are
         // going to need a duplicate whose lifetime we manage. Let's
@@ -261,7 +286,7 @@ impl ProgAttachLink {
         let prog_fd = crate::MockableFd::from_fd(prog_fd);
         let target_fd = target_fd.try_clone_to_owned()?;
         let target_fd = crate::MockableFd::from_fd(target_fd);
-        bpf_prog_attach(prog_fd.as_fd(), target_fd.as_fd(), attach_type)?;
+        bpf_prog_attach(prog_fd.as_fd(), target_fd.as_fd(), attach_type, mode.into())?;
 
         let prog_fd = ProgramFd(prog_fd);
         Ok(Self {
@@ -377,7 +402,11 @@ mod tests {
     use tempfile::tempdir;
 
     use super::{FdLink, Link, LinkMap};
-    use crate::{programs::ProgramError, sys::override_syscall};
+    use crate::{
+        generated::{BPF_F_ALLOW_MULTI, BPF_F_ALLOW_OVERRIDE},
+        programs::{CgroupAttachMode, ProgramError},
+        sys::override_syscall,
+    };
 
     #[derive(Debug, Hash, Eq, PartialEq)]
     struct TestLinkId(u8, u8);
@@ -522,5 +551,18 @@ mod tests {
         let pinned_link = fd_link.pin(&pin).expect("pin failed");
         pinned_link.unpin().expect("unpin failed");
         assert!(!pin.exists());
+    }
+
+    #[test]
+    fn test_cgroup_attach_flag() {
+        assert_eq!(u32::from(CgroupAttachMode::Single), 0);
+        assert_eq!(
+            u32::from(CgroupAttachMode::AllowOverride),
+            BPF_F_ALLOW_OVERRIDE
+        );
+        assert_eq!(
+            u32::from(CgroupAttachMode::AllowMultiple),
+            BPF_F_ALLOW_MULTI
+        );
     }
 }
--- a/aya/src/programs/lirc_mode2.rs
+++ b/aya/src/programs/lirc_mode2.rs
@@ -3,7 +3,10 @@ use std::os::fd::{AsFd, AsRawFd as _, RawFd};
 
 use crate::{
     generated::{bpf_attach_type::BPF_LIRC_MODE2, bpf_prog_type::BPF_PROG_TYPE_LIRC_MODE2},
-    programs::{load_program, query, Link, ProgramData, ProgramError, ProgramFd, ProgramInfo},
+    programs::{
+        load_program, query, CgroupAttachMode, Link, ProgramData, ProgramError, ProgramFd,
+        ProgramInfo,
+    },
     sys::{bpf_prog_attach, bpf_prog_detach, bpf_prog_get_fd_by_id},
 };
 
@@ -69,7 +72,12 @@ impl LircMode2 {
         let lircdev_fd = lircdev.as_fd().try_clone_to_owned()?;
         let lircdev_fd = crate::MockableFd::from_fd(lircdev_fd);
 
-        bpf_prog_attach(prog_fd.as_fd(), lircdev_fd.as_fd(), BPF_LIRC_MODE2)?;
+        bpf_prog_attach(
+            prog_fd.as_fd(),
+            lircdev_fd.as_fd(),
+            BPF_LIRC_MODE2,
+            CgroupAttachMode::Single.into(),
+        )?;
 
         self.data.links.insert(LircLink::new(prog_fd, lircdev_fd))
     }
--- a/aya/src/programs/sock_ops.rs
+++ b/aya/src/programs/sock_ops.rs
@@ -4,8 +4,8 @@ use std::os::fd::AsFd;
 use crate::{
     generated::{bpf_attach_type::BPF_CGROUP_SOCK_OPS, bpf_prog_type::BPF_PROG_TYPE_SOCK_OPS},
     programs::{
-        define_link_wrapper, load_program, ProgAttachLink, ProgAttachLinkId, ProgramData,
-        ProgramError,
+        define_link_wrapper, load_program, CgroupAttachMode, ProgAttachLink, ProgAttachLinkId,
+        ProgramData, ProgramError,
     },
 };
 
@@ -35,12 +35,12 @@ use crate::{
 /// # }
 /// # let mut bpf = aya::Ebpf::load(&[])?;
 /// use std::fs::File;
-/// use aya::programs::SockOps;
+/// use aya::programs::{CgroupAttachMode, SockOps};
 ///
 /// let file = File::open("/sys/fs/cgroup/unified")?;
 /// let prog: &mut SockOps = bpf.program_mut("intercept_active_sockets").unwrap().try_into()?;
 /// prog.load()?;
-/// prog.attach(file)?;
+/// prog.attach(file, CgroupAttachMode::Single)?;
 /// # Ok::<(), Error>(())
 #[derive(Debug)]
 #[doc(alias = "BPF_PROG_TYPE_SOCK_OPS")]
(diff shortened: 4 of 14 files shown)


aya/src/programs/lirc_mode2.rs
```diff
@@ -68,7 +68,8 @@ impl LircMode2 {
         let prog_fd = prog_fd.try_clone()?;
         let lircdev_fd = lircdev.as_fd().try_clone_to_owned()?;
 
-        bpf_prog_attach(prog_fd.as_fd(), lircdev_fd.as_fd(), BPF_LIRC_MODE2)?;
+        const NO_FLAGS : u32 = 0;
+        bpf_prog_attach(prog_fd.as_fd(), lircdev_fd.as_fd(), BPF_LIRC_MODE2, NO_FLAGS)?;
```
dave-tucker, line 72: Per comment below this would become `CgroupAttachFlags::default()`

aya/src/programs/sock_ops.rs
```diff
@@ -2,13 +2,27 @@
...
     programs::{
         define_link_wrapper, load_program, ProgAttachLink, ProgAttachLinkId, ProgramData,
         ProgramError,
     },
 };
 
+bitflags::bitflags! {
+    /// Flags passed to [`SockOps::attach()`].
+    #[derive(Clone, Copy, Debug, Default)]
+    pub struct SockOpsFlags: u32 {
```
dave-tucker, line 18: These aren't just SockOps flags though.
See: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L1153

This should probably live in `links.rs` and be called `CgroupAttachFlags`.

aya/src/programs/links.rs
```diff
@@ -252,14 +252,23 @@ impl ProgAttachLink {
         prog_fd: BorrowedFd<'_>,
         target_fd: BorrowedFd<'_>,
         attach_type: bpf_attach_type,
+    ) -> Result<Self, ProgramError> {
+        Self::attach_with_flags(prog_fd, target_fd, attach_type, 0)
+    }
+
+    pub(crate) fn attach_with_flags(
```
dave-tucker, line 259: I'd prefer changing the API here to add `flags: u32` instead of adding a new funciton.

aya/src/programs/sock_ops.rs
```diff
@@ -58,9 +72,16 @@ impl SockOps {
     ///
     /// The returned value can be used to detach, see [SockOps::detach].
     pub fn attach<T: AsFd>(&mut self, cgroup: T) -> Result<SockOpsLinkId, ProgramError> {
+        self.attach_with_flags(cgroup, SockOpsFlags::empty())
+    }
+
+    /// Attaches the program to the given cgroup.
+    ///
+    /// The returned value can be used to detach, see [SockOps::detach].
+    pub fn attach_with_flags<T: AsFd>(&mut self, cgroup: T, flags: SockOpsFlags) -> Result<SockOpsLinkId, ProgramError> {
```
dave-tucker, line 81: I think I'd rather break the API here and add `flags: CgroupAttachFlags`.
This makes it explicit which flags are being used.

aya/src/programs/sock_ops.rs
```diff
@@ -58,9 +72,16 @@ impl SockOps {
...
         let prog_fd = self.fd()?;
 
-        let link = ProgAttachLink::attach(prog_fd.as_fd(), cgroup.as_fd(), BPF_CGROUP_SOCK_OPS)?;
+        let link = ProgAttachLink::attach_with_flags(prog_fd.as_fd(), cgroup.as_fd(), BPF_CGROUP_SOCK_OPS, flags.bits())?;
```
dave-tucker, line 84: Aside: Given all the other Cgroup programs can be attached with bpf_link - see: https://github.com/aya-rs/aya/blob/d5414bf10c80ae8cef757f0cdf06bfdd38746daa/aya/src/programs/cgroup_device.rs#L69

I'd wager that this is likely the case for sock_ops programs too. I'll open an issue for that.

dave-tucker (changes requested):

aya/src/programs/links.rs
reyzell: Sure thing, done.

aya/src/programs/sock_ops.rs
reyzell: Sounds good.  I wasn't sure how much of an expanding change you'd prefer.  I've modified the `attach()` fn signature here and for other program types.

aya/src/programs/lirc_mode2.rs
reyzell: I've updated the fn signature to allow flags to be passed in, and updated doc tests to pass in `CgroupAttachFlags::empty()`.  Let me know if that's what you had in mind.

aya/src/programs/sock_ops.rs
reyzell: 👍

aya/src/programs/sock_ops.rs
reyzell: Good call.  Done.

reyzell:
Hello, I'm just following up on the status of this one.

aya/src/programs/links.rs
```diff
@@ -28,6 +31,25 @@ pub trait Link: std::fmt::Debug + 'static {
     fn detach(self) -> Result<(), ProgramError>;
 }
 
+bitflags::bitflags! {
+    /// Flags passed to [`SockOps::attach()`].
```
dave-tucker, line 35: This doc comment is now inaccurate

aya/src/programs/links.rs
```diff
@@ -28,6 +31,25 @@ pub trait Link: std::fmt::Debug + 'static {
...
+    #[derive(Clone, Copy, Debug, Default)]
+    pub struct CgroupAttachFlags: u32 {
+        /// Allows the eBPF program to be overridden by ones in a descendent cgroup.
+        const ALLOW_OVERRIDE = BPF_F_ALLOW_OVERRIDE;
+        /// Allows multiple eBPF programs to be attached.
+        const ALLOW_MULTI = BPF_F_ALLOW_MULTI;
+        /// Replaces an existing attachment.
+        const REPLACE = BPF_F_REPLACE;
```
dave-tucker, line 43: Have you tested this works as expected?

aya/src/programs/links.rs
```diff
@@ -28,6 +31,25 @@ pub trait Link: std::fmt::Debug + 'static {
...
+        /// Attaches the program so as to be executed first for the cgroup.
+        const BEFORE = BPF_F_BEFORE;
```
dave-tucker, line 45: And this one

aya/src/programs/links.rs
```diff
@@ -28,6 +31,25 @@ pub trait Link: std::fmt::Debug + 'static {
...
+        /// Attaches the program so as to be executed last for the cgroup.
+        const AFTER = BPF_F_AFTER;
```
dave-tucker, line 47: And this one too

aya/src/programs/links.rs
```diff
@@ -28,6 +31,25 @@ pub trait Link: std::fmt::Debug + 'static {
...
+        /// Indicates the eBPF program is referred to by its ID.
+        const ID = BPF_F_ID;
```
dave-tucker, line 49: This wouldn't make sense to include anyway since we use FD, not ID

dave-tucker (changes requested):
Almost there. See notes on the other generic flags you included in the CgroupAttachFlags.
I'm fine to have them there if you've tested that they work as expected...

I'd be genuinely surprised if all of the flags are honoured by every program type that attaches to cgroups.

I would recommend dropping them for the sake of simplicity 😆 
Otherwise, the generic flags might be better off as `struct GenericAttachFlags` it's been a while since I looked at bitflags but it would be nice to express that the flags could be either `CgroupAttachFlags` or `GenericAttachFlags`

aya/src/programs/links.rs
reyzell: Ah, true.  I've switched it to the more general comment of "Program attachment flags."  Let me know if you'd like a more descriptive comment.

aya/src/programs/links.rs
reyzell: Understood.  I've removed these additional flags, since `ALLOW_MULTI` and `ALLOW_OVERRIDE` were the focus of my change.

reyzell:
> Almost there. See notes on the other generic flags you included in the CgroupAttachFlags. I'm fine to have them there if you've tested that they work as expected...
> 
> I'd be genuinely surprised if all of the flags are honoured by every program type that attaches to cgroups.
> 
> I would recommend dropping them for the sake of simplicity 😆 

Ha, so what's funny is on the fork of Aya I started with internally, I added only the `ALLOW_OVERRIDE` and `ALLOW_MULTI` flags.  I included the full set of flags in this pull request because I assumed you'd want a more fully-featured change.  Bad guess on my part.  I've now paired this down to just the two flags.

reyzell:
### Testing

I tested with the application I'm developing by attaching the program twice to a cgroup and kicking off some network chatter.  This was performed under three different conditions: 1) before this change (i.e. using the `::empty()` flag), 2) using flag `ALLOW_OVERRIDE`, and 3) using flag `ALLOW_MULTI`.  The output logged from the application showed the multiple invocations of the BPF programs in `ALLOW_MULTI` mode, and showed the invocation was taken over by the latter program when in `ALLOW_OVERRIDE` mode.

aya/src/programs/links.rs
reyzell: Please see the **Testing** comment I've added below.  This covers the two flags that have been kept in this PR.

reyzell:
Updates: I've rebased from `aya/main` and solved clippy warnings on unused imports.

dave-tucker (approved):

reyzell:
@dave-tucker I see the lint error due to changes to the public API [a].  Are there any actions I should take here?

[a] https://github.com/aya-rs/aya/actions/runs/10159730726/job/28109127559?pr=985

tyrone-wu:
It looks like you'll need to run `cargo +nightly xtask public-api --bless` to update the public API. The public API is tracked using a `txt` file.

reyzell:
Thanks @tyrone-wu, done.  Blessed the public API.

reyzell:
Hello, @alessandrod and @dave-tucker, I'm checking in on the status of this one.  Please let me know if I can help in any way.

aya/src/programs/sock_ops.rs
alessandrod: > This makes it explicit which flags are being used.

I'm in two minds about this. For the `Cgroup*` and `SockOps` programs, I think it makes sense to have to pass `CgroupAttachFlags` to `attach()` - you're explicitly working with cgroups, so it makes sense to _have_ to think about what's going on with other cgroups.

But for `SkMsg` `SkSkb` and  `LircMode2`, I think we should default to no flags like before, and have a separate `attach_with_flags()` for the cases in which you do care about cgroups.

alessandrod (commented):
Thanks for doing this work!

Looks great, as per comment I suspect CgroupAttachFlags should be a regular enum, and some programs should have an explicit attach_with_flags, see https://github.com/aya-rs/aya/pull/985#discussion_r1716816332

aya/src/programs/sock_ops.rs
reyzell: That's a keen observation.  I've updated this change following your advice to extend `attach()` for specific program types, and adding `attach_with_flags()` where the developer does not necessarily need to think about cgroups.

aya/src/programs/links.rs
```diff
@@ -28,6 +28,17 @@ pub trait Link: std::fmt::Debug + 'static {
     fn detach(self) -> Result<(), ProgramError>;
 }
 
+bitflags::bitflags! {
```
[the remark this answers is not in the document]
reyzell: I see where you're going.  I took a look, and I believe the use of bitflags is the more usable interface in this case.  This allows the caller to easily set a combination of flags if they wish.  The bitflags API frees the developer from having to do bit manipulation themselves.

If we later choose to add support for more flags, such as `BPF_F_REPLACE`, `BPF_F_BEFORE`, and `BPF_F_AFTER`, then we would add the ability to supply the `bpf_attr` anonymous struct's three additional fields of `replace_bpf_fd`, `relative_fd`, `relative_id` – those would be separate arguments.  We would also add a way to set the `BPF_F_ID` flag to distinguish between the latter two.

I _think_ the use of bitflags now is inline with where we would want to go.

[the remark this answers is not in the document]
alessandrod: > This allows the caller to easily set a combination of flags if they wish. The bitflags API frees the developer from having to do bit manipulation themselves.

My point is that I suspect we won't ever expose the bitflag to the aya API.

> If we later choose to add support for more flags, such as BPF_F_REPLACE, BPF_F_BEFORE, and BPF_F_AFTER, then we would add the ability to supply the bpf_attr anonymous struct's three additional fields of replace_bpf_fd, relative_fd, relative_id – those would be separate arguments.

Exactly, similar to https://docs.aya-rs.dev/aya/programs/xdp/struct.xdp#method.attach_to_link where we use a REPLACE bit, but don't expose it to the API, see https://docs.aya-rs.dev/src/aya/programs/xdp.rs#233

Or in other words: you have to pass (flag, fd_or_id) and it's unlikely that we'll have a single method to do that, but more realistically we'll have attach_replace(old) attach_relative(AttachFlags::Before, fd) etc, where the bitflags become an implementation detail but are not exposed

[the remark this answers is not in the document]
reyzell: These are good points.  From what I understand, you can also attach a program with both the `ALLOW_MULTI` and `ALLOW_OVERRIDE` flags set.  A bitflags argument is useful there to allow the caller to specify empty, either flag, or set both flags.  However, to keep the Aya API simple, we could go with your idea of an enum, and perhaps with a third value that represents both (`ALLOW_MULTI_AND_OVERRIDE`...?).  If we want to avoid an `EMPTY` value on the enum, we could include both an `attach` and `attach_with_flags` function.

An alternative is to implement the flags argument as a struct with two booleans:
```Rust
struct CgroupAttachFlags {
    allow_multi: bool,
    allow_override: bool,
}
```

A caller could then set either flag, both flags, or none.  That's probably my favorite option, as it seems clean.

Do any of these options resonate with you?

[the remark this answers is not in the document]
alessandrod: from https://github.com/torvalds/linux/blob/35dfaad7188cdc043fde31709c796f5a692ba2bd/include/uapi/linux/bpf.h#L1094

```
 * BPF_F_ALLOW_OVERRIDE: If a sub-cgroup installs some bpf program,
 * the program in this cgroup yields to sub-cgroup program.
 *
 * BPF_F_ALLOW_MULTI: If a sub-cgroup installs some bpf program,
 * that cgroup program gets run in addition to the program in this cgroup.
 ```

so it seems like they're exclusive.

Given this and the docs of the other flags, I'm not convinced that at the aya level there's any need for bitflags, and bitflags are less idiomatic than regular enums so we should avoid them when possible.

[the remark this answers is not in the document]
Comment (reyzell) [reply aya/src/programs/links.rs]: You're right, good instincts.  I've updated this change to use an enum, and included a function scoped to the crate to translate the enum to its BPF value.

**Comment by reyzell** [comment]
Hmm, I see the "aya-ci / lint" job is failing on the `cargo xtask public-api` build step with the message "Error: aya public API changed; re-run with --bless."  Locally, when I run `cargo +nightly xtask public-api --bless`, I see no additional changes that need to be pushed.

I feel there's something I'm missing.

**Comment by tyrone-wu** [comment]
I think I encountered this issue before. Try updating nightly and see if blessing public-api changes.

**Comment by reyzell** [comment]
> I think I encountered this issue before. Try updating nightly and blessing public-api again.

Good call.  The public API had changed after the last update of this PR, but before the workflow approval jobs had run.  Thanks ;)

File: aya/src/programs/links.rs [hunk 12]
```diff
@@ -28,6 +28,30 @@ pub trait Link: std::fmt::Debug + 'static {
     fn detach(self) -> Result<(), ProgramError>;
 }
 
+/// Program attachment options.
+#[derive(Clone, Copy, Debug, Default)]
+pub enum CgroupAttachFlag {
```
alessandrod, line 33: nit: call this CgroupAttachMode?

aya/src/programs/lirc_mode2.rs
```diff
@@ -59,6 +62,17 @@ impl LircMode2 {
     ///
     /// The returned value can be used to detach, see [LircMode2::detach].
     pub fn attach<T: AsFd>(&mut self, lircdev: T) -> Result<LircLinkId, ProgramError> {
+        self.attach_with_flag(lircdev, CgroupAttachFlag::None)
+    }
+
+    /// Attaches the program to the given lirc device.
+    ///
+    /// The returned value can be used to detach, see [LircMode2::detach].
+    pub fn attach_with_flag<T: AsFd>(
```
alessandrod, line 71: Here and everywhere else: attach_with_cgroup_mode?

I'm not 100% convinced on mode, do you have a better name?

alessandrod (commented):
This looks good now, just not sure on the naming

alessandrod:
CI is failing because it needs rebase to pick up https://github.com/aya-rs/aya/pull/1016

aya/src/programs/lirc_mode2.rs
reyzell: Ha, now onto the hard part, naming :)

So far I think you've picked a good name.  Others I thought of are `CgroupAttach- Strategy or Policy or Cardinality or Rule`, they each have pros and cons.  ("Strategy" is a bit generic, "Policy" is good but feels more security focused, "Cardinality" is accurate but a bit long, "Rule" I kind of like.)

If we go with any of these, I'm thinking of changing the name of the default enum value, as a mode of `None` sounds odd.  Do these value names work?

```
pub enum CgroupAttachMode {
	Single,
	AllowMultiple,
	AllowOverride,
}
```

aya/src/programs/cgroup_skb.rs
```diff
@@ -87,6 +88,7 @@ impl CgroupSkb {
         &mut self,
         cgroup: T,
         attach_type: CgroupSkbAttachType,
+        flag: CgroupAttachMode,
```
alessandrod, line 91: here and everywhere else: please rename the argument name from flag to mode

aya/src/programs/links.rs
```diff
@@ -28,6 +28,30 @@ pub trait Link: std::fmt::Debug + 'static {
...
+    Single,
+
+    /// Allows the program to be overridden by one in a sub-cgroup.
+    AllowOverride,
+
+    /// Allows multiple programs to be run in the cgroup subtree.
+    AllowMultiple,
+}
+
+impl CgroupAttachMode {
```
alessandrod, line 45: I think this should be `impl From<CgroupAttachMode> for u32`

alessandrod (changes requested):
Almost there!

aya/src/programs/links.rs
reyzell: Done

aya/src/programs/cgroup_skb.rs
reyzell: Right!  Done.

alessandrod:
I just checked in the kernel and these are the cgroup-aware programs:

```
	case BPF_PROG_TYPE_CGROUP_DEVICE:
	case BPF_PROG_TYPE_CGROUP_SKB:
	case BPF_PROG_TYPE_CGROUP_SOCK:
	case BPF_PROG_TYPE_CGROUP_SOCK_ADDR:
	case BPF_PROG_TYPE_CGROUP_SOCKOPT:
	case BPF_PROG_TYPE_CGROUP_SYSCTL:
	case BPF_PROG_TYPE_SOCK_OPS:
	case BPF_PROG_TYPE_LSM:
		if (ptype == BPF_PROG_TYPE_LSM &&
		    prog->expected_attach_type != BPF_LSM_CGROUP)
			ret = -EINVAL;
		else
			ret = cgroup_bpf_prog_attach(attr, ptype, prog);
		break;
```

For LIRC_MODE2 SK_MSG and SK_SKB you can't pass any of those flags, so we should remove attach_with_cgroup_mode from them.

reyzell:
Ahh, thanks for catching that!  I should have realized that.  Fixed.

reyzell:
Applied two changes:
* Added `CgroupAttachMode` to items re-exported from `programs/mod.rs`
* Rebased from main

The following build steps pass:
```
cargo build --release
cargo +nightly xtask public-api --bless
cargo test --doc --all-features
```

alessandrod:
Can you update your nightly and re-run `cargo +nightly xtask public-api --bless`?

reyzell:
> Can you update your nightly and re-run `cargo +nightly xtask public-api --bless`?

Hmm, I made sure to do so before my last push (locally I'm up to date with the latest commit of 15eb935).  I've done so again now and don't see anything that has changed.  Something isn't clear to me about what's failing.  I'll try to dig in...

tyrone-wu:
Hmm, interesting. What's your output for `rustc +nightly -vV`?

reyzell:
I have the following:
```
> rustc +nightly -vV
rustc 1.82.0-nightly (8b3870784 2024-08-07)
binary: rustc
commit-hash: 8b3870784f69d8ed2ae4246d1b56f150e3131499
commit-date: 2024-08-07
host: x86_64-unknown-linux-gnu
release: 1.82.0-nightly
LLVM version: 19.1.0
```

tyrone-wu:
It looks you'll need to update nightly again 😅. The latest is currently `v1.83`. Perhaps that'll resolve the public-api issue?

reyzell:
You're right Tyron.  Updated!

PR #986, opened 2024-07-11
Add the ability to detach an eBPF program

reyzell:
Our user-space software agent has, until now, relied on catching the `SIGINT` signal (i.e. Ctl+C) for the application to know when to exit.  Exiting causes handles in Rust to go out of scope, which invokes `Drop` handlers and unloads eBPF programs and maps from the kernel.  However, if any unhandled signal were to cause the application to exit un-gracefully, `Drop` handlers would not be invoked, and eBPF resources would remain loaded in the kernel.  The number of loaded programs would build up over time as the application is restarted.

Gracefully handling more types of signals is a good approach, but is best effort.  If the drop logic were to fail, or an unhandle-able signal were received, resources would be left behind.  A robust approach is to perform cleanup on application startup, where previously loaded eBPF programs are identified and unloaded from the kernel.

To support that effort, this change adds the a public `detach_program()` function, which wraps the internal `bpf_prog_detach()`.

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -122,9 +122,9 @@ use crate::{
     },
     sys::{
         bpf_btf_get_fd_by_id, bpf_get_object, bpf_link_get_fd_by_id, bpf_link_get_info_by_fd,
-        bpf_load_program, bpf_pin_object, bpf_prog_get_fd_by_id, bpf_prog_get_info_by_fd,
-        bpf_prog_query, iter_link_ids, iter_prog_ids, retry_with_verifier_logs,
-        EbpfLoadProgramAttrs, SyscallError,
+        bpf_load_program, bpf_pin_object, bpf_prog_detach, bpf_prog_get_fd_by_id,
+        bpf_prog_get_info_by_fd, bpf_prog_query, iter_link_ids, iter_prog_ids,
+        retry_with_verifier_logs, EbpfLoadProgramAttrs, SyscallError,
     },
     util::{bytes_of_bpf_name, KernelVersion},
     VerifierLogLevel,
@@ -1149,6 +1149,15 @@ pub fn loaded_programs() -> impl Iterator<Item = Result<ProgramInfo, ProgramErro
             bpf_prog_get_info_by_fd(fd.as_fd(), &mut [])
         })
         .map(|result| result.map(ProgramInfo).map_err(Into::into))
+}
+
+/// Detaches the given program, unloading it from the kernel.
+pub fn detach_program(
+    prog_fd: BorrowedFd<'_>,
+    target_fd: BorrowedFd<'_>,
+    attach_type: bpf_attach_type,
+) -> Result<(), ProgramError> {
+    bpf_prog_detach(prog_fd, target_fd, attach_type).map_err(Into::into)
 }
 
 // TODO(https://github.com/aya-rs/aya/issues/645): this API is currently used in tests. Stabilize


dave-tucker:
> Our user-space software agent has, until now, relied on catching the `SIGINT` signal (i.e. Ctl+C) for the application to know when to exit. Exiting causes handles in Rust to go out of scope, which invokes `Drop` handlers and unloads eBPF programs and maps from the kernel. However, if any unhandled signal were to cause the application to exit un-gracefully, `Drop` handlers would not be invoked, and eBPF resources would remain loaded in the kernel. The number of loaded programs would build up over time as the application is restarted.

That holds true only in very specific circumstances. Any BPF syscall that returns an FD has them set CLOEXEC, so FDs will get cleaned up when your process exits - gracefully or not.

Specific cases that this doesn't work are thinks like TC (which uses netlink) etc... and my guess is that you're using one of theses types of program.

> Gracefully handling more types of signals is a good approach, but is best effort. If the drop logic were to fail, or an unhandle-able signal were received, resources would be left behind. A robust approach is to perform cleanup on application startup, where previously loaded eBPF programs are identified and unloaded from the kernel.
> 
> To support that effort, this change adds the a public `detach_program()` function, which wraps the internal `bpf_prog_detach()`.

`bpf_prog_detach()` is not a generic detach function. It's a specific detach function for links that were created with `bpf_prog_attach()` so this approach isn't going to work.

If you let me know what type of program you're having issue with I can point you in the right direction.

reyzell:
I think you're saying that once SockOps programs are updated to attach with `FdLink` (issue #987), that I'll no longer see lingering maps and programs when the user-space process terminates ungracefully.  Is that right?

dave-tucker:
> I think you're saying that once SockOps programs are updated to attach with `FdLink` (issue #987), that I'll no longer see lingering maps and programs when the user-space process terminates ungracefully. Is that right?

Yep I'm 99% sure that's the case.

viveksb007:
I am seeing similar problem for TC program. I have attached the eBPF program on TC-Egress hook and sometimes they linger around. @dave-tucker can you share the references on how to cleanup existing TC eBPF program during startup?

alessandrod:
> I think you're saying that once SockOps programs are updated to attach with `FdLink` (issue #987), that I'll no longer see lingering maps and programs when the user-space process terminates ungracefully. Is that right?

Correct, `bpf_link` fixes this, at which point the only problematic program left is TC when attached with netlink, for which we already provide an utility to detach at startup: https://docs.aya-rs.dev/aya/programs/tc/fn.qdisc_detach_program cc @viveksb007

nahuelfilipuzzi:
I think we still need this feature. Even when `bpf_link`, that is only available from Linux >=5.7, right?

Example: https://github.com/aya-rs/aya/blob/main/aya/src/programs/sock_ops.rs
```rust
 pub fn attach<T: AsFd>(
        &mut self,
        cgroup: T,
        mode: CgroupAttachMode,
    ) -> Result<SockOpsLinkId, ProgramError> {
// ...
        if KernelVersion::current().unwrap() >= KernelVersion::new(5, 7, 0) {
            let link_fd = bpf_link_create(...)
// ...
        } else {
            let link = ProgAttachLink::attach(prog_fd, cgroup_fd, attach_type, mode)?;
// ...
        }
    }
```

nahuelfilipuzzi:
This feature is this still needed if you want to use Kernel <5.7, right?

https://github.com/aya-rs/aya/blob/main/aya/src/programs/sock_ops.rs
```rust
 pub fn attach<T: AsFd>(
        &mut self,
        cgroup: T,
        mode: CgroupAttachMode,
    ) -> Result<SockOpsLinkId, ProgramError> {
/...
        if KernelVersion::current().unwrap() >= KernelVersion::new(5, 7, 0) {
            let link_fd = bpf_link_create(...)
//...
        } else {
            let link = ProgAttachLink::attach(prog_fd, cgroup_fd, attach_type, mode)?;
// ...
        }
    }
```

mehnazyunus:
@dave-tucker Is there interest in getting this merged? I agree with @nahuelfilipuzzi that it is required for kernel < 5.7.

PR #990, opened 2024-07-15
Support BPF Token

dave-tucker:
WIP PR tracking my work on BPF Token support - added in Kernel 6.9.
https://lwn.net/Articles/959350/

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/990)


dave-tucker:
This PR is functionally done. However we need a test case for it that can run in CI.
Reimplementing the kernel selftest is the best solution:  https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/prog_tests/token.c

The sequence of operations has to be EXACTLY as described in the selftests.

vadorovsky:
Superseded by https://github.com/aya-rs/aya/pull/1515

PR #991, opened 2024-07-16
docs(aya): fix typo

l2dy:
reverse-commit doesn't make sense. Should be `reserve` to match `bpf_ringbuf_reserve()`.

Reference: https://github.com/torvalds/linux/blob/master/Documentation/bpf/ringbuf.rst


ajwerner (approved):
LGTM

PR #997, opened 2024-07-24
Fix CI on macOS

tamird:
A naive attempt to resolve recent CI issues: `cargo: command not found`.


tamird:
The ubuntu build is busted because of https://github.com/llvm/llvm-project/issues/100466#issuecomment-2248889113, but this fixes the macOS problems.

PR #1000, opened 2024-07-25
aya: convert bpf programs to assembly

Billy99:
In aya/src/sys/bpf.rs, there are several simple bpf programs written as byte arrays. Rework these as assembly code and then generate the bytecode.


alessandrod:
Hello!

I don't think we should do this, for two reasons: 

- solana_rbpf implements SBPF - an eBPF dialect which is not compatible with kernel eBPF
- as it is, the patch assembles at runtime, which incurs in a performance cost. If we could fork the solana_rbpf assembler and make it work _at compile time_ that'd be great, and something I've been (idly) thinking about.

dave-tucker:
> Hello!
> 
> I don't think we should do this, for two reasons:
> 
> * solana_rbpf implements SBPF - an eBPF dialect which is not compatible with kernel eBPF
> * as it is, the patch assembles at runtime, which incurs in a performance cost. If we could fork the solana_rbpf assembler and make it work _at compile time_ that'd be great, and something I've been (idly) thinking about.

The issue here is twofold:

1. Keeping little and big endian versions of the same programs around isn't nice
2. Hand-encoding the asm instructions into byte arrays is error prone

If the solana_rbpf assembler is a no-go, that's totally fine.
I can take a pass of doing something in macros a little like what libbpf does: https://github.com/libbpf/libbpf/blob/master/src/features.c#L51-L56

alessandrod:
>  can take a pass of doing something in macros a little like what libbpf does:

something like this works, iirc there's an existing crate for it even

dave-tucker:
> > can take a pass of doing something in macros a little like what libbpf does:
> 
> something like this works, iirc there's an existing crate for it even

oh cool! if you find it let me know.

alessandrod:
> > > can take a pass of doing something in macros a little like what libbpf does:
> > 
> > 
> > something like this works, iirc there's an existing crate for it even
> 
> oh cool! if you find it let me know.

it's not the one I saw a while ago but this seems ok https://github.com/arcjustin/bpf-ins

tamird:
@dave-tucker do you know if @Billy99 plans to return to this work?

Billy99:
> @dave-tucker do you know if @Billy99 plans to return to this work?

No, I do not. Dave asked me to start this, but was going to take it in a different direction, I assumed in a different PR:
> I can take a pass of doing something in macros a little like what libbpf does: https://github.com/libbpf/libbpf/blob/master/src/features.c#L51-L56

I left this around as sample code for a head start if needed. I'm fine with closing if it's not needed for that.

tamird:
Closing doesn't throw the code away, so I'm going to do that.

Thanks!

PR #1004, opened 2024-07-29
fix(aya): Fix PerfEventArray resize logic

dave-tucker:
There was a logic bug in the previously merged patch where we set the correctly calculated max_entries size with the original.

To fix this and prevent regressions a unit test was added. This highlighted that the original map definition needs to be mutated in order for the max_entries change to be properly applied.

As such, this resize logic moved out of aya::sys into aya::maps


vadorovsky (commented):
API needs blessing

vadorovsky (approved):

PR #921, opened 2024-04-05
Tcx

astoycos:
Fixes #918 

This initial attempt works with the following example -> https://github.com/astoycos/tcxtest

Please see integration tests for example of how the new API works

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/921)

Issue #918, opened 2024-04-01
Add TCX link support 


astoycos:
Following the recent addition of [link support for TC programs](https://lwn.net/Articles/938632/) we should add the necessary bits to support this in aya as well. 

The [commit adding support to the cilium ebpf library](https://github.com/cilium/ebpf/commit/417f8a264d2e691649ede9b995fb334e6c58ca71) can sort of pave the way for us here

PR #1006, opened 2024-08-01
codegen: add `tcx_action_base` enum to bindings

astoycos:
Add tcx_action_base to pick up new TCX return codes, follow this pr up with a manual codegen actions run 

Relates to #918 and needed for #921

cc @dave-tucker


alessandrod:
kicking CI, I think this broke during the LLVM18 -> LLVM19 transition

alessandrod:
@astoycos can you rebase pls

dave-tucker:
Per: https://github.com/aya-rs/aya/pull/1009#issuecomment-2288780270
We can merge with macOS CI red since it won't work until we get LLVM19 in Homebrew 😢

dave-tucker (approved):

PR #1007, opened 2024-08-04
aya,aya-obj,integration-test: add better support in `ProgramInfo` & `MapInfo` for old kernels

tyrone-wu:
This adds support for the field availability in `ProgramInfo` and `MapInfo`.

Srry this is long, let me know and I can try to TL;DR it.

For `ProgramInfo`:
- Have `program_type()` to return `bpf_prog_type` instead of the integer representation.
- For fields that can't be `0`, return `Option<NonZero*>` type to indicate `None` as field being unavailable.
- Have `name_as_str()` use the `bpf_name()` feature probe as an additional check for whether field is available. Besides invalid unicode, `None` is returned if probe does not detect feature.
- For `mad_ids()`:
  - Now returns a result of `Option<Vec<NonZeroU32>>`, with `None` indicating field is unavailable.
  - On kernels below v4.15, the init closure that fills in the map info causes an `E2BIG` error since `check_uarg_tail_zero()` expects the entire struct to be zero bytes. I initially worked around this by having `bpf_prog_get_info_by_fd()` do an additional syscall if the first failed with `E2BIG`.
  - Now, `map_ids()` and `bpf_prog_get_info_by_fd()` both uses a new feature probe `prog_info_map_ids()` to detect if field available.
- For `gpl_compatible()`, it now returns `Option<bool>`, with `None` indicating field is unavailable. It uses a new feature probe `prog_info_gpl_compatible()` to detect for field availability.

There was a redundant export of `loaded_programs()` which gave it 2 method of access: `aya::loaded_programs` and `aya::programs::loaded_programs`. To avoid confusion, it should now only be `aya::programs::loaded_programs`.

For `MapInfo`:
- Have `map_type()` to return `bpf_map_type` instead of the integer representation.
- For fields that can't be `0`, return `Option<NonZero*>` type to indicate `None` as field being unavailable.
- For `name_as_str()`, it uses the `bpf_name()` feature probe to detect for field availability. Although the probe performs the check on program name, map name was introduced in the same commit as program name so it should suffice.

Integration tests for `ProgramInfo` and `MapInfo` are updated to use `kernel_assert`/`kernel_assert_eq` macro.
The macro first performs the assertion check. If assertion passes, continue with test. If failed, then check host's kernel version.
If host is below the feature version, then continue with test and log it in stderr (can be observed with `-- --nocapture`). Otherwise, panic 😰.

I verified that this passes on versions (i use ubuntu mainline images for my integration tests):
- 4.13.0
- 4.15.0
- 5.15.0
- 6.1.0

### DIFF
--- a/aya-log/src/lib.rs
+++ b/aya-log/src/lib.rs
@@ -150,7 +150,7 @@ impl EbpfLogger {
                 None => false,
             })
             .ok_or(Error::MapNotFound)?;
-        let map = MapData::from_id(map.id()).map_err(Error::MapError)?;
+        let map = MapData::from_id(map.id().unwrap().get()).map_err(Error::MapError)?;
 
         Self::read_logs_async(Map::PerfEventArray(map), logger)?;
 
--- a/aya-log/src/lib.rs
+++ b/aya-log/src/lib.rs
@@ -60,12 +60,11 @@ use std::{
 const MAP_NAME: &str = "AYA_LOGS";
 
 use aya::{
-    loaded_programs,
     maps::{
         perf::{AsyncPerfEventArray, Events, PerfBufferError},
         Map, MapData, MapError, MapInfo,
     },
-    programs::ProgramError,
+    programs::{loaded_programs, ProgramError},
     util::online_cpus,
     Ebpf, Pod,
 };
@@ -137,13 +136,15 @@ impl EbpfLogger {
     ) -> Result<EbpfLogger, Error> {
         let program_info = loaded_programs()
             .filter_map(|info| info.ok())
-            .find(|info| info.id() == program_id)
+            .find(|info| info.id().is_some_and(|id| id.get() == program_id))
             .ok_or(Error::ProgramNotFound)?;
+
         let map = program_info
             .map_ids()
             .map_err(Error::ProgramError)?
+            .expect("`map_ids` field in `bpf_prog_info` not available")
             .iter()
-            .filter_map(|id| MapInfo::from_id(*id).ok())
+            .filter_map(|id| MapInfo::from_id(id.get()).ok())
             .find(|map_info| match map_info.name_as_str() {
                 Some(name) => name == MAP_NAME,
                 None => false,
--- a/aya-obj/src/links.rs
+++ b/aya-obj/src/links.rs
@@ -1,0 +1,100 @@
+//! Link type bindings.
+
+use crate::{
+    generated::{bpf_attach_type, bpf_link_type},
+    InvalidTypeBinding,
+};
+
+impl TryFrom<u32> for bpf_link_type {
+    type Error = InvalidTypeBinding<u32>;
+
+    fn try_from(link_type: u32) -> Result<Self, Self::Error> {
+        use bpf_link_type::*;
+        Ok(match link_type {
+            x if x == BPF_LINK_TYPE_UNSPEC as u32 => BPF_LINK_TYPE_UNSPEC,
+            x if x == BPF_LINK_TYPE_RAW_TRACEPOINT as u32 => BPF_LINK_TYPE_RAW_TRACEPOINT,
+            x if x == BPF_LINK_TYPE_TRACING as u32 => BPF_LINK_TYPE_TRACING,
+            x if x == BPF_LINK_TYPE_CGROUP as u32 => BPF_LINK_TYPE_CGROUP,
+            x if x == BPF_LINK_TYPE_ITER as u32 => BPF_LINK_TYPE_ITER,
+            x if x == BPF_LINK_TYPE_NETNS as u32 => BPF_LINK_TYPE_NETNS,
+            x if x == BPF_LINK_TYPE_XDP as u32 => BPF_LINK_TYPE_XDP,
+            x if x == BPF_LINK_TYPE_PERF_EVENT as u32 => BPF_LINK_TYPE_PERF_EVENT,
+            x if x == BPF_LINK_TYPE_KPROBE_MULTI as u32 => BPF_LINK_TYPE_KPROBE_MULTI,
+            x if x == BPF_LINK_TYPE_STRUCT_OPS as u32 => BPF_LINK_TYPE_STRUCT_OPS,
+            x if x == BPF_LINK_TYPE_NETFILTER as u32 => BPF_LINK_TYPE_NETFILTER,
+            x if x == BPF_LINK_TYPE_TCX as u32 => BPF_LINK_TYPE_TCX,
+            x if x == BPF_LINK_TYPE_UPROBE_MULTI as u32 => BPF_LINK_TYPE_UPROBE_MULTI,
+            x if x == BPF_LINK_TYPE_NETKIT as u32 => BPF_LINK_TYPE_NETKIT,
+            _ => return Err(InvalidTypeBinding { value: link_type }),
+        })
+    }
+}
+
+impl TryFrom<u32> for bpf_attach_type {
+    type Error = InvalidTypeBinding<u32>;
+
+    fn try_from(attach_type: u32) -> Result<Self, Self::Error> {
+        use bpf_attach_type::*;
+        Ok(match attach_type {
+            x if x == BPF_CGROUP_INET_INGRESS as u32 => BPF_CGROUP_INET_INGRESS,
+            x if x == BPF_CGROUP_INET_EGRESS as u32 => BPF_CGROUP_INET_EGRESS,
+            x if x == BPF_CGROUP_INET_SOCK_CREATE as u32 => BPF_CGROUP_INET_SOCK_CREATE,
+            x if x == BPF_CGROUP_SOCK_OPS as u32 => BPF_CGROUP_SOCK_OPS,
+            x if x == BPF_SK_SKB_STREAM_PARSER as u32 => BPF_SK_SKB_STREAM_PARSER,
+            x if x == BPF_SK_SKB_STREAM_VERDICT as u32 => BPF_SK_SKB_STREAM_VERDICT,
+            x if x == BPF_CGROUP_DEVICE as u32 => BPF_CGROUP_DEVICE,
+            x if x == BPF_SK_MSG_VERDICT as u32 => BPF_SK_MSG_VERDICT,
+            x if x == BPF_CGROUP_INET4_BIND as u32 => BPF_CGROUP_INET4_BIND,
+            x if x == BPF_CGROUP_INET6_BIND as u32 => BPF_CGROUP_INET6_BIND,
+            x if x == BPF_CGROUP_INET4_CONNECT as u32 => BPF_CGROUP_INET4_CONNECT,
+            x if x == BPF_CGROUP_INET6_CONNECT as u32 => BPF_CGROUP_INET6_CONNECT,
+            x if x == BPF_CGROUP_INET4_POST_BIND as u32 => BPF_CGROUP_INET4_POST_BIND,
+            x if x == BPF_CGROUP_INET6_POST_BIND as u32 => BPF_CGROUP_INET6_POST_BIND,
+            x if x == BPF_CGROUP_UDP4_SENDMSG as u32 => BPF_CGROUP_UDP4_SENDMSG,
+            x if x == BPF_CGROUP_UDP6_SENDMSG as u32 => BPF_CGROUP_UDP6_SENDMSG,
+            x if x == BPF_LIRC_MODE2 as u32 => BPF_LIRC_MODE2,
+            x if x == BPF_FLOW_DISSECTOR as u32 => BPF_FLOW_DISSECTOR,
+            x if x == BPF_CGROUP_SYSCTL as u32 => BPF_CGROUP_SYSCTL,
+            x if x == BPF_CGROUP_UDP4_RECVMSG as u32 => BPF_CGROUP_UDP4_RECVMSG,
+            x if x == BPF_CGROUP_UDP6_RECVMSG as u32 => BPF_CGROUP_UDP6_RECVMSG,
+            x if x == BPF_CGROUP_GETSOCKOPT as u32 => BPF_CGROUP_GETSOCKOPT,
+            x if x == BPF_CGROUP_SETSOCKOPT as u32 => BPF_CGROUP_SETSOCKOPT,
+            x if x == BPF_TRACE_RAW_TP as u32 => BPF_TRACE_RAW_TP,
+            x if x == BPF_TRACE_FENTRY as u32 => BPF_TRACE_FENTRY,
+            x if x == BPF_TRACE_FEXIT as u32 => BPF_TRACE_FEXIT,
+            x if x == BPF_MODIFY_RETURN as u32 => BPF_MODIFY_RETURN,
+            x if x == BPF_LSM_MAC as u32 => BPF_LSM_MAC,
+            x if x == BPF_TRACE_ITER as u32 => BPF_TRACE_ITER,
+            x if x == BPF_CGROUP_INET4_GETPEERNAME as u32 => BPF_CGROUP_INET4_GETPEERNAME,
+            x if x == BPF_CGROUP_INET6_GETPEERNAME as u32 => BPF_CGROUP_INET6_GETPEERNAME,
+            x if x == BPF_CGROUP_INET4_GETSOCKNAME as u32 => BPF_CGROUP_INET4_GETSOCKNAME,
+            x if x == BPF_CGROUP_INET6_GETSOCKNAME as u32 => BPF_CGROUP_INET6_GETSOCKNAME,
+            x if x == BPF_XDP_DEVMAP as u32 => BPF_XDP_DEVMAP,
+            x if x == BPF_CGROUP_INET_SOCK_RELEASE as u32 => BPF_CGROUP_INET_SOCK_RELEASE,
+            x if x == BPF_XDP_CPUMAP as u32 => BPF_XDP_CPUMAP,
+            x if x == BPF_SK_LOOKUP as u32 => BPF_SK_LOOKUP,
+            x if x == BPF_XDP as u32 => BPF_XDP,
+            x if x == BPF_SK_SKB_VERDICT as u32 => BPF_SK_SKB_VERDICT,
+            x if x == BPF_SK_REUSEPORT_SELECT as u32 => BPF_SK_REUSEPORT_SELECT,
+            x if x == BPF_SK_REUSEPORT_SELECT_OR_MIGRATE as u32 => {
+                BPF_SK_REUSEPORT_SELECT_OR_MIGRATE
+            }
+            x if x == BPF_PERF_EVENT as u32 => BPF_PERF_EVENT,
+            x if x == BPF_TRACE_KPROBE_MULTI as u32 => BPF_TRACE_KPROBE_MULTI,
+            x if x == BPF_LSM_CGROUP as u32 => BPF_LSM_CGROUP,
+            x if x == BPF_STRUCT_OPS as u32 => BPF_STRUCT_OPS,
+            x if x == BPF_NETFILTER as u32 => BPF_NETFILTER,
+            x if x == BPF_TCX_INGRESS as u32 => BPF_TCX_INGRESS,
+            x if x == BPF_TCX_EGRESS as u32 => BPF_TCX_EGRESS,
+            x if x == BPF_TRACE_UPROBE_MULTI as u32 => BPF_TRACE_UPROBE_MULTI,
+            x if x == BPF_CGROUP_UNIX_CONNECT as u32 => BPF_CGROUP_UNIX_CONNECT,
+            x if x == BPF_CGROUP_UNIX_SENDMSG as u32 => BPF_CGROUP_UNIX_SENDMSG,
+            x if x == BPF_CGROUP_UNIX_RECVMSG as u32 => BPF_CGROUP_UNIX_RECVMSG,
+            x if x == BPF_CGROUP_UNIX_GETPEERNAME as u32 => BPF_CGROUP_UNIX_GETPEERNAME,
+            x if x == BPF_CGROUP_UNIX_GETSOCKNAME as u32 => BPF_CGROUP_UNIX_GETSOCKNAME,
+            x if x == BPF_NETKIT_PRIMARY as u32 => BPF_NETKIT_PRIMARY,
+            x if x == BPF_NETKIT_PEER as u32 => BPF_NETKIT_PEER,
+            _ => return Err(InvalidTypeBinding { value: attach_type }),
+        })
+    }
+}
--- a/aya-obj/src/maps.rs
+++ b/aya-obj/src/maps.rs
@@ -5,16 +5,10 @@ use core::mem;
 
 #[cfg(not(feature = "std"))]
 use crate::std;
-use crate::EbpfSectionKind;
-
-/// Invalid map type encontered
-pub struct InvalidMapTypeError {
-    /// The map type
-    pub map_type: u32,
-}
+use crate::{EbpfSectionKind, InvalidTypeBinding};
 
 impl TryFrom<u32> for crate::generated::bpf_map_type {
-    type Error = InvalidMapTypeError;
+    type Error = InvalidTypeBinding<u32>;
 
     fn try_from(map_type: u32) -> Result<Self, Self::Error> {
         use crate::generated::bpf_map_type::*;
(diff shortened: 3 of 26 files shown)


aya-obj/src/links.rs
```diff
@@ -0,0 +1,93 @@
...
+            x if x == BPF_LINK_TYPE_NETNS as u32 => BPF_LINK_TYPE_NETNS,
+            x if x == BPF_LINK_TYPE_XDP as u32 => BPF_LINK_TYPE_XDP,
+            x if x == BPF_LINK_TYPE_PERF_EVENT as u32 => BPF_LINK_TYPE_PERF_EVENT,
+            x if x == BPF_LINK_TYPE_KPROBE_MULTI as u32 => BPF_LINK_TYPE_KPROBE_MULTI,
+            x if x == BPF_LINK_TYPE_STRUCT_OPS as u32 => BPF_LINK_TYPE_STRUCT_OPS,
+            x if x == BPF_LINK_TYPE_NETFILTER as u32 => BPF_LINK_TYPE_NETFILTER,
+            x if x == BPF_LINK_TYPE_TCX as u32 => BPF_LINK_TYPE_TCX,
+            x if x == BPF_LINK_TYPE_UPROBE_MULTI as u32 => BPF_LINK_TYPE_UPROBE_MULTI,
+            x if x == BPF_LINK_TYPE_NETKIT as u32 => BPF_LINK_TYPE_NETKIT,
+            _ => __MAX_BPF_LINK_TYPE,
```
alessandrod, line 22: Technically this is not correct? Looks like this should be try_from instead?

aya/src/maps/info.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub fn from_id(id: u32) -> Result<Self, MapError> {
+        bpf_map_get_fd_by_id(id)
+            .map_err(MapError::from)
+            .and_then(|fd| Self::new_from_fd(fd.as_fd()))
+    }
+
+    /// The map type as defined by the linux kernel enum [`bpf_map_type`].
+    ///
+    /// Introduced in kernel v4.13.
+    pub fn map_type(&self) -> bpf_map_type {
```
alessandrod, line 44: we can't re-export generated types as part of the public API, or we can never
guarantee stability. I think we need our own EbpfMapType/EbpfProgramType this
issue has come up before.

aya/src/programs/info.rs
```diff
@@ -0,0 +1,281 @@
...
+impl ProgramInfo {
+    pub(crate) fn new_from_fd(fd: BorrowedFd<'_>) -> Result<Self, ProgramError> {
+        let info = bpf_prog_get_info_by_fd(fd, &mut [])?;
+        Ok(Self(info))
+    }
+
+    /// The program type as defined by the linux kernel enum [`bpf_prog_type`].
+    ///
+    /// Introduced in kernel v4.13.
+    pub fn program_type(&self) -> bpf_prog_type {
```
alessandrod, line 39: same here, can't re-export

alessandrod (commented):
Thanks for doing this work! Overall looks great, left some comments

aya/src/maps/info.rs
tyrone-wu: Could I get your thoughts on which would make the most sense for the enum name: 😅
- `EbpfProgramType`
- `EbpfProgram`
- `ProgramType`

I'm not sure if your feedback from https://github.com/aya-rs/aya/pull/959#discussion_r1716884645 applies here, since the `aya::programs::Program` enum is already being used to represent the variants of the actual program implementations.

Personally, I'm kinda leaning more towards `EbpfProgramType` or `ProgramType`, since I feel the `Type` suffix would give better distinction from the `aya::programs::Program` enum.

aya-obj/src/links.rs
tyrone-wu: You're right, it should be try_from. 👍

aya-obj/src/links.rs
```diff
@@ -0,0 +1,93 @@
...
+            x if x == BPF_TCX_EGRESS as u32 => BPF_TCX_EGRESS,
+            x if x == BPF_TRACE_UPROBE_MULTI as u32 => BPF_TRACE_UPROBE_MULTI,
+            x if x == BPF_CGROUP_UNIX_CONNECT as u32 => BPF_CGROUP_UNIX_CONNECT,
+            x if x == BPF_CGROUP_UNIX_SENDMSG as u32 => BPF_CGROUP_UNIX_SENDMSG,
+            x if x == BPF_CGROUP_UNIX_RECVMSG as u32 => BPF_CGROUP_UNIX_RECVMSG,
+            x if x == BPF_CGROUP_UNIX_GETPEERNAME as u32 => BPF_CGROUP_UNIX_GETPEERNAME,
+            x if x == BPF_CGROUP_UNIX_GETSOCKNAME as u32 => BPF_CGROUP_UNIX_GETSOCKNAME,
+            x if x == BPF_NETKIT_PRIMARY as u32 => BPF_NETKIT_PRIMARY,
+            x if x == BPF_NETKIT_PEER as u32 => BPF_NETKIT_PEER,
+            _ => __MAX_BPF_ATTACH_TYPE,
```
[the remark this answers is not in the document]
tyrone-wu: They should be update to `TryFrom` now. 🫡

aya/src/maps/info.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// Introduced in kernel v6.9.
+    #[doc(alias = "BPF_MAP_TYPE_ARENA")]
+    Arena = bpf_map_type::BPF_MAP_TYPE_ARENA as isize,
+    /// An invalid map type.
+    #[doc(hidden)]
+    Invalid = bpf_map_type::__MAX_BPF_MAP_TYPE as isize,
+}
+
+impl From<bpf_map_type> for MapType {
+    fn from(map_type: bpf_map_type) -> Self {
```
alessandrod, line 370: shouldn't this by try_from too?

aya/src/maps/info.rs
```diff
@@ -0,0 +1,410 @@
...
+    #[doc(alias = "BPF_MAP_TYPE_CGRP_STORAGE")]
+    CgrpStorage = bpf_map_type::BPF_MAP_TYPE_CGRP_STORAGE as isize,
+    /// An Arena map type.
+    ///
+    /// Introduced in kernel v6.9.
+    #[doc(alias = "BPF_MAP_TYPE_ARENA")]
+    Arena = bpf_map_type::BPF_MAP_TYPE_ARENA as isize,
+    /// An invalid map type.
+    #[doc(hidden)]
+    Invalid = bpf_map_type::__MAX_BPF_MAP_TYPE as isize,
```
alessandrod, line 366: If we do TryFrom then I don't think we need this?

aya/src/programs/info.rs
```diff
@@ -0,0 +1,524 @@
...
+    ///
+    /// Introduced in kernel v6.4.
+    #[doc(alias = "BPF_PROG_TYPE_NETFILTER")]
+    Netfilter = bpf_prog_type::BPF_PROG_TYPE_NETFILTER as isize,
+    /// An invalid program type.
+    #[doc(hidden)]
+    Invalid = bpf_prog_type::__MAX_BPF_PROG_TYPE as isize,
+}
+
+impl From<bpf_prog_type> for ProgramType {
```
alessandrod, line 484: Same here probably TryFrom?

aya/src/programs/info.rs
```diff
@@ -0,0 +1,524 @@
...
+    #[doc(alias = "BPF_PROG_TYPE_SYSCALL")]
+    Syscall = bpf_prog_type::BPF_PROG_TYPE_SYSCALL as isize,
+    /// A Netfilter program type.
+    ///
+    /// Introduced in kernel v6.4.
+    #[doc(alias = "BPF_PROG_TYPE_NETFILTER")]
+    Netfilter = bpf_prog_type::BPF_PROG_TYPE_NETFILTER as isize,
+    /// An invalid program type.
+    #[doc(hidden)]
+    Invalid = bpf_prog_type::__MAX_BPF_PROG_TYPE as isize,
```
alessandrod, line 481: 🪓

aya/src/maps/info.rs
alessandrod: ProgramType and MapType look great :)

alessandrod (commented):
Super solid work as always, just a couple of comments then we're ready to go :)

aya/src/programs/info.rs
tyrone-wu: 🫡

aya/src/maps/info.rs
tyrone-wu: 🫡

aya/src/programs/info.rs
tyrone-wu: 🫡

aya/src/maps/info.rs
tyrone-wu: You're right, `try_from` would probably be more appropriate now that you mentioned `Invalid` won't be needed. 🫡

aya/src/maps/info.rs
alessandrod: I think you forgot to remove this?

aya-obj/src/obj.rs
```diff
@@ -987,6 +1006,14 @@ pub enum ParseError {
     NoBTF,
 }
 
+/// Invalid bindings to the bpf type from the parsed/received value.
+pub struct InvalidTypeBinding<T> {
+    /// The bpf type to bind to.
+    pub binding_type: String,
```
alessandrod, line 1012: This is not being read anywhere I think, so I would remove it? Doesn't matter
much because in practice i don't think it's ever instantiated, but avoid
allocations etc :P

tyrone-wu: Make sense. 🫡

aya/src/maps/info.rs
tyrone-wu: Yep, forgot to 🪓 that. 😓

alessandrod (approved):

aya-log/src/lib.rs
```diff
@@ -137,19 +136,21 @@ impl EbpfLogger {
...
         let program_info = loaded_programs()
             .filter_map(|info| info.ok())
-            .find(|info| info.id() == program_id)
+            .find(|info| info.id().is_some_and(|id| id.get() == program_id))
             .ok_or(Error::ProgramNotFound)?;
+
         let map = program_info
             .map_ids()
             .map_err(Error::ProgramError)?
+            .expect("`map_ids` field in `bpf_prog_info` not available")
```
tamird, line 145: why is it ok to panic here?

aya-log/src/lib.rs
```diff
@@ -137,19 +136,21 @@ impl EbpfLogger {
...
             .iter()
-            .filter_map(|id| MapInfo::from_id(*id).ok())
+            .filter_map(|id| MapInfo::from_id(id.get()).ok())
             .find(|map_info| match map_info.name_as_str() {
                 Some(name) => name == MAP_NAME,
                 None => false,
             })
             .ok_or(Error::MapNotFound)?;
-        let map = MapData::from_id(map.id()).map_err(Error::MapError)?;
+        let map = MapData::from_id(map.id().unwrap().get()).map_err(Error::MapError)?;
```
tamird, line 153: why is it ok to unwrap here?

test/integration-test/src/utils.rs
```diff
@@ -70,3 +70,62 @@ impl Drop for NetNsGuard {
...
+                panic!(
+                    r#"  assertion `left == right` failed: expected host kernel v{current} to have v{feat_version} feature
+    left: {:?}
+   right: {:?}"#,
+                    $left, $right,
+                );
+            } else {
+                // Continue with tests since host is not expected to have feat
+                eprintln!(
+                    r#"ignoring assertion at {}:{}
```
tamird, line 120: we shouldn't ignore the assertion - we should instead assert that the feature is indeed missing.

aya/src/programs/info.rs
```diff
@@ -0,0 +1,548 @@
...
+    /// Introduced in kernel v4.15.
+    pub fn map_ids(&self) -> Result<Option<Vec<NonZeroU32>>, ProgramError> {
+        if FEATURES.prog_info_map_ids() {
+            let mut map_ids = vec![0u32; self.0.nr_map_ids as usize];
+            bpf_prog_get_info_by_fd(self.fd()?.as_fd(), &mut map_ids)?;
+
+            Ok(Some(
+                map_ids
+                    .into_iter()
+                    .map(|id| NonZeroU32::new(id).unwrap())
```
tamird, line 133: this should be .expect with a string that mentions the version detection above.

aya-log/src/lib.rs
tyrone-wu: Sorry, I definitely shouldn't be panic in the library.

aya-log/src/lib.rs
tyrone-wu: Sorry, I should have mapped that into an error. I'll have that resolved soon.

aya-log/src/lib.rs
```diff
@@ -137,19 +136,21 @@ impl EbpfLogger {
     ) -> Result<EbpfLogger, Error> {
         let program_info = loaded_programs()
             .filter_map(|info| info.ok())
-            .find(|info| info.id() == program_id)
+            .find(|info| info.id().is_some_and(|id| id.get() == program_id))
```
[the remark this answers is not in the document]
tyrone-wu: Ahhhhh you're right. 😓

aya/src/programs/info.rs
tyrone-wu: Now that I think about it, I think I might of turned the API into an unwrapping nightmare with all the `Option<NonZero*>`. 😵

I'm currently thinking of having fields that were introduced in the same kernel version as the struct not be options (just the direct value). What do you think? 

For `map_ids()`, it looks like it's reasonable to assume the map id would never be zero based on what I traced:
- https://elixir.bootlin.com/linux/v4.15/source/kernel/bpf/syscall.c#L1590: how map IDs are getting filled in `bpf_prog_get_info_by_fd()`
- https://elixir.bootlin.com/linux/v4.15/source/kernel/bpf/verifier.c#L4735: how map info is filled from `bpf_verifier_env` into `bpf_prog_aux`
- https://elixir.bootlin.com/linux/v4.15/source/kernel/bpf/verifier.c#L4249: how map is set in `bpf_verifier_env` before it's verified

Some of the code is shifted in kernel v6.10, but the logic should still be the same. https://elixir.bootlin.com/linux/v6.10/source/kernel/bpf/verifier.c#L18565

cc'ing @alessandrod for thoughts as well

test/integration-test/src/utils.rs
tyrone-wu: So I did some experimenting with `bpf_obj_get_info_by_fd()` after examining the code some more. I tested that I can use `bpf_check_uarg_tail_zero()` to check if a field gives `E2BIG`, however, it would require exposing `bpf_obj_get_info_by_fd()` for the `test` crate to access.

aya/src/programs/info.rs
alessandrod: > I'm currently thinking of having fields that were introduced in the same kernel version as the struct not be options (just the direct value). What do you think?

Yep this sounds good.

> For map_ids(), it looks like it's reasonable to assume the map id would never be zero based on what I traced:

I don't think that any bpf ids can be 0 regardless of type (map, program, etc)?

In fact I'd probably change all these to return Option<u32>. I think NonZero* is nice to accept as input to enforce safety, but as a return value in these utils seems more annoying than anything else.

marysaka: > I don't think that any bpf ids can be 0 regardless of type (map, program, etc)?

It should not be possible yes

> In fact I'd probably change all these to return Option. I think NonZero* is nice to accept as input to enforce safety, but as a return value in these utils seems more annoying than anything else.

Yeah I do agree with you, I would prefer `Option` over `NonZero*` on returns too.

aya-obj/src/links.rs
```diff
@@ -0,0 +1,100 @@
...
+    InvalidTypeBinding,
+};
+
+impl TryFrom<u32> for bpf_link_type {
+    type Error = InvalidTypeBinding<u32>;
+
+    fn try_from(link_type: u32) -> Result<Self, Self::Error> {
+        use bpf_link_type::*;
+        Ok(match link_type {
+            x if x == BPF_LINK_TYPE_UNSPEC as u32 => BPF_LINK_TYPE_UNSPEC,
```
marysaka, line 14: Kind of a bit nit here but I would instead transform link_type to `bpf_link_type::Type` that would remove all the as u32 boilerplate around :smile:

aya-obj/src/links.rs
```diff
@@ -0,0 +1,100 @@
...
+        })
+    }
+}
+
+impl TryFrom<u32> for bpf_attach_type {
+    type Error = InvalidTypeBinding<u32>;
+
+    fn try_from(attach_type: u32) -> Result<Self, Self::Error> {
+        use bpf_attach_type::*;
+        Ok(match attach_type {
```
marysaka, line 38: Same comment as earlier but this time with `bpf_attach_type::Type`

aya-obj/src/maps.rs
```diff
@@ -31,7 +25,6 @@ impl TryFrom<u32> for crate::generated::bpf_map_type {
             x if x == BPF_MAP_TYPE_LRU_HASH as u32 => BPF_MAP_TYPE_LRU_HASH,
             x if x == BPF_MAP_TYPE_LRU_PERCPU_HASH as u32 => BPF_MAP_TYPE_LRU_PERCPU_HASH,
             x if x == BPF_MAP_TYPE_LPM_TRIE as u32 => BPF_MAP_TYPE_LPM_TRIE,
-            x if x == BPF_MAP_TYPE_BLOOM_FILTER as u32 => BPF_MAP_TYPE_BLOOM_FILTER,
```
marysaka, line 34: Why was that removed?

aya-obj/src/maps.rs
```diff
@@ -42,7 +35,6 @@ impl TryFrom<u32> for crate::generated::bpf_map_type {
             x if x == BPF_MAP_TYPE_CGROUP_STORAGE_DEPRECATED as u32 => {
                 BPF_MAP_TYPE_CGROUP_STORAGE_DEPRECATED
             }
-            x if x == BPF_MAP_TYPE_CGRP_STORAGE as u32 => BPF_MAP_TYPE_CGRP_STORAGE,
```
marysaka, line 45: Why was that removed?

aya-obj/src/programs/types.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::{
+    generated::bpf_prog_type::{self, *},
+    InvalidTypeBinding,
+};
+
+impl TryFrom<u32> for bpf_prog_type {
+    type Error = InvalidTypeBinding<u32>;
+
+    fn try_from(prog_type: u32) -> Result<Self, Self::Error> {
+        Ok(match prog_type {
```
marysaka, line 12: Same comment as earlier but this time with `bpf_prog_type::Type`

aya/src/programs/info.rs
```diff
@@ -0,0 +1,548 @@
...
+    /// Introduced in kernel v4.15.
+    pub fn name_as_str(&self) -> Option<&str> {
+        let name = std::str::from_utf8(self.name()).ok();
+        if let Some(name_str) = name {
+            if FEATURES.bpf_name() || !name_str.is_empty() {
+                return name;
+            }
+        }
+        None
+    }
```
marysaka, line 161:
```suggestion
    pub fn name_as_str(&self) -> Option<&str> {
        let name = std::str::from_utf8(self.name()).ok()?;

        if FEATURES.bpf_name() || !name.is_empty() {
            return Some(name);
        }

        None
    }
```

aya/src/util.rs
```diff
@@ -177,6 +178,12 @@ impl KernelVersion {
     }
 }
 
+impl Display for KernelVersion {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}.{}.{}", self.major, self.minor, self.patch)
+    }
+}
+
```
marysaka, line 186: Nice thing to see added :+1:

marysaka (commented):
Overall, I agree with @alessandrod, we shouldn't use `Option<NonZero*>` as return values but `Option<IntegerType>` instead (0 mapping to None)

Some code pattern could also be simplified I feel here and there but the logic related to those are usually small so that's fine in any cases.

Nice to see all the new comments around kernel version here, I didn't dig too much if the version were appropriate everywhere but if you want I can look more closely at it later :+1:

tamird:
> Overall, I agree with @alessandrod, we shouldn't use `Option<NonZero*>` as return values but `Option<IntegerType>` instead (0 mapping to None)

Why is that better? If we know they can't be zero it's better to communicate that in the type.

alessandrod:
> Why is that better? If we know they can't be zero it's better to communicate that in the type.

Because API ergonomics matter. 

`NonZero<T>` doesn't Deref to T, so I have to call `.get()`.  Having to call get seems worse than having extra information there's no use for, especially since `NonZero<T>::new(v)` return `Option<NonZero<T>>`, so I have to match _and_ get. 

```
// This _does_ provide useful information - I can't call f(0)
fn f(x: NonZeroU32) {
}

// Option tells me whether the map_id is there or not. Then what use is it to me to know that the map id won't be 0? 
// This would be useful if I were to pass the return value to something else, but we have stronger wrappers for things like ProgramIds etc.
fn map_id(&self) -> Option<NonZeroU32> {
}

```

tamird:
> > Why is that better? If we know they can't be zero it's better to communicate that in the type.
> 
> Because API ergonomics matter.
> 
> `NonZero<T>` doesn't Deref to T, so I have to call `.get()`. Having to call get seems worse than having extra information there's no use for, especially since `NonZero<T>::new(v)` return `Option<NonZero<T>>`, so I have to match _and_ get.
> 
> ```
> // This _does_ provide useful information - I can't call f(0)
> fn f(x: NonZeroU32) {
> }
> 
> // Option tells me whether the map_id is there or not. Then what use is it to me to know that the map id won't be 0? 
> // This would be useful if I were to pass the return value to something else, but we have stronger wrappers for things like ProgramIds etc.
> fn map_id(&self) -> Option<NonZeroU32> {
> }
> ```

Then why expose the scalar at all? If the value doesn't matter, make it opaque.

alessandrod:
> Then why expose the scalar at all? If the value doesn't matter, make it opaque.

We should definitely have types for all these IDs, but babysteps. Even without stronger ids this PR was an improvement over what was there before, and removing NonZeroU32 is an improvement over what's there, imo.

aya-obj/src/maps.rs
tyrone-wu: There was a duplicate match for `bloom_filter` line 58.

aya-obj/src/maps.rs
tyrone-wu: Also duplicate `cgro_storage` match on line 60.

aya-obj/src/links.rs
tyrone-wu: Could you elaborate on `::Type`? I'm not familiar with the usage of that syntax. 🤔

marysaka: Those `BPF_LINK_TYPE_` values have a type of `bpf_link_type::Type` (that you can find defined [here for x86_64](https://github.com/aya-rs/aya/blob/635ed3baed5442c1364a360d7234b72c4ffe3fd8/ebpf/aya-ebpf-bindings/src/x86_64/bindings.rs#L698))

So the idea would be to cast link_type to `bpf_link_type::Type` instead of casting each  `BPF_LINK_TYPE_` values to u32.

tamird: @tyrone-wu did you want to address this one?

test/integration-test/src/utils.rs
tamird: That seems OK to me. Can you revisit or file an issue?

aya/src/sys/bpf.rs
```diff
@@ -694,6 +703,62 @@ pub(crate) fn is_prog_name_supported() -> bool {
     bpf_prog_load(&mut attr).is_ok()
 }
 
+/// Tests whether `nr_map_ids` & `map_ids` fields in `bpf_prog_info` is available.
+pub(crate) fn is_info_map_ids_supported() -> bool {
```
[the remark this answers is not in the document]
tamird: reminder that we might want to deduplicate here.

tamird (commented):
Just pinging some leftovers.

aya/src/sys/bpf.rs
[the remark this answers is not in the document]
tamird: https://github.com/aya-rs/aya/pull/1165

aya-obj/src/links.rs
tamird: I think this was done.

test/integration-test/src/utils.rs
tyrone-wu: I can revisit/ 👍

aya-log/src/lib.rs
tamird: https://github.com/aya-rs/aya/commit/02d1db5fc043fb7af90c14d13de6419ec5b9bcb5

aya-log/src/lib.rs
tamird: https://github.com/aya-rs/aya/commit/02d1db5fc043fb7af90c14d13de6419ec5b9bcb5

test/integration-test/src/utils.rs
tamird: https://github.com/aya-rs/aya/pull/1219

aya/src/programs/info.rs
tamird: https://github.com/aya-rs/aya/commit/02d1db5fc043fb7af90c14d13de6419ec5b9bcb5

aya-obj/src/links.rs
tamird: This wasn't done, but the suggestion is UB. Casting an out-of-range integer to a Rust enum is UB.

aya-obj/src/links.rs
tamird: Same as above, doing this is UB.

PR #1028, opened 2024-09-11
ci: pin rust to nightly-2024-09-09 until public-api fixed

Billy99:
The public-api (cargo xtask public-api) is failing with the latest rust nightly package. Temporarily pin nightly to last known working version until the issue can be resolved.


alessandrod:
for context https://discord.com/channels/855676609003651072/855676609003651075/1285234572061048844

PR #1029, opened 2024-09-13
Fix linux kernel version in gh ci runners

davibe:
This updates the linux kernels of the CI linux runners since at least one of the current ones has become unavailable as seem in this build https://github.com/aya-rs/aya/actions/runs/10843802034/job/30091556395


alessandrod (approved):

alessandrod (approved):

alessandrod:
Merging on red since the lint failure is being addressed in another PR

PR #1038, opened 2024-09-25
aya: add guardrails for valid combinations of perf_event `type` and `config` fields

tyrone-wu:
Add guardrails for when setting event `type` and `config` for perf_event program. The added `PerfEventConfig` enum now defines the event `type` and `config` of interest.

 Idiomatic Rust types are added for:
- `perf_hw_id` => `HardwareEvent`
- `perf_sw_ids` => `SoftwareEvent`
- `perf_hw_cache_id` => `HwCacheEvent`
- `perf_hw_cache_op_id` => `HwCacheOp`

The motivation behind this is mainly for the `type` and `config` fields of `bpf_link_info.perf_event.event` (which I'm working on in another branch). I plan to use these newly added enums `bpf_link_info` metadata. https://elixir.bootlin.com/linux/v6.10/source/include/uapi/linux/bpf.h#L6714

One thing to note is that although `Breakpoint`/`PERF_TYPE_BREAKPOINT` variant exists, it is not fully implemented at the moment (due to some additional fields need in `perf_event_attr` like `bp_type`, etc.). It's only usage currently will be in retrieving link info.


tyrone-wu:
I just realized that if `PERF_TYPE_BREAKPOINT` get implemented, then it can't be reused in link info due to the extra fields...

aya/src/programs/perf_event.rs
```diff
@@ -23,22 +21,200 @@ use crate::{
...
+        /// This value can extracted from `/sys/bus/event_source/devices/*/format/`, where the
+        /// `config:<value>` indicates the bit position to set.
+        ///
+        /// For example, `config:3` => `config = 1 << 3`.
+        config: u64,
+    },
+}
+
+/// The "generalized" hardware CPU events provided by the kernel.
+#[repr(u64)]
```
tamird, line 84: why do we need this repr?

tamird (approved):
LGTM

tamird (changes requested):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1038)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @alessandrod and @tyrone-wu)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IiIKK47gKnMG9C8Xv):*
nit: get this down to 50 columns plz (just the subject)
___
*[`aya/src/programs/perf_event.rs` line 89 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IjZI90AS74LKxgycX:-O8IjZI90AS74LKxgycY:bxsjc2v) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L89)):*
> ```XML
>     /// The total CPU cycles.
>     #[doc(alias = "PERF_COUNT_HW_CPU_CYCLES")]
>     CpuCycles = perf_hw_id::PERF_COUNT_HW_CPU_CYCLES as isize,
> ```

it'd be better to use repr(foo) if it means we can avoid these casts (especially from an unsigned type to this signed type).
___
*[`aya/src/programs/perf_event.rs` line 327 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8Iif8f7sMtlM68_Fg_:-O8Iif8f7sMtlM68_Fga:b-i0d18e) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L327)):*
> ```XML
> 
>         let (event_type, config) = match perf_type {
>             PerfEventConfig::Hardware(hw_event) => (PERF_TYPE_HARDWARE as u32, hw_event as u64),
> ```

would it be possible to avoid these `as` casts? In other words can we align the types we use such that we match the types of these constants? More precisely we should use the type `perf_type_id` instead of u32 and

tyrone-wu (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1038)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IiIKK47gKnMG9C8Xv):*

nit: get this down to 50 columns plz (just the subject)
</blockquote></details>

:thumbsup: :thumbsup: 
___
*[`aya/src/programs/perf_event.rs` line 89 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IjZI90AS74LKxgycX:-O8JY9600SeCNZbku2d_:b-91xh4m) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L89)):*

it'd be better to use repr(foo) if it means we can avoid these casts (especially from an unsigned type to this signed type).
</blockquote></details>

Correct me if I'm wrong, but I believe casting is unavoidable since they're being coerced from enum type. `isize` would let the compiler handle its layout, so I don't think it would differ with a different `repr` since casting happens either way.
___
*[`aya/src/programs/perf_event.rs` line 327 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8Iif8f7sMtlM68_Fg_:-O8NU6XZ3BWXdTpLlxRk:b-yss7n6) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L327)):*

would it be possible to avoid these `as` casts? In other words can we align the types we use such that we match the types of these constants? More precisely we should use the type `perf_type_id` instead of u32 and
</blockquote></details>

:saluting_face: The `PerfEventConfig::Pmu` variant not belonging to `perf_type_id`, but I was sorta able to align it for a single `u32` cast.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1038)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @alessandrod and @tyrone-wu)

___
*[`aya/src/programs/perf_event.rs` line 89 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IjZI90AS74LKxgycX:-O8SMemy8wv8uikHQ_dg:b-6zwyna) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L89)):*
<details><summary><i>Previously, tyrone-wu (Tyrone Wu) wrote…</i></summary><blockquote>

Correct me if I'm wrong, but I believe casting is unavoidable since they're being coerced from enum type. `isize` would let the compiler handle its layout, so I don't think it would differ with a different `repr` since casting happens either way.
</blockquote></details>

Could we use https://github.com/rust-num/num-derive to derive ToPrimitive on `perf_hw_id` and then use that instead of these `as` casts?

tyrone-wu (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1038)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`aya/src/programs/perf_event.rs` line 89 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IjZI90AS74LKxgycX:-O8dBurM0nnSF9zunpfw:bx518w8) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L89)):*

Could we use https://github.com/rust-num/num-derive to derive ToPrimitive on `perf_hw_id` and then use that instead of these `as` casts?
</blockquote></details>

I gave `num-derive` a try , it turns out the functions implemented from `ToPrimitive` aren't public :frowning_face:. And also, the functions (`.to_u32`, etc.) return `Option`s. https://docs.rs/num-traits/0.2.19/src/num_traits/cast.rs.html#17

There is https://github.com/illicitonion/num_enum, however, it looks like it just implements `From<repr(num)>` based on provided `#[repr()]`, which wouldn't be too different from `as`. :/

tamird (changes requested):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1038)* status: all files reviewed, 7 unresolved discussions (waiting on @alessandrod and @tyrone-wu)

___
*[`aya/src/programs/perf_event.rs` line 89 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8IjZI90AS74LKxgycX:-O9kGCZx1HpPHqB1fwHb:bdjohk5) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L89)):*
<details><summary><i>Previously, tyrone-wu (Tyrone Wu) wrote…</i></summary><blockquote>

I gave `num-derive` a try , it turns out the functions implemented from `ToPrimitive` aren't public :frowning_face:. And also, the functions (`.to_u32`, etc.) return `Option`s. https://docs.rs/num-traits/0.2.19/src/num_traits/cast.rs.html#17

There is https://github.com/illicitonion/num_enum, however, it looks like it just implements `From<repr(num)>`, which wouldn't be too different `as`. :/
</blockquote></details>

Implementing From is better than `as` because it guarantees that there's no precision loss. Let's use num_enum?
___
*[`aya/src/programs/perf_event.rs` line 327 at r1](https://reviewable.io/reviews/aya-rs/aya/1038#-O8Iif8f7sMtlM68_Fg_:-O9kHblxCgKpZnyrBxhP:b-cadf1i) ([raw file](https://github.com/aya-rs/aya/blob/8f8210c7d6817893caf9a7e2c63820f24e70be76/aya/src/programs/perf_event.rs#L327)):*
<details><summary><i>Previously, tyrone-wu (Tyrone Wu) wrote…</i></summary><blockquote>

:saluting_face: The `PerfEventConfig::Pmu` variant not belonging to `perf_type_id`, but I was sorta able to align it for a single `u32` cast.
</blockquote></details>

I'd love to see these casts go away
___
*[`aya/src/programs/perf_event.rs` line 26 at r3](https://reviewable.io/reviews/aya-rs/aya/1038#-O9kH6g815n3nhuH16Vc:-O9kH6g815n3nhuH16Vd:b-37hzl9) ([raw file](https://github.com/aya-rs/aya/blob/c5be26776b5649707fd24b99bd82a1b9c8b77260/aya/src/programs/perf_event.rs#L26)):*
> ```XML
> /// The type of perf event and their respective configuration.
> #[doc(alias = "perf_type_id")]
> #[derive(Debug, Clone)]
> ```

should this be Copy too?
___
*[`aya/src/programs/perf_event.rs` line 274 at r3](https://reviewable.io/reviews/aya-rs/aya/1038#-O9kGzpoAbNfXr3YnvUN:-O9kGzpoAbNfXr3YnvUO:b4rv38l) ([raw file](https://github.com/aya-rs/aya/blob/c5be26776b5649707fd24b99bd82a1b9c8b77260/aya/src/programs/perf_event.rs#L274)):*
> ```XML
> /// #     Ebpf(#[from] aya::EbpfError)
> /// # }
> /// use aya::{
> ```

since you're shuffling, put this at the top?
___
*[`aya/src/programs/perf_event.rs` line 288 at r3](https://reviewable.io/reviews/aya-rs/aya/1038#-O9kHA4r59q-Ay0IwhMS:-O9kHA4r59q-Ay0IwhMT:b-8909ic) ([raw file](https://github.com/aya-rs/aya/blob/c5be26776b5649707fd24b99bd82a1b9c8b77260/aya/src/programs/perf_event.rs#L288)):*
> ```XML
> /// for cpu in online_cpus().map_err(|(_, error)| error)? {
> ///     prog.attach(
> ///         perf_type.clone(),
> ```

consider removing clone this if you make the type Copy
___
*[`aya/src/programs/perf_event.rs` line 310 at r3](https://reviewable.io/reviews/aya-rs/aya/1038#-O9kG_L_4otUV-lb4H00:-O9kG_L_4otUV-lb4H01:b8ynvoc) ([raw file](https://github.com/aya-rs/aya/blob/c5be26776b5649707fd24b99bd82a1b9c8b77260/aya/src/programs/perf_event.rs#L310)):*
> ```XML
>     /// Attaches to the given perf event.
>     ///
>     /// The [`perf_type`](PerfEventConfig) defines the event `type` and `config` of interest.
> ```

did you want want to add the word "argument"? you have it in the next sentence.
___
*[`aya/src/programs/perf_event.rs` line 344 at r3](https://reviewable.io/reviews/aya-rs/aya/1038#-O9kHMrJA6uJIfTpQmZy:-O9kHMrJA6uJIfTpQmZz:bhx7t3i) ([raw file](https://github.com/aya-rs/aya/blob/c5be26776b5649707fd24b99bd82a1b9c8b77260/aya/src/programs/perf_event.rs#L344)):*
> ```XML
>                     PerfEventConfig::Raw { event_id } => (PERF_TYPE_RAW, event_id),
>                     PerfEventConfig::Breakpoint => (PERF_TYPE_BREAKPOINT, 0),
>                     PerfEventConfig::Pmu { .. } => unreachable!(), // not possible due to earlier match
> ```

yuck. can we avoid this? the only shared code is `as u32` =/

tamird:
@tyrone-wu interested in getting this over the finish line?

tamird:
Weird, I pushed this branch and I can see that in https://github.com/tyrone-wu/aya/tree/aya/perf-event-type-config but the PR hasn't updated.

ver-nyan:
> Addressed my own comments. @tyrone-wu does this look reasonable to you?

Looks good to me. 👍 Maybe once `const into` is stabilized https://github.com/rust-lang/rust/issues/143773 (which might take a while), the conversions can probaly be simplified.

tamird:
@alessandrod does the API look reasonable to you?

PR #1039, opened 2024-09-25
ci: use distro's clang, llvm

tamird:


tamird:
Because homebrew now has LLVM 19 this will fix our CI under #1022 can get us out from under the mercy of homebrew.

Issue #1024, opened 2024-09-09
Remove instances of KernelVersion::current().unwrap()


alessandrod:
In a bunch of places we do `if KernelVersion::current().unwrap() >= KernelVersion::new(....) {`

In theory, we should always know the kernel version so the unwrap is understandable. However it seems that distributions really like to mess with how the kernel version is reported and we've had this unwrap fail in more than one case (some version of ubuntu and recently proxmox).

In the cases where we have `if version { ... } else { ... }` we should handle the case where current() fails, log a warning, and probably do the _else_ thing which is usually whatever is supported by older kernels.

pythops:
I have started working on this one. 
What do you mean by `log a warning` ? would `eprintln!` with a message work or should we include tracing crate ?

alessandrod:
> I have started working on this one.

Awesome thanks!

> What do you mean by `log a warning` ? would `eprintln!` with a message work or should we include tracing crate ?

We use the `log` crate, so you can
```
use log::warn;

warn!("some warning message");
```

PR #1042, opened 2024-09-27
handle unwrap while getting current kernel version

pythops:
Closes https://github.com/aya-rs/aya/issues/1024 

There is still 2 places where we still `unwrap` though. To handle it there might require changes to some functions that I am not sure if they should be part of this PR

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1042)

### DIFF
--- a/aya/src/programs/cgroup_device.rs
+++ b/aya/src/programs/cgroup_device.rs
@@ -1,7 +1,8 @@
 //! Cgroup device programs.
 
-use log::warn;
 use std::os::fd::AsFd;
+
+use log::warn;
 
 use crate::{
     generated::{bpf_attach_type::BPF_CGROUP_DEVICE, bpf_prog_type::BPF_PROG_TYPE_CGROUP_DEVICE},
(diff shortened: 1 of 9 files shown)


tamird:
I won't speak for @alessandrod but IMO logging errors is not what we want to do here. We should instead propagate those errors, probably.

alessandrod:
> I won't speak for @alessandrod but IMO logging errors is not what we want to do here. We should instead propagate those errors, probably.

IMO it's not worth polluting the API for a condition that should effectively never happen. At some point, we'll fix our errors so that they don't have 2000 variants and users can do exhaustive matches. Doing an exhaustive match on "the kernel version can't be detected" is unnecessary cognitive overhead. We should always be able to detect the kernel version and we are. There are some obscure edge cases with distros very few people use (proxmox) and in that case I think it's sufficient to warn and execute the most conservative code. Warnings are shown by default, so they're unlikely to go undetected.

pythops:
Any review guys :pray:

tamird:
This has merge conflicts.

PR #1043, opened 2024-09-27
perf: cache `nr_cpus` in a thread_local

0xdeafbeef:
image

image

### DIFF
--- a/aya/src/util.rs
+++ b/aya/src/util.rs
@@ -196,7 +196,15 @@ pub fn online_cpus() -> Result<Vec<u32>, (&'static str, io::Error)> {
 ///
 /// See `/sys/devices/system/cpu/possible`.
 pub fn nr_cpus() -> Result<usize, (&'static str, io::Error)> {
-    read_cpu_ranges(POSSIBLE_CPUS).map(|cpus| cpus.len())
+    thread_local! {
+        // TODO(https://github.com/rust-lang/rust/issues/109737): Use
+        // `std::cell::OnceCell` when `get_or_try_init` is stabilized.
+        static CACHE: once_cell::unsync::OnceCell<usize> = const { once_cell::unsync::OnceCell::new() };
+    }
+    CACHE.with(|cell| {
+        cell.get_or_try_init(|| read_cpu_ranges(POSSIBLE_CPUS).map(|cpus| cpus.len()))
+            .copied()
+    })
 }
 
 fn read_cpu_ranges(path: &'static str) -> Result<Vec<u32>, (&'static str, io::Error)> {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ netns-rs = { version = "0.1", default-features = false }
 nix = { version = "0.29.0", default-features = false }
 num_enum = { version = "0.7", default-features = false }
 object = { version = "0.36", default-features = false }
+once_cell = { version = "1.20.1", default-features = false }
 proc-macro-error = { version = "1.0", default-features = false }
 proc-macro2 = { version = "1", default-features = false }
 public-api = { version = "0.38.0", default-features = false }
--- a/aya/Cargo.toml
+++ b/aya/Cargo.toml
@@ -21,6 +21,7 @@ bytes = { workspace = true }
 libc = { workspace = true }
 log = { workspace = true }
 object = { workspace = true, features = ["elf", "read_core", "std", "write"] }
+once_cell = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["rt"], optional = true }
 


aya/src/util.rs
```diff
@@ -200,9 +201,25 @@ pub fn online_cpus() -> Result<Vec<u32>, io::Error> {
 
 /// Get the number of possible cpus.
 ///
+/// Caches the value into thread local value on the first successful access
 /// See `/sys/devices/system/cpu/possible`.
 pub fn nr_cpus() -> Result<usize, io::Error> {
-    Ok(possible_cpus()?.len())
+    // per https://www.kernel.org/doc/Documentation/core-api/cpu_hotplug.rst
```
tamird, line 207: Why a thread local rather than a global static? Why not https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html?

0xdeafbeef: It doesn't have overhead on synchronization at all. Initialization is pretty cheap, so calling it `num thread` times is not a problem. Also, `OnceLock` has an MSRV of 1.70 which is incompatible with aya's MSRV. And I'm not sure how to cache fallible operations given that `io::Error` is not clonable:)

tamird: Probably want https://doc.rust-lang.org/stable/std/cell/struct.OnceCell.html#method.get_or_try_init which is still unstable.

aya/src/util.rs
```diff
@@ -200,9 +201,25 @@ pub fn online_cpus() -> Result<Vec<u32>, io::Error> {
...
+    // it's safe to assume that this value won't change
+    thread_local! {
+        pub static NUM_CPUS: Cell<Option<usize>> = const {Cell::new(None)};
```
tamird, line 210: remove the `pub`? it's meaningless here

aya/src/util.rs
```diff
@@ -200,9 +201,25 @@ pub fn online_cpus() -> Result<Vec<u32>, io::Error> {
 
 /// Get the number of possible cpus.
 ///
+/// Caches the value into thread local value on the first successful access
 /// See `/sys/devices/system/cpu/possible`.
 pub fn nr_cpus() -> Result<usize, io::Error> {
-    Ok(possible_cpus()?.len())
```
tamird, line 205: Why not cache in `possible_cpus`? Surely the most expensive thing is file system interaction.

aya/src/util.rs
0xdeafbeef: done

aya/src/util.rs
0xdeafbeef: It's used only inside nr_cpus, but I can change it to 

```rs
fn possible_cpus() -> Result<&'static [u32], io::Error>
```
Cloning vector seems costly

aya/CHANGELOG.md
```diff
@@ -182,1097 +182,708 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Chore
 
- - <csr-id-b3e7ef741c5b8d09fc7dc8302576f8174be75ff4/> Use the cargo workspace package table
```
tamird, line 185: Why?

aya/src/util.rs
tamird: Yes, changing the signature is fine.

0xdeafbeef: > Yes, changing the signature is fine.

```rust
pub(crate) fn possible_cpus() -> Result<&'static [u32], io::Error> {
    thread_local! {
        static CPUS_LIST: RefCell<Option<Vec<u32>>> = const {RefCell::new(None)};
    }

    CPUS_LIST.with(|cell| {
        if let Some(cpus) = &*cell.borrow() {
            Ok(cpus.as_slice())
        } else {
            let data = fs::read_to_string(POSSIBLE_CPUS)?;
            let cpus = parse_cpu_ranges(data.trim()).map_err(|_| {
                io::Error::new(
                    io::ErrorKind::Other,
                    format!("unexpected {POSSIBLE_CPUS} format"),
                )
            })?;

            *cell.borrow_mut() = Some(cpus.clone());

            Ok(cpus.as_slice())
        }
    })
}
```

It won't compile :(
image

So we should either use `OnceLock` or just cache `u32`. I'm voting for 2 variant

aya/CHANGELOG.md
0xdeafbeef: Oops, zed just reformatted this without asking me, reverted

aya/src/util.rs
tamird: I sent https://github.com/aya-rs/aya/pull/1045 in which I remove `possible_cpus`. This can then cache the u32 without seeming strange.

0xdeafbeef: @tamird should I rebase on it?

0xdeafbeef: Yep, looks good to me :)

> Rebased this, does it look reasonable to you?

PR #1048, opened 2024-10-07
fix(aya): XdpFlags default value use XdpFlags::SKB_MODE

cppcoffee:
`XdpFlags` type is u32, use `#derive(Default)` u32 default value is 0, `XdpFlags` does not define a value of 0.

to implement `Default` value use `XdpFlags::SKB_MODE`.

### DIFF
--- a/aya/src/programs/xdp.rs
+++ b/aya/src/programs/xdp.rs
@@ -42,7 +42,7 @@ pub enum XdpError {
 
 bitflags::bitflags! {
     /// Flags passed to [`Xdp::attach()`].
-    #[derive(Clone, Copy, Debug, Default)]
+    #[derive(Clone, Copy, Debug)]
     pub struct XdpFlags: u32 {
         /// Skb mode.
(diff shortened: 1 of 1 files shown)


tamird:
It's a `bitflags`. `0` is just fine, it means none of the flags are set.

alessandrod:
XdpFlags::default() seems more idiomatic than making people pass 0? But I thought the bitflags macro already implemented ::default()?

cppcoffee:
The example code uses `XdpFlags::default()` and appears to see that it doesn't run.

https://aya-rs.dev/book/start/hello-xdp/#user-space-component

So setting the default value to `XdpFlags::SKB_MODE` will work for most scenarios.

tamird:
> XdpFlags::default() seems more idiomatic than making people pass 0? But I thought the bitflags macro already implemented ::default()?

That is what I was saying. The implementation of default returns zero.

alessandrod:
Ooh I see, @cppcoffee wants to set it to SKB_MODE so that it works more often by default. No I agree that the default should be 0.

PR #1053, opened 2024-10-10
codegen: add `bpfcgroup_iter_order` `NFPROTO*` `nf_inet_hooks`

tyrone-wu:
Adds the following to codegen:
- `bpf_cgroup_iter_order`: used in `bpf_link_info.iter.group.order`
- `NFPROTO_*`: used in `bpf_link_info.netfilter.pf`
- `nf_inet_hooks`: used in `bpf_link_info.netfilter.hooknum`

`linux/netfilter.h` is included in `aya-obj/include/linux_wrapper.h` for `NFPROTO_.*` and `nf_inet_hooks` to generate.

This is all that's left for `bpf_link_info`. 🙃

### DIFF
--- a/xtask/src/codegen/aya.rs
+++ b/xtask/src/codegen/aya.rs
@@ -109,6 +109,10 @@ fn codegen_bindings(opts: &SysrootOptions, libbpf_dir: &Path) -> Result<(), anyh
         // NETLINK
         "ifinfomsg",
         "tcmsg",
+        // ITER
+        "bpf_cgroup_iter_order",
+        // NETFILTER
+        "nf_inet_hooks",
     ];
 
     let vars = [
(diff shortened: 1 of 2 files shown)


xtask/src/codegen/aya.rs
```diff
@@ -211,7 +217,10 @@ fn codegen_bindings(opts: &SysrootOptions, libbpf_dir: &Path) -> Result<(), anyh
                 .constified_enum("BTF_VAR_.*")
                 .constified_enum("IFLA_.*")
                 .constified_enum("TCA_.*")
-                .constified_enum("BPF_RINGBUF_.*");
+                .constified_enum("BPF_RINGBUF_.*")
```
dave-tucker, line 220:
```suggestion
```
Looks like this is a duplicate of the line above?

dave-tucker (dismissed):

xtask/src/codegen/aya.rs
dave-tucker: or line below even - either way I'm seeing 2 of them lol

tyrone-wu: ahh, thanks for catching that 😵

alessandrod (approved):

PR #1057, opened 2024-10-15
WIP: auto-attach tracepoints based on SEC() data

Ablu:
This is mostly a conversation starter. The work originated in me asking about this on Discord [1].

I can try to flesh this out to a full solution, but I have some questions on what the ideal solution looks like:

Question 1: Currently I am mostly seeing these `SEC()` annotations in C bpf programs. The automatic template generator of the aya-ebpf tracepoints does not seem to populate the fields (tough it looks like they are supported!). Should we populate those fields there too?
libbpf-tools generates those .skel.h headers that allow automating the attach. I find that a nice QoL improvement over manually having to attach all the probes.

Question 2: I assume we also want to expose public getters for these fields?

Question 3: Do we want to extend this to all of the other "well-known" `SEC()` annotations that libbpf-tools uses? This would allow us to have a central `auto_attach()` that simply iterates over all programs and attaches them all?

Open TODOs:
- tests
- better documentation
- review contribution guidelines in detail

[1] https://discord.com/channels/855676609003651072/855676609003651075/1288759816612741120

### DIFF
--- a/aya-obj/src/obj.rs
+++ b/aya-obj/src/obj.rs
@@ -249,7 +249,10 @@ pub enum ProgramSection {
     URetProbe {
         sleepable: bool,
     },
-    TracePoint,
+    TracePoint {
+        category: Option<String>,
+        name: Option<String>,
+    },
     SocketFilter,
     Xdp {
         frags: bool,
@@ -330,7 +333,11 @@ impl FromStr for ProgramSection {
                 },
             },
             "tp_btf" => BtfTracePoint,
-            "tracepoint" | "tp" => TracePoint,
+            "tracepoint" | "tp" => {
+                let category = pieces.next().map(|s| s.to_string());
+                let name = pieces.next().map(|s| s.to_string());
+                TracePoint { category, name }
+            }
             "socket" => SocketFilter,
             "sk_msg" => SkMsg,
             "sk_skb" => {
@@ -2014,7 +2021,7 @@ mod tests {
         assert_matches!(
             obj.parse_section(fake_section(
                 EbpfSectionKind::Program,
-                "tracepoint/foo",
+                "tracepoint/cat/name",
                 bytes_of(&fake_ins()),
                 None
             )),
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -30,10 +30,11 @@ use crate::{
         Object, ParseError, ProgramSection,
     },
     programs::{
-        BtfTracePoint, CgroupDevice, CgroupSkb, CgroupSkbAttachType, CgroupSock, CgroupSockAddr,
-        CgroupSockopt, CgroupSysctl, Extension, FEntry, FExit, KProbe, LircMode2, Lsm, PerfEvent,
-        ProbeKind, Program, ProgramData, ProgramError, RawTracePoint, SchedClassifier, SkLookup,
-        SkMsg, SkSkb, SkSkbKind, SockOps, SocketFilter, TracePoint, UProbe, Xdp,
+        trace_point::TracePointAttachInfo, BtfTracePoint, CgroupDevice, CgroupSkb,
+        CgroupSkbAttachType, CgroupSock, CgroupSockAddr, CgroupSockopt, CgroupSysctl, Extension,
+        FEntry, FExit, KProbe, LircMode2, Lsm, PerfEvent, ProbeKind, Program, ProgramData,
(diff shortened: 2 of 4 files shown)


alessandrod:
> This is mostly a conversation starter. The work originated in me asking about this on Discord [1].

Hey! Thanks for the PR!

 
> I can try to flesh this out to a full solution, but I have some questions on what the ideal solution looks like:
> 
> Question 1: Currently I am mostly seeing these `SEC()` annotations in C bpf programs. The automatic template generator of the aya-ebpf tracepoints does not seem to populate the fields (tough it looks like they are supported!). Should we populate those fields there too? libbpf-tools generates those .skel.h headers that allow automating the attach. I find that a nice QoL improvement over manually having to attach all the probes.

For things where it makes sense - for example tracepoints that in most cases have exactly 1 attach target - yes I think it makes sense. 


> 
> Question 2: I assume we also want to expose public getters for these fields?

Yes, see CgroupSkb where we expose already https://docs.rs/aya/latest/aya/programs/cgroup_skb/struct.CgroupSkb.html#method.expected_attach_type

> 
> Question 3: Do we want to extend this to all of the other "well-known" `SEC()` annotations that libbpf-tools uses?

This for sure - we want to be as compatible with libbpf as possible.

> This would allow us to have a central `auto_attach()` that simply iterates over all programs and attaches them all?

This is a separate question, and I think it requires a bit more thinking, eg:

```
let ebpf = Ebpf::load(...);
ebpf.auto_attach()?; // what happens to the programs that _can't_ be auto attached? 
```

I think it's counter intuitive that some programs wouldn't get attached. Not sure what the solution would be. What does libbpf do?

Ablu:
> For things where it makes sense - for example tracepoints that in most cases have exactly 1 attach target - yes I think it makes sense.

I would add k(ret)probe's to the list then :). I can do some research on which programs can be auto-attached in libbpf.

> I think it's counter intuitive that some programs wouldn't get attached. Not sure what the solution would be. What does libbpf do?

The code is here: https://github.com/libbpf/libbpf/blob/7984737fbf3b2a14a86321387bb62abb16cfc4ed/src/libbpf.c#L12365-L12404

So it looks like some conventions to prevent autoattachment: https://github.com/libbpf/libbpf/blob/7984737fbf3b2a14a86321387bb62abb16cfc4ed/src/libbpf.c#L747-L757 and some program types are simply not autoattached. Otherwise, a fail to auto-attach will simply cause an error.

That overall seems like reasonable to me!

alessandrod:
> > For things where it makes sense - for example tracepoints that in most cases have exactly 1 attach target - yes I think it makes sense.
> 
> I would add k(ret)probe's to the list then :). I can do some research on which programs can be auto-attached in libbpf.

Yep for sure - ideally we get to 100% libbpf compatibility.

> > I think it's counter intuitive that some programs wouldn't get attached. Not sure what the solution would be. What does libbpf do?
> 
> The code is here: https://github.com/libbpf/libbpf/blob/7984737fbf3b2a14a86321387bb62abb16cfc4ed/src/libbpf.c#L12365-L12404
> 
> So it looks like some conventions to prevent autoattachment: https://github.com/libbpf/libbpf/blob/7984737fbf3b2a14a86321387bb62abb16cfc4ed/src/libbpf.c#L747-L757 and some program types are simply not autoattached. Otherwise, a fail to auto-attach will simply cause an error.
> 
> That overall seems like reasonable to me!

I don't like that in the case of `auto_attach()` failure some things might remain attached, but I don't have a solution, and I like being "more compatible" with libbpf. If we do implement `auto_attach()`, we should also add `program.links()` and `program.links_mut()` so one can retrieve the attached links and detach.

Ablu:
I gave this a bit more of a try over the last days... But I keep hitting a few things that cause some frustration and leave me with no particular nice solution:

- I end up with a lot of duplication across `ProgramSection` and `Program`
- I end up with a lot of duplication between the individual entries within the section enum structs... To support the `?kprobe/foo` opt-out of auto-attaching I need to put some `no_auto_attach` flag into each struct. I feel like it may be better a property on ALL `ProgramSection`s (maybe along with `sleepable`?).
- A lot of boilerplate code to actually route the auto_attach to all the `Program` variants.

I am starting to wonder if moving the `Program` to a trait makes sense... Then they could be treated as equals without a lot of the enum variant routing...

Has this ever been explored? Or am I overthinking this and the duplication is not too bad?

PR #1059, opened 2024-10-16
ebpf: bpf_get_current_comm() use `comm.len()` instead of fixed number

cppcoffee:
Avoid changing the length of the `comm` variable and missing a change in another place.


qjerome:
This change does nothing. `comm` is an array and for arrays `len` function is `const` so in this precise case it will be `16u32`

PR #1061, opened 2024-10-18
wip errors

dave-tucker:


tyrone-wu:
I was wondering if the errors from `aya/src/utils.rs` should also be apart of this. They're technically not ebpf specific errors, but I think some parts would align with `FileError`.

dave-tucker:
> I was wondering if the errors from `aya/src/utils.rs` should also be apart of this. They're technically not ebpf specific errors, but I think some parts would align with `FileError`.

Good call! I missed those on the first pass through.
Once I've got agreement on the general direction this PR is taking I'll go ahead and clean those up too

PR #1065, opened 2024-10-21
HashMap: Impl. safe get_corrupt_u32()

luisgerhorst:
Just a minimal patch to see if there is any need for discussion here. If needed and if you agree with this approach, similar safe functions could likely be implemented for other collections / scalar types.


kaivogelgesang:
Since `get_corrupt_u32` returns a reference (`&u32`), could this create some dangling pointer that could later cause memory corruption?

PR #1067, opened 2024-10-23
ebpf: fixed wrong bpf_get_stackid flags type

cppcoffee:
The `flags` parameter of the `bpf_get_stackid` function is of type `u64`.


dave-tucker (changes requested):
These bindings are autogenerated and shouldn't be changed directly as they'll get overwritten on next codegen.

Based on the commit message it's not clear what you're trying to achieve.
The helper `bpf_get_stackid` returns a [`long` (not a ulong)](https://docs.ebpf.io/linux/helper-function/bpf_get_stackid/)
That's an `i64` in Rust, which is the type returned by [`get_stack_id`](https://docs.aya-rs.dev/aya_ebpf/maps/stack_trace/struct.stacktrace#method.get_stackid) or if calling the helper directly.

Therefore I think this is unnecessary

cppcoffee:
`get_stackid` flags are of u64 type, but got `BPF_F_USER_STACK` is u32.

```
let stack_id = unsafe { STACK_TRACES.get_stackid(ctx, BPF_F_USER_STACK as u64)? };
```

PR #1092, opened 2024-11-18
fix unsupported maps warning message

### LINKED ISSUE #1081: Unsupported map type warning message doesn't go away when unsupported type maps are allowed
I was trying to load an unsupported map type and I got the following error:

```console
The map my_task_storage is of type BPF_MAP_TYPE_TASK_STORAGE which is currently unsupported in Aya, use `allow_unsupported_maps()` to load it anyways
```

and my program crashes with something like:

```console
Unsupported map type found 29
```

and it's OK. 

Then I added `allow_unsupported_maps()` to `BpfLoader` it loads the program without crashing but prints again:

```console
The map my_task_storage is of type BPF_MAP_TYPE_TASK_STORAGE which is currently unsupported in Aya, use `allow_unsupported_maps()` to load it anyways
```

banditopazzo:
fix #1081 

the proposed solution is to remove the warning log and instead provide more meaningful information directly within the error itself.

without considering the issue, right now we have just a warning log (which contains the important information), but this is misleading since it actually leads to an error. I believe this can cause confusion because the critical details should be part of the error itself. For this reason I consolidated all the relevant information into the error message.

Additionally, I updated the `map_type` field type in `Unsupported` error from`u32` to `bpf_map_type`, as an unsupported map has a specific map type, which is distinct from a generic `u32` used in cases like the `InvalidMapType` error.

PS: Regarding `InvalidMapType` and other map-related errors, I believe they should also include the names of the maps.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1092)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -695,23 +695,17 @@ impl<'a> EbpfLoader<'a> {
             .collect();
         let maps = maps
             .drain()
-            .map(parse_map)
+            .map(|data| parse_map(data, *allow_unsupported_maps))
             .collect::<Result<HashMap<String, Map>, EbpfError>>()?;
-
-        if !*allow_unsupported_maps {
-            maps.iter().try_for_each(|(_, x)| match x {
-                Map::Unsupported(map) => Err(EbpfError::MapError(MapError::Unsupported {
-                    map_type: map.obj().map_type(),
-                })),
-                _ => Ok(()),
-            })?;
-        };
 
         Ok(Ebpf { maps, programs })
     }
 }
 
-fn parse_map(data: (String, MapData)) -> Result<(String, Map), EbpfError> {
+fn parse_map(
+    data: (String, MapData),
+    allow_unsupported_maps: bool,
+) -> Result<(String, Map), EbpfError> {
     let (name, map) = data;
     let map_type = bpf_map_type::try_from(map.obj().map_type()).map_err(MapError::from)?;
     let map = match map_type {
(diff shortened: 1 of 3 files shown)


tamird (changes requested):
Mostly looks good. Please run `CARGO_CFG_BPF_TARGET_ARCH=x86_64 cargo +nightly xtask public-api --bless --target x86_64-unknown-linux-gnu` to regenerate the API file.

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1092#-:-OC98E6x2FzqNifRo1Z9:b7iyv3t)* status: all files reviewed, 1 unresolved discussion (waiting on @banditopazzo)

___
*[`aya/src/maps/mod.rs` line 198 at r1](https://reviewable.io/reviews/aya-rs/aya/1092#-OC97hri8UzbI0ID_Dea:-OC97hri8UzbI0ID_Deb:bym89c2) ([raw file](https://github.com/aya-rs/aya/blob/b87361f0dd3b8d2208a0deda10e6845ff80469f9/aya/src/maps/mod.rs#L198)):*
> ```XML
> 
>     /// Unsupported Map type
>     #[error("the map {name} is of type {map_type:#?} which is currently unsupported in Aya, use `allow_unsupported_maps()` to load it anyways")]
> ```

can we make this a little shorter? seems way too long.

tamird (changes requested):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1092)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod and @banditopazzo)

___
*[`aya/src/maps/mod.rs` line 198 at r2](https://reviewable.io/reviews/aya-rs/aya/1092#-OCEHPCE7Wr8MIeUs8FZ:-OCEHPCF78esxPTb0gMG:b2xnbyu) ([raw file](https://github.com/aya-rs/aya/blob/3f5ce23f8704d28162cbc40a2191dc54d31fd4cd/aya/src/maps/mod.rs#L198)):*
> ```XML
> 
>     /// Unsupported Map type
>     #[error("map {name} is of type {map_type:#?}, unsupported in Aya; use `allow_unsupported_maps()` to load it")]
> ```

Why do we need `{:#?}` rather than `{:?}`?

```rust
#[error("type of {name} ({map_type:?}) is unsupported; see `EbpfLoader::allow_unsupported_maps`")]
```



_Code quote:_
```XML
 #[error("map {name} is of type {map_type:#?}, unsupported in Aya; use `allow_unsupported_maps()` to load it")]
```

tamird (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1092)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

tamird:
Thanks!

PR #1106, opened 2024-12-02
aya-ebpf-macros: remove aya-ebpf version

### LINKED ISSUE #1050: cyclic dependency breaks cargo smart-release
`aya-ebpf-macros` dev-depends on `aya-ebpf`

https://github.com/aya-rs/aya/blob/0e70838b0f896ec73a4c7a51cab29625e7c87f9a/aya-ebpf-macros/Cargo.toml#L20-L21

`aya-ebpf` depends on `aya-ebpf-macros` since it re-exports those macros.

https://github.com/aya-rs/aya/blob/0e70838b0f896ec73a4c7a51cab29625e7c87f9a/ebpf/aya-ebpf/Cargo.toml#L11-L13

This gets us into trouble since we can't `cargo publish` aya-ebpf since `aya-ebpf-macros` hasn't yet been published.

The "fix" is to remove the dev-dependency, `cargo publish` with `--allow-dirty`. Then resume releasing.

tamird:
The presence of this version specification causes cargo-smart-release to
trip on the circular dependency.

Fixes #1050.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1106)

### DIFF
--- a/aya-ebpf-macros/Cargo.toml
+++ b/aya-ebpf-macros/Cargo.toml
@@ -18,4 +18,4 @@ quote = { workspace = true }
 syn = { workspace = true, default-features = true, features = ["full"] }
 
 [dev-dependencies]
-aya-ebpf = { path = "../ebpf/aya-ebpf", version = "^0.1.1", default-features = false }
+aya-ebpf = { path = "../ebpf/aya-ebpf", default-features = false }


dave-tucker:
CI seems to disagree about it being unused 🤔

tamird:
Oh, it's a dependency for doctests 🤦  that's why running clippy locally didn't catch it.

tamird:
@dave-tucker seems this issue is related: https://github.com/rust-lang/rust/issues/58700 and one solution is to move documentation to the main crate from the macro crate. This is pretty inconvenient in our case because we write https://github.com/aya-rs/aya/blob/3d8cb08b7f067a5d57a7268d0b53131b527f4583/ebpf/aya-ebpf/src/lib.rs#L30

Maybe this bug should go upstream to https://github.com/Byron/cargo-smart-release?

aya-ebpf-macros/Cargo.toml
```diff
@@ -16,6 +16,3 @@ proc-macro2 = { workspace = true }
 proc-macro-error = { workspace = true }
 quote = { workspace = true }
 syn = { workspace = true, default-features = true, features = ["full"] }
-
-[dev-dependencies]
-aya-ebpf = { path = "../ebpf/aya-ebpf", version = "^0.1.1", default-features = false }
```
Byron, line 21: Remove the `version = *` field and the problems will go away. `gitoxide` also does this, and also went through the trouble with the `version` field.

tamird: Thanks! Done.

dave-tucker (approved):
Awesome! LGTM

PR #1109, opened 2024-12-03
Use Set (instead of Map) to hold links

tamird:
This uses hashbrown instead of std because the latter relies on Borrow
which requires a reference; hashbrown's Equivalent is more flexible.

Consider this an RFC, I'm not sure it's worth it.


vadorovsky (approved):
I will let @dave-tucker decide if it's worth merging. But for me, the code looks good, the explanation makes sense and it doesn't break anything, so +1

PR #1115, opened 2024-12-12
ci: download gen_init_cpio with authentication

tamird:
See commit message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1115)


vadorovsky (approved):
Nice (also because it's not `exec`ing curl :smile: )

PR #1116, opened 2024-12-12
ci: download gen_init_cpio with authentication

tamird:
API rate limiting seems broken - let's hope this gets around it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1116)


tamird:
Of course. Github's raw endpoint doesn't respect authorization headers.

Folded into #1115.

PR #1139, opened 2025-01-21
Fix aya-ebpf-* riscv64 build

hack3ric:
bpf_target_arch should be riscv64, but target triple starts with riscv64gc (and possibly riscv64 followed by any combination of extensions).

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1139)


vadorovsky (approved):
Thanks!

PR #1143, opened 2025-01-22
integration-test: allow llvm-objcopy env override

ignacya:
- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1143)


tamird:
It would be good to explain why you need this in the commit message.

tamird (changes requested):
Please amend the commit message to explain why you need this.

PR #1173, opened 2025-02-07
feat(aya): do not load BTF from system in EbpfLoader initialization

### LINKED ISSUE #1172: EbpfLoader should not load BTF from sys by default
# Origin
I'm writing a very simple tracepoint eBPF program, which does not need BTF.
But I notice that while the program is running, the memory usage is far more large than I expected 

Image

# Investigation
After running the memory profile, I found the program contains lots of fragmental allocations.
and it doesn't return memory to OS

Memory exactly used:
Image

Memory allocation in full lifetime:
Image

Here, the `Btf::from_sys_fs()` is even called twice.
Image

`EbpfLoader` doesn't give any chance to prevent BTF loading, it load BTF at the struct initialization.
https://github.com/aya-rs/aya/blob/ade2e2a739dbe0b1c44b02c19fed8343121dd00a/aya/src/bpf.rs#L157

and `Ebpf::load`causes the BTF double loading
https://github.com/aya-rs/aya/blob/ade2e2a739dbe0b1c44b02c19fed8343121dd00a/aya/src/bpf.rs#L879-L883

# Fixing
After I remove the `Btf::from_sys_fs()` in the `EbpfLoader`, the memory usage after calling `Ebpf::load` reduce to half.

Once I manually build `Ebpf` from `EbpfLoader`, the heap allocation is reduced to 150 KiB

Image

# More
Although not my use case, there is still a lot of room for optimization in BTF loading, such as arena (like bumpalo?) and using stack allocation instead of `Vec` abusing.

zkonge:
fix #1172

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1173)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -154,7 +154,7 @@ impl<'a> EbpfLoader<'a> {
     /// Creates a new loader instance.
     pub fn new() -> Self {
         Self {
-            btf: Btf::from_sys_fs().ok().map(Cow::Owned),
+            btf: None,
             map_pin_path: None,
             globals: HashMap::new(),
             max_entries: HashMap::new(),


dave-tucker (changes requested):
Thanks for your contribution. I'm not sure this is the right fix for the issue you reported.

For the general case, it's better to have BTF to avoid issues with program types not working, failing relocations etc... Especially given how important BTF is these days.

Did you try this: https://docs.aya-rs.dev/aya/struct.ebpfloader#method.btf

Passing in None will disable BTF.

dave-tucker:
Closing as I believe the comment I added here is the correct workaround to the issue. Feel free to re-open if you'd like to discuss further.

PR #1182, opened 2025-02-26
aya: Fix PerCpuHashMap NotFound

keskydi:
PerCpuHashMap was never returning MapError::KeyNotFound because
bpf_map_lookup_elem_per_cpu was replacing Ok(None) with Ok(Some(zeroed_value)).

Update bpf_map_lookup_elem_per_cpu to map the Option value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1182)

PR #1186, opened 2025-02-27
aya: encode bpf(2) contract

tamird:
...and generally be stricter about information we throw away.

Inspired by #1182.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,9 @@ name: aya-ci
 
 on:
   push:
-    branches:
-      - main
+    branches-ignore: 'create-pull-request/**'
 
   pull_request:
-    branches:
-      - main
 
   schedule:
     - cron: 00 4 * * *
--- a/.github/workflows/gen.yml
+++ b/.github/workflows/gen.yml
@@ -2,8 +2,7 @@ name: codegen
 
 on:
   push:
-    branches:
-      - main
+    branches-ignore: 'create-pull-request/**'
 
   schedule:
     - cron: 00 4 * * *
--- a/aya-log-parser/Cargo.toml
+++ b/aya-log-parser/Cargo.toml
@@ -11,5 +11,8 @@ edition.workspace = true
 [dependencies]
 aya-log-common = { path = "../aya-log-common", version = "^0.1.14", default-features = false }
 
+[dev-dependencies]
+assert_matches = { workspace = true }
+
 [lib]
 path = "src/lib.rs"
--- a/aya-log-parser/src/lib.rs
+++ b/aya-log-parser/src/lib.rs
@@ -128,11 +128,10 @@ pub fn parse(format_string: &str) -> Result<Vec<Fragment>, String> {
 
 #[cfg(test)]
 mod test {
+    use assert_matches::assert_matches;
+
     use super::*;
 
-    // TODO(https://github.com/rust-lang/rust-clippy/issues/13885): narrow this to just the specific
-    // strings when that doesn't trip the lint.
-    #[allow(clippy::literal_string_with_formatting_args)]
     #[test]
     fn test_parse() {
         assert_eq!(
@@ -160,9 +159,9 @@ mod test {
                 }),
             ])
         );
-        assert!(parse("foo {:}").is_err());
-        assert!(parse("foo { bar").is_err());
-        assert!(parse("foo } bar").is_err());
-        assert!(parse("foo { bar }").is_err());
+        assert_matches!(parse("foo {:}"), Err(_));
+        assert_matches!(parse("foo { bar"), Err(_));
+        assert_matches!(parse("foo } bar"), Err(_));
+        assert_matches!(parse("foo { bar }"), Err(_));
     }
 }
--- a/aya-log/src/lib.rs
+++ b/aya-log/src/lib.rs
@@ -468,12 +468,12 @@ fn log_buf(mut buf: &[u8], logger: &dyn Log) -> Result<(), ()> {
     let mut line = None;
     let mut num_args = None;
 
-    for _ in 0..LOG_FIELDS {
+    for () in std::iter::repeat_n((), LOG_FIELDS) {
         let (RecordFieldWrapper(tag), value, rest) = try_read(buf)?;
 
         match tag {
             RecordField::Target => {
-                target = Some(str::from_utf8(value).map_err(|_| ())?);
+                target = Some(str::from_utf8(value).map_err(|std::str::Utf8Error { .. }| ())?);
             }
             RecordField::Level => {
                 level = Some({
@@ -488,16 +488,24 @@ fn log_buf(mut buf: &[u8], logger: &dyn Log) -> Result<(), ()> {
                 })
             }
             RecordField::Module => {
-                module = Some(str::from_utf8(value).map_err(|_| ())?);
+                module = Some(str::from_utf8(value).map_err(|std::str::Utf8Error { .. }| ())?);
             }
             RecordField::File => {
-                file = Some(str::from_utf8(value).map_err(|_| ())?);
+                file = Some(str::from_utf8(value).map_err(|std::str::Utf8Error { .. }| ())?);
             }
             RecordField::Line => {
-                line = Some(u32::from_ne_bytes(value.try_into().map_err(|_| ())?));
+                line = Some(u32::from_ne_bytes(
+                    value
+                        .try_into()
+                        .map_err(|std::array::TryFromSliceError { .. }| ())?,
+                ));
             }
             RecordField::NumArgs => {
-                num_args = Some(usize::from_ne_bytes(value.try_into().map_err(|_| ())?));
+                num_args = Some(usize::from_ne_bytes(
+                    value
+                        .try_into()
+                        .map_err(|std::array::TryFromSliceError { .. }| ())?,
+                ));
             }
         }
 
@@ -506,7 +514,7 @@ fn log_buf(mut buf: &[u8], logger: &dyn Log) -> Result<(), ()> {
 
     let mut full_log_msg = String::new();
     let mut last_hint: Option<DisplayHintWrapper> = None;
-    for _ in 0..num_args.ok_or(())? {
+    for () in std::iter::repeat_n((), num_args.ok_or(())?) {
         let (ArgumentWrapper(tag), value, rest) = try_read(buf)?;
 
         match tag {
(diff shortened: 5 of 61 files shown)


dave-tucker (changes requested):
this is great. thank you. some minor comments. the CI one is blocking IMO but the rest are mainly nits.

> Reviewed 2 of 2 files at r1, 9 of 9 files at r2, 39 of 40 files at r3.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186#-:-OKCY8l465648HqF_6os:b-vt5yw3)* status: 40 of 56 files reviewed, 6 unresolved discussions (waiting on @alessandrod and @vadorovsky)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHF1CDvAy1VDyK8Kx:-OKCHF1CDvAy1VDyK8Ky:b-uouvh8):*
Except when it isn't. We're limited in free GH action minutes and ISTR we have been historically very close to that limit in the past.
___
*[`.github/workflows/gen.yml` line 5 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCIkBg7zpzKre5JP8O:-OKCIkBg7zpzKre5JP8P:b-l8q7tg) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/gen.yml#L5)):*
> ```YAML
> on:
>   push:
>     branches:
> ```

I'm unsure what benefit there is to running the codegen job against other branches?
Isn't this just going to create a lot of noise?
___
*[`aya/src/sys/bpf.rs` line 1090 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCS6ZfAZsPYJdbRcAu:-OKCS6ZfAZsPYJdbRcAv:bq3k7jr) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/sys/bpf.rs#L1090)):*
> ```XML
> }
> 
> fn sys_bpf_unit(cmd: bpf_cmd, attr: &mut bpf_attr) -> io::Result<()> {
> ```

since we have `fd_sys_bpf` I suppose this would be `unit_sys_bpf` although I'm not really keen on the naming. while it would be a slightly bigger change, did you consider using functions we could use for mapping instead - perhaps `sysresult_to_fd` and `sysresult_to_unit`

```
sys_bpf(...).map(sysresult_to_fd)
```
___
*[`aya/src/maps/lpm_trie.rs` line 177 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCOUAE5ttKbszDORPI:-OKCOUAE5ttKbszDORPJ:bisjj9k) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/maps/lpm_trie.rs#L177)):*
> ```XML
>                 io_error,
>             })
>             .map_err(Into::into)
> ```

here an everywhere else we call `.map_err(Into::into)`. can we call `.into()` in the `.map_err` closure above to avoid calling `.map_err` twice?
___
*[`.github/workflows/ci.yml` line 6 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHdIjBmocCvTy0SLq:-OKCHdIkDK5RJJnUt3Au:byoho59) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/ci.yml#L6)):*
> ```YAML
>   push:
>     branches:
>       - main
> ```

You need to exclude branches that are used by automation to create pull requests since it's wasteful to test them twice (once for the push to branch, once for the PR). Bear in mind we're also limited to how many runners can be used concurrently, so setting off many jobs at the same time isn't ideal either.
___
*[`aya/src/maps/bloom_filter.rs` line 137 at r2](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCLnMy2hICBfif_WtQ:-OKCLnMy2hICBfif_WtR:buxi270) ([raw file](https://github.com/aya-rs/aya/blob/b118e2a219f3ba7c2edb3665a6b2af5ff237b507/aya/src/maps/bloom_filter.rs#L137)):*
> ```XML
>         let map = new_map(new_obj_map());
> 
>         let _: BloomFilter<_, u32> = BloomFilter::new(&map).unwrap();
> ```

Do we really need the `let _ `?
This would be more compact as `BloomFilter::<_, u32>::new(&map).unwrap()`

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186)* status: 40 of 56 files reviewed, 6 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @vadorovsky)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHF1CDvAy1VDyK8Kx:-OKCYa3lCcmL-Y848OOX:b-qa7rai):*

Except when it isn't. We're limited in free GH action minutes and ISTR we have been historically very close to that limit in the past.
</blockquote></details>

I can reframe the rationale as "if I'm pushing a branch I probably want CI" which was really what happened here. WDYT?
___
*[`.github/workflows/ci.yml` line 6 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHdIjBmocCvTy0SLq:-OKCYnMH1uT_pmHwx9YT:bsvjn8t) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/ci.yml#L6)):*

You need to exclude branches that are used by automation to create pull requests since it's wasteful to test them twice (once for the push to branch, once for the PR). Bear in mind we're also limited to how many runners can be used concurrently, so setting off many jobs at the same time isn't ideal either.
</blockquote></details>

It seems to me that this is a niche case, automation generates these branches something like once a week on average, or less. We had 2 this week.
___
*[`.github/workflows/gen.yml` line 5 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCIkBg7zpzKre5JP8O:-OKCZ23PBYR6pFWjxKOD:b-s7exr8) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/gen.yml#L5)):*

I'm unsure what benefit there is to running the codegen job against other branches?
Isn't this just going to create a lot of noise?
</blockquote></details>

You tell me. FWIW I always push my branches to my fork, which mitigates all the concerns about waste counting against aya's quota.
___
*[`aya/src/maps/bloom_filter.rs` line 137 at r2](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCLnMy2hICBfif_WtQ:-OKCZB8ECxPF2TgZMa2e:b-6b79nm) ([raw file](https://github.com/aya-rs/aya/blob/b118e2a219f3ba7c2edb3665a6b2af5ff237b507/aya/src/maps/bloom_filter.rs#L137)):*

Do we really need the `let _ `?
This would be more compact as `BloomFilter::<_, u32>::new(&map).unwrap()`
</blockquote></details>

But what if the return type changes? In the end state I'd like to enable https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-results which would require this.
___
*[`aya/src/maps/lpm_trie.rs` line 177 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCOUAE5ttKbszDORPI:-OKCZLEBDXRyYluB0D4f:b-w1m8xv) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/maps/lpm_trie.rs#L177)):*

here an everywhere else we call `.map_err(Into::into)`. can we call `.into()` in the `.map_err` closure above to avoid calling `.map_err` twice?
</blockquote></details>

Yes. I find this form clearer because it's separating the concerns of generating the error vs converting it to the right type. WDYT?
___
*[`aya/src/sys/bpf.rs` line 1090 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCS6ZfAZsPYJdbRcAu:-OKCZauA8HO_2HpfElBS:b-kndv5q) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/sys/bpf.rs#L1090)):*

since we have `fd_sys_bpf` I suppose this would be `unit_sys_bpf` although I'm not really keen on the naming. while it would be a slightly bigger change, did you consider using functions we could use for mapping instead - perhaps `sysresult_to_fd` and `sysresult_to_unit`

```
sys_bpf(...).map(sysresult_to_fd)
```
</blockquote></details>

I hadn't, and we could. Is that better?

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186)* status: 2 of 61 files reviewed, 5 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @vadorovsky)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHF1CDvAy1VDyK8Kx:-OKGvBCt0RROmOkri_lK:b-qj92pf):*

I can reframe the rationale as "if I'm pushing a branch I probably want CI" which was really what happened here. WDYT?
</blockquote></details>

Reworded.
___
*[`aya/src/sys/bpf.rs` line 1090 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCS6ZfAZsPYJdbRcAu:-OKGzUiiASTw_Dr9t4Wz:bs9maxf) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/sys/bpf.rs#L1090)):*

I hadn't, and we could. Is that better?
</blockquote></details>

Renamed to `unit_sys_bpf`.

dave-tucker (changes requested):
> Reviewed 60 of 60 files at r12, 9 of 9 files at r13.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186)* status: 5 of 61 files reviewed, 2 unresolved discussions (waiting on @alessandrod, @tamird, and @vadorovsky)

___
*[`.github/workflows/ci.yml` line 6 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHdIjBmocCvTy0SLq:-OKRCnSoCoCKT3ZOtmHW:b-pau9j6) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/ci.yml#L6)):*

It seems to me that this is a niche case, automation generates these branches something like once a week on average, or less. We had 2 this week.
</blockquote></details>

This seems reasonable no?
```yaml
on:
  push:
    branches-ignore:
       - 'create-pull-request/**'
```
It achieves what you want (runs on branches) with the exception of those create by automation for which PRs are going to be opened anyway.
___
*[`.github/workflows/gen.yml` line 5 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCIkBg7zpzKre5JP8O:-OKR7oYZ8LryKfygJlaa:bwtnd8t) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/gen.yml#L5)):*

You tell me. FWIW I always push my branches to my fork, which mitigates all the concerns about waste counting against aya's quota.
</blockquote></details>

This isn't about wastage on forks. The scheduled codegen workflow create a branch, which triggers the `on: push` condition, which will then trigger this workflow again.

I suppose we could remove `on: push` entirely, or use the same syntax as I proposed in the other comment to exclude branches created by automation.
___
*[`aya/src/maps/lpm_trie.rs` line 177 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCOUAE5ttKbszDORPI:-OKRANCW0oXwlpVetF2e:b-3sdtpx) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/maps/lpm_trie.rs#L177)):*

Yes. I find this form clearer because it's separating the concerns of generating the error vs converting it to the right type. WDYT?
</blockquote></details>

Initially I found it more (too?) verbose, but given we are nesting errors this way I'm not against it.
If you find it clearer, that's fine by me.
___
*[`aya/src/sys/bpf.rs` line 1090 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCS6ZfAZsPYJdbRcAu:-OKRArSu0KECoq_D0-s0:b-7gxv85) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/sys/bpf.rs#L1090)):*

Renamed to `unit_sys_bpf`.
</blockquote></details>

I feel it's better given each wrapper is just calling `sys_bpf(...).map(...)` anyway, and we don't gain anything by hiding that as an implementation detail of a function.

I would also not be surprised if we find out later that there are exceptions to the rules in man bpf(2) requiring slightly different behaviour in some cases :wink: 

I don't feel strongly enough about it to block this PR so happy to resolve the discussion. I can always follow up with this change later and we can see if it is actually any better.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186)* status: 2 of 61 files reviewed, 2 unresolved discussions (waiting on @alessandrod, @dave-tucker, and @vadorovsky)

___
*[`.github/workflows/ci.yml` line 6 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCHdIjBmocCvTy0SLq:-OKRYVyh7aFMrjR2SzpA:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/ci.yml#L6)):*

This seems reasonable no?
```yaml
on:
  push:
    branches-ignore:
       - 'create-pull-request/**'
```
It achieves what you want (runs on branches) with the exception of those create by automation for which PRs are going to be opened anyway.
</blockquote></details>

Done.
___
*[`.github/workflows/gen.yml` line 5 at r1](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCIkBg7zpzKre5JP8O:-OKRYWkrECpYBfXMtXqk:b5hlfp0) ([raw file](https://github.com/aya-rs/aya/blob/39e40ba5c7fe027486ec79c8b411d169660acf12/.github/workflows/gen.yml#L5)):*

This isn't about wastage on forks. The scheduled codegen workflow create a branch, which triggers the `on: push` condition, which will then trigger this workflow again.

I suppose we could remove `on: push` entirely, or use the same syntax as I proposed in the other comment to exclude branches created by automation.
</blockquote></details>

Sounds good - went with the other syntax you suggested here and there.
___
*[`aya/src/sys/bpf.rs` line 1090 at r3](https://reviewable.io/reviews/aya-rs/aya/1186#-OKCS6ZfAZsPYJdbRcAu:-OKRYgzeDbTqd9GsrW13:bdx6l9e) ([raw file](https://github.com/aya-rs/aya/blob/e94d28f249e79abf1926b72095737904ba1c011f/aya/src/sys/bpf.rs#L1090)):*

I feel it's better given each wrapper is just calling `sys_bpf(...).map(...)` anyway, and we don't gain anything by hiding that as an implementation detail of a function.

I would also not be surprised if we find out later that there are exceptions to the rules in man bpf(2) requiring slightly different behaviour in some cases :wink: 

I don't feel strongly enough about it to block this PR so happy to resolve the discussion. I can always follow up with this change later and we can see if it is actually any better.
</blockquote></details>

There are 15 callers, so I'm inclined to leave it as-is for now. As you say, can always change later. Thanks!

dave-tucker (approved):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1186)* status: 0 of 61 files reviewed, all discussions resolved (waiting on @alessandrod and @vadorovsky)

PR #1187, opened 2025-03-02
Add bounds check on `SkBuff::load_bytes`

djahandarie:
There have been some reports on the Discord that people run into a verifier issue with use of `load_bytes`:
* https://discord.com/channels/855676609003651072/1307248077744377886/1307248077744377886
* https://discord.com/channels/855676609003651072/1301565441256198225/1301565441256198225
* https://discord.com/channels/855676609003651072/855676609003651075/1345517171370885160 (mine)

It seems to be that it thinks it's possible that the length could be zero. At first I tried resolving it with a `len !=0` check but that didn't work. Adding `.max(1)` did work, but that's arguably not the right semantics as it should probably error if the program really passed in 0, not just arbitrarily read one byte.

alessandrod told me about the existence of `check_bounds_signed`, which I've used to add a check here. This seems to also work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1187)

### DIFF
--- a/ebpf/aya-ebpf/src/programs/sk_buff.rs
+++ b/ebpf/aya-ebpf/src/programs/sk_buff.rs
@@ -10,7 +10,7 @@ use aya_ebpf_bindings::helpers::{
 };
 use aya_ebpf_cty::c_long;
 
-use crate::{bindings::__sk_buff, EbpfContext};
+use crate::{bindings::__sk_buff, check_bounds_signed, EbpfContext};
 
 pub struct SkBuff {
     pub skb: *mut __sk_buff,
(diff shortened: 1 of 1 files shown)


tamird:
Looks like voodoo to me, will let @alessandrod have a look.

alessandrod:
as per discord, can we add an integration test please

ebpf/aya-ebpf/src/programs/sk_buff.rs
```diff
@@ -90,6 +90,10 @@ impl SkBuff {
         let len = usize::try_from(self.len()).map_err(|core::num::TryFromIntError { .. }| -1)?;
         let len = len.checked_sub(offset).ok_or(-1)?;
         let len = len.min(dst.len());
+        let in_bounds = check_bounds_signed(len as c_long, 0, dst.len() as c_long + 1);
```
dave-tucker, line 93: This can be done without a let binding.

```suggestion
        if !check_bounds_signed(len as c_long, 0, dst.len() as c_long + 1) {
            return Err(-1);
        }
```

Also why `dst.len() as c_long + 1` - surely that's out of bounds?

RyanAlameddine: `check_bounds_signed` is inclusive on both sides, so this code is wrong. Pushing a new PR with the correct fix and an integration test soon.

RyanAlameddine:
I believe this is resolved here: https://github.com/aya-rs/aya/pull/1218

dave-tucker:
Closing in favour of #1218

PR #1194, opened 2025-03-03
Bump edition to 2024

tamird:
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1194)


Cargo.toml
```diff
@@ -52,7 +52,8 @@ authors = ["Aya Contributors"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/aya-rs/aya"
 homepage = "https://aya-rs.dev"
-edition = "2021"
+rust-version = "1.85.0"
```
dave-tucker, line 55: I know we had talked about setting an MSRV in the past, but are there specific features that drive the need for this (OwnedFd was one that came to mind)?
Adding the rationale for the version requirement in the README would be useful so we can direct users there if they have questions.

ebpf/aya-ebpf/src/helpers.rs
```diff
@@ -375,7 +364,7 @@ pub unsafe fn bpf_probe_read_user_str(src: *const u8, dest: &mut [u8]) -> Result
 /// [core::str::from_utf8_unchecked]:
 ///
 /// ```no_run
-/// # #![expect(dead_code)]
+/// # #![expect(static_mut_refs)]
```
dave-tucker, line 367: This isn't the fix. Maps haven't needed to be `static mut` for a long time.
The map in the examples can be `static`

tamird: Thanks for pointing that out! Done.

Cargo.toml
dave-tucker: Yes I know that. But this is a breaking change for users, so please update `BREAKING_CHANGES.md` and elaborate a little more on what we gain by moving to the 2024 edition there.

tamird: Done.

aya/BREAKING-CHANGES.md
```diff
@@ -17,6 +20,15 @@ history and changelog. We also tag PRs on github with a [breaking change] label.
   - BTF types have moved to the `aya-obj` crate.
   - `aya::PerfEvent::attach` and `detach` signatures have changed.
 
+## v0.13.0
```
dave-tucker, line 23: Same here.

dave-tucker (changes requested):
Last nit, but otherwise this LGTM, thanks for taking this on.

aya/BREAKING-CHANGES.md
```diff
@@ -8,6 +8,9 @@ history and changelog. We also tag PRs on github with a [breaking change] label.
 
 ## Summary
 
+- [v0.13.0](#v0130)
```
dave-tucker, line 11:
```suggestion
- [Unreleased](#unreleased)
```

We're already at v0.13.1 so next version is likely to be 0.14.0

tamird: Done.

aya/BREAKING-CHANGES.md
tamird: Done.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1194)* status: 0 of 212 files reviewed, 4 unresolved discussions (waiting on @dave-tucker)

dave-tucker (approved):
LGTM

PR #1196, opened 2025-03-03
aya: remove panics on indeterminate kernel version

tamird:
Cache the current kernel version in a thread-local while I'm here.

Closes https://github.com/aya-rs/aya/issues/1024.
Closes https://github.com/aya-rs/aya/pull/1042.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1196)

### DIFF
--- a/aya/src/sys/bpf.rs
+++ b/aya/src/sys/bpf.rs
@@ -151,7 +151,7 @@ pub(crate) fn bpf_load_program(
 
     if let Some(name) = &aya_attr.name {
         let name_bytes = name.to_bytes();
-        let len = cmp::min(name_bytes.len(), u.prog_name.len());
+        let len = cmp::min(name_bytes.len(), u.prog_name.len() - 1); // Ensure NULL termination.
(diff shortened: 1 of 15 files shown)


aya/src/sys/bpf.rs
```diff
@@ -151,7 +151,7 @@ pub(crate) fn bpf_load_program(
 
     if let Some(name) = &aya_attr.name {
         let name_bytes = name.to_bytes();
-        let len = cmp::min(name_bytes.len(), u.prog_name.len());
+        let len = cmp::min(name_bytes.len(), u.prog_name.len() - 1); // Ensure NULL termination.
```
alessandrod, line 153: I don't think that TASK_COMM has to be NULL terminated?

alessandrod (approved):

aya/src/sys/bpf.rs
tamird: I tried that but got integration test failures :(

PR #1200, opened 2025-03-04
Improve AF_XDP test

tamird:
This is #846. I'm recreating it because I am unable to push changes to that PR
due to branch protection (the author created the PR using their main branch).

cc @arctic-alpaca

Closes #846.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1200)

### DIFF
--- a/ebpf/aya-ebpf/src/programs/xdp.rs
+++ b/ebpf/aya-ebpf/src/programs/xdp.rs
@@ -21,12 +21,6 @@ impl XdpContext {
         unsafe { (*self.ctx).data_end as usize }
     }
 
-    /// Return the interface of the index as determined by the OS.
-    #[inline]
-    pub fn ingress_ifindex(&self) -> usize {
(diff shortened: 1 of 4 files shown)


ebpf/aya-ebpf/src/programs/xdp.rs
```diff
@@ -38,6 +32,18 @@ impl XdpContext {
     pub fn metadata_end(&self) -> usize {
         self.data()
     }
+
```
dave-tucker, line 35: Not sure why this moved?

dave-tucker (approved):
LGTM. Small nit on the move of the ingress_ifindex function

ebpf/aya-ebpf/src/programs/xdp.rs
tamird: I put it in the same order as the fields in `ctx`

PR #1204, opened 2025-03-04
fix rustfmt (again)

tamird:
- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1204)


tamird:
This finally doesn't generate a formatting diff. I tried using a macro to define all the generated modules but then rustfmt is unable to follow them. Instead I always declare all the architecture-specific modules but only reexport the selected one.

PR #1208, opened 2025-03-06
xtask: Use reqwest to fetch `gen_init_cpio.c`

heiher:
Since `octorust` depends on an old version of `ring`, some newer Rust targets cannot compile with it. This patch switches to using `reqwest` to fetch `gen_init_cpio.c` to make it work on more targets.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1208)

### DIFF
--- a/xtask/src/run.rs
+++ b/xtask/src/run.rs
@@ -10,7 +10,6 @@ use std::{
 };
 
 use anyhow::{Context as _, Result, anyhow, bail};
-use base64::engine::Engine as _;
 use cargo_metadata::{Artifact, CompilerMessage, Message, Target};
 use clap::Parser;
 use xtask::{AYA_BUILD_INTEGRATION_BPF, Errors};
@@ -28,12 +27,6 @@ enum Environment {
         /// The cache directory in which to store intermediate artifacts.
         #[clap(long)]
         cache_dir: PathBuf,
-
-        /// The Github API token to use if network requests to Github are made.
-        ///
-        /// This may be required if Github rate limits are exceeded.
-        #[clap(long)]
-        github_api_token: Option<String>,
 
         /// The kernel images to use.
         ///
@@ -180,7 +173,6 @@ pub fn run(opts: Options) -> Result<()> {
         }
         Environment::VM {
             cache_dir,
-            github_api_token,
             kernel_image,
         } => {
             // The user has asked us to run the tests on a VM. This is involved; strap in.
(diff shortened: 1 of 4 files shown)


xtask/src/run.rs
```diff
@@ -206,31 +198,17 @@ pub fn run(opts: Options) -> Result<()> {
...
-                let client = octorust::Client::new(
-                    String::from("aya-xtask-integration-test-run"),
-                    github_api_token.map(octorust::auth::Credentials::Token),
-                )?;
-                let octorust::Response {
-                    status: _,
-                    headers: _,
-                    body: octorust::types::ContentFile { mut content, .. },
-                } = tokio::runtime::Builder::new_current_thread()
+                let content = tokio::runtime::Builder::new_current_thread()
```
dave-tucker, line 201: I'd rather use: https://docs.rs/reqwest/latest/reqwest/blocking/index.html

xtask/src/run.rs
```diff
@@ -29,12 +28,6 @@ enum Environment {
         #[clap(long)]
         cache_dir: PathBuf,
 
-        /// The Github API token to use if network requests to Github are made.
-        ///
-        /// This may be required if Github rate limits are exceeded.
-        #[clap(long)]
-        github_api_token: Option<String>,
```
dave-tucker, line 36: This served a purpose. We frequently were getting rate limited in CI. Please reinstate it.

xtask/src/run.rs
```diff
@@ -206,31 +198,17 @@ pub fn run(opts: Options) -> Result<()> {
...
                     .build()
                     .unwrap()
-                    .block_on(client.repos().get_content_file(
-                        "torvalds",
-                        "linux",
-                        "usr/gen_init_cpio.c",
-                        "master",
-                    ))
+                    .block_on(async {
+                        reqwest::get("https://github.com/torvalds/linux/raw/refs/heads/master/usr/gen_init_cpio.c")
```
dave-tucker, line 206: You'll want to use https://docs.rs/reqwest/latest/reqwest/blocking/struct.Client.html
so we can set headers.

[This doc](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28) suggests it would be wise to:

1. Set the header for the expect API version
2. Set the auth header if `--github-api-token` was provided

Furthermore, I would suggest using the same API endpoint as `octorust` was using due to the aforementioned issues with rate limiting.

https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28

You'll then want to parse the reponse with [`serde_json`](https://docs.rs/serde_json/latest/serde_json/#operating-on-untyped-json-values)

dave-tucker (changes requested):
I understand the motivation that's driving this change, however, I'd rather keep functionality the same as the existing solution with `octorust`. See comments on how I feel this can be accomplished.

tamird:
Alternatively we could wait for https://github.com/oxidecomputer/third-party-api-clients/pull/99.

heiher:
Thanks for your review and comments. It looks like `octorust` may upgrade its dependencies next week, so we can still use the current implementation.

PR #1211, opened 2025-03-06
programs: allow program names to be static strings

tamird:
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1211)


dave-tucker (approved):
Thanks @tamird, much nicer than having to use `.to_string()`.

PR #1224, opened 2025-03-20
chore(*): set clippy unused_trait_names = warn

dave-tucker:
We have previously tried to import traits anonymously where possible but enforcing this manually was hard.

Since Rust 1.83 clippy can now enforce this for us.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1224)


tamird:
Let's mark it warn, we deny warnings in CI. No need to make this annoying in development.

tamird (approved):

PR #1225, opened 2025-03-21
chore(aya): Fix clippy unused cfg_attr

dave-tucker:
clippy complains that cfg_attr is applied to the macro invocation and therefore will not be expanded. Attempted to fix this (and clean up macro usage) by adding a meta matcher to these macros.

However, evaluating the expanded output shows that cfg_attr gets ignored?

```rust
 #[cfg(any(feature = "async_tokio", feature = "async_std"))]
    impl<'a> TryFrom<&'a mut Map> for AsyncPerfEventArray<&'a mut MapData> {
        type Error = MapError;
        fn try_from(map: &'a mut Map) -> Result<Self, Self::Error> {
            match map {
                Map::PerfEventArray(map_data) => Self::new(map_data),
                map =>
                    Err(MapError::InvalidMapType { map_type: map.map_type() }),
            }
        }
    }
```

This suppresses the clippy error for now, and the cfg_attr is only for docs so it's not a big issue, but it would be nice to get this fixed someday.

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -438,28 +438,30 @@ macro_rules! impl_try_from_map {
     // rather than the repeated idents used later because the macro language does not allow one
     // repetition to be pasted inside another.
     ($ty_param:tt {
-        $($ty:ident $(from $($variant:ident)|+)?),+ $(,)?
+        $($(#[$meta:meta])* $ty:ident $(from $($variant:ident)|+)?),+ $(,)?
     }) => {
-        $(impl_try_from_map!(<$ty_param> $ty $(from $($variant)|+)?);)+
+        $(impl_try_from_map!($(#[$meta])* <$ty_param> $ty $(from $($variant)|+)?);)+
     };
     // Add the "from $variant" using $ty as the default if it is missing.
-    (<$ty_param:tt> $ty:ident) => {
-        impl_try_from_map!(<$ty_param> $ty from $ty);
+    ($(#[$meta:meta])* <$ty_param:tt> $ty:ident) => {
+        impl_try_from_map!($(#[$meta])* <$ty_param> $ty from $ty);
     };
     // Dispatch for each of the lifetimes.
     (
-        <($($ty_param:ident),*)> $ty:ident from $($variant:ident)|+
+        $(#[$meta:meta])* <($($ty_param:ident),*)> $ty:ident from $($variant:ident)|+
     ) => {
-        impl_try_from_map!(<'a> ($($ty_param),*) $ty from $($variant)|+);
-        impl_try_from_map!(<'a mut> ($($ty_param),*) $ty from $($variant)|+);
-        impl_try_from_map!(<> ($($ty_param),*) $ty from $($variant)|+);
+        impl_try_from_map!($(#[$meta])* <'a> ($($ty_param),*) $ty from $($variant)|+);
+        impl_try_from_map!($(#[$meta])* <'a mut> ($($ty_param),*) $ty from $($variant)|+);
+        impl_try_from_map!($(#[$meta])* <> ($($ty_param),*) $ty from $($variant)|+);
     };
     // An individual impl.
     (
+        $(#[$meta:meta])*
         <$($l:lifetime $($m:ident)?)?>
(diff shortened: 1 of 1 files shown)


tamird (approved):
Upstream is fixing the false positive in https://github.com/rust-lang/rust/pull/138794, but this looks good anyway.

aya/src/maps/mod.rs
```diff
@@ -487,11 +489,8 @@ impl_try_from_map!(() {
     SockMap,
     StackTraceMap,
     XskMap,
-});
-
-#[cfg(any(feature = "async_tokio", feature = "async_std"))]
-#[cfg_attr(docsrs, doc(cfg(any(feature = "async_tokio", feature = "async_std"))))]
-impl_try_from_map!(() {
+    #[cfg(any(feature = "async_tokio", feature = "async_std"))]
+    #[cfg_attr(docsrs, doc(cfg(any(feature = "async_tokio", feature = "async_std"))))] // FIXME: This doesn't get expanded
```
tamird, line 493: Does this work before this change? I think it does.

tamird (changes requested):

aya/src/maps/mod.rs
dave-tucker: It doesn't AFAICT. See: https://docs.aya-rs.dev/aya/maps/perf/struct.asyncperfeventarray#impl-TryFrom%3C%26Map%3E-for-AsyncPerfEventArray%3C%26MapData%3E
If it worked, that function would be annotated `Available on crate features async_tokio or async_std only.`

dave-tucker: Verified this change with `cargo expand --all-features -p aya aya::maps`:

**before**:
```rust
    impl<'a> TryFrom<&'a Map> for AsyncPerfEventArray<&'a MapData> {
        type Error = MapError;
        fn try_from(map: &'a Map) -> Result<Self, Self::Error> {
            match map {
                Map::PerfEventArray(map_data) => Self::new(map_data),
                map =>
                    Err(MapError::InvalidMapType { map_type: map.map_type() }),
            }
        }
    }
```

**after**:
```rust
    #[cfg(any(feature = "async_tokio", feature = "async_std"))]
    impl<'a> TryFrom<&'a Map> for AsyncPerfEventArray<&'a MapData> {
        type Error = MapError;
        fn try_from(map: &'a Map) -> Result<Self, Self::Error> {
            match map {
                Map::PerfEventArray(map_data) => Self::new(map_data),
                map =>
                    Err(MapError::InvalidMapType { map_type: map.map_type() }),
            }
        }
    }
 ```

Comment (dave-tucker) [reply aya/src/maps/mod.rs]: It does not work now.

I've done some more digging. If you run this command on main you get a warning:

```console
$ RUSTFLAGS='--cfg docsrs' cargo expand --all-features -p aya aya::maps > expanded.rs
    Checking aya v0.13.1 (/home/dave/dev/aya-rs/aya/aya)
warning: unused attribute `doc`
   --> aya/src/maps/mod.rs:493:20
    |
493 | #[cfg_attr(docsrs, doc(cfg(any(feature = "async_tokio", feature = "async_std"))))]
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the built-in attribute `doc` will be ignored, since it's applied to the macro invocation `impl_try_from_map`
   --> aya/src/maps/mod.rs:494:1
    |
494 | impl_try_from_map!(() {
    | ^^^^^^^^^^^^^^^^^
    = note: `#[warn(unused_attributes)]` on by default
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
```

Running this command with this commit gives no warnings.
And In the after section in the comment above, you can see that the expanded code has the correct attributes.
```
    #[cfg(any(feature = "async_tokio", feature = "async_std"))]
    #[doc(cfg(any(feature = "async_tokio", feature = "async_std")))]
```

tamird: Sorry by now I meant after this change. Sounds like it does work.

tamird (approved):

Issue #1226, opened 2025-03-23
Allow pinning nightly version in `aya-build`


thomaseizinger:
I'd like to pin the specific nightly version that is used for compiling my eBPF program with `aya-build`: https://github.com/aya-rs/aya/blob/9eecbe9d0e9dc1fdbbc87d41512d4202e26d4687/aya-build/src/lib.rs#L62

When using this in CI, the current setup will always download the latest nightly toolchain because just "nightly" is different every day so it installs a new one on every CI run.

tamird:
We could make it an optional argument. Would you like to send a PR?

thomaseizinger:
Happy to send a PR yeah.

PR #1228, opened 2025-03-24
feat(aya-build): allow setting Rust version

thomaseizinger:
Resolves: #1226

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1228)

### DIFF
--- a/aya-build/src/lib.rs
+++ b/aya-build/src/lib.rs
@@ -22,7 +22,7 @@ use cargo_metadata::{Artifact, CompilerMessage, Message, Package, Target};
 /// prevent their use for the time being.
 ///
 /// [bindeps]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html?highlight=feature#artifact-dependencies
-pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
+pub fn build_ebpf(packages: impl IntoIterator<Item = Package>, toolchain: Toolchain) -> Result<()> {
     let out_dir = env::var_os("OUT_DIR").ok_or(anyhow!("OUT_DIR not set"))?;
     let out_dir = PathBuf::from(out_dir);
 
@@ -57,9 +57,11 @@ pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
         // changes to the binaries too, which gets us the rest of the way.
         println!("cargo:rerun-if-changed={dir}");
 
-        let mut cmd = Command::new("cargo");
+        let mut cmd = Command::new("rustup");
         cmd.args([
-            "+nightly",
+            "run",
+            toolchain.as_str(),
+            "cargo",
             "build",
             "--package",
             &name,
@@ -146,4 +148,25 @@ pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
         }
     }
     Ok(())
+}
+
+/// The toolchain to use for building eBPF programs.
+#[derive(Default)]
+pub enum Toolchain<'a> {
+    /// The latest nightly toolchain i.e. `nightly`.
+    #[default]
+    Nightly,
+    /// A custom toolchain e.g. `nightly-2021-01-01`.
+    ///
+    /// The toolchain specifier is passed to `rustup run` and therefore should _not_ have a preceding `+`.
+    Custom(&'a str),
+}
+
+impl<'a> Toolchain<'a> {
+    fn as_str(&self) -> &'a str {
+        match self {
+            Toolchain::Nightly => "nightly",
+            Toolchain::Custom(toolchain) => toolchain,
+        }
+    }
 }
--- a/test/integration-test/build.rs
+++ b/test/integration-test/build.rs
@@ -182,7 +182,7 @@ fn main() -> Result<()> {
             }
         }
 
-        aya_build::build_ebpf([integration_ebpf_package])?;
+        aya_build::build_ebpf([integration_ebpf_package], aya_build::Toolchain::default())?;
     } else {
         for (src, build_btf) in C_BPF {
             let dst = out_dir.join(src).with_extension("o");
--- a/xtask/public-api/aya-build.txt
+++ b/xtask/public-api/aya-build.txt
@@ -1,3 +1,30 @@
 pub mod aya_build
 pub use aya_build::cargo_metadata
-pub fn aya_build::build_ebpf(packages: impl core::iter::traits::collect::IntoIterator<Item = cargo_metadata::Package>) -> anyhow::Result<()>
+pub enum aya_build::Toolchain<'a>
+pub aya_build::Toolchain::Custom(&'a str)
+pub aya_build::Toolchain::Nightly
+impl<'a> core::default::Default for aya_build::Toolchain<'a>
+pub fn aya_build::Toolchain<'a>::default() -> aya_build::Toolchain<'a>
+impl<'a> core::marker::Freeze for aya_build::Toolchain<'a>
+impl<'a> core::marker::Send for aya_build::Toolchain<'a>
+impl<'a> core::marker::Sync for aya_build::Toolchain<'a>
+impl<'a> core::marker::Unpin for aya_build::Toolchain<'a>
+impl<'a> core::panic::unwind_safe::RefUnwindSafe for aya_build::Toolchain<'a>
+impl<'a> core::panic::unwind_safe::UnwindSafe for aya_build::Toolchain<'a>
+impl<T, U> core::convert::Into<U> for aya_build::Toolchain<'a> where U: core::convert::From<T>
+pub fn aya_build::Toolchain<'a>::into(self) -> U
+impl<T, U> core::convert::TryFrom<U> for aya_build::Toolchain<'a> where U: core::convert::Into<T>
+pub type aya_build::Toolchain<'a>::Error = core::convert::Infallible
+pub fn aya_build::Toolchain<'a>::try_from(value: U) -> core::result::Result<T, <T as core::convert::TryFrom<U>>::Error>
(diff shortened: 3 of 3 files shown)


aya-build/src/lib.rs
```diff
@@ -22,7 +22,10 @@ use cargo_metadata::{Artifact, CompilerMessage, Message, Package, Target};
 /// prevent their use for the time being.
 ///
 /// [bindeps]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html?highlight=feature#artifact-dependencies
-pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
+pub fn build_ebpf(
+    packages: impl IntoIterator<Item = Package>,
+    nightly_version: Option<String>,
```
tamird, line 27: Why not &str?

thomaseizinger: No particular reason, can change it to a reference.

aya-build/src/lib.rs
```diff
@@ -22,7 +22,10 @@ use cargo_metadata::{Artifact, CompilerMessage, Message, Package, Target};
 /// prevent their use for the time being.
 ///
 /// [bindeps]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html?highlight=feature#artifact-dependencies
-pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
+pub fn build_ebpf(
+    packages: impl IntoIterator<Item = Package>,
+    nightly_version: Option<&str>,
```
tamird, line 27: This isn't going to work the way you want, I think. The caller is going to have to pass `+nightly-2025-03-24`, or something like that.

I think we should introduce an enum here which has a Default impl that selects latest, with a variant that takes a strict, and then we create the `+nightly-{date}` from it.

thomaseizinger: > This isn't going to work the way you want, I think. The caller is going to have to pass `+nightly-2025-03-24`, or something like that.

Yeah that is what I am doing in my repo. Why wouldn't that work?

> I think we should introduce an enum here which has a Default impl that selects latest, with a variant that takes a strict, and then we create the `+nightly-{date}` from it.

Is the effort of modelling that out worth it? Not every date maps to a valid nightly version. Not every nightly version contains necessarily all components.

I am a huge proponent of type-safe APIs in general. In this case however, I wonder if we are making life unnecessarily difficult for ourselves? It is pretty easy to read the code and figure out what you can pass here and sometimes, simple passthrough is the best option because it gives you the most control.

Also, it is part of the 2025 goals of Rust to stabilise some -Zbuild-std, meaning we might not need this API for very long (hopefully).

Can go either way though. If you have particular requirements for how you want this solved, let me know so we can reduce the number of iterations on this PR :)
I simply want to control the nightly version and would like to not have to fork aya for that.

tamird: If you just want to control the nightly, can't you alias some nightly version to "nightly" locally?

You seem to be overly focused on "this works for me" but this API is not suitable for public consumption - it isn't obvious that the caller must stick a `+` in front.

thomaseizinger: > If you just want to control the nightly, can't you alias some nightly version to "nightly" locally?

I am not aware that rustup can do that.

> You seem to be overly focused on "this works for me" but this API is not suitable for public consumption - it isn't obvious that the caller must stick a `+` in front.

I think the problem is worth solving for everyone. I think +nightly is a bad default actually that because +nightly means something different every day, making your CI non-deterministic :)

I agree that having to pass the + isn't great. Should we just call the variable nightly-date instead? Do we want to enforce in the type system that it is a valid date or is a string good enough?

tamird: We definitely want to preserve the ability to use naked "+nightly" - this is very useful for *us*, the aya developers, because we actually do want to learn of regressions as soon as they happen.

I think a string is good enough, it doesn't even have to be a date. See https://rust-lang.github.io/rustup/concepts/toolchains.html#custom-toolchains.

That's why I suggested an enum for some well-known names (the only useful one in practice will be nightly until BPF is a tier 2 platform).

thomaseizinger: > I think we should introduce an enum here which has a Default impl that selects latest, with a variant that takes a strict, and then we create the `+nightly-{date}` from it.

If we just pass `+nightly` that will basically mean latest for the definition of "latest installed". That could mean 3 months ago on a local system that never updated the nightly.

I don't think it is worth going beyond that and actually enforce the latest nightly of the day this is compiled on (also bad for reproducibility).

```rust
enum NightlyVersion {
    /// Maps to `+nightly`
    Latest,
    /// Maps to `+nightly-{date}`
    Date(String)
}
```

tamird: Why not just

```rust
#[derive(Default)]
enum Toolchain<'a> {
   #[default]
   Nightly,
   Custom(&'a str),
}
```

thomaseizinger: Or maybe:

```
PinnedDate { year: u16, month: u8, day: u8 }
```

Not sure if bringing in a full date/time library is worth it here?

thomaseizinger:
@tamird How do you handle breaking changes? Do I need to bump the crate version right away or are you going to do that later upon release?

aya-build/src/lib.rs
```diff
@@ -147,3 +147,21 @@ pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
     }
     Ok(())
 }
+
+#[derive(Default)]
+pub enum Toolchain<'a> {
+    /// Use the latest nightly toolchain to compile to eBPF kernel, e.g. `+nightly`.
```
tamird, line 153: s/e.g./i.e./

e.g. = exempli gratia = for example
i.e. = id est = that is

this should be i.e., for custom e.g. is fine

aya-build/src/lib.rs
```diff
@@ -147,3 +147,21 @@ pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
...
+    #[default]
+    Nightly,
+    /// Uses a custom toolchain to compile to eBPF kernel, e.g. `+nightly-2021-01-01`.
```
tamird, line 156: please clarify whether the leading plus is required or not. my preference is "not"

aya-build/src/lib.rs
tamird, line 153: instead of repeating "to eBPF kernel", put this part on the enum's doc comment. also consider calling it EbpfToolchain and drop the word "kernel", it doesn't mean anything int his context.

test/integration-test/build.rs
```diff
@@ -182,7 +182,7 @@ fn main() -> Result<()> {
             }
         }
 
-        aya_build::build_ebpf([integration_ebpf_package])?;
+        aya_build::build_ebpf([integration_ebpf_package], aya_build::Toolchain::Nightly)?;
```
tamird, line 185: nit: use Default::default()

aya-build/src/lib.rs
```diff
@@ -59,7 +59,7 @@ pub fn build_ebpf(packages: impl IntoIterator<Item = Package>) -> Result<()> {
 
         let mut cmd = Command::new("cargo");
         cmd.args([
-            "+nightly",
+            toolchain.as_str(),
```
tamird, line 62: let's add the + ourselves, or use `rustup run <toolchain.as_str()> cargo ...`

tamird (changes requested):
You'll need to regenerate the public API files: `CARGO_CFG_BPF_TARGET_ARCH=x86_64 cargo +nightly xtask public-api --bless --target x86_64-unknown-linux-gnu`

tamird (approved):
LGTM. @dave-tucker can you have a look? I guess we want a changelog entry, and then we'll need to update aya-template and book which build against HEAD.

dave-tucker (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1228#-:-OMQxxlRCe5tx-HL2CZb:bnfp4nl)* status: 2 of 3 files reviewed, 8 unresolved discussions (waiting on @tamird and @thomaseizinger)

dave-tucker:
@tamird ISTR that changelogs are autogenerated by cargo-release.
Breaking changes however should go in a `BREAKING-CHANGES.md` which we have to curate ourselves.
I'd recommend we merge this as-is and we can handle the version bump + breaking changes file + release ourselves.
Note: I've also noticed that you need to add `github:aya-rs/owners` as an owner on https://crates.io/crates/aya-build

test/integration-test/build.rs
tamird: why not Default::default()?

tamird (changes requested):
> Reviewed 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1228)* status: all files reviewed, 3 unresolved discussions (waiting on @dave-tucker and @thomaseizinger)

___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/aya-rs/aya/1228#-OMQx6lT7VpUTkXp3Usm:-OMT6DfZ18p4i__27cvp:b7qzpfr):*

nit. This should be `aya-build: Allow setting Rust nightly version`.
or `feat(aya-build): Allow setting Rust nightly version`
</blockquote></details>

(please fix)
___
*[`-- commits` line 13 at r5](https://reviewable.io/reviews/aya-rs/aya/1228#-OMT6G3-Dz6Bfq0Ys5nz:-OMT6G3-Dz6Bfq0Ys5o-:bz9kfj):*
please squash these commits

tamird (approved):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1228)* status: all files reviewed, 1 unresolved discussion (waiting on @dave-tucker and @thomaseizinger)

tamird:
Thanks!

PR #1248, opened 2025-04-10
Add support for kernel module split BTF

### LINKED ISSUE #300: Add support for kernel module "split" BTFs
These BTFs reference types and strings from the "base" vmlinux BTF, with most of the implementation outlined at https://lore.kernel.org/bpf/105c48d56a550af6e0008b4b5867eb51764d41c9.camel@nvidia.com/T/#m08e1ff97bbae4448b14ea9b0641d4cd60f110a9a (aka: no proper documentation).

roblabla:
The kernel splits kernel modules into dedicated BTF files, using "split BTF mode". In split BTF mode, each split BTF may reference the "base BTF" (AKA the vmlinux BTF), but may not reference each-other. The way it works is fairly simple: you can combine a base BTF and split BTF by concatenating their strings and types together, creating a working merged BTF.

This commit adds support for merging one base BTF and multiple split BTF into a single merged BTF, allowing eBPFs to manipulate structs that come from kernel modules. It also reworks `Btf::from_sys_fs` to make use of this capacity.

Fixes #300


tamird:
Thanks for the PR! I notice there aren't any new tests. Have you thought about how we might test this?

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1248)* status: 0 of 7 files reviewed, 14 unresolved discussions

___
*[`aya-obj/src/btf/btf.rs` line 274 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7GkS5b1HtpTy1lZ_:-ONV7GkS5b1HtpTy1lZa:b-itxtni) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L274)):*
> ```XML
>     pub fn merge_split_btfs(base_btf: &Btf, split_btfs: &[Btf]) -> Btf {
>         // Create a Btf from all the btfs.
>         let mut out_btf = base_btf.clone();
> ```

nit: pass `base_btf` by value, let the caller decide how to manage this cloning
___
*[`aya-obj/src/btf/btf.rs` line 276 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV8TPL2ackZxYM0_US:-ONV8TPL2ackZxYM0_UT:b-fnwtpw) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L276)):*
> ```XML
>         let mut out_btf = base_btf.clone();
> 
>         let mut rebase_info = BtfRebaseInfo {
> ```

instead of making this `mut`, consider creating rebase_info in the loop.
___
*[`aya-obj/src/btf/btf.rs` line 282 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV97Fe-m4O53nYM5YJ:-ONV97Fe-m4O53nYM5YK:b-b9zsit) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L282)):*
> ```XML
>             types_new_offset: out_btf.types.types.len() as u32,
>         };
>         for btf in split_btfs {
> ```

can you destructure this? it's quite hard to see that all the proper fields are being used



_Code quote:_
```XML
btf 
```
___
*[`aya-obj/src/btf/btf.rs` line 327 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9EZbEXhpa014JQjl:-ONV9EZbEXhpa014JQjm:bjl0tt1) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L327)):*
> ```XML
>     }
> 
>     /// Loads BTF metadata from `/sys/kernel/btf`.
> ```

please add back the trailing `/`
___
*[`aya-obj/src/btf/btf.rs` line 337 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9KHP9_PYL0Sk4_lC:-ONV9KHP9_PYL0Sk4_lD:b-k0qa8o) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L337)):*
> ```XML
>             .flatten()
>             .filter_map(|v| v.ok())
>             .filter(|v| v.file_name() != "vmlinux");
> ```

this is discarding errors. why is that OK?



_Code quote:_
```XML
        let dir_iter = std::fs::read_dir("/sys/kernel/btf")
            .ok()
            .into_iter()
            .flatten()
            .filter_map(|v| v.ok())
            .filter(|v| v.file_name() != "vmlinux");
```
___
*[`aya-obj/src/btf/btf.rs` line 341 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9QRf3tWvy8nVL1VK:-ONV9QRf3tWvy8nVL1VL:bb7r2f0) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L341)):*
> ```XML
>             match entry.file_type() {
>                 Ok(v) if !v.is_file() => continue,
>                 Err(_err) => continue,
> ```

why?
___
*[`aya-obj/src/btf/btf.rs` line 346 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9SF45FlSA0H6StjZ:-ONV9SF45FlSA0H6Stj_:bsjjuqp) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L346)):*
> ```XML
>             match Btf::parse_file(entry.path(), Endianness::default()) {
>                 Ok(v) => split_btfs.push(v),
>                 // Ignore errors - the goal is to enhance the base BTF with as
> ```

This isn't explaining why ignoring errors is ok.
___
*[`aya-obj/src/btf/btf.rs` line 352 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9WOr5VoLhUcCGiPy:-ONV9WOr5VoLhUcCGiPz:b-q443pn) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L352)):*
> ```XML
>         }
> 
>         if split_btfs.len() > 0 {
> ```

do you need this conditional? why?
___
*[`aya-obj/src/btf/types.rs` line 16 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7RdBC_B9sZcp6yH1:-ONV7RdBC_B9sZcp6yH2:bg1i7fh) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L16)):*
> ```XML
>     /// Offsets below this value are considered to be part of the "base BTF",
>     /// and as such should not be relocated.
>     pub(crate) str_rebase_from: u32,
> ```

can we use usize for all of these? it's an internal type, i think usize would be more convenient
___
*[`aya-obj/src/btf/types.rs` line 21 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7WI-BDuDay3Nhq8G:-ONV7WI03nLEaFxqn9qm:b-q3cvim) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L21)):*
> ```XML
>     /// Indices below this value are considered to be part of the "base BTF",
>     /// and as such should not be relocated.
>     pub(crate) types_rebase_from: u32,
> ```

these can probably be `pub` (rather than `pub(crate)`)
___
*[`aya-obj/src/btf/types.rs` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7dLm3kB_RslkHFGY:-ONV7dLm3kB_RslkHFGZ:bjbp6yf) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L23)):*
> ```XML
>     pub(crate) types_rebase_from: u32,
>     /// The new starting offset for strings.
>     pub(crate) str_new_offset: u32,
> ```

consider extracting an inner type and storing 2 of them: one for strings, one for types. This would reduce the duplication in the doc comments.
___
*[`aya-obj/src/btf/types.rs` line 30 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7lC09JTrM6xxsChn:-ONV7lC09JTrM6xxsCho:b-oavvw7) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L30)):*
> ```XML
> impl BtfRebaseInfo {
>     fn rebase_str(&self, str_offset: u32) -> u32 {
>         if str_offset < self.str_rebase_from {
> ```

this can be a bit pithier:

match str_offset.checked_sub(self.str_rebase_from) {
  None => str_offset,
  Some(str_offset) => self.str_new_offset + str_offset
}

this should be a method on the type I suggested extracting above. Then you can also write `let Self { rebase_from, new_offset } = self` at the top instead of writing `self.xxx`.
___
*[`aya-obj/src/btf/types.rs` line 91 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV8BM6EW-bvF8Jva2_:-ONV8BM6EW-bvF8Jva2a:brprq53) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L91)):*
> ```XML
>     }
> 
>     pub(crate) fn rebase(&self, rebase_info: &BtfRebaseInfo) -> Self {
> ```

in all these methods: could you destructure `self`?
___
*[`aya-obj/src/btf/types.rs` line 92 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV8L0j1uV5hehC1i57:-ONV8L0j1uV5hehC1i58:b-kncwfr) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L92)):*
> ```XML
> 
>     pub(crate) fn rebase(&self, rebase_info: &BtfRebaseInfo) -> Self {
>         Fwd {
> ```

can this be `Self {` instead of Fwd? here and everywhere.

roblabla (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1248)* status: 0 of 7 files reviewed, 14 unresolved discussions (waiting on @tamird)

___
*[`aya-obj/src/btf/btf.rs` line 337 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9KHP9_PYL0Sk4_lC:-ONVKwjaC7_RPen4ZcoZ:b-qlpoed) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L337)):*

this is discarding errors. why is that OK?
</blockquote></details>

So, if we successfully read `/sys/kernel/btf/vmlinux` and fail on some other module BTF, it'd kinda suck to be completely locked out of using the working BTFs just because one module doesn't work. Especially since the vast majority of eBPFs out there won't need those split BTF to load properly. So I made this function ignore the broken modules/skip the BTF it cannot read.

I can definitely change it to be stricter.
___
*[`aya-obj/src/btf/btf.rs` line 341 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9QRf3tWvy8nVL1VK:-ONVLb-lDsBWWJsKVkRd:b2nreh6) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L341)):*

why?
</blockquote></details>

Same as above.
___
*[`aya-obj/src/btf/btf.rs` line 346 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9SF45FlSA0H6StjZ:-ONVLcuE8kycmiSZuoQU:bbgbnmg) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L346)):*

This isn't explaining why ignoring errors is ok.
</blockquote></details>

Same as above, I can put the reasoning in the comment.
___
*[`aya-obj/src/btf/btf.rs` line 352 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9WOr5VoLhUcCGiPy:-ONVLhiGC1HFN2YSt9tv:b-l09s7n) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L352)):*

do you need this conditional? why?
</blockquote></details>

This is to avoid the cloning done in merge_split_btfs. It's a simple (if perhaps premature) optimization.
___
*[`aya-obj/src/btf/types.rs` line 16 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7RdBC_B9sZcp6yH1:-ONVLqZgDgb_ux-lAfTP:bmbzulv) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L16)):*

can we use usize for all of these? it's an internal type, i think usize would be more convenient
</blockquote></details>

u32 is the datatype used all over the place in `BtfType`, so using a u32 makes a whole lot more sense here. If I use usize, then I'll end up needing to do casting all over the place, which would definitely be less convenient.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1248)* status: 0 of 7 files reviewed, 13 unresolved discussions (waiting on @roblabla)

___
*[`aya-obj/src/btf/btf.rs` line 337 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9KHP9_PYL0Sk4_lC:-ONVQ3Yw2z68jPN68rAq:b-p0ep0w) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L337)):*
<details><summary><i>Previously, roblabla (Robin Lambertz) wrote…</i></summary><blockquote>

So, if we successfully read `/sys/kernel/btf/vmlinux` and fail on some other module BTF, it'd kinda suck to be completely locked out of using the working BTFs just because one module doesn't work. Especially since the vast majority of eBPFs out there won't need those split BTF to load properly. So I made this function ignore the broken modules/skip the BTF it cannot read.

I can definitely change it to be stricter.
</blockquote></details>

I think we should be stricter. There may be certain kinds of errors we'd want to allow and silence, but the default position is that we don't discard errors.
___
*[`aya-obj/src/btf/btf.rs` line 352 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9WOr5VoLhUcCGiPy:-ONVQBpR0hh64r5pYByW:bftphcf) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L352)):*
<details><summary><i>Previously, roblabla (Robin Lambertz) wrote…</i></summary><blockquote>

This is to avoid the cloning done in merge_split_btfs. It's a simple (if perhaps premature) optimization.
</blockquote></details>

See my other comment; if you make that function take its argument by value you'd avoid needing this special case.
___
*[`aya-obj/src/btf/types.rs` line 16 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7RdBC_B9sZcp6yH1:-ONVQ1me13MNwxRdewW-:b2nhktp) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L16)):*
<details><summary><i>Previously, roblabla (Robin Lambertz) wrote…</i></summary><blockquote>

u32 is the datatype used all over the place in `BtfType`, so using a u32 makes a whole lot more sense here. If I use usize, then I'll end up needing to do casting all over the place, which would definitely be less convenient.
</blockquote></details>

OK.

roblabla (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1248)* status: 0 of 7 files reviewed, 12 unresolved discussions (waiting on @tamird)

___
*[`aya-obj/src/btf/btf.rs` line 276 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV8TPL2ackZxYM0_US:-ONVYkfQ2y5hcZQ-misO:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L276)):*

instead of making this `mut`, consider creating rebase_info in the loop.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 282 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV97Fe-m4O53nYM5YJ:-ONVYm3EEGJTIdWghyYF:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L282)):*

can you destructure this? it's quite hard to see that all the proper fields are being used
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 327 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9EZbEXhpa014JQjl:-ONVYrbmD3k9Guf-WQKy:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L327)):*

please add back the trailing `/`
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 337 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9KHP9_PYL0Sk4_lC:-ONVYsPK85SKKVYnW97c:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L337)):*

I think we should be stricter. There may be certain kinds of errors we'd want to allow and silence, but the default position is that we don't discard errors.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 341 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9QRf3tWvy8nVL1VK:-ONVYtXOBu4VkKxqeSDR:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L341)):*
<details><summary><i>Previously, roblabla (Robin Lambertz) wrote…</i></summary><blockquote>

Same as above.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/btf.rs` line 346 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9SF45FlSA0H6StjZ:-ONVYxjt3CoUc5YwTnAN:bxcuudp) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L346)):*
<details><summary><i>Previously, roblabla (Robin Lambertz) wrote…</i></summary><blockquote>

Same as above, I can put the reasoning in the comment.
</blockquote></details>

Done, doesn't ignore errors anymore.
___
*[`aya-obj/src/btf/btf.rs` line 352 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV9WOr5VoLhUcCGiPy:-ONVZ0Nx6p4rWwGdGT3N:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/btf.rs#L352)):*

See my other comment; if you make that function take its argument by value you'd avoid needing this special case.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/types.rs` line 21 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7WI-BDuDay3Nhq8G:-ONVZ2yh8Va05DfacW_r:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L21)):*

these can probably be `pub` (rather than `pub(crate)`)
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/types.rs` line 23 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7dLm3kB_RslkHFGY:-ONVZ3XtCbMuO9srVNgm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L23)):*

consider extracting an inner type and storing 2 of them: one for strings, one for types. This would reduce the duplication in the doc comments.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/types.rs` line 30 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV7lC09JTrM6xxsChn:-ONVZ6Q02WIxcuxwZYlY:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L30)):*

this can be a bit pithier:

match str_offset.checked_sub(self.str_rebase_from) {
  None => str_offset,
  Some(str_offset) => self.str_new_offset + str_offset
}

this should be a method on the type I suggested extracting above. Then you can also write `let Self { rebase_from, new_offset } = self` at the top instead of writing `self.xxx`.
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/types.rs` line 91 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV8BM6EW-bvF8Jva2_:-ONVZI6h4E1uXQe-lw4C:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L91)):*

in all these methods: could you destructure `self`?
</blockquote></details>

Done.
___
*[`aya-obj/src/btf/types.rs` line 92 at r2](https://reviewable.io/reviews/aya-rs/aya/1248#-ONV8L0j1uV5hehC1i57:-ONVZ9Q00G24IsjzI3xm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/c58433b0e9c5a45b6c0d0c1a798e81b9dd052de8/aya-obj/src/btf/types.rs#L92)):*

can this be `Self {` instead of Fwd? here and everywhere.
</blockquote></details>

Done.

roblabla:
> We shouldn't be loading all the module BTFs in from_sys_fs() since that's default. There's a recent issue about Aya memory usage and this will get out of hand pretty quickly. Instead we should be loading either as opt-in for specific kmods (via BpfLoader)

I feel pretty strongly that since `from_sys_fs` is the default, it should be optimized for making things "just work", and so it should instead be opt-out if the user wants to reduce memory usage. For context, what prompted me to work on this is that I found out the following code failed silently:

```
if (bpf_core_type_exists(struct btrfs_inode)) {
    // do something
} else {
    // do something else
}
```

Ergo this would always take the `else` branch, due to Aya not loading the kernel module BTFs. This is a silent failure that is pretty unintuitive and hard to debug.

Also, your mileage may vary here, but from what I can see, vmlinux is usually the biggest BTF by multiple orders of magnitude, with the module BTF being much, much smaller. On my current kernel, the `/sys/kernel/btf` folder is 5.8MiB, where vmlinux takes 4.9MiB of those. So the additional module BTF is on the order of 10%. This of course will vary from distro to distro, but I don't think an extra MiB or two of memory in exchange for making sure all eBPFs program work without fuss is a bad default ^^.

> or by detecting that we need to via checking which modules we need through kallsyms when attaching probes.

That would probably be the best of both worlds here, yes. 

> We need relocation tests to ensure that relocations are working correctly for structs etc.. that are in module BTF

Yes, I'm working on those.

> I'm also not convinced that mutating the Base BTF is the best approach either. In my branch I opted to instead create a composite type CompositeBTF which comprised of a base BTF + module BTF. That's closer to how it's implemented in libbpf

So I tried something along those lines first, but quickly gave up as I realized it required changing quite a few abstraction in the relocation code. In particular, it means that `type_id` is no longer a unique identifier of a type. And similarly, getting the strings associated with a type requires knowing where that type originated from. None of those are necessarily hard problems, but the changes ended up being more involved than I wanted, so I figured merging the BTFs back into a traditional single-file BTF was just easier.

> , but also makes it a little easier to test things out - for example, you can easily write unit tests for type resolution to show that you check module btf types before searching in the base btf.

I'm not sure I understand what this means. For starters, I don't understand why you'd want to check module BTF types before searching the base BTF? As far as I understand, you shouldn't have any conflicts here (if you did, that would cause ambiguity - what if you had conflicts betwen different module BTFs?). Generally speaking, I think testing this approach is rather straightforward: you can just create a base BTF object and multiple "split" BTFs, merge them, and make sure everything still resolves properly.

dave-tucker:
> > We shouldn't be loading all the module BTFs in from_sys_fs() since that's default. There's a recent issue about Aya memory usage and this will get out of hand pretty quickly. Instead we should be loading either as opt-in for specific kmods (via BpfLoader)
> 
> I feel pretty strongly that since `from_sys_fs` is the default, it should be optimized for making things "just work", and so it should instead be opt-out if the user wants to reduce memory usage. For context, what prompted me to work on this is that I found out the following code failed silently:
> 
> ```
> if (bpf_core_type_exists(struct btrfs_inode)) {
>     // do something
> } else {
>     // do something else
> }
> ```
> 
> Ergo this would always take the `else` branch, due to Aya not loading the kernel module BTFs. This is a silent failure that is pretty unintuitive and hard to debug.

I agree that the silent failure there in unfortunate - in most cases I've seen related to missing split-BTF we poison the instruction and then the program fails to load, but obviously here that's not possible.

The fact remains though with the API you propose (changing `from_sys_fs` to load all the module BTF) there is no option to opt-out of that at all. This is what I was proposing, which also describes the changes to the default behaviour:
https://github.com/dave-tucker/aya/commit/88b98ec3a002b7082c0e225bf993bcf39fc4bb77#diff-7705012b274301237c1a099d2fa7eb4691aa4d262cc909a344fc9c9eb7d18d07R351-R373

> Also, your mileage may vary here, but from what I can see, vmlinux is usually the biggest BTF by multiple orders of magnitude, with the module BTF being much, much smaller. On my current kernel, the `/sys/kernel/btf` folder is 5.8MiB, where vmlinux takes 4.9MiB of those. So the additional module BTF is on the order of 10%. This of course will vary from distro to distro, but I don't think an extra MiB or two of memory in exchange for making sure all eBPFs program work without fuss is a bad default ^^.
> 
> > or by detecting that we need to via checking which modules we need through kallsyms when attaching probes.
> 
> That would probably be the best of both worlds here, yes.

Feel free to pull in the first commit from https://github.com/dave-tucker/aya/tree/split-btf which adds kallsysms parsing.

> > We need relocation tests to ensure that relocations are working correctly for structs etc.. that are in module BTF
> 
> Yes, I'm working on those.
> 
> > I'm also not convinced that mutating the Base BTF is the best approach either. In my branch I opted to instead create a composite type CompositeBTF which comprised of a base BTF + module BTF. That's closer to how it's implemented in libbpf
> 
> So I tried something along those lines first, but quickly gave up as I realized it required changing quite a few abstraction in the relocation code. In particular, it means that `type_id` is no longer a unique identifier of a type. And similarly, getting the strings associated with a type requires knowing where that type originated from. None of those are necessarily hard problems, but the changes ended up being more involved than I wanted, so I figured merging the BTFs back into a traditional single-file BTF was just easier.

type_id's not being unique across modules is a "feature" of module BTF unfortunately.
either way, the type_id is still unique across any composite btf (base + module).

> > , but also makes it a little easier to test things out - for example, you can easily write unit tests for type resolution to show that you check module btf types before searching in the base btf.
> 
> I'm not sure I understand what this means. For starters, I don't understand why you'd want to check module BTF types before searching the base BTF? As far as I understand, you shouldn't have any conflicts here (if you did, that would cause ambiguity - what if you had conflicts betwen different module BTFs?). Generally speaking, I think testing this approach is rather straightforward: you can just create a base BTF object and multiple "split" BTFs, merge them, and make sure everything still resolves properly.

AIUI you need this for forward declarations. The base BTF could declare `FWD foo` where `STRUCT foo` resides in a module BTF. When resolving type by name you'd want to find `STRUCT foo` from the module BTF first since the main BTF doesn't know what type it's supposed to be. It may be totally valid for 2 modules to have different declarations of the same FWD decl - which comes back to why having single `CompositeBTF` might be better than merging all the BTFs together in to one.

Basically, I'd rather we stick as close to what libbpf does as possible.

roblabla:
> AIUI you need this for forward declarations. The base BTF could declare FWD foo where STRUCT foo resides in a module BTF. When resolving type by name you'd want to find STRUCT foo from the module BTF first since the main BTF doesn't know what type it's supposed to be. It may be totally valid for 2 modules to have different declarations of the same FWD decl - which comes back to why having single CompositeBTF might be better than merging all the BTFs together in to one.

This still feels off to me. Rather than this, wouldn't it make more sense to keep looking if we hit a FWD to see if we can find a STRUCT? This way, the behavior is consistent whether we are in a split BTF scenario or not.

dave-tucker:
> > AIUI you need this for forward declarations. The base BTF could declare FWD foo where STRUCT foo resides in a module BTF. When resolving type by name you'd want to find STRUCT foo from the module BTF first since the main BTF doesn't know what type it's supposed to be. It may be totally valid for 2 modules to have different declarations of the same FWD decl - which comes back to why having single CompositeBTF might be better than merging all the BTFs together in to one.
> 
> This still feels off to me. Rather than this, wouldn't it make more sense to keep looking if we hit a FWD to see if we can find a STRUCT? This way, the behavior is consistent whether we are in a split BTF scenario or not.

🤷 ultimately it comes back to 👇 

> I'd rather we stick as close to what libbpf does as possible

Or follow prior-art... the same pattern is used in cilium/ebpf too: https://github.com/cilium/ebpf/pull/1300/files#diff-82332939f914df35856d29498006ce845e5341b42f3b550179a4fa1c8a7dddcf

It seems having `Merge()` for BTF was considered and then discounted in https://github.com/cilium/ebpf/issues/705
The discussion calls out a few cases that we need to also consider so it's worth a read.

tamird:
@roblabla do you plan to resume work on this? Please let me know how I can help.

Issue #1252, opened 2025-04-18
Accessing "unset" value in array yields value with all zeros


thomaseizinger:
I am not sure if this is common eBPF knowledge but I've just spent a couple of hours chasing down a bug that was rather surprising to find out.

I am using an [`Array`](https://docs.rs/aya/latest/aya/maps/array/struct.Array.html) with a single element to share a `Config` struct between user space and kernel space. Due to a bug in my user space program, that config never got initialised.

The type-signature of `Array::get` currently returns an `Option`:

https://github.com/aya-rs/aya/blob/0237e36dbeb882dccc4887fadc3e0dd0324e2eb7/ebpf/aya-ebpf/src/maps/array.rs#L50-L54

It appears though that instead of returning `None` if a certain index has not been set, this returns a value `T` that instead has all its values set to `0`.

This this expected / known? If yes, then I'd argue the type-signature should be changed and we should not be returning an `Option` here.

tamird:
The Option, I think, indicates whether or not you are in-bounds.

thomaseizinger:
In that case, I think the API is quite misleading considering that similar operations on `Vec` or slices in Rust behave very differently.

I guess we can't detect whether the slot has been set to 0 or is unset? Should we change the API then to return a `Result` with an explicit error that the index is out of bounds?

tamird:
Don't you think `Option` better matches the C API? https://docs.ebpf.io/linux/helper-function/bpf_map_lookup_elem/

For arrays this is a bit confusing, but for maps it does behave the way you expect.

thomaseizinger:
And how much value does it have to match the C API? I am using Rust so that I don't have to deal with (so many) C APIs!

If there is an opportunity to make the API better through good use of the Rust type-system, then I think we should. Most Rust data structures panic if you access something out of bounds. We can't do that so an error seems like the next closest thing.

The user space part of `Array` actually already has the API I am suggesting: https://docs.rs/aya/latest/aya/maps/array/struct.Array.html#method.get

Aligning the APIs of these two types as much as possible seems more useful than matching the C API.

tamird:
Sure, want to send a PR?

thomaseizinger:
> Most Rust data structures panic if you access something out of bounds. We can't do that so an error seems like the next closest thing.

I have to retract this, that is actually not true. The API of `slice::get` is in fact what we have today in `Array::get`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get

[...]

PR #1253, opened 2025-04-19
aya-ebpf: introduce `array::OutOfBounds` error

thomaseizinger:
This attempts to address #1252 by adding a dedicated `OutOfBounds` error and updating the documentation of several functions.

Resolves: #1252


tamird (changes requested):
You'll need to run `CARGO_CFG_BPF_TARGET_ARCH=x86_64 cargo +nightly xtask public-api --bless --target x86_64-unknown-linux-gnu` to fix the build failures (please put the resulting change in the first commit).

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1253#-:-OONguvN6EPXh99SHDal:b-iqdzp4)* status: 0 of 7 files reviewed, 1 unresolved discussion

___
*[`test/integration-ebpf/src/bpf_probe_read.rs` line 71 at r1](https://reviewable.io/reviews/aya-rs/aya/1253#-OONgkLr3AEvHefeyXzk:-OONgkLr3AEvHefeyXzl:bow6fxq) ([raw file](https://github.com/aya-rs/aya/blob/9f3a944f9b94cedbfb86a42d33ae353c971d6400/test/integration-ebpf/src/bpf_probe_read.rs#L71)):*
> ```XML
>         KERNEL_BUFFER
>             .get_ptr(0)
>             .ok()
> ```

this is a compelling argument against this change. now to retain ergonomics we're having to do this error-discarding nonsense. is it better? I'll let @alessandrod adjudicate.

thomaseizinger:
> this is a compelling argument against this change. now to retain ergonomics we're having to do this error-discarding nonsense. is it better? I'll let @alessandrod adjudicate.

In my eBPF program, I've introduced a dedicated `Error` enum that gets returned in several places and mapped to a return code in the entry-point function. Handling a case like this would therefore just require `?` which is quite ergonomic.

tamird:
> > this is a compelling argument against this change. now to retain ergonomics we're having to do this error-discarding nonsense. is it better? I'll let @alessandrod adjudicate.
> 
> In my eBPF program, I've introduced a dedicated `Error` enum that gets returned in several places and mapped to a return code in the entry-point function. Handling a case like this would therefore just require `?` which is quite ergonomic.
> 
> In other words, I agree this particular case here is unergonomic. That is due to how the test is set up and not intrinsic to the new API.

My beef here isn't ergonomics, it's that the error is being discarded. In your program, couldn't you trivially wrap functions in something that converts from Option to Result with the error type of your choice, and then use `?`? It's worth noting that `?` does work in functions that return Option, when used on Option values.

thomaseizinger:
> My beef here isn't ergonomics, it's that the error is being discarded.

I am not sure I am following. You say your beef isn't ergonomics but then previously you said:

> now to retain ergonomics we're having to do this error-discarding nonsense.

Which to me sounds like the ergonomics are important? The ergonomics in the test being bad are not a result of this change but due to error handling not being set up properly in the test like it would be in a real application. I am happy to change that but I'd like to achieve consensus on it first before I invest the time.

> In your program, couldn't you trivially wrap functions in something that converts from Option to Result with the error type of your choice, and then use `?`? It's worth noting that `?` does work in functions that return Option, when used on Option values.

Yes I could do that. But that would make this change pointless because we again don't have an unambiguous API as to what the failure modes of accessing an index are. I thought we agree in #1252 that aligning the eBPF API with user-space is good?

tamird:
I am 👎 on this. Perhaps @alessandrod can weigh in.

PR #1255, opened 2025-04-20
aya-ebpf: put mem{set,move,cpy} behind cfg(target_arch = "bpf")

### LINKED ISSUE #1254: aya_ebpf overrides C library `memcpy()` function with an implementation with incorrect return value
I've been debugging why the test suite of `mitmproxy-linux` crashes under under Gentoo Sandbox. Long story short, it turned out that aya_ebpf somehow overrides the libc `memcpy()` function with a Rust implementation:

https://github.com/aya-rs/aya/blob/0237e36dbeb882dccc4887fadc3e0dd0324e2eb7/ebpf/aya-ebpf/src/lib.rs#L73-L77

Note that while the C implementation returns a pointer to `dest`, the Rust implementation doesn't — so the crashing program was getting `(void*)1` instead of the `dest` pointer.

<details>
<summary>Dire details</summary>

```
hread 8.1 "mitmproxy_linux" hit Breakpoint 1.7, init_env_entries (prefixes=prefixes@entry=0x7ffff7f800c0 <sbcontext+32>, 
    env=env@entry=0x7ffff7f76972 "SANDBOX_READ", prefixes_env=prefixes_env@entry=0x7fffffffed5f "/")
    at /usr/src/debug/sys-apps/sandbox-2.46/sandbox-2.46/libsandbox/libsandbox.c:220
220             buffer = xstrdup(prefixes_env);
(gdb) s
strdup (s=0x7fffffffed5f "/") at /usr/src/debug/sys-apps/sandbox-2.46/sandbox-2.46/libsandbox/memory.c:149
149     {
(gdb) n
156             len = strlen(s);
(gdb) n
157             ret = malloc(len + 1);
(gdb) n
158             if (!ret)
(gdb) p ret
$17 = 0x7ffff7fbd010 ""
(gdb) n
160             return memcpy(ret, s, len + 1);
(gdb) s
memcpy (__dest=0x7ffff7fbd010, __src=0x7fffffffed5f, __len=2) at /usr/include/bits/string_fortified.h:29
29        return __builtin___memcpy_chk (__dest, __src, __len,
(gdb) s
aya_ebpf::memcpy (dest=0x7ffff7fbd010, src=0x7fffffffed5f, n=2) at src/lib.rs:76
76          copy_forward(dest, src, n);
(gdb) l
71          }
72      }
73
74      #[no_mangle]
75      pub unsafe extern "C" fn memcpy(dest: *mut u8, src: *mut u8, n: usize) {
76          copy_forward(dest, src, n);
77      }
78
79      #[no_mangle]
80      pub unsafe extern "C" fn memmove(dest: *mut u8, src: *mut u8, n: usize) {
```
</details>

tamird:
Closes #1254.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1255)

### DIFF
--- a/xtask/public-api/aya-ebpf.txt
+++ b/xtask/public-api/aya-ebpf.txt
@@ -2833,6 +2833,3 @@ pub fn aya_ebpf::programs::tracepoint::TracePointContext::as_ptr(&self) -> *mut 
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::xdp::XdpContext
 pub fn aya_ebpf::programs::xdp::XdpContext::as_ptr(&self) -> *mut core::ffi::c_void
 pub fn aya_ebpf::check_bounds_signed(value: i64, lower: i64, upper: i64) -> bool
-#[no_mangle] pub unsafe c fn aya_ebpf::memcpy(dest: *mut u8, src: *mut u8, n: usize)
-#[no_mangle] pub unsafe c fn aya_ebpf::memmove(dest: *mut u8, src: *mut u8, n: usize)
-#[no_mangle] pub unsafe c fn aya_ebpf::memset(s: *mut u8, c: aya_ebpf_cty::ad::c_int, n: usize)
--- a/ebpf/aya-ebpf/src/lib.rs
+++ b/ebpf/aya-ebpf/src/lib.rs
@@ -27,7 +27,7 @@ use core::{ffi::c_void, ptr::NonNull};
 
 pub use aya_ebpf_cty as cty;
 pub use aya_ebpf_macros as macros;
-use cty::{c_int, c_long};
+use cty::c_long;
 use helpers::{
     bpf_get_current_comm, bpf_get_current_pid_tgid, bpf_get_current_uid_gid, bpf_map_delete_elem,
     bpf_map_lookup_elem, bpf_map_update_elem,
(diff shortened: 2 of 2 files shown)


vadorovsky (approved):

xtask/public-api/aya-ebpf.txt
```diff
@@ -2833,6 +2833,3 @@ pub fn aya_ebpf::programs::tracepoint::TracePointContext::as_ptr(&self) -> *mut
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::xdp::XdpContext
 pub fn aya_ebpf::programs::xdp::XdpContext::as_ptr(&self) -> *mut core::ffi::c_void
 pub fn aya_ebpf::check_bounds_signed(value: i64, lower: i64, upper: i64) -> bool
-#[no_mangle] pub unsafe c fn aya_ebpf::memcpy(dest: *mut u8, src: *mut u8, n: usize)
```
alessandrod, line 2836: this means we're generating the API for the wrong arch right?

tamird: Absolutely, but we always have. We are generating the API for x86_64-unknown-linux-gnu.

Perhaps we can do better in a follow up, but that's where we're at presently.

PR #1286, opened 2025-07-02
aya: `construct_debuglink_path()` should be infallible

bertschingert:
`construct_debuglink_path()` returns a `Result`, but it doesn't actually do anything that can fail. This means callers must do a useless `unwrap()` to get at the return value.

This fixes the signature to reflect that the function is infallible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1286)


tamird:
LGTM. Thanks!

PR #1294, opened 2025-07-15
aya-log: make logging zero-copy

tamird:
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294)

### DIFF
--- a/aya-log-common/src/lib.rs
+++ b/aya-log-common/src/lib.rs
@@ -2,12 +2,10 @@
 
 use core::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr},
-    num::{NonZeroUsize, TryFromIntError},
+    num::TryFromIntError,
 };
 
 use num_enum::IntoPrimitive;
-
-pub const LOG_FIELDS: usize = 6;
 
 pub type LogValueLength = u16;
 
@@ -67,6 +65,8 @@ impl_formatter_for_types!(
     }
 );
 
+impl<const N: usize> LowerHexFormatter for &[u8; N] {}
+
 pub trait UpperHexFormatter {}
 impl_formatter_for_types!(
     UpperHexFormatter: {
@@ -75,6 +75,8 @@ impl_formatter_for_types!(
         &[u8]
     }
 );
+
+impl<const N: usize> UpperHexFormatter for &[u8; N] {}
 
 pub trait IpFormatter {}
 impl IpFormatter for IpAddr {}
@@ -93,8 +95,8 @@ impl UpperMacFormatter for [u8; 6] {}
 
 #[repr(u8)]
 #[derive(Copy, Clone, Debug, IntoPrimitive)]
-pub enum RecordField {
-    Target = 1,
+pub enum RecordFieldKind {
+    Target,
     Level,
     Module,
     File,
@@ -106,7 +108,7 @@ pub enum RecordField {
 /// programs to userspace.
 #[repr(u8)]
 #[derive(Copy, Clone, Debug, IntoPrimitive)]
-pub enum Argument {
+pub enum ArgumentKind {
     DisplayHint,
 
     I8,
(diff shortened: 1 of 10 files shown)


ajwerner (approved):
Neat!

> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294#-:-OVFGQnq1FHkVeu28T23:b-83v4r2)* status: all files reviewed, 3 unresolved discussions (waiting on @alessandrod and @vadorovsky)

___
*[`aya-log-ebpf-macros/src/expand.rs` line 166 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFCknN3T0Ls0eR7vWt:-OVFCknN3T0Ls0eR7vWu:bue76hk) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L166)):*
> ```XML
>             use ::aya_log_ebpf::macro_support::{Header, Field, Argument, AYA_LOGS};
> 
>             let #num_args = match u32::try_from(#len) {
> ```

You know len statically in the macro, just check there and make it a u32 ident?
___
*[`aya-log-ebpf-macros/src/expand.rs` line 180 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFDQrp0EgYX9x2VvJU:-OVFDQrp0EgYX9x2VvJV:b-hneiz7) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L180)):*
> ```XML
> 
>             #(
>                 let #tmp = #values;
> ```

I feel like the more Tamir way to write this would have been the following. Any reason you prefer the splatting and then tmp?

```rust
#( 
    let #idents = {
        let (#kind, #value) = #values.as_argument();
        Field::new(kind, value)?
    };
)*
```
___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVF8e821oXabRhmpzmn:b-n31b8f) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*
> ```XML
>             )*
> 
>             let mut #size = 0;
> ```

It's rather amazing that this works and always verifies. I wonder how much of this is LLVM evaluating into constants ahead of time and how much is symbolic execution. I wonder if there's a size cutoff for a large number of arguments where the behavior changes. If you have the compiled bytecode of the log integration test ebpf program handy, I'd be curious to look at it. Did you have any trouble with the verifier on this?
___
*[`ebpf/aya-log-ebpf/src/lib.rs` line 17 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFGXiXB7GyBhYTY5Fx:-OVFGXiXB7GyBhYTY5Fy:b-ri9p2p) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/ebpf/aya-log-ebpf/src/lib.rs#L17)):*
> ```XML
>     };
> 
>     const LOG_BUF_CAPACITY: LogValueLength = 8192;
> ```

I think this constant is sort of meaningless now. The maximum log size ends up being the maximum message size for the ringbuffer given the capacity it's configured with (which can be overridden prior to loading). I wonder how bad the error message is when you hit it -- I suspect it'll just be a verifier failure.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @ajwerner, @alessandrod, and @vadorovsky)

___
*[`aya-log-ebpf-macros/src/expand.rs` line 166 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFCknN3T0Ls0eR7vWt:-OVFNu8e7uHASpIpowm9:bz5msft) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L166)):*

You know len statically in the macro, just check there and make it a u32 ident?
</blockquote></details>

Yeah, okay - done.
___
*[`aya-log-ebpf-macros/src/expand.rs` line 180 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFDQrp0EgYX9x2VvJU:-OVFNxZ36aETCB3Nnits:bj7mjhu) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L180)):*

I feel like the more Tamir way to write this would have been the following. Any reason you prefer the splatting and then tmp?

```rust
#( 
    let #idents = {
        let (#kind, #value) = #values.as_argument();
        Field::new(kind, value)?
    };
)*
```
</blockquote></details>

The return of `Argument::as_argument` is borrowed from `self`:
```
fn as_argument(&self) -> (ArgumentKind, impl AsRef<[u8]>);
```
(there's a lifetime hidden in that `impl AsRef<[u8]>`)
___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVFOzGkAK_GDdz7Betm:b-kenwnf) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*

It's rather amazing that this works and always verifies. I wonder how much of this is LLVM evaluating into constants ahead of time and how much is symbolic execution. I wonder if there's a size cutoff for a large number of arguments where the behavior changes. If you have the compiled bytecode of the log integration test ebpf program handy, I'd be curious to look at it. Did you have any trouble with the verifier on this?
</blockquote></details>

I had some trouble when I tried to do fancier things like chaining iterators rather than the closure passing you see in the current version. 

Yep, here it is: https://drive.google.com/file/d/1fHIO5Xq7tvGaDrcvvx_Gde7nwEBGCQz8/view?usp=sharing
___
*[`ebpf/aya-log-ebpf/src/lib.rs` line 17 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFGXiXB7GyBhYTY5Fx:-OVFOJ3G-Pbvyc7MwlBb:b-4cidss) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/ebpf/aya-log-ebpf/src/lib.rs#L17)):*

I think this constant is sort of meaningless now. The maximum log size ends up being the maximum message size for the ringbuffer given the capacity it's configured with (which can be overridden prior to loading). I wonder how bad the error message is when you hit it -- I suspect it'll just be a verifier failure.
</blockquote></details>

Yeah, this constant is utterly meaningless, I'll remove it.

As for the size being configurable: of course you're right and it is, but we don't currently expose an API to do so specifically for aya-log. You could obviously reach into the program and configure it, though.

Regarding the verifier failure, I'm not sure - I would think that if you tried to log something too big you'd hit one of the `?`s in the macro expansion and just silently not log anything. Don't you think?

ajwerner (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294#-:-OVG28Wp5hetNTwxtEht:bnfp4nl)* status: 2 of 7 files reviewed, all discussions resolved (waiting on @alessandrod and @vadorovsky)

___
*[`aya-log-ebpf-macros/src/expand.rs` line 180 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVFDQrp0EgYX9x2VvJU:-OVFxKiV00_astQjZjOQ:b-py1dax) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L180)):*

The return of `Argument::as_argument` is borrowed from `self`:
```
fn as_argument(&self) -> (ArgumentKind, impl AsRef<[u8]>);
```
(there's a lifetime hidden in that `impl AsRef<[u8]>`)
</blockquote></details>

Ah, yes
___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVFzNmq3FYfoJglwlBw:b-ljd6ti) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*

I had some trouble when I tried to do fancier things like chaining iterators rather than the closure passing you see in the current version. 

Yep, here it is: https://drive.google.com/file/d/1fHIO5Xq7tvGaDrcvvx_Gde7nwEBGCQz8/view?usp=sharing
</blockquote></details>

It's incredible. Compilers are so dope. It literally does boil down to completely eliminating the size computation into static values in the assembly and inline instructions to copy data into the buffer.

For the first call (commentary mine):
```
       1:	18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00	r1 = 0x0 ll // map pointer to be filled in
       3:	b7 02 00 00 54 00 00 00	r2 = 0x54 // #size 
       4:	b7 03 00 00 00 00 00 00	r3 = 0x0 // flags
       5:	85 00 00 00 83 00 00 00	call 0x83 // bpf_ringbuf_reserve
;     #[inline]
       6:	15 00 a4 00 00 00 00 00	if r0 == 0x0 goto +0xa4 <test_log+0x558>  // failure check
// fill in the buffer with exact offset
```

https://gist.github.com/ajwerner/3b7cca9336c7f913d2982ae7e9a38ebd

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294)* status: 2 of 7 files reviewed, all discussions resolved (waiting on @ajwerner, @alessandrod, and @vadorovsky)

___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVIZ549AsA-l0Lk2Uxk:b-uopanh) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*

It's incredible. Compilers are so dope. It literally does boil down to completely eliminating the size computation into static values in the assembly and inline instructions to copy data into the buffer.

For the first call (commentary mine):
```
       1:	18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00	r1 = 0x0 ll // map pointer to be filled in
       3:	b7 02 00 00 54 00 00 00	r2 = 0x54 // #size 
       4:	b7 03 00 00 00 00 00 00	r3 = 0x0 // flags
       5:	85 00 00 00 83 00 00 00	call 0x83 // bpf_ringbuf_reserve
;     #[inline]
       6:	15 00 a4 00 00 00 00 00	if r0 == 0x0 goto +0xa4 <test_log+0x558>  // failure check
// fill in the buffer with exact offset
```

https://gist.github.com/ajwerner/3b7cca9336c7f913d2982ae7e9a38ebd
</blockquote></details>

This got me wondering what would happen if I added a test that tried to log a value whose size was not a compile-time constant.

I wrote [this](https://gist.github.com/tamird/bec42098434b80905617b13fb0bdbe31) and the result is verifier failure:

```
2720: (b7) r1 = 1024                  ; R1=1024
; let buf = if len < buf.len() {
2721: (2d) if r1 > r7 goto pc+1 2723: R0=map_value(off=0,ks=4,vs=1032,imm=0) R1=1024 R6=map_value(off=0,ks=4,vs=1032,imm=0) R7=scalar(umax=1023,var_off=(0x0; 0x3ff)) R8=115 R9=19 R10=fp0 fp-8=mmmm????
; info!(&ctx, "variable length buffer: {:x}", buf);                                                                                         
2723: (bf) r2 = r7                    ; R2_w=scalar(id=29,umax=1023,var_off=(0x0; 0x3ff)) R7=scalar(id=29,umax=1023,var_off=(0x0; 0x3ff))
2724: (07) r2 += 99                   ; R2_w=scalar(umin=99,umax=1122,var_off=(0x0; 0x7ff))
2725: (b7) r8 = 0                     ; R8_w=0
; fun(ringbuf, size, flags)
2726: (18) r1 = 0xffff5f3d823cc200    ; R1_w=map_ptr(off=0,ks=0,vs=0,imm=0)
2728: (b7) r3 = 0                     ; R3_w=0
2729: (85) call bpf_ringbuf_reserve#131
R2 is not a known constant'
```

maybe this is the reason we didn't implement support for DSTs when we first wrote the ring buffer abstraction?

ajwerner (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294)* status: 2 of 10 files reviewed, 1 unresolved discussion (waiting on @alessandrod and @vadorovsky)

___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVIgIPL8OnCLLFK4eRj:b77yrbp) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*

This got me wondering what would happen if I added a test that tried to log a value whose size was not a compile-time constant.

I wrote [this](https://gist.github.com/tamird/bec42098434b80905617b13fb0bdbe31) and the result is verifier failure:

```
2720: (b7) r1 = 1024                  ; R1=1024
; let buf = if len < buf.len() {
2721: (2d) if r1 > r7 goto pc+1 2723: R0=map_value(off=0,ks=4,vs=1032,imm=0) R1=1024 R6=map_value(off=0,ks=4,vs=1032,imm=0) R7=scalar(umax=1023,var_off=(0x0; 0x3ff)) R8=115 R9=19 R10=fp0 fp-8=mmmm????
; info!(&ctx, "variable length buffer: {:x}", buf);                                                                                         
2723: (bf) r2 = r7                    ; R2_w=scalar(id=29,umax=1023,var_off=(0x0; 0x3ff)) R7=scalar(id=29,umax=1023,var_off=(0x0; 0x3ff))
2724: (07) r2 += 99                   ; R2_w=scalar(umin=99,umax=1122,var_off=(0x0; 0x7ff))
2725: (b7) r8 = 0                     ; R8_w=0
; fun(ringbuf, size, flags)
2726: (18) r1 = 0xffff5f3d823cc200    ; R1_w=map_ptr(off=0,ks=0,vs=0,imm=0)
2728: (b7) r3 = 0                     ; R3_w=0
2729: (85) call bpf_ringbuf_reserve#131
R2 is not a known constant'
```

maybe this is the reason we didn't implement support for DSTs when we first wrote the ring buffer abstraction?
</blockquote></details>

Correct. For writing to ringbufs you need the offsets and the buffer capacity to be a statically known size. If you want the underlying allocation to be variable size, you can get access to a statically sized sub-array with [bpf_ringbuf_reserve_dynptr](https://docs.ebpf.io/linux/helper-function/bpf_ringbuf_reserve_dynptr/) and then [bpf_dynptr_data](https://docs.ebpf.io/linux/helper-function/bpf_dynptr_data/) but that's not available until 5.19. If you want to do a dynamically sized write, you have to use `bpf_dynptr_write`. 
___
*[`test/integration-ebpf/src/log.rs` line 113 at r10](https://reviewable.io/reviews/aya-rs/aya/1294#-OVIhPPuDcr3EEYMHHbN:-OVIhPPuDcr3EEYMHHbO:bp5vyrk) ([raw file](https://github.com/aya-rs/aya/blob/3847281b15dbd61a9f8e137f4db10471682822ca/test/integration-ebpf/src/log.rs#L113)):*
> ```XML
>         &buf[..]
>     };
>     info!(&ctx, "variable length buffer: {:x}", buf);
> ```

comment that this doesn't work but does verify?

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294)* status: 2 of 10 files reviewed, 4 unresolved discussions (waiting on @ajwerner, @alessandrod, and @vadorovsky)

___
*[`aya-log/src/lib.rs` line 797 at r10](https://reviewable.io/reviews/aya-rs/aya/1294#-OVIhzCtBgzpTuvCH_tE:-OVIjz96BLdOGUYo1N_i:bwv1apm) ([raw file](https://github.com/aya-rs/aya/blob/3847281b15dbd61a9f8e137f4db10471682822ca/aya-log/src/lib.rs#L797)):*

I think that getting a slice with the full range should work?
</blockquote></details>

This is not equivalent. The equivalent would be `let buf = buf.get_mut(size..size+slice.len())?;` but then my overflow spidey sense starts tingling and I think it's not worth it.
___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVInjWU5PAaTSEfBUy1:b-8rat87) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*

Correct. For writing to ringbufs you need the offsets and the buffer capacity to be a statically known size. If you want the underlying allocation to be variable size, you can get access to a statically sized sub-array with [bpf_ringbuf_reserve_dynptr](https://docs.ebpf.io/linux/helper-function/bpf_ringbuf_reserve_dynptr/) and then [bpf_dynptr_data](https://docs.ebpf.io/linux/helper-function/bpf_dynptr_data/) but that's not available until 5.19. If you want to do a dynamically sized write, you have to use `bpf_dynptr_write`.
</blockquote></details>

Ah, I see. I hadn't appreciated that. I think I will rework this PR by just allocating a fixed size buffer from the RB (the old 8k constant) and write to it in the manner previously used to write to the scratch buffer. Most of the refactoring in this PR can be shelved -- calculating the size in advance of writing doesn't really buy us much.
___
*[`aya-log-ebpf-macros/src/expand.rs` line 202 at r10](https://reviewable.io/reviews/aya-rs/aya/1294#-OVIhj559kM4a2CjiX4c:-OVIkDFf0waSktQvFcrH:b-hzznqy) ([raw file](https://github.com/aya-rs/aya/blob/3847281b15dbd61a9f8e137f4db10471682822ca/aya-log-ebpf-macros/src/expand.rs#L202)):*

I think that getting a slice with the full range should work?
</blockquote></details>

Same as the other.
___
*[`test/integration-ebpf/src/log.rs` line 113 at r10](https://reviewable.io/reviews/aya-rs/aya/1294#-OVIhPPuDcr3EEYMHHbN:-OVIkqGt-M-XrVUOJTdf:b-1r4gwg) ([raw file](https://github.com/aya-rs/aya/blob/3847281b15dbd61a9f8e137f4db10471682822ca/test/integration-ebpf/src/log.rs#L113)):*

comment that this doesn't work but does verify?
</blockquote></details>

Well, it does *not* verify!

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294)* status: 2 of 10 files reviewed, 4 unresolved discussions (waiting on @ajwerner, @alessandrod, and @vadorovsky)

___
*[`aya-log-ebpf-macros/src/expand.rs` line 185 at r3](https://reviewable.io/reviews/aya-rs/aya/1294#-OVF8e821oXabRhmpzmm:-OVZW4Lm46ew23ykE4Ce:bu0o1eg) ([raw file](https://github.com/aya-rs/aya/blob/eeee7966c61f43428b496aa3065d4d7e54568e9b/aya-log-ebpf-macros/src/expand.rs#L185)):*
> What'd be great is a way to enforce that these sizes can be const computed for all types that are loggable and maybe also we need to seal what you can log?

Yeah, these are both good things to call out. I'll definitely adding the seal and then I'll look into enforcing compile-time size calculation. The latter might be hard because [trait functions cannot be declared const](https://doc.rust-lang.org/error_codes/E0379.html). Certainly the north star here is that we should be able to remove the new ringbuf methods I added since true DSTs cannot be used.

tamird (commented):
Thanks to some pair programming with @ajwerner this now works and is efficient! Hurray!

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1294#-:-OZg8-4jCea-9-l0YYtp:b9p3w2t)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @ajwerner, @alessandrod, and @vadorovsky)

___
*[`test/integration-ebpf/src/log.rs` line 112 at r10](https://reviewable.io/reviews/aya-rs/aya/1294#-OVIhHYEDTCaogJS_3V9:-OZg7wh5COiIsOrP-k0i:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/3847281b15dbd61a9f8e137f4db10471682822ca/test/integration-ebpf/src/log.rs#L112)):*

Or perhaps even `&buf[..len]` if we can guarantee that `len` never exceeds the buffer length.
</blockquote></details>

Done.
___
*[`test/integration-ebpf/src/log.rs` line 113 at r10](https://reviewable.io/reviews/aya-rs/aya/1294#-OVIhPPuDcr3EEYMHHbN:-OZg7hZT1j_95qOnwtia:b-h155di) ([raw file](https://github.com/aya-rs/aya/blob/3847281b15dbd61a9f8e137f4db10471682822ca/test/integration-ebpf/src/log.rs#L113)):*

Well, it does *not* verify!
</blockquote></details>

Resolving.

PR #1297, opened 2025-07-17
aya-log: simplify EbpfLogger as_raw_fd implementation

cppcoffee:
Replace struct pattern matching with direct field access in the as_raw_fd method, making the code more concise and cleaner.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1297)


tamird:
This is not simpler. It's a stylistic change that appears to have been generated by an LLM.

PR #1298, opened 2025-07-17
aya-log: suppress dead_code warnings for wrapper structs

cppcoffee:
Fixes compiler warnings about RecordFieldWrapper and ArgumentWrapper never being constructed by adding #[expect(dead_code)] attributes.

Solve the error message as follows:

```shell
# cargo build
warning: struct `RecordFieldWrapper` is never constructed
  --> aya-log/src/lib.rs:81:8
   |
81 | struct RecordFieldWrapper(RecordField);
   |        ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: struct `ArgumentWrapper` is never constructed
  --> aya-log/src/lib.rs:84:8
   |
84 | struct ArgumentWrapper(Argument);
   |        ^^^^^^^^^^^^^^^
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1298)

### DIFF
--- a/aya-log/src/lib.rs
+++ b/aya-log/src/lib.rs
@@ -76,9 +76,11 @@ use aya_log_common::{Argument, DisplayHint, LOG_FIELDS, Level, LogValueLength, R
 use log::{Log, Record, error};
 use thiserror::Error;
 
+#[expect(dead_code)]
 #[derive(Copy, Clone)]
 #[repr(transparent)]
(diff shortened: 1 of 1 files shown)


tamird:
This doesn't work. You appear to have generated this change using an LLM so I doubt you'll do the necessary deep investigation here.

PR #1299, opened 2025-07-17
aya-log: simplify map name matching using direct Option comparison

cppcoffee:
Replace match statement with direct Option comparison for cleaner code

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1299)


tamird:
This is better, thank you. CI is slightly broken until https://github.com/cargo-public-api/cargo-public-api/pull/807 is merged and released, but I'll go ahead and merge this anyway.

PR #1300, opened 2025-07-17
aya-log: Use `None` instead of wildcard in `Format` impls

cppcoffee:
The `Format` implementations for various types were using a wildcard `_` to match the `None` case on `Option<DisplayHint>`.

This is incorrect as it would also match any `Some(...)` variants that were not explicitly handled, leading to unexpected behavior.

This commit changes the wildcard `_` to an explicit `None` match to ensure that only the `None` case is handled, making the matching more explicit and correct.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1300)


tamird:
This is better, thank you. CI is slightly broken until https://github.com/cargo-public-api/cargo-public-api/pull/807 is merged and released, but I'll go ahead and merge this anyway.

PR #1304, opened 2025-07-20
aya-ebpf: replace match statement with then_some for cleaner code

cppcoffee:
Replace verbose match statement with more idiomatic then_some().ok_or() pattern for handling boolean return values in assign method.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1304)


tamird:
Doesn't seem useful.

PR #1312, opened 2025-08-04
aya: fix handle null bytes in paths when creating CString

cppcoffee:
Add a new InvalidPath error variant to MapError enum that wraps std::ffi::NulError

This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1312)

### DIFF
--- a/aya/src/maps/info.rs
+++ b/aya/src/maps/info.rs
@@ -118,8 +118,8 @@ impl MapInfo {
     pub fn from_pin<P: AsRef<Path>>(path: P) -> Result<Self, MapError> {
         use std::os::unix::ffi::OsStrExt as _;
 
-        // TODO: avoid this unwrap by adding a new error variant.
-        let path_string = CString::new(path.as_ref().as_os_str().as_bytes()).unwrap();
+        let path_string =
+            CString::new(path.as_ref().as_os_str().as_bytes()).map_err(MapError::from)?;
         let fd = bpf_get_object(&path_string).map_err(|io_error| SyscallError {
             call: "BPF_OBJ_GET",
             io_error,
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -174,6 +174,10 @@ pub enum MapError {
     #[error(transparent)]
     SyscallError(#[from] SyscallError),
 
+    /// An error indicating that an interior nul byte was found.
+    #[error(transparent)]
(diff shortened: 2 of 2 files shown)


aya/src/maps/info.rs
```diff
@@ -118,8 +118,8 @@ impl MapInfo {
     pub fn from_pin<P: AsRef<Path>>(path: P) -> Result<Self, MapError> {
         use std::os::unix::ffi::OsStrExt as _;
 
-        // TODO: avoid this unwrap by adding a new error variant.
-        let path_string = CString::new(path.as_ref().as_os_str().as_bytes()).unwrap();
+        // Convert path to CString, returning InvalidPath error if it contains null bytes
```
tamird: Useless comment, please remove.

aya/src/maps/mod.rs
```diff
@@ -198,6 +198,10 @@ pub enum MapError {
         /// The map type
         map_type: bpf_map_type,
     },
+
+    /// Invalid path encountered when creating a CString
+    #[error("path contains null bytes and cannot be converted to CString")]
```
tamird: Do we need to mention path? Can this be a general conversion error?

tamird (changes requested):

cppcoffee:
It seems that `NulError` comes with path information, so we can use it directly.

Test code:

```rust
use std::ffi::{CString, NulError};

fn main() {
    let x: NulError = CString::new(b"f\0oo".to_vec()).unwrap_err();
    println!("{:?}", x);
}
```

Output:

```
NulError(1, [102, 0, 111, 111])
```

aya/src/maps/mod.rs
```diff
@@ -174,6 +174,10 @@ pub enum MapError {
     #[error(transparent)]
     SyscallError(#[from] SyscallError),
 
+    /// An error indicating that an interior nul byte was found.
+    #[error(transparent)]
+    NulError(#[from] std::ffi::NulError),
```
alessandrod: This is not the right way to go. This exposes an implementation detail (we convert to CString), to the aya API.

cppcoffee: Should we use `#[error("invalid path '{path}' to create CString {source}")]` as the error message returned from the aya API?

cppcoffee:
introduced `InvalidCString` error as failed to create `CString` due to an invalid C-style string.

Issue #1309, opened 2025-08-01
calling .next() on RingBuf in userspace yields zero length RingBufItem


lblebasheer:
My project consists of two crates that use aya. The first one (producer) is attached to a kprobe and submits a sample into a pinned RingBuf each time the kprobe fires. The second crate (consumer) polls the same pinned map from userspace and generates output. The whole setup works fine when when both programs are started. The consumer doesn't see any new samples and produces no output when there are no new samples pushed by the producer. The problem arises when I terminate the consumer with Ctrl-C and then start it again. When I do that, each time .next() is called on the RingBuf, it yields a zero length item, instead of returning None,  when there are no new samples pushed by the producer. It does this no matter how many time I restart the consumer, until I `rm` the pinned map and then restart the consumer. I tried to make a minimal reproducer which I will include here.

running on kernel based on SuSE distro: 5.14.21-150500.55.97_13.0.78

```
[DEBUG aya::bpf] BPF Feature Detection: Features {
        bpf_name: true,
        bpf_probe_read_kernel: true,
        bpf_perf_link: true,
        bpf_global_data: true,
        bpf_cookie: true,
        cpumap_prog_id: true,
        devmap_prog_id: true,
        prog_info_map_ids: true,
        prog_info_gpl_compatible: true,
        btf: Some(
            BtfFeatures {
                btf_func: true,
                btf_func_global: true,
                btf_datasec: true,
                btf_float: true,
                btf_decl_tag: true,
                btf_type_tag: true,
                btf_enum64: false,
            },
        ),
    }
```

### Producer

producer
```rust
use aya::programs::KProbe;
#[rustfmt::skip]
use log::{debug, warn};
use async_channel;
use smol::block_on;

fn main() -> anyhow::Result<()> {
    env_logger::init();

    // Bump the memlock rlimit. This is needed for older kernels that don't use the
    // new memcg based accounting, see https://lwn.net/Articles/837122/
    let rlim = libc::rlimit {
        rlim_cur: libc::RLIM_INFINITY,
        rlim_max: libc::RLIM_INFINITY,
    };
    let ret = unsafe { libc::setrlimit(libc::RLIMIT_MEMLOCK, &rlim) };
    if ret != 0 {
        debug!("remove limit on locked memory failed, ret is: {ret}");
    }

    // This will include your eBPF object file as raw bytes at compile-time and load it at
    // runtime. This approach is recommended for most real-world use cases. If you would
[...]

PR #1318, opened 2025-08-11
aya/maps/ring_buf: fix producer position initialization

lblebasheer:
The RingBuf caches the last value it read of the producer so it doesn't
need to constantly contend on the actual producer cache line if lots of
messages have yet to be consumed. It was bogus to initialize this cache
at 0. This patch initializes it properly and adds testing.

Fixes: #1309

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318)

### DIFF
--- a/test/integration-test/src/tests/ring_buf.rs
+++ b/test/integration-test/src/tests/ring_buf.rs
@@ -1,6 +1,7 @@
 use std::{
     mem,
     os::fd::AsRawFd as _,
+    path::Path,
     sync::{
         Arc,
         atomic::{AtomicBool, Ordering},
@@ -19,6 +20,7 @@ use aya::{
 use aya_obj::generated::BPF_RINGBUF_HDR_SZ;
 use integration_common::ring_buf::Registers;
 use rand::Rng as _;
+use scopeguard::defer;
 use tokio::io::{Interest, unix::AsyncFd};
 
 struct RingBufTest {
@@ -35,14 +37,25 @@ const RING_BUF_MAX_ENTRIES: usize = 512;
 
 impl RingBufTest {
     fn new() -> Self {
+        Self::new_with_mutators(|_loader| {}, |_bpf| {})
+    }
+
+    // Allows the test to mutate the EbpfLoader before it loads the object file from disk, and to
+    // mutate the loaded Ebpf object after it has been loaded from disk but before it is loaded
+    // into the kernel.
+    fn new_with_mutators<'loader>(
+        loader_fn: impl FnOnce(&mut EbpfLoader<'loader>),
+        bpf_fn: impl FnOnce(&mut Ebpf),
+    ) -> Self {
         const RING_BUF_BYTE_SIZE: u32 =
             (RING_BUF_MAX_ENTRIES * (mem::size_of::<u64>() + BPF_RINGBUF_HDR_SZ as usize)) as u32;
 
         // Use the loader API to control the size of the ring_buf.
-        let mut bpf = EbpfLoader::new()
-            .set_max_entries("RING_BUF", RING_BUF_BYTE_SIZE)
-            .load(crate::RING_BUF)
-            .unwrap();
+        let mut loader = EbpfLoader::new();
+        loader.set_max_entries("RING_BUF", RING_BUF_BYTE_SIZE);
+        loader_fn(&mut loader);
+        let mut bpf = loader.load(crate::RING_BUF).unwrap();
+        bpf_fn(&mut bpf);
         let ring_buf = bpf.take_map("RING_BUF").unwrap();
         let ring_buf = RingBuf::try_from(ring_buf).unwrap();
         let regs = bpf.take_map("REGISTERS").unwrap();
@@ -411,4 +424,71 @@ impl WriterThread {
         done.store(true, Ordering::Relaxed);
         thread.join().unwrap();
     }
+}
+
+// This tests that a ring buffer can be pinned and then re-opened and attached to a subsequent
+// program. It ensures that the producer position is properly synchronized between the two
+// programs, and that no unread data is lost.
+#[tokio::test(flavor = "multi_thread")]
+#[test_log::test]
+async fn ring_buf_pinned() {
+    let pin_path =
+        Path::new("/sys/fs/bpf/").join(format!("ring_buf_{}", rand::rng().random::<u64>()));
+
+    let RingBufTest {
+        mut ring_buf,
+        regs: _,
+        _bpf,
+    } = RingBufTest::new_with_mutators(
+        |_loader| {},
+        |bpf| {
+            let ring_buf = bpf.map_mut("RING_BUF").unwrap();
+            ring_buf.pin(&pin_path).unwrap();
+        },
+    );
+    defer! { std::fs::remove_file(&pin_path).unwrap() }
+
+    // Write a few items to the ring buffer.
+    let to_write_before_reopen = [2, 4, 6, 8];
+    for v in to_write_before_reopen {
+        ring_buf_trigger_ebpf_program(v);
+    }
+    let (to_read_before_reopen, to_read_after_reopen) = to_write_before_reopen.split_at(2);
+    for v in to_read_before_reopen {
+        let item = ring_buf.next().unwrap();
+        let item: [u8; 8] = item.as_ref().try_into().unwrap();
+        assert_eq!(item, v.to_ne_bytes());
+    }
+    drop(ring_buf);
+    drop(_bpf);
+
+    // Reopen the ring buffer using the pinned map.
+    let RingBufTest {
+        mut ring_buf,
+        regs: _,
+        _bpf,
+    } = RingBufTest::new_with_mutators(
+        |loader| {
+            loader.map_pin_path("RING_BUF", &pin_path);
+        },
+        |_bpf| {},
+    );
+    let to_write_after_reopen = [10, 12];
+
+    // Write some more data.
+    for v in to_write_after_reopen {
+        ring_buf_trigger_ebpf_program(v);
+    }
+    // Read both the data that was written before the ring buffer was reopened and the data that
+    // was written after it was reopened.
+    for v in to_read_after_reopen
+        .iter()
+        .chain(to_write_after_reopen.iter())
+    {
+        let item = ring_buf.next().unwrap();
+        let item: [u8; 8] = item.as_ref().try_into().unwrap();
+        assert_eq!(item, v.to_ne_bytes());
+    }
+    // Make sure there is nothing else in the ring buffer.
+    assert_matches!(ring_buf.next(), None);
 }
--- a/aya/src/maps/ring_buf.rs
+++ b/aya/src/maps/ring_buf.rs
@@ -303,6 +303,10 @@ impl ProducerData {
         let len = page_size + 2 * usize::try_from(byte_size).unwrap();
         let mmap = MMap::new(fd, len, PROT_READ, MAP_SHARED, offset.try_into().unwrap())?;
 
+        // The producer position may be non-zero if the map is being loaded from a pin, or the map
+        // has been used previously; load the initial value of the producer position from the mmap.
+        let pos_cache = load_producer_pos(&mmap);
+
         // byte_size is required to be a power of two multiple of page_size (which implicitly is a
         // power of 2), so subtracting one will create a bitmask for values less than byte_size.
         debug_assert!(byte_size.is_power_of_two());
@@ -310,7 +314,7 @@ impl ProducerData {
         Ok(Self {
             mmap,
             data_offset: page_size,
-            pos_cache: 0,
+            pos_cache,
             mask,
         })
     }
@@ -322,7 +326,6 @@ impl ProducerData {
             ref mut pos_cache,
             ref mut mask,
         } = self;
-        let pos = unsafe { mmap.ptr().cast().as_ref() };
         let mmap_data = mmap.as_ref();
         let data_pages = mmap_data.get(*data_offset..).unwrap_or_else(|| {
             panic!(
@@ -331,7 +334,7 @@ impl ProducerData {
                 mmap_data.len()
             )
         });
-        while data_available(pos, pos_cache, consumer) {
+        while data_available(mmap, pos_cache, consumer) {
             match read_item(data_pages, *mask, consumer) {
                 Item::Busy => return None,
                 Item::Discard { len } => consumer.consume(len),
@@ -347,17 +350,15 @@ impl ProducerData {
         }
 
         fn data_available(
-            producer: &AtomicUsize,
-            cache: &mut usize,
+            producer: &MMap,
+            producer_cache: &mut usize,
             consumer: &ConsumerPos,
         ) -> bool {
             let ConsumerPos { pos: consumer, .. } = consumer;
-            if consumer == cache {
-                // This value is written using Release by the kernel [1], and should be read with
-                // Acquire to ensure that the prior writes to the entry header are visible.
-                //
-                // [1]: https://github.com/torvalds/linux/blob/eb26cbb1/kernel/bpf/ringbuf.c#L447-L448
-                *cache = producer.load(Ordering::Acquire);
+            // Refresh the producer position cache if it appears that the consumer is caught up
+            // with the producer position.
+            if consumer == producer_cache {
+                *producer_cache = load_producer_pos(producer);
             }
 
             // Note that we don't compare the order of the values because the producer position may
@@ -369,7 +370,7 @@ impl ProducerData {
             // producer position has wrapped around.
             //
             // [1]: https://github.com/torvalds/linux/blob/4b810bf0/kernel/bpf/ringbuf.c#L434-L440
-            consumer != cache
+            consumer != producer_cache
         }
 
         fn read_item<'data>(data: &'data [u8], mask: u32, pos: &ConsumerPos) -> Item<'data> {
@@ -402,4 +403,13 @@ impl ProducerData {
             }
         }
     }
+}
+
+// Loads the producer position from the shared memory mmap.
+fn load_producer_pos(producer: &MMap) -> usize {
+    // This value is written using Release by the kernel [1], and should be read with
+    // Acquire to ensure that the prior writes to the entry header are visible.
+    //
+    // [1]: https://github.com/torvalds/linux/blob/eb26cbb1/kernel/bpf/ringbuf.c#L447-L448
+    unsafe { producer.ptr().cast::<AtomicUsize>().as_ref() }.load(Ordering::Acquire)
 }


test/integration-test/src/tests/ring_buf.rs
```diff
@@ -27,11 +27,18 @@ struct RingBufTest {
...
+    ring_buf: RingBuf<MapData>,
+    regs: PerCpuArray<MapData, Registers>,
+}
+
 // Note that it is important for this test that RING_BUF_MAX_ENTRIES ends up creating a ring buffer
 // that is exactly a power-of-two multiple of the page size. The synchronous test will fail if
 // that's not the case because the actual size will be rounded up, and fewer entries will be dropped
 // than expected.
 const RING_BUF_MAX_ENTRIES: usize = 512;
+const RING_BUF_PIN_PATH: &str = "/sys/fs/bpf/RING_BUF";
```
ajwerner: nit: this is an odd name to use with the file having all caps.

test/integration-test/src/tests/ring_buf.rs
```diff
@@ -69,7 +76,45 @@ impl RingBufTest {
...
+impl PinnedRingBufTest {
+    fn new() -> Self {
+        const RING_BUF_BYTE_SIZE: u32 =
+            (RING_BUF_MAX_ENTRIES * (mem::size_of::<u64>() + BPF_RINGBUF_HDR_SZ as usize)) as u32;
+
+        let mut bpf = EbpfLoader::new()
+            .set_max_entries("RING_BUF", RING_BUF_BYTE_SIZE)
+            .load(crate::RING_BUF_PINNED)
+            .unwrap();
+        // We assume the map has been pinned as part of the loading process
```
ajwerner: nit: period at the end of sentences

test/integration-test/src/tests/ring_buf.rs
ajwerner: My preference would be that you just make up a random name for your ringbuf so the tests do not interfere with each other and can be run in parallel. Perhaps pass a path to `PinnedRingBufTest::new`?

lblebasheer:
> Thanks for taking a stab at this! Sorry for the slow turnaround, I was traveling. I'll try to keep it to under 24h from here on out.

No problem at all and thank you for taking the time to review. I tried my best to carefully (re)read your comments, so apologies if I'm missing the point you're trying to make.

> Right now the test consumes everything before closing the old ring buffer, so it'd expect that the right thing to happen would be to initializer the consumer to the producer position. In the general case, that's not definitely the right thing.

I went back to re-read how the code works currently. I don't think i'm updating the consumer position at all when a ring buffer is re-opened. Correct me if I'm wrong, but my understanding was that `pos_cache` remembered the last position of the producer since it was last updated from the mmap. With this patch, `pos_cache` is initialized to the last consumer position. Setting `pos_cache` to the consumer position is not "correct", as in, it wasn't a previous producer position. But by doing that we ensure that `pos_cache` is updated to the current producer position the first time we call `.next()` while keeping `pos_cache`'s value pointing to actual valid data.

Also, depending on this discussion, I can add a case where data is added to the ringbuf while the consumer is stopped and before it is restarted.

> This can be fine for a first pass if you don't know the answer and don't feel prepared to push on it, but I'd love to understand the semantics of what's going on here. It seems like what we're saying is that the consumer progress is thrown away. That's sort of sad. Are those the semantics we get from the kernel?

I did a simple experiment to check this in the context of my project. So I stopped my consumer while the producer (which adds items to the ring buffer with a sequence number) was still running. When I restarted the consumer, it picked up where it left off, continuing with the next sequence number item, which I think means that the consumer position handling is correct. If i run the producer long enough while the consumer is stopped, the producer correctly fails to add new items to the ringbuf, and the consumer reads the items that were added until the buffer got full.

test/integration-test/src/tests/ring_buf.rs
lblebasheer: changed this in the latest push, but I get this warning for the ebpf part

```
warning: integration-test@0.1.0:    = note: `#[warn(non_upper_case_globals)]` on by default
warning: integration-test@0.1.0: help: convert the identifier to upper case
warning: integration-test@0.1.0:    |
warning: integration-test@0.1.0: 14 - static pinned_ringbuf_reuse_test: RingBuf = RingBuf::pinned(0, 0);
warning: integration-test@0.1.0: 14 + static PINNED_RINGBUF_REUSE_TEST: RingBuf = RingBuf::pinned(0, 0);
```

test/integration-test/src/tests/ring_buf.rs
lblebasheer: fixed

test/integration-test/src/tests/ring_buf.rs
lblebasheer: I made an attempt to generate a random name for the pinned map. That wasn't too hard for the test case, but I ran into a roadblock when I realized I don't know how to rename a map that's defined in the ebpf part. For example, how do I change the name of a map declared like this at run time?

```
#[map]
static pinned_ringbuf_reuse_test: RingBuf = RingBuf::pinned(0, 0);
```

lblebasheer:
Hi @ajwerner. Coming back to this. I was out sick for 2 weeks. Did the response above to your comment about semantics make sense?

test/integration-test/src/tests/ring_buf.rs
ajwerner: Two options:
* Simplest:     [map_pin_path](https://docs.rs/aya/latest/aya/struct.EbpfLoader.html#method.map_pin_path) and create a unique directory
* Less simple: I think you can have it defined in the ebpf program as not being pinned, initialize the loader with the pinned ringbuf.

ajwerner (commented):
Ah yeah, this makes sense. Sorry I was being dense! This is looking good. If you set the map_pin_path on the loader to a unique directory name, I'll be very pleased!

> @ajwerner reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318#-:-OZLJMdh5C9ckblGtut8:b-ltz6gf)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @lblebasheer)

lblebasheer:
> Two options:
> 
> Simplest: [map_pin_path](https://docs.rs/aya/latest/aya/struct.EbpfLoader.html#method.map_pin_path) and create a unique directory
> Less simple: I think you can have it defined in the ebpf program as not being pinned, initialize the loader with the pinned ringbuf.

Thanks for the tip on letting the loader decide the pin path. I made an attempt at the second option. Happy to make more changes as you see fit. I could create a completely random name for the pinned map if thats more appropriate.

tamird (changes requested):
> @tamird reviewed 3 of 5 files at r1, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @ajwerner and @lblebasheer)

___
*[`test/integration-ebpf/Cargo.toml` line 97 at r2](https://reviewable.io/reviews/aya-rs/aya/1318#-OXVEdWF0bTfOWghKrXy:-OXVEdWG3b247F37icGU:b-h3dfyg) ([raw file](https://github.com/aya-rs/aya/blob/f2d6ba817f424a4d6bc95bc2e0f0d31184941670/test/integration-ebpf/Cargo.toml#L97)):*
> ```TOML
> [[bin]]
> name = "ring_buf_pinned"
> path = "src/ring_buf_pinned.rs"
> ```

please remove this (see other comment)



_Code quote:_
```TOML
[[bin]]
name = "ring_buf_pinned"
path = "src/ring_buf_pinned.rs"
```
___
*[`test/integration-test/src/lib.rs` line 58 at r2](https://reviewable.io/reviews/aya-rs/aya/1318#-OXVElej8ot7EHqvW1VN:-OXVElej8ot7EHqvW1VO:bkle8id) ([raw file](https://github.com/aya-rs/aya/blob/f2d6ba817f424a4d6bc95bc2e0f0d31184941670/test/integration-test/src/lib.rs#L58)):*
> ```XML
>     XDP_SEC => "xdp_sec",
>     UPROBE_COOKIE => "uprobe_cookie",
>     RING_BUF_PINNED => "ring_buf_pinned",
> ```

keep this sorted, put it near RING_BUF
___
*[`test/integration-test/src/tests/ring_buf.rs` line 33 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVO1vh-JAF_cR3Hr3G:-OZVO1vh-JAF_cR3Hr3H:b-luohiz) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-test/src/tests/ring_buf.rs#L33)):*
> ```XML
> }
> 
> struct PinnedRingBufTest {
> ```

I may be missing something but the duplication in this file seems significant. Is it possible to modify this such that all the existing test cases _also_ run against a pinned RB?
___
*[`test/integration-ebpf/src/ring_buf_pinned.rs` line 1 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVNWxM8b0jZHFpEtxJ:-OZVNWxM8b0jZHFpEtxK:bdchlox) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-ebpf/src/ring_buf_pinned.rs#L1)):*
> ```XML
> #![no_std]
> ```

this is exactly the same as `ring_buf.rs`, isn't it? Couldn't you just use that same program from your new userspace test?

actually I think this file is dead code; your userspace test already uses `ring_buf.rs`, not this. Please delete this.

lblebasheer (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @ajwerner and @tamird)

___
*[`test/integration-ebpf/Cargo.toml` line 97 at r2](https://reviewable.io/reviews/aya-rs/aya/1318#-OXVEdWF0bTfOWghKrXy:-OZVWcR46F1ENM73kX8a:bebrt1w) ([raw file](https://github.com/aya-rs/aya/blob/f2d6ba817f424a4d6bc95bc2e0f0d31184941670/test/integration-ebpf/Cargo.toml#L97)):*

please remove this (see other comment)
</blockquote></details>

Will do
___
*[`test/integration-ebpf/src/ring_buf_pinned.rs` line 1 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVNWxM8b0jZHFpEtxJ:-OZVUHFe1IoJU82bmgku:b-oxu1yd) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-ebpf/src/ring_buf_pinned.rs#L1)):*

this is exactly the same as `ring_buf.rs`, isn't it? Couldn't you just use that same program from your new userspace test?

actually I think this file is dead code; your userspace test already uses `ring_buf.rs`, not this. Please delete this.
</blockquote></details>

Will do
sorry I missed that. hadn't realized I circled back to making it the same as `ring_buf.rs`.
___
*[`test/integration-test/src/lib.rs` line 58 at r2](https://reviewable.io/reviews/aya-rs/aya/1318#-OXVElej8ot7EHqvW1VN:-OZVW2c43ruqegkGk44T:bebrt1w) ([raw file](https://github.com/aya-rs/aya/blob/f2d6ba817f424a4d6bc95bc2e0f0d31184941670/test/integration-test/src/lib.rs#L58)):*

actually, please remove this - it is unused.
</blockquote></details>

Will do
___
*[`test/integration-test/src/tests/ring_buf.rs` line 33 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVO1vh-JAF_cR3Hr3G:-OZVUkwi-A-KUCp7PsXi:b-d5vbd2) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-test/src/tests/ring_buf.rs#L33)):*

I may be missing something but the duplication in this file seems significant. Is it possible to modify this such that all the existing test cases _also_ run against a pinned RB?
</blockquote></details>

The duplication has been bothering me too. I initially considered adding some sort of flag that decided whether to use a pinned map or not. I think that will require updating all the tests to add that parameter. Having all the tests run against a pinned map sounds easier, but I haven't really thought about any test coverage we would lose with respect to pinned vs. non-pinned maps, so I'll rely on your advice here.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @ajwerner and @lblebasheer)

___
*[`test/integration-test/src/tests/ring_buf.rs` line 33 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVO1vh-JAF_cR3Hr3G:-OZVmlHj10dQpVhs_ZGe:b-gwaa7x) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-test/src/tests/ring_buf.rs#L33)):*
<details><summary><i>Previously, lblebasheer wrote…</i></summary><blockquote>

The duplication has been bothering me too. I initially considered adding some sort of flag that decided whether to use a pinned map or not. I think that will require updating all the tests to add that parameter. Having all the tests run against a pinned map sounds easier, but I haven't really thought about any test coverage we would lose with respect to pinned vs. non-pinned maps, so I'll rely on your advice here.
</blockquote></details>

We already make extensive use of the `test_case` macros, couldn't you just add another dimension so that each test runs as it does now and again with a pinned map?

Concretely you'd add a boolean to each test function, and to `RingBufTest::new`. Does that make sense?

lblebasheer (commented):
I think I broke something with this change. The test doesn't seem to fail anymore with the index handling fix removed. So I think whats happening is that the first time the ebpf program is loaded, a map is initialized as part of loaded and then the loader can pin that map. I think the problem is that, when I load the program the second time, it initializes a new map (non-pinned) and the now the pinned map from the first load is a distinct map separate from the newly initialized one. hope that makes sense. i think i can fall back to using a `map_pin_path` instead of resorting to something more complex.

> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318#-:-OZVgDH44tq91DuY4S0R:b-gsltum)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @ajwerner)

___
*[`test/integration-test/src/tests/ring_buf.rs` line 33 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVO1vh-JAF_cR3Hr3G:-OZVydy8BBwCPbxp7tia:b718zfn) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-test/src/tests/ring_buf.rs#L33)):*

We already make extensive use of the `test_case` macros, couldn't you just add another dimension so that each test runs as it does now and again with a pinned map?

Concretely you'd add a boolean to each test function, and to `RingBufTest::new`. Does that make sense?
</blockquote></details>

i think so. so the thing is that the test case for this PR is unique in that it loads the ebpf program and opens the same map twice, closing it in-between. referring to the comment to @ajwerner above, my understanding is that maps defined in the ebpf program are created at load time. so if i want to use the same map on the second load of the program, i'd need to go back to using the separate `ring_buf_pinned.rs` ebpf program that defines it as pinned. that is unless @ajwerner has more insight into how it can be done with the map defined as not pinned in the ebpf part.

given that, if I add a `pinned: bool` to `RingBugTest::new`, i think it would only make sense to be `true` for this one test. We could make `ring_buf.rs` use a pinned map and have all the tests run against a pinned map as you originally suggested, but I think we'd lose the ability to run them in parallel since they'll share the name in the filesystem namespace, unless we can rename maps at load time or something similar.

tamird (commented):
> @tamird reviewed 8 of 8 files at r17, 5 of 5 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318)* status: 9 of 12 files reviewed, 3 unresolved discussions (waiting on @ajwerner, @alessandrod, and @lblebasheer)

___
*[`aya/src/bpf.rs` line 315 at r15](https://reviewable.io/reviews/aya-rs/aya/1318#-O_jsX7z1BZJaFVq_Nbt:-Oa5aoKkDJYapaiNGHKG:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/092622e8720786f0d850e03d1f8d98f5a7f35918/aya/src/bpf.rs#L315)):*

should this mention what happens if called multiple times with the same name?
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/ring_buf.rs` line 446 at r16](https://reviewable.io/reviews/aya-rs/aya/1318#-O_juGojBxq6kF4F_Txo:-Oa5apo76Tg2Z3Voim63:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/00baeda0b4f1f170588aa3dec29cb7108ab83f03/test/integration-test/src/tests/ring_buf.rs#L446)):*

this could unconditionally expect success if placed after the `new_with_mutators` call below, no?
</blockquote></details>

Done.

alessandrod (changes requested):
looks good, but see API name comment

tamird:
> looks good, but see API name comment

What comment?

alessandrod:
> > looks good, but see API name comment
> 
> What comment?

Submitted from vscode but GraphQL giving errors...

aya/src/bpf.rs
```diff
@@ -301,6 +308,37 @@ impl<'a> EbpfLoader<'a> {
...
+    ///
+    /// # let mut loader = aya::EbpfLoader::new();
+    /// # let pin_path = Path::new("/sys/fs/bpf/my-pinned-map");
+    /// let bpf = loader
+    ///     .set_map_pin_path("map", pin_path)
+    ///     .load_file("file.o")?;
+    /// # Ok::<(), aya::EbpfError>(())
+    /// ```
+    ///
+    pub fn set_map_pin_path<P: Into<Cow<'a, Path>>>(
```
alessandrod: surely we can't have map_pin_path() and set_map_pin_path()

maybe rename existing map_pin_path() to default_map_pin_path(), and call this map_pin_path()?

ajwerner: The `set_` prefix is used for `set_global` and `set_max_entries`, so it felt fitting here. I don't have strong feelings. I've added a commit to rename `map_pin_path` to `default_map_pin_directory` to clarify that it's not the path to any particular map but rather is a directory in which maps will be referenced by name, and then I left `set_map_pin_path`. If you would like me to remove the `set_` prefix, I'm happy to.

ajwerner (commented):
Sorry for the slow turnaround. See if y'all like this.

I'm also not sure about what popped out from `CARGO_CFG_BPF_TARGET_ARCH=x86_64 cargo +nightly xtask public-api --bless --target x86_64-unknown-linux-gnu`, so I pulled that into its own commit. 

> @ajwerner reviewed 1 of 2 files at r16, 5 of 6 files at r19.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318#-:-OajyJBBBb_0KXAd5sEh:b-oovy9z)* status: 4 of 12 files reviewed, 3 unresolved discussions (waiting on @alessandrod and @tamird)

___
*[`aya/src/maps/ring_buf.rs` line 306 at r16](https://reviewable.io/reviews/aya-rs/aya/1318#-O_jtVxFEoy-PISXXlSt:-OajwO4a0E1Ckmu7J1Q3:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/00baeda0b4f1f170588aa3dec29cb7108ab83f03/aya/src/maps/ring_buf.rs#L306)):*

consider elaborating on this -- currently it just repeats the code but doesn't explain the circumstances under which there would be meaningful data to read.
</blockquote></details>

Done.
___
*[`aya/src/maps/ring_buf.rs` line 408 at r16](https://reviewable.io/reviews/aya-rs/aya/1318#-O_jtiC_6ltCHWJs8li_:-Oajy8jy-kVasJE8y1cY:b-5zk7xj) ([raw file](https://github.com/aya-rs/aya/blob/00baeda0b4f1f170588aa3dec29cb7108ab83f03/aya/src/maps/ring_buf.rs#L408)):*

awkward because "the shared memory map" isn't in the API
</blockquote></details>

Reworked the API to take the mmap. It centralizes the unsafe and feels like an improvement.
___
*[`test/integration-ebpf/src/ring_buf_pinned.rs` line 1 at r6](https://reviewable.io/reviews/aya-rs/aya/1318#-OZVNWxM8b0jZHFpEtxJ:-OajwuF27Lp5Bf6UH0pR:b-cw6svg) ([raw file](https://github.com/aya-rs/aya/blob/47c93e4fb0da7e2a7efb10694115582cf6f04af8/test/integration-ebpf/src/ring_buf_pinned.rs#L1)):*

Done.
</blockquote></details>

Thank you. Thank works. That being said, I saw a bunch of changes I don't think are expected. They're sitting in the first commit.

tamird:
The spurious API changes should go away if you `cargo update` - something changed in the hashbrown crate, I think.

ajwerner (commented):
Alright, should now be good to go. @alessandrod PTAL at your convenience. I'm more than happy to iterate on the method names until you're pleased.

> The spurious API changes should go away if you `cargo update` - something changed in the hashbrown crate, I think.

Thank you. Commit removed.

> @ajwerner reviewed 4 of 5 files at r28, 3 of 3 files at r29, 2 of 2 files at r30, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318#-:-OaniUSB6rHwzlZE7rwK:b-lu0mit)* status: 9 of 12 files reviewed, 3 unresolved discussions (waiting on @alessandrod and @tamird)

aya/src/bpf.rs
alessandrod: it's not used for bpf extension and others tho. Since this is a builder and we don't provide getters, we should probably not have the set prefix anywhere?

ajwerner:
@alessandrod when you find a moment to review the naming choices, I'd appreciate it very much. Looking to get this off my list of in-flight things.

alessandrod:
> @alessandrod when you find a moment to review the naming choices, I'd appreciate it very much. Looking to get this off my list of in-flight things.

see https://github.com/aya-rs/aya/pull/1318#discussion_r2404425273

tamird (approved):
> @tamird reviewed 6 of 6 files at r37, 3 of 3 files at r38, 2 of 2 files at r39, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318)* status: all files reviewed, 1 unresolved discussion (waiting on @alessandrod and @lblebasheer)

aya/src/bpf.rs
ajwerner: Sorry for missing this. I removed this `set_` prefix in this PR and added a CHANGELOG entry for the breaking rename. I'll send another patch with other renames. I propose `set_global` -> `override_global` and `set_max_entries`->`map_max_entries`, but leaving it for a separate PR.

alessandrod: Yep those renames sound good! We can probably also deprecate the old names instead of removing them right away?

alessandrod (approved):
typo in changelog else it's good to go, thanks!

aya/src/bpf.rs
ajwerner: PR inbound.

ajwerner (commented):
Thanks for the reviews, folks.

> typo in changelog else it's good to go, thanks!

Fixed.

> @ajwerner reviewed 5 of 6 files at r37, 1 of 6 files at r40.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1318#-:-Oc5dBfr3fHx3tDNfy8-:b8pcsgb)* status: 8 of 13 files reviewed, 1 unresolved discussion (waiting on @alessandrod and @tamird)

lblebasheer:
Thank you all and especially @ajwerner for all the work on this.

Issue #816, opened 2023-10-15
Missing Cargo.lock


kriomant:
It's not easy to set up environment for eBPF development.
Some kernels have bugs, some package version miss features, some rust versions produce invalid code. So I fixed most versions to be able to reproduce working environment.

However, when I install 'aya-tool', despite I use exact git revision, all it's dependencies are resolved afresh which often causes failure.
Last case was fresh 'anstream' dependency which stopped building with rust version I use. I want to use `cargo install --locked` here, but 'aye-rs' repository miss 'Cargo.lock', so it won't work.

Can you please add 'Cargo.lock'? Note that it won't change current behavior for most users, because lock file is only used when '--locked' options is given and current documentation doesn't use it.

alessandrod:
Which kernels have bugs? Which package versions miss features? Which rust versions produce invalid code?

kriomant:
Sorry, it was just an explanation why I fix all the versions I can.

The real problem is that `cargo install --git https://github.com/aya-rs/aya -- aya-tool` installs different aya-tool binary every time it is called. I can fix version of aya-tool itself with `--git` and `--rev`. But I can't fix versions of aya-tool dependencies, because there is no corresponding Cargo.lock in the repo.

Last time I tried to reproduce my env, 'anstream' dependency was updated and it can no longer be compiled with rust version I use. And updating rust version is a pain.

tamird:
Seems like we're not going to do this.

PR #1320, opened 2025-08-17
add Cargo.lock

cppcoffee:
Fixed https://github.com/aya-rs/aya/issues/816


tamird:
I don't think we want this.

PR #1325, opened 2025-08-23
aya: change StackTrace ID type from u32 to i64

cppcoffee:
Fixed https://github.com/aya-rs/aya/issues/1052

`StackTrace::get_stackid()` return type is `i64`;
`StackTraceMap::get()` parameter requirements should also be standardized to `i64` type to avoid confusion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1325)


cppcoffee:
I understand, thank you for your description.

PR #1337, opened 2025-09-05
Do not make the static mut to prevent verification failure

addisoncrump:
#1335 was incomplete as the static variable was declared mut, putting it into a section that the verifier considers mutable and thus breaking the check. This change simply unmarks it `mut` and adds a corresponding test which should fail if the verifier does not recognise that the loop is not executed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1337)

### DIFF
--- a/test/integration-test/src/tests/log.rs
+++ b/test/integration-test/src/tests/log.rs
@@ -224,3 +224,40 @@ fn log_level_only_error_warn() {
 
     assert_eq!(records.next(), None);
 }
+
+#[test_log::test]
+fn log_level_prevents_verif_fail() {
+    let level = aya_log::Level::Warn as u8;
+    let mut bpf = EbpfLoader::new()
+        .set_global(aya_log::LEVEL, &level, true /* must_exist */)
+        .load(crate::LOG)
+        .unwrap();
+
+    let mut captured_logs = Vec::new();
+    let logger = TestingLogger {
+        log: Mutex::new(|record: &Record| {
+            captured_logs.push(CapturedLog {
+                body: format!("{}", record.args()).into(),
+                level: record.level(),
+                target: record.target().to_string().into(),
+            });
+        }),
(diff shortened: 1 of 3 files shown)


addisoncrump:
> start writing commit messages

No worries, I wanted to get a stable state and then flatten to a single commit once approved :sweat_smile: It's just not in my normal workflow (since I use `-m`).

vadorovsky (approved):
Thanks! :slightly_smiling_face:

test/integration-test/src/tests/log.rs
```diff
@@ -224,3 +224,39 @@ fn log_level_only_error_warn() {
 
     assert_eq!(records.next(), None);
 }
+
+#[test_log::test]
+fn log_level_prevents_verif_fail() {
```
tamird, line 229: You might just remove the other test.

addisoncrump: Since they check for different behaviour (one is checking the correctness of logging w/ different levels, the other for verifier behaviour), I figured it would be best to keep them separate.

test/integration-test/src/tests/log.rs
```diff
@@ -224,3 +224,39 @@ fn log_level_only_error_warn() {
...
+        .try_into()
+        .unwrap();
+    prog.load().unwrap();
+    prog.attach("trigger_ebpf_program", "/proc/self/exe", None, None)
+        .unwrap();
+
+    trigger_ebpf_program();
+    logger.flush();
+
+    let mut records = captured_logs.iter();
```
tamird, line 260: micronit: newline after this to match other tests

tamird (approved):

tamird:
Fixed this up. Merging.

addisoncrump:
Just saw this, thanks.

PR #1344, opened 2025-09-15
aya-ebpf: Add BPF_F_ADJ_ROOM_ENCAP_L2 "macro"

aureliar8:
This function is needed to properly add a L2 header when using `bpf_skb_adjust_room` [1]. As it is originally a C macro, it isn't automatically generated in the `bindings` crate

[1]: https://docs.ebpf.io/linux/helper-function/bpf_skb_adjust_room/

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1344)

### DIFF
--- a/ebpf/aya-ebpf/src/lib.rs
+++ b/ebpf/aya-ebpf/src/lib.rs
@@ -128,6 +128,19 @@ mod intrinsics {
     }
 }
 
+/// Builds a flag for [`SkBuffContext::adjust_room`](programs::SkBuffContext::adjust_room)
+/// that defines L2 encapsulation, using `len` as the inner MAC header length.
+///
+/// Equivalent to the [`BPF_F_ADJ_ROOM_ENCAP_L2`][uapi-bpf-adj-room-encap-l2] macro
+/// in the Linux user-space API.
+///
+/// [uapi-bpf-adj-room-encap-l2]: https://github.com/torvalds/linux/blob/v6.17/include/uapi/linux/bpf.h#L6181
+#[doc(alias = "BPF_F_ADJ_ROOM_ENCAP_L2")]
+#[inline(always)]
+pub const fn bpf_f_adj_room_encap_l2(len: u64) -> u64 {
+    (len & bindings::BPF_ADJ_ROOM_ENCAP_L2_MASK as u64) << bindings::BPF_ADJ_ROOM_ENCAP_L2_SHIFT
+}
+
 /// Check if a value is within a range, using conditional forms compatible with
 /// the verifier.
 #[inline(always)]
--- a/xtask/public-api/aya-ebpf.txt
+++ b/xtask/public-api/aya-ebpf.txt
@@ -3460,4 +3460,5 @@ pub fn aya_ebpf::programs::xdp::XdpContext::gid(&self) -> u32
 pub fn aya_ebpf::programs::xdp::XdpContext::pid(&self) -> u32
 pub fn aya_ebpf::programs::xdp::XdpContext::tgid(&self) -> u32
 pub fn aya_ebpf::programs::xdp::XdpContext::uid(&self) -> u32
+pub const fn aya_ebpf::bpf_f_adj_room_encap_l2(len: u64) -> u64
 pub fn aya_ebpf::check_bounds_signed(value: i64, lower: i64, upper: i64) -> bool


aureliar8:
While looking at the generated bindings, I was also very surprised to see that `BPF_F_ADJ_ROOM_xxx` are `c_uint` / `u32`
https://github.com/aya-rs/aya/blob/0c2cc1f85b8ae06ac1fb40d0d76a8426d31f57de/ebpf/aya-ebpf-bindings/src/x86_64/bindings.rs#L1554-L1563

Even though the kernel definitions for them is a `ULL` https://elixir.bootlin.com/linux/v6.16.4/source/include/uapi/linux/bpf.h#L6132

Shouldn't they be mapped to u64 ?

tamird:
The usual workaround is to add a function (in C) that returns the macro value. As for the other "constants": feel free to try to changing the bindgen options.

aureliar8:
> The usual workaround is to add a function (in C) that returns the macro value

I'm not sure what you mean by that, as I see no C file that I could modify in this repo ? Unless you mean adding this function in the kernel source code ?

I'll have a look at the bindgen options

tamird:
Yes currently we point bindgen at libbpf headers. You'd have to introduce a new headset that `#include`s the libbpf headers *and* defines a function that returns this macro, then point bindgen at it.

tamird:
Wait, that's not right. That won't work without actual ffi. Maybe ask upstream to figure out why this macro can't generate a rust definition?

ebpf/aya-ebpf/src/lib.rs
```diff
@@ -113,6 +113,15 @@ mod intrinsics {
 #[cfg(target_arch = "bpf")]
 pub use intrinsics::*;
 
+// A reimplementation of the BPF_F_ADJ_ROOM_ENCAP_L2(len) macro of the kernel, to use to construct
+// flags to pass to bpf_skb_adjust_room.
+// https://elixir.bootlin.com/linux/v6.16.4/source/include/uapi/linux/bpf.h#L6149
+#[inline(always)]
+#[allow(non_snake_case)]
+pub fn BPF_F_ADJ_ROOM_ENCAP_L2(len: u64) -> u64 {
```
tamird: Let's make this const

ebpf/aya-ebpf/src/lib.rs
```diff
@@ -113,6 +113,15 @@ mod intrinsics {
 #[cfg(target_arch = "bpf")]
 pub use intrinsics::*;
 
+// A reimplementation of the BPF_F_ADJ_ROOM_ENCAP_L2(len) macro of the kernel, to use to construct
+// flags to pass to bpf_skb_adjust_room.
+// https://elixir.bootlin.com/linux/v6.16.4/source/include/uapi/linux/bpf.h#L6149
```
vadorovsky: First of all, let's make it a proper docstring (`///` instead of `//`).

Second of all, let's describe what the function does first. Currently, this comment doesn't explain that at all and it implicitly requires people to already be familiar with kernel/libbpf API. Being an equivalent of `BPF_F_ADJ_ROOM_ENCAP_L2` is something we can mention later as a less important detail. :slightly_smiling_face: 

Last but not least, we should encourage people to use the `adjust_room` method from `SkBuffContext` instead of the raw binding.

I would phrase it the following way:

```suggestion
/// Builds a flag for [`SkBuffContext::adjust_room`] that defines L2 encapsulation,
/// using `len` as the inner MAC header length.
///
/// Equivalent to the [`BPF_F_ADJ_ROOM_ENCAP_L2`][uapi-bpf-adj-room-encap-l2] macro
/// in the Linux user-space API.
///
/// [uapi-bpf-adj-room-encap-l2]: https://elixir.bootlin.com/linux/v6.16.4/source/include/uapi/linux/bpf.h#L6149
```

ebpf/aya-ebpf/src/lib.rs
```diff
@@ -113,6 +113,15 @@ mod intrinsics {
...
+#[inline(always)]
+#[allow(non_snake_case)]
```
vadorovsky: I would actually prefer to stick to the Rust naming conventions and just call it `bpf_d_adj_room_encap_l2`. I think we should limit C style names only to bindings (and also - we should try to wrap as many bindings as possible and limit their direct usage in actual programs).

We could still keep a doc alias though:

```
#[doc(alias = "BPF_F_ADJ_ROOM_ENCAP_L2")]
```

So when someone types the C name, rust-analyzer's autocomplete still points to this function.

aureliar8: Are you sure about following the Rust naming convention ? It's going to look a bit odd to have this function in lowercase while other bindings are in upper case. 
For example the code to perform geneve encapsulation would be the following with the current casing
```rust
let additional_len = EthHdr::LEN + Ipv4Hdr::LEN + UdpHdr::LEN + GeneveHdr::LEN;
ctx.adjust_room(
  additional_len as i32,
  BPF_ADJ_ROOM_MAC, 
  | BPF_F_ADJ_ROOM_ENCAP_L2_ETH as u64 
  | BPF_F_ADJ_ROOM_ENCAP_L2(EthHdr::LEN) 
  | BPF_F_ADJ_ROOM_ENCAP_L3_IPV4 as u64 
  | BPF_F_ADJ_ROOM_ENCAP_L4_UDP as u64, 
)
``` 

With your casing proposal it'd look like that:
```rust
let additional_len = EthHdr::LEN + Ipv4Hdr::LEN + UdpHdr::LEN + GeneveHdr::LEN;
ctx.adjust_room(
  additional_len as i32,
  BPF_ADJ_ROOM_MAC, 
  | BPF_F_ADJ_ROOM_ENCAP_L2_ETH as u64 
  | bpf_f_adj_room_encap_l2(EthHdr::LEN) 
  | BPF_F_ADJ_ROOM_ENCAP_L3_IPV4 as u64 
  | BPF_F_ADJ_ROOM_ENCAP_L4_UDP as u64, 
)
```

And actually I'm also wondering is the definition of `BPF_F_ADJ_ROOM_ENCAP_L2` shouldn't live in the `aya-ebpf-bindings` crate. Because it seems weird  that`BPF_F_ADJ_ROOM_ENCAP_L2` lives in a different crate than 
`BPF_F_ADJ_ROOM_ENCAP_L2_ETH`, `BPF_F_ADJ_ROOM_ENCAP_L3_IPV4` etc...

vadorovsky: Yeah, I think it looks alright. Constants are constants, function calls are function calls. That's how you would write it in idiomatic Rust if there was no bias given by the C API.

You could always keep the function call as the last element of the `|` chain to group the constants together:

```rust
ctx.adjust_room(
  additional_len as i32,
  BPF_ADJ_ROOM_MAC, 
  | BPF_F_ADJ_ROOM_ENCAP_L2_ETH as u64,
  | BPF_F_ADJ_ROOM_ENCAP_L3_IPV4 as u64 
  | BPF_F_ADJ_ROOM_ENCAP_L4_UDP as u64,
  | bpf_f_adj_room_encap_l2(EthHdr::LEN)
)
```

aureliar8: Done :) 

The netifly CI seems to have a problem. Should I rebase or do something or is it fine to keep it like that ?

tamird: never hurts to rebase. please squash the commits as well

ebpf/aya-ebpf/src/lib.rs
tamird: can this function be const?

tamird (commented):
> @tamird reviewed 2 of 2 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1344)* status: all files reviewed, 4 unresolved discussions (waiting on @aureliar8 and @vadorovsky)

___
*[`ebpf/aya-ebpf/src/lib.rs` line 123 at r10](https://reviewable.io/reviews/aya-rs/aya/1344#-Odxhclj149mr51VNF0L:-OdxhclkByz28p65x5Ia:b-ljnduh) ([raw file](https://github.com/aya-rs/aya/blob/04c0ba92de4f2ed68cc8ef6c3128f8dc20d106a1/ebpf/aya-ebpf/src/lib.rs#L123)):*
> ```XML
> /// in the Linux user-space API.
> ///
> /// [uapi-bpf-adj-room-encap-l2]: https://elixir.bootlin.com/linux/v6.16.4/source/include/uapi/linux/bpf.h#L6149
> ```

would you mind making this a github link instead of elixir?

ebpf/aya-ebpf/src/lib.rs
aureliar8: Yes sorry !

aureliar8:
> would you mind making this a github link instead of elixir?

Done :)

tamird (approved):
> @tamird reviewed 2 of 2 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1344)* status: all files reviewed, 3 unresolved discussions (waiting on @aureliar8 and @vadorovsky)

___
*[`-- commits` line 8 at r10](https://reviewable.io/reviews/aya-rs/aya/1344#-OdxroxPETgk8gyu4rI6:-OdxroxPETgk8gyu4rI7:beu1fy9):*
can you restore this? we do not retain the PR description, so if you care about this being in `git log` it must be here in the commit



_Code quote:_
```
  This macro is needed to properly add a L2 header when using
  bpf_skb_adjust_room [1]. As it is originally a C macro, it isn't
  automatically generated in the `bindings` crate

  [1]: https://docs.ebpf.io/linux/helper-function/bpf_skb_adjust_room/
```

PR #1350, opened 2025-09-23
aya-log: add `#[must_use]` attribute to `EbpfLogger`

thomaseizinger:
First attempt at documenting the flushing requirement on the `EbpfLogger`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1350)


aya-log/src/lib.rs
```diff
@@ -95,6 +95,7 @@ unsafe impl Pod for DisplayHintWrapper {}
 /// Log messages generated by `aya_log_ebpf` using the [log] crate.
 ///
 /// For more details see the [module level documentation](crate).
+#[must_use = "The logger needs to be continuously flushed to actually receive log messages in userspace."]
```
tamird: The more important thing is that dropping it will cause program loading to fail because the map FD will have been closed.

Both are important (futures are must_use because they do nothing unless polled) but I think the critical thing is that the logger must not be dropped even if logs are never consumed.

PR #1365, opened 2025-10-10
perf_event: support hardware breakpoints

FridayOrtiz:
Implement `PerfEventConfig::Breakpoint`, allowing users to attach
hardware breakpoints. Generate `HW_BREAKPOINT_*` and
`struct bpf_perf_event_data` in support of this feature and update the type
of`PerfEventContext` accordingly.

Add a test exercising R, W, RW, and X breakpoints. Note that R
breakpoints are unsupported on x86, and this is asserted in the test.

Extend the VM integration test harness and supporting infrastructure
(e.g. `download_kernel_images.sh`) to download kernel debug packages and
mount `System.map` in initramfs. This is needed (at least) on the aarch
6.1 Debian kernel which was not compiled with `CONFIG_KALLSYMS_ALL=y`
for some reason, and the locations of globals are not available in
kallsyms. To attach breakpoints to these symbols in the test pipeline,
we need to read them from System.map and apply the KASLR offset to get
their real address. The `System.map` file is not provided in the kernel
package by default, so we need to extract it from the corresponding
debug package. The KASLR offset is computed using `gunzip` which appears
in kallsyms on all Debian kernels tested.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1365)

### DIFF
--- a/.github/scripts/download_kernel_images.sh
+++ b/.github/scripts/download_kernel_images.sh
@@ -8,6 +8,12 @@ if [ "$#" -lt 3 ]; then
   exit 1
 fi
 
+escape_regex() {
+  # Escape characters that have special meaning in extended regular expressions so that
+  # we can safely interpolate package names into grep patterns.
+  printf '%s\n' "$1" | sed 's/[][(){}.^$*+?|\\-]/\\&/g'
+}
+
 OUTPUT_DIR=$1
 ARCHITECTURE=$2
 shift 2
@@ -25,6 +31,25 @@ for VERSION in "${VERSIONS[@]}"; do
     exit 1
   }
   FILES+=("$match")
+
+  # The debug package contains the actual System.map. Debian has transitioned
+  # between -dbg and -dbgsym suffixes, so match either for the specific kernel
+  # we just selected.
+  kernel_basename=$(basename "$match")
+  kernel_prefix=${kernel_basename%%_*}
+  kernel_suffix=${kernel_basename#${kernel_prefix}_}
+  base_prefix=${kernel_prefix%-unsigned}
+
+  base_prefix_regex=$(escape_regex "$base_prefix")
+  kernel_suffix_regex=$(escape_regex "$kernel_suffix")
+
+  DEBUG_REGEX="${base_prefix_regex}-dbg(sym)?_${kernel_suffix_regex}"
+  debug_match=$(printf '%s\n' "$URLS" | grep -E "$DEBUG_REGEX" | sort -V | tail -n1) || {
+    printf 'Failed to locate debug package matching %s\n%s\nVERSION=%s\nREGEX=%s\n' \
+      "$kernel_basename" "$URLS" "$VERSION" "$DEBUG_REGEX" >&2
+    exit 1
+  }
+  FILES+=("$debug_match")
 done
 
 # Note: `--etag-{compare,save}` are not idempotent until curl 8.9.0 which included
--- a/aya-ebpf-macros/src/perf_event.rs
+++ b/aya-ebpf-macros/src/perf_event.rs
@@ -29,7 +29,7 @@ impl PerfEvent {
             #[unsafe(no_mangle)]
             #[unsafe(link_section = "perf_event")]
             #vis fn #fn_name(ctx: *mut ::core::ffi::c_void) -> u32 {
-               let _ = #fn_name(::aya_ebpf::programs::PerfEventContext::new(ctx));
+               let _ = #fn_name(::aya_ebpf::programs::PerfEventContext::new(ctx.cast()));
                return 0;
 
                #item
@@ -60,7 +60,7 @@ mod tests {
             #[unsafe(no_mangle)]
             #[unsafe(link_section = "perf_event")]
             fn foo(ctx: *mut ::core::ffi::c_void) -> u32 {
-               let _ = foo(::aya_ebpf::programs::PerfEventContext::new(ctx));
+               let _ = foo(::aya_ebpf::programs::PerfEventContext::new(ctx.cast()));
                return 0;
 
                fn foo(ctx: PerfEventContext) -> i32 {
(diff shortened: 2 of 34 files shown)


FridayOrtiz:
@tamird 
> Have you thought about how you'd test this?

I just pushed a test up to the branch. It's failing in CI on the aarch64 6.1 kernel because it's unable to read kallsyms for some reason. I can't replicate the failure locally. Any insight?

> can we push this cast to the caller?

This was how I originally had it, but the other program types handle it for the caller (e.g., SkBuff, ProbeContext). As far as I can tell this is always the correct context type for this program type (https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_PERF_EVENT/), the bindings would just have to be re-generated for kernels older than 4.17. Is that the issue?

>I think https://github.com/aya-rs/aya/pull/1038 takes us in the direction of being able to detect this at compile time. Do we want to base this PR on that one?
> do we need this check, or can we just let the syscall fail? otherwise, do we want to make this API more type-safe? again see https://github.com/aya-rs/aya/pull/1038.

Yeah I think that approach makes more sense, I'll pull #1038 into this branch and base the interface around that.

tamird:
> I just pushed a test up to the branch. It's failing in CI on the aarch64 6.1 kernel because it's unable to read kallsyms for some reason. I can't replicate the failure locally. Any insight?

The error handling in that function is non-existent, so it is not possible to determine why it fails. Perhaps you could make it a bit stricter.

FridayOrtiz:
@tamird 
> The error handling in that function is non-existent, so it is not possible to determine why it fails. Perhaps you could make it a bit stricter.

For some reason the 6.1 aarch64 debian kernel was built without defining `CONFIG_KALLSYMS_ALL`, so few non-text symbols were available. I changed the tests to grab the corresponding debug package and copy `System.map` from that, then use it to determine the address to set the breakpoint at if the symbol can't be found in kallsyms. 

I also incorporated #1038 to clean up the interface, which looks like it _just_ got merged. 😅

FridayOrtiz:
@tamird 

> Can you remove this now that it should be verifiable at compile time?

No problem, it was left in by mistake.

> this API ... sucks

Take it up with the kernel maintainers, I copied it from the man page and source. It should look a bit more sensible now. The caller still needs to have some idea of length for non-execute breakpoints, but they only have valid choices now.

> why is this u64?

The kernel API expects u64, check any of the `aya-obj/src/generated/linux_bindings_<arch>.rs` files for `bp_len`. I'm enforcing it here to try to match what #1038 did. 

> instead of this boolean and passing the entire config below, why don't you pass Option<Breakpoint>? you should grab it in the big match above.

I tried both ways and this way felt vaguely better, not a problem to switch it back.

> did this need to change?

I don't think so, rust-analyzer complained about importing both `perf_attach` (function) and `perf_attach` (module) at some point in development so I just fully qualified it and forgot about it. I'll change it back.

> thinking out loud: all this information is available to us at compile time, and we already have a build script in this crate. Can we put all this ceremony in there, and just bake the magic number into this binary? then we would avoid baking these files into the image.

Probably, yes. I'm unconsciously biased towards getting this information at runtime because of my work which is why I wrote it this way in the first place. For transparency, I don't think I have time to rewrite this part and after tomorrow I'm not going to be able to look at this PR for a few weeks.

> the comment says modprobe_path, but that's not here. can you elaborate on that comment so the connection is clear? i guess this file is a pseudo thing that is backed by the symbol modprobe_path?

That's correct. I'll make the comments more clear and reference the source. Here's where that's actually configured in sysctl: https://elixir.bootlin.com/linux/v6.1.155/source/kernel/sysctl.c#L1770

> by the way, is it necessary to actually read it, or can we just open it? finally, please avoid let _ unless you expect an error. We prefer strict code over lax code because it is more informative to the reader.

The file needs to actually be read. I can unwrap the error so it fails if the read is unsuccessful for any reason. The test should fail anyway because the breakpoint won't fire if the read fails, but it'll be more clear why.

> hold up, something isn't right here. you can't just use the same regex, you need to use the exact version that was picked above in $match.

Yeah, I'll tighten this up.

tamird:
@FridayOrtiz could you please reply in reviewable?

FridayOrtiz:
I'm not really comfortable connecting my GH to some random appspot app with ~1000 users, sorry.

--
Friday Ortiz | https://ortiz.sh/

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

tamird (approved):
> @tamird reviewed 1 of 27 files at r1, 6 of 35 files at r8, 1 of 1 files at r10, 35 of 35 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1365)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @alessandrod)

___
*[`test/integration-test/src/tests/perf_event_bp.rs` line 20 at r6](https://reviewable.io/reviews/aya-rs/aya/1365#-ObhQvMQ8VY0tbX5q58B:-OdfDClK6O7kkrczXayd:ban8ud5) ([raw file](https://github.com/aya-rs/aya/blob/38e3f7fce2d6bc7147f05ef7cd0538538cc3d0ce/test/integration-test/src/tests/perf_event_bp.rs#L20)):*

> Probably, yes. I'm unconsciously biased towards getting this information at runtime because of my work which is why I wrote it this way in the first place. For transparency, I don't think I have time to rewrite this part and after tomorrow I'm not going to be able to look at this PR for a few weeks.

A few weeks isn't that long -- I'd prefer we get this right. It's can be hard to remove stuff from the image once added.
</blockquote></details>

I've reversed course on this since it's preferable to have the test binaries preserve the property that they can be built on machine A and run on machine B.

tamird:
I'm planning to merge this after #1385 -- @FridayOrtiz if you'd like to look it over, that would be great. Thanks again.

tamird:
@codex review

PR #1368, opened 2025-10-15
feature: pointer logging

ChristianJacobsen:
This PR adds support for logging pointers in eBPFs.

~~You will have to use either `:p`, `:x`, or `:X` when logging a pointer type.
The `Argument` trait has been implemented for _any_ `*const T` or `*mut T`.
I might have added more than strictly necessary, i.e. I could have made use of `ArgumentKind::Usize`, but I found it more prudent to make it a dedicated `Pointer` kind instead.~~

~~I've added the pointer logging test like you asked in Discord @tamird.~~

~~I have _not_ blessed the new public API, so the `lint` workflow fails. I have a patch locally that blesses it, but I wanted your opinion on whether I should first, @tamird.~~

Pointers can be logged with the `:p` hint and has `0x` prefixed like in regular stdlib.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1368)

### DIFF
--- a/aya-log-common/src/lib.rs
+++ b/aya-log-common/src/lib.rs
@@ -97,6 +97,10 @@ impl LowerMacFormatter for [u8; 6] {}
 pub trait UpperMacFormatter {}
 impl UpperMacFormatter for [u8; 6] {}
 
+pub trait PointerFormatter {}
+impl<T> PointerFormatter for *const T {}
+impl<T> PointerFormatter for *mut T {}
+
 #[repr(u8)]
 #[derive(Copy, Clone, Debug, IntoPrimitive)]
 pub enum RecordFieldKind {
@@ -144,6 +148,8 @@ pub enum ArgumentKind {
 
     Bytes,
     Str,
+
+    Pointer,
 }
 
 /// All display hints
@@ -162,6 +168,8 @@ pub enum DisplayHint {
     LowerMac,
     /// `:MAC`
     UpperMac,
+    /// `:p`
+    Pointer,
 }
 
 mod sealed {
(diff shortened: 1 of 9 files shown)


tamird (approved):
> @tamird reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1368)* status: all files reviewed, 1 unresolved discussion (waiting on @ChristianJacobsen)

___
*[`aya-log/src/lib.rs` line 780 at r3](https://reviewable.io/reviews/aya-rs/aya/1368#-Obg3FO2D0Ta6l4I248S:-Obg3FO2D0Ta6l4I248T:bess61d) ([raw file](https://github.com/aya-rs/aya/blob/40e2bafe75792c2140d5fbbd4cee4fbe86026219/aya-log/src/lib.rs#L780)):*
> ```XML
>                     .try_into()
>                     .map_err(|std::array::TryFromSliceError { .. }| ())?;
>                 let ptr = usize::from_ne_bytes(value) as *const c_void;
> ```

nit: use `*const ()` to avoid the c_void import

ChristianJacobsen (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1368)* status: all files reviewed, 1 unresolved discussion (waiting on @ChristianJacobsen)

___
*[`aya-log/src/lib.rs` line 780 at r3](https://reviewable.io/reviews/aya-rs/aya/1368#-Obg3FO2D0Ta6l4I248S:-ObgBzHf4BmLjIg8mcSw:b-1a99k8) ([raw file](https://github.com/aya-rs/aya/blob/40e2bafe75792c2140d5fbbd4cee4fbe86026219/aya-log/src/lib.rs#L780)):*

nit: use `*const ()` to avoid the c_void import
</blockquote></details>

Will do. Just a brain fart on my side.

PR #1380, opened 2025-10-29
aya-build: restore interfaces

tamird:
- **aya-build: use OsString::into_string**
- **aya-build: clarify naming**
- **aya-build: simplify**
- **aya-build: read AYA_BPF_TARGET_ARCH**
- **aya-build: guess `bpf_target_arch` from `HOST`**

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1380)

### DIFF
--- a/aya-build/src/lib.rs
+++ b/aya-build/src/lib.rs
@@ -16,16 +16,11 @@ pub struct Package<'a> {
     pub root_dir: &'a str,
 }
 
-fn target_arch() -> Cow<'static, str> {
-    const TARGET_ARCH: &str = "CARGO_CFG_TARGET_ARCH";
-    let target_arch = env::var_os(TARGET_ARCH).unwrap_or_else(|| panic!("{TARGET_ARCH} not set"));
-    let target_arch = target_arch
-        .into_string()
-        .unwrap_or_else(|err| panic!("OsString::into_string({TARGET_ARCH}): {err:?}"));
+fn target_arch_fixup(target_arch: Cow<'_, str>) -> Cow<'_, str> {
     if target_arch.starts_with("riscv64") {
         "riscv64".into()
     } else {
-        target_arch.into()
+        target_arch
     }
 }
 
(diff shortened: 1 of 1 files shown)


vadorovsky (changes requested):
The diff looks good to me, but

* You need to change this to `AYA_BPF_TARGET_ARCH` to make the build green.
  * https://github.com/aya-rs/aya/blob/3a3c45100942c903c21e69595eff9e4289910835/.github/workflows/ci.yml#L181
* Perhaps we could also build `integration-ebpf` the same way (the same place should work), so we prove that @alessandrod's use case works.

tamird:
> The diff looks good to me, but
> 
> * You need to change this to `AYA_BPF_TARGET_ARCH` to make the build green.
>   
>   * https://github.com/aya-rs/aya/blob/3a3c45100942c903c21e69595eff9e4289910835/.github/workflows/ci.yml#L181
> * Perhaps we could also build `integration-ebpf` the same way (the same place should work), so we prove that @alessandrod's use case works.

Nah I don't want to use `AYA_BPF_TARGET_ARCH` there because of the caching reasons, but I fixed it to use `RUSTFLAGS`.

aya-build/src/lib.rs
```diff
@@ -202,16 +203,45 @@ impl<'a> Toolchain<'a> {
...
+    // not ergonomic. In contrast to RUSTFLAGS this mechanism reuses the target
+    // cache for all values, producing many more invalidations.
+    const AYA_BPF_TARGET_ARCH: &str = "AYA_BPF_TARGET_ARCH";
+    println!("cargo:rerun-if-env-changed={AYA_BPF_TARGET_ARCH}");
+
+    const HOST: &str = "HOST";
+    println!("cargo:rerun-if-env-changed={HOST}");
+
+    if std::env::var_os(BPF_TARGET_ARCH).is_none() {
+        let host = std::env::var_os(HOST).unwrap_or_else(|| panic!("{HOST} not set"));
```
alessandrod, line 224: shouldn't we check CARGO_CFG_TARGET_ARCH before HOST for the cross compilation case? If I'm compiling on x86_64 for aarch64, I need bpf_target_arch=aarch64 or things won't work

tamird: we just aren't checking `CARGO_CFG_TARGET_ARCH` at all here anymore. If you're cross compiling you are gonna set `AYA_BPF_TARGET_ARCH` because `CARGO_CFG_TARGET_ARCH` is going to be "bpfel".

alessandrod: as per discord convo, this code only gets called in the --target=bpf* build.rs

PR #1381, opened 2025-11-02
aya-build: allow setting features to be enabled in ebpf by aya-build

timwoocker:
Right now when building an ebpf program using aya-build, you can't enable or disable any features. To change this, I introduced another Package attribute which takes a Vec of features to enable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1381)

### DIFF
--- a/aya-build/src/lib.rs
+++ b/aya-build/src/lib.rs
@@ -11,9 +11,12 @@ use std::{
 use anyhow::{Context as _, Result, anyhow};
 use cargo_metadata::{Artifact, CompilerMessage, Message, Target};
 
+#[derive(Default)]
 pub struct Package<'a> {
     pub name: &'a str,
     pub root_dir: &'a str,
+    pub no_default_features: bool,
+    pub features: &'a [&'a str],
 }
 
 fn target_arch_fixup(target_arch: Cow<'_, str>) -> Cow<'_, str> {
@@ -62,7 +65,13 @@ pub fn build_ebpf<'a>(
     let bpf_target_arch = target_arch_fixup(bpf_target_arch.into());
     let target = format!("{target}-unknown-none");
 
-    for Package { name, root_dir } in packages {
+    for Package {
+        name,
+        root_dir,
+        no_default_features,
+        features,
+    } in packages
+    {
         // We have a build-dependency on `name`, so cargo will automatically rebuild us if `name`'s
         // *library* target or any of its dependencies change. Since we depend on `name`'s *binary*
         // targets, that only gets us half of the way. This stanza ensures cargo will rebuild us on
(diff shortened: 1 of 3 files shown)


aya-build/src/lib.rs
```diff
@@ -14,6 +14,7 @@ use cargo_metadata::{Artifact, CompilerMessage, Message, Target};
 pub struct Package<'a> {
     pub name: &'a str,
     pub root_dir: &'a str,
+    pub enabled_features: Vec<&'a str>,
```
tamird: Can this be a slice?

timwoocker: Ahh you mean `&'a [&'a str]`? Yeah that should probably be fine as well

timwoocker: Updated in below commit

tamird:
Can you explain the motivation for this? Do you actually have a BPF program that you build with different sets of features?

timwoocker:
Yes, I have an XDP program which behaves differently on different setups because of different NIC capabilities and required features.

Basically, some NICs support hardware UDP flow steering to steer packets to a specific RX queue. In this case, my XDP program just redirects all incoming packets on that queue to my user-space without checking anything.
If the NIC does not support hardware flow steering, I need to parse the UDP packet, lookup an XDP socket in a HashMap and forward the packet to that specific socket.

I used to solve this by just creating two XDP programs using `#[xdp]` but now I need to configure even more dynamically configurable features like whether the XDP program should redirect specific multicast packets, etc..
Since this is about high-performance networking, I don't want unnecessary "if-else" blocks to check for enabled features in my XDP program during runtime. Creating indivudual XDP programs for each optional feature would mean I'd need to create 2^(n-1) XDP programs (where n is the number of features).

Using features I can enable different features at compile time while only having a single XDP program.

tamird:
Thanks for explaining. Do we need a treatment for default features here, too?

timwoocker:
We could add another bool field `no_default_features` which passes the `--no-default-features` flag.
I don't like the fact that changes like this change the API for creating an instance of the `Package` type. Maybe we should have a constructor which only sets the required attributes (`name`, `root_dir`) and then individual setters for additional attributes like `enabled_features` and `no_default_features`?
That way in the future if you were to add new fields to `Package`, the API wouldn't break.

tamird:
> We could add another bool field `no_default_features` which passes the `--no-default-features` flag.

Let me noodle on this.

> I don't like the fact that changes like this change the API for creating an instance of the `Package` type. Maybe we should have a constructor which only sets the required attributes (`name`, `root_dir`) and then individual setters for additional attributes like `enabled_features` and `no_default_features`? That way in the future if you were to add new fields to `Package`, the API wouldn't break.

We could also derive `Default` and let users write `...Default::default()`.

tamird:
Thanks @timwoocker !

PR #1386, opened 2025-11-09
perf_event: minor API improvements

tamird:
- **perf_event: push down type safety**
- **perf_event: simplify PerfEventScope**

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1386)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,29 +119,26 @@ jobs:
           RUST_BACKTRACE: full
         run: |
           set -euxo pipefail
-          cargo hack test --all-targets --feature-powerset \
+          cargo hack test --all-targets \
             --exclude aya-ebpf \
             --exclude aya-ebpf-bindings \
             --exclude aya-log-ebpf \
             --exclude integration-ebpf \
             --exclude integration-test \
-            --exclude xtask \
-            --workspace
+            --feature-powerset
 
       - name: Doctests
         env:
           RUST_BACKTRACE: full
         run: |
           set -euxo pipefail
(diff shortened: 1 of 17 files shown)


tamird:
@codex review

aya/src/programs/perf_event.rs
```diff
@@ -276,23 +270,32 @@ pub enum HwCacheResult {
...
 pub enum SamplePolicy {
     /// Period
     Period(u64),
     /// Frequency
     Frequency(u64),
 }
 
+/// Wakeup Policy
+#[derive(Debug, Clone, Copy)]
+pub enum WakeupPolicy {
```
alessandrod: does this have to be pub?

tamird: Not as of this commit, but it should probably be part of the attach API. Which way would you rather I go? (not pub or add to the API)

alessandrod: non pub pls, unclear that this should be API, can't think of a case where you want to wake up on bytes

tamird: You might want to wake up on every N events though rather than every single event.

alessandrod: yeah I just realized with this PR that we might be waking up at each event but that seems to horrible to be true, I'm gonna check the kernel but no need to stop this PR in the meantime

tamird: Done.

tamird: lmao yes horrible indeed

PR #1396, opened 2025-11-15
doc(aya): document tcx link pinning for atomic program replacement

### LINKED ISSUE #1395: Support for pinning SchedClassifier links after attachment
## Use Case

We're building a network flow monitoring tool that attaches eBPF TC programs to network interfaces in Kubernetes. When pods restart (frequently in cloud-native environments), we need to atomically replace old eBPF programs with new ones without packet drops or traffic interruption.

This is a critical requirement for production Kubernetes deployments where:

- Pods restart frequently (rolling updates, autoscaling, node drains)
- Zero-downtime is expected
- Manual intervention is impossible (ephemeral, automated environments)

## Current Limitation - No Workaround Exists

The current Aya API doesn't support pinning `SchedClassifier` links after attachment:

```rust
// Current API - returns only the link ID, not the link object
let link_id: SchedClassifierLinkId = program
    .attach_with_options(interface, attach_type, options)?;

// Cannot pin the link because we don't have the SchedClassifierLink object
// link_id.pin("/sys/fs/bpf/my_link")?;  // ❌ No such method
```

The `SchedClassifierLink` objects are stored internally in `ProgramData` and are not accessible after attachment. This prevents us from:

1. **Pinning links to bpffs** for persistence across process restarts
2. **Using `attach_to_link()` for atomic replacement** - while this method exists, we can't obtain the old link to replace it

### TCX Mode (kernel >= 6.6)

- Orphaned programs from crashed/restarted processes remain attached
- Cannot be removed without the original link_id
- TCX allows multiple programs, so new attachments succeed, but orphans waste resources
- Kernel eventually GCs orphans when old process FDs close (not immediate)

### Netlink Mode (kernel < 6.6)

- Can use `qdisc_detach_program()` to remove by name (workaround exists)
- But this doesn't provide atomic replacement (brief window with no program attached)

## Proposed Solution

Add support for accessing and pinning links after attachment. Here are a few potential API designs:

### Option 1: Return the link object directly

```rust
pub fn attach_with_options(
    &mut self,
    interface: &str,
    attach_type: TcAttachType,
    options: TcAttachOptions,
) -> Result<SchedClassifierLink, ProgramError>
```

Then users could pin it:

```rust
let link = program.attach_with_options(interface, attach_type, options)?;
link.pin("/sys/fs/bpf/my_link")?;
```

### Option 2: Add a `pin` method that accepts link_id

```rust
pub fn pin_link(
    &mut self,
    link_id: SchedClassifierLinkId,
    path: &str,
) -> Result<(), ProgramError>
```

### Option 3: Expose the link object via the link_id

```rust
pub fn get_link(&self, link_id: SchedClassifierLinkId) -> Option<&SchedClassifierLink>
```

## Desired Workflow

With link pinning support, we could implement atomic replacement:

```rust
// On startup, check for orphaned pinned link from previous instance
let old_link = if Path::new(pin_path).exists() {
    Some(PinnedLink::from_pin(pin_path)?)
} else {
    None
};

// Either atomically replace old program or attach new
let link = if let Some(old) = old_link {
    // Atomic replacement - no packet drops!
    program.attach_to_link(old.try_into()?)?
} else {
    // First attachment
    program.attach_with_options(interface, attach_type, options)?
};

// Pin for next restart
link.pin(pin_path)?;
```

## Benefits

1. **Zero-downtime restarts** - atomic replacement with no packet loss
2. **Proper cleanup** - can always detach/replace orphaned programs
3. **Better resource management** - no orphaned programs consuming kernel resources
4. **Kubernetes-friendly** - handles frequent pod restarts gracefully without manual intervention
5. **Works for both TCX and netlink modes** - unified approach
6. **Production-ready** - fully automated, no external tools or scripts required

## Impact on Production Users

**Without this feature, production Kubernetes users must choose between:**

1. **Accept resource leaks** - orphaned programs accumulate with every restart
2. **Accept packet drops** - use netlink mode with detach/reattach (non-atomic)
3. **Don't use eBPF TC in Kubernetes** - this is currently the safest option

This limitation makes eBPF TC programs unsuitable for production Kubernetes deployments, despite TCX being specifically designed for modern kernels.

### Manual bpftool pinning - ❌ Impossible in Kubernetes

```bash
# This requires manual intervention after EVERY pod restart
bpftool link show | grep tcx
bpftool link pin id <LINK_ID> /sys/fs/bpf/mermin/links/eth0_ingress
```

**Problem**: In Kubernetes, pods are ephemeral and restart automatically. There's no opportunity for manual commands between pod starts. Init containers can't pin links from the *previous* pod (they're in a new pod).

### Lifecycle hooks with bpftool - ❌ Race conditions

```yaml
lifecycle:
  postStart:
    exec: ["bpftool", "link", "pin", "..."]
```

**Problem**:

1. The link doesn't exist until *after* mermin attaches (chicken-and-egg)
2. Requires parsing bpftool output (fragile, error-prone)
3. No guarantee the old link is still available when new pod starts
4. Adds external dependency (bpftool must be in container)

### Accept orphaned programs - ❌ Wastes resources, breaks monitoring

**Problem**:

1. Every pod restart leaves orphaned programs consuming kernel memory
2. After N restarts, you have N orphaned programs (resource leak)
3. Monitoring becomes confused (which program is actually processing packets?)
4. Kernel eventually GCs orphans, but timing is unpredictable

### Netlink mode fallback - ❌ Not available on modern kernels

**Problem**: Kernel >= 6.6 (GKE, AKS, EKS standard) requires TCX. Netlink is being deprecated.

## Why This Can't Be Worked Around in Userspace

The fundamental issue is that Aya hides the link object after attachment:

```rust
// After attach, we only have an ID, not the link object
let link_id: SchedClassifierLinkId = program.attach_with_options(...)?;

// The actual SchedClassifierLink is stored in ProgramData<SchedClassifierLink>
// but there's no API to access it: program.data.links is private
// The FdLinkId contains RawFd but it's pub(crate) - not accessible
// Even iter_link_ids() is private in aya::sys
```

Every potential workaround hits the same wall: we can't access the link object to pin it.

## Current Non-Solutions

- **Netlink mode**: Use `qdisc_detach_program()` by name (works but has packet drop window during replacement)
- **TCX mode**: Accept orphaned programs until kernel GC (resource leak, confusing monitoring)
- **Manual pinning**: Impossible in automated/ephemeral environments like Kubernetes
- **External tooling**: Can't reliably correlate links to programs across pod restarts

None of these provide atomic replacement in production Kubernetes environments.

## Related Code

- `SchedClassifier::attach_to_link()` exists but is unusable without pinned links
- `PinnedLink::from_pin()` works for loading pinned links
- `SchedClassifierLink` has the internal structure but isn't exposed after attachment

## References

- Aya version: 0.13.1

Would this enhancement be acceptable? I would be happy to contribute a PR if there's agreement on the API design.

svencowart:
Add documentation and integration test for SchedClassifier link pinning, which enables zero-downtime program updates in production environments.

- Add link pinning example to SchedClassifier::attach_with_options() showing atomic replacement workflow for TCX mode (kernel >= 6.6).
- Add pin_tcx_link() integration test verifying link persistence across program unload and atomic replacement capability.

Closes #1395

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1396)

### DIFF
--- a/aya/src/programs/tc.rs
+++ b/aya/src/programs/tc.rs
@@ -205,12 +205,55 @@ impl SchedClassifier {
     ///
     /// The returned value can be used to detach, see [SchedClassifier::detach].
     ///
+    /// # Link Pinning (TCX mode, kernel >= 6.6)
+    ///
+    /// Links can be pinned to bpffs for atomic replacement across process restarts.
+    ///
+    /// ```no_run
+    /// # use std::{io, path::Path};
+    ///
+    /// # use aya::{
+    /// #     programs::{
+    /// #         LinkOrder, SchedClassifier, TcAttachType,
+    /// #         links::{FdLink, LinkError, PinnedLink},
+    /// #         tc::TcAttachOptions,
+    /// #     },
+    /// #     sys::SyscallError,
+    /// # };
+    ///
+    /// # let mut bpf = aya::Ebpf::load(&[])?;
+    /// # let prog: &mut SchedClassifier = bpf.program_mut("prog").unwrap().try_into()?;
+    /// # prog.load()?;
+    /// let pin_path = "/sys/fs/bpf/my_link";
+    ///
+    /// let link_id = match PinnedLink::from_pin(pin_path) {
+    ///     Ok(old) => {
+    ///         let link = FdLink::from(old).try_into()?;
+    ///         prog.attach_to_link(link)?
+    ///     }
+    ///     Err(LinkError::SyscallError(SyscallError { io_error, .. }))
+    ///         if io_error.kind() == io::ErrorKind::NotFound =>
+    ///     {
+    ///         prog.attach_with_options(
+    ///             "eth0",
+    ///             TcAttachType::Ingress,
+    ///             TcAttachOptions::TcxOrder(LinkOrder::default()),
+    ///         )?
+    ///     }
+    ///     Err(e) => return Err(e.into()),
+    /// };
+    ///
+    /// let link = prog.take_link(link_id)?;
+    /// let fd_link: FdLink = link.try_into()?;
+    /// fd_link.pin(pin_path)?;
+    /// # Ok::<(), Box<dyn std::error::Error>>(())
+    /// ```
+    ///
     /// # Errors
     ///
     /// [`TcError::NetlinkError`] is returned if attaching fails. A common cause
     /// of failure is not having added the `clsact` qdisc to the given
-    /// interface, see [`qdisc_add_clsact`]
-    ///
+    /// interface, see [`qdisc_add_clsact`].
     pub fn attach_with_options(
         &mut self,
         interface: &str,
--- a/test/integration-test/src/tests/load.rs
+++ b/test/integration-test/src/tests/load.rs
@@ -6,11 +6,13 @@ use aya::{
     maps::Array,
     pin::PinError,
     programs::{
-        FlowDissector, KProbe, ProbeKind, Program, ProgramError, TracePoint, UProbe, Xdp, XdpFlags,
+        FlowDissector, KProbe, LinkOrder, ProbeKind, Program, ProgramError, SchedClassifier,
+        TcAttachType, TracePoint, UProbe, Xdp, XdpFlags,
         flow_dissector::{FlowDissectorLink, FlowDissectorLinkId},
         kprobe::{KProbeLink, KProbeLinkId},
         links::{FdLink, LinkError, PinnedLink},
         loaded_links, loaded_programs,
+        tc::TcAttachOptions,
         trace_point::{TracePointLink, TracePointLinkId},
         uprobe::{UProbeLink, UProbeLinkId},
         xdp::{XdpLink, XdpLinkId},
@@ -391,6 +393,58 @@ fn pin_link() {
 
     // finally when new_link is dropped we're detached
     drop(new_link);
+    assert_unloaded(program_name);
+}
+
+#[test_log::test]
+fn pin_tcx_link() {
+    // TCX links require kernel >= 6.6
+    let kernel_version = KernelVersion::current().unwrap();
+    if kernel_version < KernelVersion::new(6, 6, 0) {
+        eprintln!("skipping pin_tcx_link test on kernel {kernel_version:?}");
+        return;
+    }
+
+    use crate::utils::NetNsGuard;
+    let _netns = NetNsGuard::new();
+
+    let program_name = "tcx_next";
+    let pin_path = "/sys/fs/bpf/aya-tcx-test-lo";
+    let mut bpf = Ebpf::load(crate::TCX).unwrap();
+    let prog: &mut SchedClassifier = bpf.program_mut(program_name).unwrap().try_into().unwrap();
+    prog.load().unwrap();
+
+    let link_id = prog
+        .attach_with_options(
+            "lo",
+            TcAttachType::Ingress,
+            TcAttachOptions::TcxOrder(LinkOrder::default()),
+        )
+        .unwrap();
+    let link = prog.take_link(link_id).unwrap();
+    assert_loaded(program_name);
+
+    let fd_link: FdLink = link.try_into().unwrap();
+    fd_link.pin(pin_path).unwrap();
+
+    // Because of the pin, the program is still attached
+    prog.unload().unwrap();
+    assert_loaded(program_name);
+
+    // Load a new program and atomically replace the old one using attach_to_link
+    let mut bpf = Ebpf::load(crate::TCX).unwrap();
+    let prog: &mut SchedClassifier = bpf.program_mut(program_name).unwrap().try_into().unwrap();
+    prog.load().unwrap();
+
+    let old_link = PinnedLink::from_pin(pin_path).unwrap();
+    let link = FdLink::from(old_link).try_into().unwrap();
+    let _link_id = prog.attach_to_link(link).unwrap();
+
+    assert_loaded(program_name);
+
+    // Clean up: remove the stale pin file and drop the bpf instance (which drops the program and link)
+    remove_file(pin_path).unwrap();
+    drop(bpf);
     assert_unloaded(program_name);
 }
 


tamird:
Could you address the CI failures?

svencowart:
> Could you address the CI failures?

Yup, all done.

tamird (commented):
Thanks! A few comments.

> @tamird reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1396#-:-OeXpAdfAchJe6y374LN:b-jqv8pm)* status: all files reviewed, 6 unresolved discussions (waiting on @svencowart)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/aya-rs/aya/1396#-OeXp7vt0z19mnGAUyrc:-OeXp7vt0z19mnGAUyrd:b-2vgsrw):*
please add periods at the end s of these list items



_Code quote:_
```
  - Add link pinning example to SchedClassifier::attach_with_options()
    showing atomic replacement workflow for TCX mode (kernel >= 6.6)
  - Add pin_tcx_link() integration test verifying link persistence
    across program unload and atomic replacement capability
```
___
*[`aya/src/programs/tc.rs` line 220 at r1](https://reviewable.io/reviews/aya-rs/aya/1396#-OeXoWD124sjNxwWY1Pw:-OeXoWD124sjNxwWY1Px:b-r9hpgt) ([raw file](https://github.com/aya-rs/aya/blob/8af3b7553cb28ffedc7891d5f8a725205752ae3a/aya/src/programs/tc.rs#L220)):*
> ```XML
>     /// let pin_path = "/sys/fs/bpf/my_link";
>     ///
>     /// let link_id = if Path::new(pin_path).exists() {
> ```

This is a classic TOCTOU race. Can we instead write as handling the error from from_pin?
___
*[`aya/src/programs/tc.rs` line 222 at r1](https://reviewable.io/reviews/aya-rs/aya/1396#-OeXobE-8QXVAeSlKdVy:-OeXobE-8QXVAeSlKdVz:b-qj0wht) ([raw file](https://github.com/aya-rs/aya/blob/8af3b7553cb28ffedc7891d5f8a725205752ae3a/aya/src/programs/tc.rs#L222)):*
> ```XML
>     /// let link_id = if Path::new(pin_path).exists() {
>     ///     let old = PinnedLink::from_pin(pin_path)?;
>     ///     prog.attach_to_link(FdLink::from(old).try_into()?)?  // atomic replacement
> ```

could you please extract a local variable for the result of the intermediate fallible operation?
___
*[`aya/src/programs/tc.rs` line 225 at r1](https://reviewable.io/reviews/aya-rs/aya/1396#-OeXp3rG5Ck8IrslSrl2:-OeXp3rG5Ck8IrslSrl3:b-20ubxu) ([raw file](https://github.com/aya-rs/aya/blob/8af3b7553cb28ffedc7891d5f8a725205752ae3a/aya/src/programs/tc.rs#L225)):*
> ```XML
>     /// } else {
>     ///     prog.attach_with_options("eth0", TcAttachType::Ingress,
>     ///                               TcAttachOptions::TcxOrder(LinkOrder::default()))?
> ```

would you mind double checking that this formatting is what rustfmt would do?



_Code quote:_
```XML
    ///     prog.attach_with_options("eth0", TcAttachType::Ingress,
    ///                               TcAttachOptions::TcxOrder(LinkOrder::default()))?
```
___
*[`aya/src/programs/tc.rs` line 238 at r1](https://reviewable.io/reviews/aya-rs/aya/1396#-OeXnuWT74IHnpiRrLQn:-OeXnuWT74IHnpiRrLQo:b-tb7cfr) ([raw file](https://github.com/aya-rs/aya/blob/8af3b7553cb28ffedc7891d5f8a725205752ae3a/aya/src/programs/tc.rs#L238)):*
> ```XML
>     /// [`TcError::NetlinkError`] is returned if attaching fails. A common cause
>     /// of failure is not having added the `clsact` qdisc to the given
>     /// interface, see [`qdisc_add_clsact`]
> ```

mind adding a period?
___
*[`test/integration-test/src/tests/load.rs` line 400 at r1](https://reviewable.io/reviews/aya-rs/aya/1396#-OeXoxdyBG3bvnxqnDpL:-OeXoxdyBG3bvnxqnDpM:bdcyfrx) ([raw file](https://github.com/aya-rs/aya/blob/8af3b7553cb28ffedc7891d5f8a725205752ae3a/test/integration-test/src/tests/load.rs#L400)):*
> ```XML
> 
> #[test_log::test]
> fn pin_tcx_link() {
> ```

did you want to use `attach_to_link` here? I thought that was the main thing you wanted to demonstrate here.

svencowart:
@tamird it seems to me the failing CI jobs have nothing to do with my changes. Am I right about that?

tamird:
Yep #1409 should fix it

tamird:
@codex review

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1396)* status: all files reviewed, 1 unresolved discussion (waiting on @svencowart).

___
*[`aya/src/programs/tc.rs` line 218 at r6](https://reviewable.io/reviews/aya-rs/aya/1396#-Oh9_7M64EdC6TduryAW:-Oh9_7M64EdC6TduryAX:b-5z9rib) ([raw file](https://github.com/aya-rs/aya/blob/d1a2f5d370532b3a02025d868ee15b96f232a7ea/aya/src/programs/tc.rs#L218)):*
> ```XML
>     /// #     programs::{
>     /// #         LinkError, LinkOrder, SchedClassifier, TcAttachType,
>     /// #         links::{FdLink, PinnedLink},
> ```

looks like LinkError should be on this line rather than above

tamird:
Hi @svencowart are you planning to finish this up?

svencowart:
> Hi @svencowart are you planning to finish this up?

@tamird yes. Sorry I was on vacation over the holidays. I'm getting back into it this week.

tamird:
Neat. Please note we prefer linear history around here, please rebase rather than merge main into your branch.

svencowart:
Me too on the linear history - I'll adjust it. I made a mistake and clicked on "update branch" within the GH PR UI and didn't realize that's what it would do.

tamird:
Thanks!

PR #1417, opened 2025-12-15
aya: add multi-uprobe attach support

swananan:
This PR tackles #992 and focuses on multi-attach uprobes.

Main talking points:

1. Refactors symbol resolution to batch lookups over a single object::File mmap, since repeated mmap/parse per location was too expensive for multi-attach workloads. I haven’t split out a dedicated symbol resolver yet because that felt heavier, but I’m open to doing this.

2. Introduces `#[uprobe(multi)]` expectations, add attach-type checks, and emit friendly `ProgramNotMulti`/`UProbeMultiNotSupported` errors. To keep `compiled_for_multi` in sync, `UProbe` can no longer rely on the shared  `impl_from_prog_info!` macro, so it now hand-rolls `from_program_info` (open to suggestions for a more reusable pattern).

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1417)

PR #1421, opened 2025-12-17
uprobe: bundle attach location+cookie via UProbeAttachPoint

swananan:
This follows the #1417 review discussion:

By bundling `location + cookie` into a `UProbeAttachPoint` we get a more idiomatic `Into<_>` entry point, keep the one-to-one relationship enforced by the type system, and make it easier to extend `attach` with multi-location support without introducing parallel arrays or a brand new API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1421)


swananan:
@tamird thanks for the review, I have addressed all the comments. I misused build pattern here, since pub fields are enough.

PR #1372, opened 2025-10-22
aya: Support loading programs with ksyms

altugbozkurt07:
Collect extern symbols from object BTF and fix up the .ksyms datasec so the resulting BTF is acceptable to the kernel.

Resolve typed extern vars and kfuncs through kernel BTF. Fall back to `/proc/kallsyms` for typeless vars, then apply the corresponding extern relocations before program load.

Expose the object-level ksym APIs in aya-obj via `Object::resolve_externs()`, `Object::relocate_externs()`, and `KsymsError`.

Weak extern handling follows libbpf-style behavior. Unresolved weak var references are null-patched, unresolved weak kfunc calls are poisoned, and unresolved strong extern relocations are rejected.

Add integration coverage for typed ksyms, typed kfuncs, kallsyms-backed ksyms, and missing strong-symbol failures.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1372)

PR #1422, opened 2025-12-17
Add a check to skip .ksyms/.kconfig from BTF section fixups

grantseltzer:
When I was trying to use a [custom kfunc](https://eunomia.dev/tutorials/43-kfuncs/) I was met with the following error when trying to load my bpf object:

```
Error: BTF error: Unable to determine the size of section `.ksyms`

Caused by:
    Unable to determine the size of section `.ksyms`

```

libbpf (and [cilium/ebpf](https://github.com/cilium/ebpf/blob/32e841a02106c2faa9f5c65ea25836c6d58d3965/btf/btf.go#L257)) ignore .ksyms and .kconfig when patching missing info in Datasecs. This is because these two sections are external, and won't have any information to gather in ELF sections. Aya should do the same.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1422)

### DIFF
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -501,12 +501,16 @@ impl Btf {
         symbol_offsets: &HashMap<String, u64>,
         features: &BtfFeatures,
     ) -> Result<(), BtfError> {
+        debug!("fixup_and_sanitize: starting with {} BTF types", self.types.types.len());
         let enum64_placeholder_id = OnceCell::new();
         let filler_var_id = OnceCell::new();
         let mut types = mem::take(&mut self.types);
         for i in 0..types.types.len() {
             let t = &mut types.types[i];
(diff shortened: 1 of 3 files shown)


grantseltzer:
Changed this to a draft as after further testing there appears to be more linking fixes needed. Actively working on it.

vadorovsky:
There is ongoing work on supporting ksyms - #1372 - that already contains a fix similar to yours, but also relocates the ksyms using kernel's BTF. Ler's maybe try to get that one in?

grantseltzer:
> There is ongoing work on supporting ksyms - #1372 - that already contains a fix similar to yours, but also relocates the ksyms using kernel's BTF. Ler's maybe try to get that one in?

Ah that's great, thank you for pointing that out. I'll keep my eye on that PR

PR #1407, opened 2025-11-30
Add licese for aya-obj

danielmellado:
This commit adds a link for the needed licenses in aya-obj for fedora
packaging.

Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1407)

PR #1426, opened 2025-12-24
build: add symlink to root licenses in aya-log

Qwor01:
Added a symlink from project root into aya-log in order for aya-log to be able to package in fedora

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1426)


aya-log/LICENSE-MIT
```diff
@@ -0,0 +1 @@
+LICENSE-MIT
```
tamird, line 1: Does this work? Shouldn't it be pointing up a level?

Qwor01: It should work like this, please review this previous pr as an example:
https://github.com/aya-rs/aya/pull/1407

tamird: In that PR the links point to `../LICENSE-MIT`

Qwor01: Yes, sorry i see the problem now, i'll close this pr and open a new one with the correct change, thank you

PR #1434, opened 2025-12-31
Add batch_lookup_and_delete method to PerCpuHashMap

arthur-zhang:
- Add batch_lookup_and_delete method to PerCpuHashMap that retrieves and removes multiple entries in a single syscall
  - Implement low-level bpf_map_lookup_and_delete_batch_per_cpu syscall wrapper in sys/bpf.rs
  - Provide efficient batch operations for per-CPU hash maps, improving performance when processing multiple entries (requires kernel 5.6+)

  What Changed

  New API (aya/src/maps/hash_map/per_cpu_hash_map.rs:141):
  - Added batch_lookup_and_delete() method that accepts batch size, optional cursor, and flags
  - Returns tuple of (keys, values, out_batch) for cursor-based iteration
  - Includes comprehensive documentation with usage examples

  System Call Implementation (aya/src/sys/bpf.rs:260):
  - Added bpf_map_lookup_and_delete_batch_per_cpu() to handle the BPF syscall
  - Properly handles per-CPU value alignment and memory layout
  - Returns ENOENT when no entries remain

  Testing:
  - Added test for empty map scenario
  - Added test with mock entries to verify correct key/value retrieval and cursor handling

  Test plan

  - Verify existing tests pass with cargo test
  - Check that the new batch API correctly handles empty maps
  - Confirm batch operations work with multiple entries across CPUs
  - Validate cursor-based iteration for large datasets
  - Test on kernel 5.6+ to ensure syscall compatibility

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1434)

### DIFF
--- a/aya/src/maps/hash_map/per_cpu_hash_map.rs
+++ b/aya/src/maps/hash_map/per_cpu_hash_map.rs
@@ -10,8 +10,13 @@ use crate::{
     maps::{
         IterableMap, MapData, MapError, MapIter, MapKeys, PerCpuValues, check_kv_size, hash_map,
     },
-    sys::{SyscallError, bpf_map_lookup_elem_per_cpu, bpf_map_update_elem_per_cpu},
+    sys::{
+        SyscallError, bpf_map_lookup_and_delete_batch_per_cpu, bpf_map_lookup_elem_per_cpu,
+        bpf_map_update_elem_per_cpu,
+    },
 };
+
+type BatchResult<K, V> = (Vec<K>, Vec<PerCpuValues<V>>, Option<K>);
 
 /// Similar to [`HashMap`](crate::maps::HashMap) but each CPU holds a separate value for a given key. Typically used to
 /// minimize lock contention in eBPF programs.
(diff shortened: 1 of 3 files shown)


arthur-zhang:
ENOENT in  batch  means：no more data, but may have items in current result. @tamird 

i see the similar logic in ebpf-go `/cilium/ebpf@v0.17.3/map.go`

```
func (m *Map) batchLookupPerCPU(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
	count, err := sliceLen(keysOut)
	if err != nil {
		return 0, fmt.Errorf("keys: %w", err)
	}

	valueBuf := make([]byte, count*int(m.fullValueSize))
	valuePtr := sys.NewSlicePointer(valueBuf)

	n, sysErr := m.batchLookupCmd(cmd, cursor, count, keysOut, valuePtr, opts)
	if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
		return 0, err
	}

	err = unmarshalBatchPerCPUValue(valuesOut, count, int(m.valueSize), valueBuf)
	if err != nil {
		return 0, err
	}

	return n, sysErr
}


```

tamird (changes requested):
> @tamird reviewed 1 file, made 6 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1434)* status: 1 of 3 files reviewed, 9 unresolved discussions (waiting on @arthur-zhang).

___
*[`aya/src/sys/bpf.rs` line 310 at r2](https://reviewable.io/reviews/aya-rs/aya/1434#-Oi42noZ4Z57_vy36pXR:-Oi7q9rQ4MhoxhCcxKZk:b-ccf82e) ([raw file](https://github.com/aya-rs/aya/blob/5e08a9d2dc9ac64f599ece661b146b6f482c8fc0/aya/src/sys/bpf.rs#L310)):*

what does ENOENT mean?
</blockquote></details>

shouldn't we be signaling this out to the caller somehow?
___
*[`-- commits` line 1 at r4](https://reviewable.io/reviews/aya-rs/aya/1434#-Oi7s-ME8jSlJII487Eh:-Oi7s-ME8jSlJII487Ei:b-a7wlzs):*
The PR description mentions testing against an old kernel, but this commit doesn't have any integration tests. Could you add some?
___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/aya-rs/aya/1434#-Oi7p8679nMvFrQm6dsW:-Oi7p8679nMvFrQm6dsX:bj34owo):*
Please follow the advice in https://cbea.ms/git-commit/
___
*[`aya/src/maps/hash_map/per_cpu_hash_map.rs` line 19 at r4](https://reviewable.io/reviews/aya-rs/aya/1434#-Oi7s9NF6xeby4z98Trv:-Oi7s9NF6xeby4z98Trw:bwhc89h) ([raw file](https://github.com/aya-rs/aya/blob/95dca492c781d75f48a433963f4547d806cd8205/aya/src/maps/hash_map/per_cpu_hash_map.rs#L19)):*
> ```XML
> };
> 
> type BatchResult<K, V> = (Vec<K>, Vec<PerCpuValues<V>>, Option<K>);
> ```

This type seems very raw, I don't think this is the API we want -- for a start, the first two vectors being always the same size is odd, can we instead have a single `Vec<K, PerCpuValues<V>>`?

I also think exposing the "next key" doesn't make for a great API. This should be exposed as a sort of iterator -- the next key should be private.
___
*[`aya/src/sys/bpf.rs` line 262 at r4](https://reviewable.io/reviews/aya-rs/aya/1434#-Oi7pL-x78F31ptHpXw5:-Oi7pL-x78F31ptHpXw6:be3deed) ([raw file](https://github.com/aya-rs/aya/blob/95dca492c781d75f48a433963f4547d806cd8205/aya/src/sys/bpf.rs#L262)):*
> ```XML
> /// Result of a batch lookup operation on a per-CPU map.
> pub(crate) struct PerCpuBatch<K: Pod, V: Pod> {
>     /// Vector of retrieved keys
> ```

please add a period after each of these
___
*[`aya/src/sys/bpf.rs` line 288 at r4](https://reviewable.io/reviews/aya-rs/aya/1434#-Oi7pTjlD-80GPByDMbI:-Oi7pTjlD-80GPByDMbJ:bvwvyre) ([raw file](https://github.com/aya-rs/aya/blob/95dca492c781d75f48a433963f4547d806cd8205/aya/src/sys/bpf.rs#L288)):*
> ```XML
> /// - out_batch: Optional cursor for the next batch
> ///
> /// # Introduced in kernel v5.6
> ```

can you please add a citation?

PR #1439, opened 2026-01-09
aya-build: try to build when rustup is not found

famfo:
In some environments, rustup is not availible but cargo and its targets are. This changes aya-build to try to continue building if rustup is not found, even with stable Rust. If the Rust channel differs from the selected toolchain, the user will be warned but a build will still be attempted.

Fixes: #1329

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1439)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ public-api = { version = "0.50.0", default-features = false }
 quote = { version = "1", default-features = false }
 rand = { version = "0.9", default-features = false }
 rbpf = { version = "0.4.0", default-features = false }
+rustc_version = { version = "0.4.1", default-features = false }
 rustdoc-json = { version = "0.9.0", default-features = false }
 rustup-toolchain = { version = "0.1.5", default-features = false }
 rustversion = { version = "1.0.0", default-features = false }
--- a/aya-build/Cargo.toml
+++ b/aya-build/Cargo.toml
@@ -16,3 +16,5 @@ workspace = true
 [dependencies]
 anyhow = { workspace = true, default-features = true }
 cargo_metadata = { workspace = true }
+rustc_version = { workspace = true }
+which = { workspace = true, features = ["real-sys"] }
--- a/aya-build/src/lib.rs
+++ b/aya-build/src/lib.rs
@@ -1,7 +1,7 @@
 use std::{
     borrow::Cow,
     env,
-    ffi::OsString,
+    ffi::{OsStr, OsString},
     fs,
     io::{BufRead as _, BufReader},
     path::PathBuf,
@@ -10,6 +10,8 @@ use std::{
 
 use anyhow::{Context as _, Result, anyhow};
 use cargo_metadata::{Artifact, CompilerMessage, Message, Target};
+use rustc_version::Channel;
+use which::which;
 
 #[derive(Default)]
 pub struct Package<'a> {
@@ -77,6 +79,41 @@ pub fn build_ebpf<'a>(
     let bpf_target_arch = target_arch_fixup(bpf_target_arch.into());
     let target = format!("{target}-unknown-none");
 
+    const RUSTUP: &str = "rustup";
+    let rustup = which(RUSTUP);
+    let prefix: &[_] = match rustup.as_ref() {
+        Ok(rustup) => &[
+            rustup.as_os_str(),
+            OsStr::new("run"),
+            OsStr::new(toolchain.as_str()),
+        ],
+        Err(err) => {
+            println!("cargo:warning=which({RUSTUP})={err}; proceeding with current toolchain");
+            &[]
+        }
+    };
+
+    let cmd = |program| match prefix {
+        [] => Command::new(program),
+        [wrapper, args @ ..] => {
+            let mut cmd = Command::new(wrapper);
+            cmd.args(args).arg(program);
+            cmd
+        }
+    };
+
+    let rustc_version::VersionMeta {
+        semver: _,
+        commit_hash: _,
+        commit_date: _,
+        build_date: _,
+        channel,
+        host: _,
+        short_version_string: _,
+        llvm_version: _,
+    } = rustc_version::VersionMeta::for_command(cmd("rustc"))
+        .context("failed to get rustc version meta")?;
+
     for Package {
         name,
         root_dir,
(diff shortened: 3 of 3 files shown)


famfo:
If desired, I'd be happy to write an additional `build-test` for this.

famfo:
Should I include the changes from `cargo xtask public-api --bless` in the commit?

tamird:
Yes

tamird (changes requested):
> @tamird reviewed all commit messages and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1439)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @famfo).

___
*[`aya-build/src/lib.rs` line 69 at r1](https://reviewable.io/reviews/aya-rs/aya/1439#-OiY60rIDzJzRWUFUSj-:-OiY60rIDzJzRWUFUSj0:b-uqw9jd) ([raw file](https://github.com/aya-rs/aya/blob/2065d42115fab7c0c1185ea125d131aca5ca75a1/aya-build/src/lib.rs#L69)):*
> ```XML
>     let target = format!("{target}-unknown-none");
> 
>     let has_rustup = which("rustup").is_ok();
> ```

instead of using `is_ok` here, match on the value below (use the absolute path as input to `Command::new`)
___
*[`aya-build/src/lib.rs` line 92 at r1](https://reviewable.io/reviews/aya-rs/aya/1439#-OiY6dvS37B4Px6w7GBb:-OiY6dvS37B4Px6w7GBc:b7jwkmh) ([raw file](https://github.com/aya-rs/aya/blob/2065d42115fab7c0c1185ea125d131aca5ca75a1/aya-build/src/lib.rs#L92)):*
> ```XML
>                 "build",
>                 "--package",
>                 name,
> ```

these are repeated in the `else` branch



_Code quote:_
```XML
                "build",
                "--package",
                name,
```
___
*[`aya-build/src/lib.rs` line 99 at r1](https://reviewable.io/reviews/aya-rs/aya/1439#-OiY6HZI3WhxgYoH87di:-OiY6HZI3WhxgYoH87dj:bpbpau6) ([raw file](https://github.com/aya-rs/aya/blob/2065d42115fab7c0c1185ea125d131aca5ca75a1/aya-build/src/lib.rs#L99)):*
> ```XML
>             cmd
>         } else {
>             let channel = rustc_version::version_meta()?.channel;
> ```

why are we calling this in a loop?
___
*[`aya-build/src/lib.rs` line 100 at r1](https://reviewable.io/reviews/aya-rs/aya/1439#-OiY6OJG2KRLaTWVdrin:-OiY6OJG2KRLaTWVdrio:b-95q2jv) ([raw file](https://github.com/aya-rs/aya/blob/2065d42115fab7c0c1185ea125d131aca5ca75a1/aya-build/src/lib.rs#L100)):*
> ```XML
>         } else {
>             let channel = rustc_version::version_meta()?.channel;
>             if toolchain != channel {
> ```

yoda condition, should be `channel != toolchain`
___
*[`aya-build/src/lib.rs` line 102 at r1](https://reviewable.io/reviews/aya-rs/aya/1439#-OiY6Rq1BhwmjzjRhu3O:-OiY6Rq1BhwmjzjRhu3P:byt4xsd) ([raw file](https://github.com/aya-rs/aya/blob/2065d42115fab7c0c1185ea125d131aca5ca75a1/aya-build/src/lib.rs#L102)):*
> ```XML
>             if toolchain != channel {
>                 println!(
>                     "cargo:warning=rustup was not found and the selected toolchain ({}) differs from the toolchain currently in use ({channel:?}), attempting to build anyways",
> ```

anyways isn't a word
___
*[`aya-build/src/lib.rs` line 252 at r1](https://reviewable.io/reviews/aya-rs/aya/1439#-OiY6jac0TrlN7kGk2pu:-OiY6jac0TrlN7kGk2pv:b-5uv7v6) ([raw file](https://github.com/aya-rs/aya/blob/2065d42115fab7c0c1185ea125d131aca5ca75a1/aya-build/src/lib.rs#L252)):*
> ```XML
>         self == &Self::Nightly && other == &rustc_version::Channel::Nightly
>     }
> }
> ```

please remove this and just do it inline, this is more confusing than it's worth



_Code quote:_
```XML
impl PartialEq<rustc_version::Channel> for Toolchain<'_> {
    fn eq(&self, other: &rustc_version::Channel) -> bool {
        self == &Self::Nightly && other == &rustc_version::Channel::Nightly
    }
}
```

tamird (changes requested):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1439)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @famfo).

___
*[`aya-build/src/lib.rs` line 222 at r3](https://reviewable.io/reviews/aya-rs/aya/1439#-OiYf7ph9JXDsSx_kICk:-OiYf7ph9JXDsSx_kICl:b673evf) ([raw file](https://github.com/aya-rs/aya/blob/c3d16aebf2e2bdaf39a0aa46affe19fcbafec2ec/aya-build/src/lib.rs#L222)):*
> ```XML
> 
> /// The toolchain to use for building eBPF programs.
> #[derive(Default, PartialEq)]
> ```

revert plz, then you can remove the API changes

famfo:
> _[`aya-build/src/lib.rs` line 222 at r3](https://reviewable.io/reviews/aya-rs/aya/1439#-OiYf7ph9JXDsSx_kICk:-OiYf7ph9JXDsSx_kICl:b673evf) ([raw file](https://github.com/aya-rs/aya/blob/c3d16aebf2e2bdaf39a0aa46affe19fcbafec2ec/aya-build/src/lib.rs#L222)):_
> 
> > ```
> > /// The toolchain to use for building eBPF programs.
> > #[derive(Default, PartialEq)]
> > ```
> 
> revert plz, then you can remove the API changes

That would complicate the toolchain check to an `if let` or string comparison, which one do you prefer?

vadorovsky:
Is your use case packaging or working with custom toolchains build from source (but without using rustup)? If so, I think a better way would be providing a way to just disable aya-build and let the package manifests perform the build of eBPF crate manually, like:

```shell
# Build eBPF crate.
cargo build --target bpfel-unknown-none -p foo-ebpf
# Build user-space crate with aya-build disabled.
AYA_BUILD_DISABLE=1 cargo build -p foo 
```

The caller would have full control over which cargo toolchain is called and with which flags. I've been voicing that idea multiple times, I guess I need to finally send a PR.

I'm worried that if we start customizing aya-build for different use cases, we will end up writing a complex build system and spending maintenance effort on users reporting different issues on different configurations. I would prefer aya-build to stay small and focus purely on local development with rustup and nightly, where it "just works" and is convenient for 99% developers. Using different toolchains is where complexity starts and IMO it's better to be not opinionated.

famfo:
I am not using rustup to manage my toolchains, completely disabling aya-build would be annoying as well because I'd have to write a custom build script again.

vadorovsky:
OK, your PR is small enough that I'm fine with it (after rebase)

famfo:
> We don't need to do this if `rustup` is on the PATH, right? In other words, this should be a lazy thing declared up here but called in the `else` branch below.

That would either require some `Option` or executing it in every loop, not sure if that is much better.

tamird:
I pushed a commit with what I had in mind. How does it look to you?

tamird:
> I did not try it but looks promising. Thanks!

Would be great if you could try it :)

famfo:
> I pushed a commit with what I had in mind. How does it look to you?

Ah, I get the idea now, lost the bigger picture slightly. Not sure when I have time to clean up the code properly and retest in the next few days though.

tamird:
What cleanup is needed? Can you test as-is?

famfo:
Yes, it does work as-is.

PR #1457, opened 2026-01-23
refactor(aya-ebpf): make btf_maps libbpf-compatible

### LINKED ISSUE #1455: BTF map definitions emitted by aya are not libbpf-compatible
Hi Aya maintainers,

While working with BTF-defined maps (`#[btf_map]`) in aya-generated eBPF programs, I ran into an interoperability issue with loaders that expect libbpf-style BTF, in particular github.com/cilium/ebpf.

Aya currently emits BTF map definitions wrapped in one or more single-field struct layers (notably `__0` → `value`) before the actual map definition whereas libbpf (and tools built on top of it) expect the map definition struct to be either directly exposed or wrapped only minimally.

This results in loaders failing with errors such as:

```
load BTF maps: map RESULT: unrecognized field __0 in BTF map definition
```

I investigated the issue on the consumer side and confirmed that unwrapping these layers makes aya-generated BTF maps load correctly. However, maintainers of cilium/ebpf raised valid concerns about adding aya-specific logic and non-reproducible Rust-generated ELF testdata on their side (cf [PR](https://github.com/cilium/ebpf/pull/1941)).

Given this, I wanted to ask:

- Is emitting libbpf-compatible BTF for map definitions something that Aya would consider?
- Alternatively, could this be made opt-in (e.g. a compatibility mode for `#[btf_map]`)?
- Or is the current layout considered stable and intentional, with consumers expected to adapt?

I’m happy to help investigate or prototype changes on the Aya side if that would be useful.

Thanks a lot for your work on Aya!

Brskt:
## Summary

- Remove the `btf_map_def!` macro which wrapped fields in `UnsafeCell`, creating nested BTF structures that libbpf couldn't parse
- Convert `Array`, `RingBuf`, and `SkStorage` to use flat `#[repr(C)]` structs with direct pointer fields
- Add C BPF program and Rust test to verify BTF map definitions are compatible with libbpf's parser

Closes #1455

## Test plan

- [x] Existing integration tests pass (array, sk_storage)
- [x] New btf_maps test verifies C program with BTF maps loads correctly
- [x] BTF structure verified with `bpftool btf dump` - produces flat anonymous structs
- [x] Verified libbpf (via bpftool) correctly parses the BTF map definitions

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1457)

### DIFF
--- a/ci/headers/argp.h
+++ b/ci/headers/argp.h
@@ -1,0 +1,42 @@
+#ifndef AYA_CI_STUB_ARGP_H
+#define AYA_CI_STUB_ARGP_H
+
+#include <stdio.h>
+
+typedef int error_t;
+
+struct argp_option {
+  const char *name;
+  int key;
+  const char *arg;
+  int flags;
+  const char *doc;
+  int group;
+};
+
+struct argp_state;
+
+typedef error_t (*argp_parser_t)(int key, char *arg, struct argp_state *state);
+
+struct argp {
+  const struct argp_option *options;
+  argp_parser_t parser;
+  const char *args_doc;
+  const char *doc;
+  const void *children;
+  void *help_filter;
+  const char *argp_domain;
+};
+
+struct argp_state {
+  const char *name;
+};
+
+#define OPTION_ARG_OPTIONAL 0x1
+#define ARGP_HELP_SEE 0x40
+#define ARGP_ERR_UNKNOWN 1
+
+int argp_help(const struct argp *argp, FILE *stream, unsigned int flags,
+              char *name);
+
+#endif /* AYA_CI_STUB_ARGP_H */
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,13 @@ jobs:
 
       - run: cargo +nightly fmt --all -- --check
 
+      - name: Install libbpf-rs prerequisites
+        run: |
+          set -euxo pipefail
+
+          sudo apt update
+          sudo apt -y install autopoint
+
       - run: ./clippy.sh
 
       # On the `aya-rs/aya` repository, regenerate the public API on a schedule.
@@ -102,7 +109,19 @@ jobs:
         with:
           target: ${{ matrix.arch }}
 
+      - name: Install libbpf-rs prerequisites
+        run: |
+          set -euxo pipefail
+
+          sudo apt update
+          sudo apt -y install autopoint
+
       - name: Build
+        env:
+          # Fix for 32-bit ARM: Ubuntu 24.04 GCC defines _TIME_BITS=64 as a built-in,
+          # which requires _FILE_OFFSET_BITS=64 (glibc requirement). We must unset
+          # _TIME_BITS first because the built-in is seen before -D flags by libbpf-sys.
+          CFLAGS_armv7_unknown_linux_gnueabi: "-U_TIME_BITS -D_FILE_OFFSET_BITS=64"
         run: |
           set -euxo pipefail
           cargo hack build --all-targets --feature-powerset \
@@ -262,7 +281,12 @@ jobs:
           sudo udevadm trigger --name-match=kvm || true # kvm is not available on arm64.
 
           sudo apt update
+          # linux-libc-dev provides kernel UAPI headers for libbpf-sys cross-compilation.
+          #
+          # See https://github.com/libbpf/libbpf-sys/issues/137.
           sudo apt -y install \
+            autopoint \
+            linux-libc-dev \
             liblzma-dev \
             lynx \
             musl-tools \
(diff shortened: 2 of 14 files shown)


ebpf/aya-ebpf/src/btf_maps/mod.rs
```diff
@@ -18,43 +18,3 @@ impl AyaBtfMapMarker {
         Self(PhantomData)
     }
 }
-
-#[macro_export]
```
tamird, line 22: lol why was this exported

I think we should keep this macro to avoid all the duplication - we should just change what it does. WDYT?

ebpf/aya-ebpf/src/btf_maps/sk_storage.rs
```diff
@@ -1,23 +1,31 @@
...
+    btf_maps::AyaBtfMapMarker,
     helpers::generated::{bpf_sk_storage_delete, bpf_sk_storage_get},
     programs::sock_addr::SockAddrContext,
 };
 
-btf_map_def!(SkStorageDef, BPF_MAP_TYPE_SK_STORAGE);
-
-// TODO(https://github.com/rust-lang/rust/issues/76560): this should be:
-//
-// { F | BPF_F_NO_PREALLOC as usize }.
```
vadorovsky, line 18: I think this TODO is still relevant, @tamird?

ebpf/aya-ebpf/src/btf_maps/mod.rs
vadorovsky: I agree, we could still keep the macro. And sorry for the export. :melting_face:

Brskt: Done! I've updated the PR to keep the macro but modified it to generate flat `#[repr(C)]` structs directly.

ebpf/aya-ebpf/src/btf_maps/sk_storage.rs
tamird: yes.

Brskt: Reverted.

Brskt:
The libbpf compatibility test requires `libelf-dev` to be installed. I made it optional (behind `libbpf-test` feature) so the CI passes for now. To enable the test in CI, `libelf-dev` would need to be added to the workflow. Is there another way to test libbpf compatibility?

Brskt:
The `vendored` feature requires `gettext` (for autopoint) to build libbpf from source. What do u think?

tamird:
I'm trying to get CI working, but we have a blocker: this PR makes it impossible to run clippy on macOS because of the libbpf dependency. So I think we do need to go back to the optional feature and built it only on Linux.

tamird:
Looks like we also have compilation problems on arm in general.

vadorovsky:
This error:

```
  checking for library containing argp_parse... no

  --- stderr
  configure: WARNING: compiler doesn't generate build-id by default
  configure: error: in `/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libbpf-sys-1.6.2+v1.6.2/elfutils':
  configure: error: failed to find argp_parse
  See `config.log' for more details

  thread 'main' (6277) panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libbpf-sys-1.6.2+v1.6.2/build.rs:347:5:
  make failed
```

occurs because musl doesn't have argp. Usually that can be solved by installing [argp-standalone](https://github.com/ericonr/argp-standalone), that's packaged by distros supporting musl (e.g. [Alpine](https://pkgs.alpinelinux.org/package/edge/main/aarch64/argp-standalone). [Gentoo](https://packages.gentoo.org/packages/sys-libs/argp-standalone)), but I can't find it on Ubuntu and Debian. [Homebrew packages is there](https://formulae.brew.sh/formula/argp-standalone) - perhaps that will work.

AFK now, but I can try in the afternoon.

tamird:
I'm fine with you squashing.

There is one unsolved problem: clippy.sh won't work on Mac, we need to share the cross compilation hacks from CI with it.

tamird:
LGTM. @vadorovsky to take a final pass and merge if LGTY

ci/headers/argp.h
```diff
@@ -0,0 +1,42 @@
+#ifndef AYA_CI_STUB_ARGP_H
```
vadorovsky, line 1: neat!

tamird:
@Brskt thanks for this PR! We also run aya's integration tests in bpf-linker, and those tests are now failing due to the missing libbpf-rs dependencies. Would you mind rehabilitating that infrastructure? https://github.com/aya-rs/bpf-linker/actions/runs/21425176833 -- thanks!

Brskt:
> @Brskt thanks for this PR! We also run aya's integration tests in bpf-linker, and those tests are now failing due to the missing libbpf-rs dependencies. Would you mind rehabilitating that infrastructure? https://github.com/aya-rs/bpf-linker/actions/runs/21425176833 -- thanks!

Sure! This PR should fix the aya integration tests aya-rs/bpf-linker#340

PR #1465, opened 2026-01-28
ci: Cancel stale PR runs automatically

vadorovsky:
Add a workflow-level concurrency group so PR workflows share a `<workflow>-pr-<number>` group and GitHub cancels any in-progress run when a new push arrives. Non-PR events fall back to the ref name, which keeps push/cron runs isolated while still preventing multiple runs per ref from piling up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1465)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions: {}
 
 jobs:


.github/workflows/ci.yml
```diff
@@ -11,6 +11,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number || github.ref }}
```
tamird: I think github.ref is sufficient? This is a pretty sophisticated thing.

> The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request that were not merged, this is the pull request merge branch. If the pull request was merged, this is the head branch. For workflows triggered by release, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>. For pull requests events except pull_request_target that were not merged, it is refs/pull/<pr_number>/merge. pull_request_target events have the ref from the base branch. For tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

.github/workflows/ci.yml
```diff
@@ -11,6 +11,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: ${{ github.ref }}
```
tamird: sorry I think you want

  group: ${{ github.workflow }}-${{ github.ref }}

i should've been clearer

PR #1467, opened 2026-01-29
aya-{common,ebpf}: Add spin lock support

vadorovsky:
Add a new `aya-common` crate with a `SpinLock` struct, that wraps `bpf_spin_lock` and allows to use it in BPF maps.

Add an extension trait `EbpfSpinLock` that provides a `lock` method, which is a wrapper over `bpf_spin_lock` helper. It returns a `SpinLockGuard` that calls `bpf_spin_unlock` helper once dropped.

Test that functionality with a simple XDP counter program.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1467)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,9 +77,11 @@ jobs:
         run: |
           set -euxo pipefail
           cargo +nightly hack miri test --all-targets --feature-powerset \
+            --exclude aya-common \
             --exclude aya-ebpf \
             --exclude aya-ebpf-bindings \
             --exclude aya-log-ebpf \
+            --exclude integration-common \
             --exclude integration-ebpf \
             --exclude integration-test \
             --workspace
@@ -129,9 +131,11 @@ jobs:
         run: |
           set -euxo pipefail
           cargo hack build --all-targets --feature-powerset \
+            --exclude aya-common \
             --exclude aya-ebpf \
             --exclude aya-ebpf-bindings \
             --exclude aya-log-ebpf \
+            --exclude integration-common \
             --exclude integration-ebpf \
             --exclude xtask \
             --workspace
@@ -142,9 +146,11 @@ jobs:
         run: |
           set -euxo pipefail
           cargo hack test --all-targets \
+            --exclude aya-common \
             --exclude aya-ebpf \
             --exclude aya-ebpf-bindings \
             --exclude aya-log-ebpf \
+            --exclude integration-common \
             --exclude integration-ebpf \
             --exclude integration-test \
             --feature-powerset
@@ -155,9 +161,11 @@ jobs:
         run: |
           set -euxo pipefail
           cargo hack test --doc \
+            --exclude aya-common \
             --exclude aya-ebpf \
             --exclude aya-ebpf-bindings \
             --exclude aya-log-ebpf \
+            --exclude integration-common \
             --exclude integration-ebpf \
             --exclude integration-test \
             --feature-powerset
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
     "aya",
     "aya-build",
+    "aya-common",
     "aya-log",
     "aya-log-common",
     "aya-log-parser",
--- a/aya-common/Cargo.toml
+++ b/aya-common/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+description = "Library shared across eBPF and user-space"
+documentation = "https://docs.rs/aya-common"
+keywords = ["bpf", "ebpf", "common"]
+name = "aya-common"
+version = "0.1.0"
+
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+aya-ebpf-bindings = { version = "^0.1.2", path = "../ebpf/aya-ebpf-bindings" }
+
+[features]
+user = ["aya-ebpf-bindings/user"]
--- a/aya-common/src/lib.rs
+++ b/aya-common/src/lib.rs
@@ -1,0 +1,9 @@
+#![cfg_attr(
+    target_arch = "bpf",
+    expect(unused_crate_dependencies, reason = "compiler_builtins")
+)]
+#![no_std]
+
+mod spin_lock;
+
+pub use spin_lock::SpinLock;
--- a/aya-common/src/spin_lock.rs
+++ b/aya-common/src/spin_lock.rs
@@ -1,0 +1,2 @@
+/// A spin lock that can be used to protect shared data in eBPF maps
+pub type SpinLock = aya_ebpf_bindings::bindings::bpf_spin_lock;
--- a/ebpf/aya-ebpf-bindings/Cargo.toml
+++ b/ebpf/aya-ebpf-bindings/Cargo.toml
@@ -13,7 +13,11 @@ repository.workspace = true
 workspace = true
 
 [dependencies]
+aya = { path = "../../aya", version = "^0.13.1", optional = true }
 aya-ebpf-cty = { version = "^0.2.3", path = "../aya-ebpf-cty" }
 
 [build-dependencies]
 aya-build = { version = "^0.1.3", path = "../../aya-build" }
+
+[features]
+user = ["dep:aya"]
--- a/ebpf/aya-ebpf-bindings/src/lib.rs
+++ b/ebpf/aya-ebpf-bindings/src/lib.rs
@@ -118,4 +118,7 @@ pub mod bindings {
         pub id: ::aya_ebpf_cty::c_uint,
         pub pinning: ::aya_ebpf_cty::c_uint,
     }
+
+    #[cfg(feature = "user")]
+    unsafe impl aya::Pod for bpf_spin_lock {}
 }
--- a/ebpf/aya-ebpf/Cargo.toml
+++ b/ebpf/aya-ebpf/Cargo.toml
@@ -14,6 +14,7 @@ rust-version.workspace = true
 workspace = true
 
 [dependencies]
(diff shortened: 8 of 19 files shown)


vadorovsky:
Couldn't re-open #1356, because of a force push / rebase. All comments from there are addressed.

vadorovsky (commented):
> @vadorovsky made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1467)* status: 5 of 16 files reviewed, 9 unresolved discussions (waiting on @tamird).

___
*[`aya-common/src/spin_lock.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAnnNx2v-epRnnqbgN:-OkAu2hA2AOdolPkRyLU:b-c6aj5f) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L7)):*
> why does it hold a uint?

Because it's used by the helper.
https://elixir.bootlin.com/linux/v6.18.6/source/kernel/bpf/helpers.c#L289

Do I really have to comment on that? I would expect people to look at the kernel sources if they're curious about why certain kernel have certain layout. We don't explain how every single binding works.

> what does it mean to copy it?

1. Pod requires Copy.
2. It lets people derive Copy (if they want) on the types they want to store in maps. The lack of it would prevent people from deriving Copy on types that have a spin lock.
___
*[`aya-common/src/spin_lock.rs` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAo-pw1cb2r-GSR9Bk:-OkAtKAl5QIB8ORUbCzW:bqnlbpf) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L14)):*

spinlock is pod?
</blockquote></details>

Yes - it's meant to be used in BPF maps. If we don't implement it ourselves, users will have to do it - which would be a horrible user experience.
___
*[`ebpf/aya-ebpf/src/spin_lock.rs` line 24 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAofEg1lF-PzxsZ4Rw:-OkAtYqa2lKVMqCCRrPm:b-vl1j7w) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/ebpf/aya-ebpf/src/spin_lock.rs#L24)):*

hmm, why an extension trait? why can't userspace get a wrapper type that exposes methods? we don't want anyone else implementing this, right?
</blockquote></details>

eBPF spin locks can't be user in the userspace.

tamird (commented):
> @tamird reviewed 1 file and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1467)* status: 6 of 16 files reviewed, 9 unresolved discussions (waiting on @vadorovsky).

___
*[`aya-common/src/spin_lock.rs` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAo-pw1cb2r-GSR9Bk:-OkB2mIg6-CTjEGNYUaB:bcl0uih) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L14)):*

Yes - it's meant to be used in BPF maps. If we don't implement it ourselves, users will have to do it - which would be a horrible user experience.
</blockquote></details>

I guess I'm surprised we want to allow this; copying the spinlock seems like asking for undefined behavior
___
*[`ebpf/aya-ebpf/src/spin_lock.rs` line 24 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAofEg1lF-PzxsZ4Rw:-OkB2KWeE-T-d9zDjy5U:b-x8kiwp) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/ebpf/aya-ebpf/src/spin_lock.rs#L24)):*

eBPF spin locks can't be user in the userspace.
</blockquote></details>

sorry, i meant it the other way around, but i guess the reason is explained on the type definition in spin_lock.rs.

can you please seal this trait and comment on it that it exists because we want methods only in eBPF and can't use a wrapper type?

vadorovsky (commented):
> @vadorovsky made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1467)* status: 6 of 16 files reviewed, 9 unresolved discussions (waiting on @tamird).

___
*[`aya-common/src/spin_lock.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAnnNx2v-epRnnqbgN:-OkFU5JTAl1VHwD0UlUW:bybf75f) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L7)):*

OK, I see. Is it also required that the inner field is unnamed?
</blockquote></details>

No, the unnamed field is not a requirement - just my choice, since we aren't really interested in accessing it, and naming it would make clippy/rustc complain that the field is unused. We just need to ensure that the layout and alignment of this struct is same as the kernel expects.
___
*[`aya-common/src/spin_lock.rs` line 8 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAnxLN8jHiRRQl4-y-:-OkFUJXt-U9GGIBJWMDb:bj4ws24) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L8)):*

does this name need to be pub, given the below?
</blockquote></details>

Public `type` aliases require the aliased item to be also public.
___
*[`aya-common/src/spin_lock.rs` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAo-pw1cb2r-GSR9Bk:-OkFURxy1YZ7O-EUcvYC:bfgyven) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L14)):*

I guess I'm surprised we want to allow this; copying the spinlock seems like asking for undefined behavior
</blockquote></details>

Yeah, giving it a thought, I do agree that it's a shitty and confusing API. I can't think of a legitimate reason why Pod would require `Copy` (or even `Clone`) and the explanation why we ended up with putting these derives carelessly is likely "because bindgen applies all these derives" or "because I've seen these derives in bytemuck".

I'll try to get rid of that requirement.
___
*[`ebpf/aya-ebpf/src/spin_lock.rs` line 33 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAoqTiDltPmF84f76H:-OkFYIOdDsqKNxWFWShf:b-k7fwn0) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/ebpf/aya-ebpf/src/spin_lock.rs#L33)):*

shouldn't this be self.spin_lock? can you please add some helpers that take the correct pointer type so we don't footgun ourselves with void*?
</blockquote></details>

The problem is that the autogenerated helper has its own `bpf_

tamird (commented):
> @tamird made 3 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1467)* status: 6 of 16 files reviewed, 8 unresolved discussions (waiting on @vadorovsky).

___
*[`aya-common/src/spin_lock.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAnnNx2v-epRnnqbgN:-OkIovfIACWloL-dIzoW:b-k0th6z) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L7)):*

No, the unnamed field is not a requirement - just my choice, since we aren't really interested in accessing it, and naming it would make clippy/rustc complain that the field is unused. We just need to ensure that the layout and alignment of this struct is same as the kernel expects.
</blockquote></details>

You can give it a name starting with an underscore and the lint will be silent. That gives an opportunity to name it something explanatory.
___
*[`aya-common/src/spin_lock.rs` line 14 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAo-pw1cb2r-GSR9Bk:-OkIpkt9E2Rwugn498CW:b-1wpqvy) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/aya-common/src/spin_lock.rs#L14)):*

Yeah, giving it a thought, I do agree that it's a shitty and confusing API. I can't think of a legitimate reason why Pod would require `Copy` (or even `Clone`) and the explanation why we ended up with putting these derives carelessly is likely "because bindgen applies all these derives" or "because I've seen these derives in bytemuck".

I'll try to get rid of that requirement.
</blockquote></details>

Well I guess the reason is that e.g. `BPF_MAP_LOOKUP_ELEM` copies the bytes into user-supplied memory. So you need a way to indicate that it's safe to do that. So I guess you're just supposed to be OK with copying spin locks into userspace.
___
*[`ebpf/aya-ebpf/src/spin_lock.rs` line 33 at r2](https://reviewable.io/reviews/aya-rs/aya/1467#-OkAoqTiDltPmF84f76H:-OkIqNYv3Ny0QDHUkUdH:bqyinwb) ([raw file](https://github.com/aya-rs/aya/blob/c9707265dd6401bf971cd1bd8d5a5c13c4184b6e/ebpf/aya-ebpf/src/spin_lock.rs#L33)):*

The problem is that the autogenerated helper has its own `bpf_
</blockquote></details>

ah i just assumed it takes void* -- but the solution i propose still stands: we want our own helper that takes our pointer type so we can have some type safety. otherwise `.cast()` will just accept whatever

PR #1469, opened 2026-02-01
fix: check length in SkBuff's load_bytes function

### LINKED ISSUE #1207: Fix or examples required for socket filter byte loading
In socket filter programs, I cannot use the load_bytes function without getting a invalid zero-length read error in the verifier.

Code:
```let mut dst = [0;32];
let res = ctx.load_bytes(2, &mut dst);```

Error:
```18: (85) call bpf_skb_load_bytes#26
R4 invalid zero-sized read: u64=[0,31]
verification time 524 usec
stack depth 32+0
processed 190 insns (limit 1000000) max_states_per_insn 2 total_states 5 peak_states 5 mark_read 4```
I have tried adding a condition which early returns if ctx.len() < 1, but this did not resolve anything.

hsqStephenZhang:
Fix: https://github.com/aya-rs/aya/issues/1207

Should i create a new file for the eBPF program? Though that seems odd in current integration-test's file structure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1469)

### DIFF
--- a/ebpf/aya-ebpf/src/programs/sk_buff.rs
+++ b/ebpf/aya-ebpf/src/programs/sk_buff.rs
@@ -86,6 +86,10 @@ impl SkBuff {
         let len = usize::try_from(self.len()).map_err(|core::num::TryFromIntError { .. }| -1)?;
         let len = len.checked_sub(offset).ok_or(-1)?;
         let len = len.min(dst.len());
+        if len == 0 {
+            return Ok(0);
+        }
+
         let len_u32 = u32::try_from(len).map_err(|core::num::TryFromIntError { .. }| -1)?;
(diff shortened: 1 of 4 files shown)


hsqStephenZhang:
sorry i did not check it on earlier versions. seems that it works on linux 6.12, but fails on 5.10 and 6.1

hsqStephenZhang:
kernel has done a lot of improvement on verifier in 6.x including tnum and range tracking, so it's a reasonable observation that it succeeded on 6.12 but not on earlier ones.

hsqStephenZhang:
I'm still trying to figure out at which kernel version it's fixed.

Issue #953, opened 2024-05-26
aya: bloom_filter: contains is not working


jkoeppeler:
Hi,
I am having trouble to successfully execute the example code of the bloom_filter map:
```
// snippet
let mut bloom_filter = BloomFilter::try_from(bpf.map_mut("BLOOM_FILTER").unwrap())?;
bloom_filter.insert(1, 0)?;
assert!(bloom_filter.contains(&1, 0).is_ok());
``` 
The assert states that the item is not found. I want to use the bloom_filter it in an XDP program.

Steps to reproduce:
```
cargo generate --name myapp -d program_type=xdp https://github.com/aya-rs/aya-template
```
eBPF program:
```
#![no_std]
#![no_main]

use aya_ebpf::{bindings::xdp_action, macros::{map,xdp}, programs::XdpContext, maps::bloom_filter::BloomFilter};
use aya_log_ebpf::info;
#[map]
static mut MYFILTER: BloomFilter<u32> = BloomFilter::with_max_entries(1024,0);

#[xdp]
pub fn myapp(ctx: XdpContext) -> u32 {
    match try_myapp(ctx) {
        Ok(ret) => ret,
        Err(_) => xdp_action::XDP_ABORTED,
    }
}

fn try_myapp(ctx: XdpContext) -> Result<u32, u32> {
    Ok(xdp_action::XDP_PASS)
}

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    unsafe { core::hint::unreachable_unchecked() }
}
```

Userspace program:
```
use anyhow::Context;
use aya::programs::{Xdp, XdpFlags};
use aya::maps::bloom_filter::BloomFilter;
use aya::{include_bytes_aligned, Bpf};
use aya_log::BpfLogger;
use clap::Parser;
use log::{info, warn, debug};
use tokio::signal;

#[derive(Debug, Parser)]
struct Opt {
    #[clap(short, long, default_value = "enp2s0")]
    iface: String,
}

#[tokio::main]
async fn main() -> Result<(), anyhow::Error> {
    let opt = Opt::parse();

    env_logger::init();

    // Bump the memlock rlimit. This is needed for older kernels that don't use the
    // new memcg based accounting, see https://lwn.net/Articles/837122/
    let rlim = libc::rlimit {
        rlim_cur: libc::RLIM_INFINITY,
        rlim_max: libc::RLIM_INFINITY,
    };
[...]

PR #1462, opened 2026-01-27
maps: Add integration test for `BloomFilter`

vadorovsky:
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1462)

PR #1470, opened 2026-02-02
test: integration: add bloom filter tests

jkoeppeler:
This commit proves the problem in the bloom filter implementation in issue #953.
Calling contains from user space will always fail and does not work properly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1470)

### DIFF
--- a/test/integration-ebpf/Cargo.toml
+++ b/test/integration-ebpf/Cargo.toml
@@ -29,6 +29,10 @@ name = "array"
 path = "src/array.rs"
 
 [[bin]]
+name = "bloom_filter"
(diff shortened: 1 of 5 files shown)


jkoeppeler:
Implemented the expected-to-fail integration test for the bloom_filter map. As described in issue #953 the user space implementation of the contains function does not work properly. Let me know if I should improve something :)

Thanks!

tamird:
Possible dup of #1462?

jkoeppeler:
seems like it (at least partially).... Will close this. Thanks

PR #1471, opened 2026-02-02
fix: fix bloom_filter contains pointer dereference

hsqStephenZhang:
Fix https://github.com/aya-rs/aya/issues/953

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1471)


hsqStephenZhang:
Just saw https://github.com/aya-rs/aya/pull/1470

PR #1474, opened 2026-02-09
ebpf: Add helper for safer loading of globals

### LINKED ISSUE #1472: Safety of globals
Currently, the documentation for global variables states that they should be placed in a `static`, be `Pod` and loaded with `read_volatile`.
While the exact semantics of statics being mutated externally are not fully resolved, these values should ideally be placed in an `UnsafeCell` to inhibit assumptions that they will not change.
Additionally, it is easy for users to load from the static directly without a volatile read, which would most likely be optimized out.
Introducing a type like the following would be helpful to ensure code is safe.
```rs
pub struct Global<T>(UnsafeCell<T>);
unsafe impl<T: Sync> Sync for Global<T> {}

impl<T: Pod> Global<T> {
  fn load(&self) -> T {
    self.0.get().read_volatile()
  }
}
```
relevant discussion: https://github.com/rust-lang/unsafe-code-guidelines/issues/397

jamie-osec:
Closes #1472

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1474)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -245,26 +245,13 @@ impl<'a> EbpfLoader<'a> {
     ///
     /// If the `must_exist` argument is `true`, [`EbpfLoader::load`] will fail with [`ParseError::SymbolNotFound`] if the loaded object code does not contain the variable.
     ///
-    /// From Rust eBPF, a global variable can be defined as follows:
-    ///
-    /// ```no_run
-    /// #[unsafe(no_mangle)]
-    /// static VERSION: i32 = 0;
-    /// ```
-    ///
-    /// Then it can be accessed using `core::ptr::read_volatile`:
-    ///
-    /// ```no_run
-    /// # #[unsafe(no_mangle)]
-    /// # static VERSION: i32 = 0;
-    /// # fn try_test() {
-    /// let version = unsafe { core::ptr::read_volatile(&VERSION) };
-    /// # }
-    /// ```
+    /// From Rust eBPF, a global variable can be defined using `EbpfGlobal` - please refer to the `aya-ebpf` documentation.
     ///
     /// The type of a global variable must be `Pod` (plain old data), for instance `u8`, `u32` and
     /// all other primitive types. You may use custom types as well, but you must ensure that those
     /// types are `#[repr(C)]` and only contain other `Pod` types.
+    ///
+    /// The type used here and in the eBPF must be the same (or have a compatible layout).
     ///
     /// From C eBPF, you would annotate a global variable as `volatile const`.
     ///
--- a/ebpf/aya-ebpf/src/lib.rs
+++ b/ebpf/aya-ebpf/src/lib.rs
@@ -46,7 +46,10 @@ pub mod helpers;
 pub mod maps;
 pub mod programs;
 
-use core::ptr::{self, NonNull};
+use core::{
+    mem::MaybeUninit,
+    ptr::{self, NonNull},
+};
 
 pub use aya_ebpf_cty as cty;
 pub use aya_ebpf_macros as macros;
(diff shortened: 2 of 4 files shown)


aya/src/bpf.rs
```diff
@@ -245,20 +245,15 @@ impl<'a> EbpfLoader<'a> {
...
-    /// Then it can be accessed using `core::ptr::read_volatile`:
+    /// static VERSION: BpfGlobal<i32> = BpfGlobal::new();
     ///
-    /// ```no_run
-    /// # #[unsafe(no_mangle)]
-    /// # static VERSION: i32 = 0;
-    /// # fn try_test() {
-    /// let version = unsafe { core::ptr::read_volatile(&VERSION) };
+    /// # fn loadit() {
+    /// let version = unsafe { VERSION.load() };
```
tamird: Does `load` need to be unsafe?

ebpf/aya-ebpf/src/lib.rs
```diff
@@ -177,3 +181,54 @@ fn lookup<K, V>(def: *mut c_void, key: &K) -> Option<NonNull<V>> {
...
+    /// `T` must match the data type that was used by the loader.
+    #[inline]
+    pub unsafe fn load(&self) -> T {
+        unsafe { ptr::read_volatile(self.value.value.get() as *const T) }
+    }
+}
+
+/// [`UnsafeCell`] but [`Sync`].
+///
+/// Copy of the standard library's unstable `SyncUnsafeCell`
```
tamird: Can you link to the tracking issue with a TODO?

ebpf/aya-ebpf/src/lib.rs
```diff
@@ -177,3 +181,54 @@ fn lookup<K, V>(def: *mut c_void, key: &K) -> Option<NonNull<V>> {
     let key = ptr::from_ref(key);
     NonNull::new(unsafe { bpf_map_lookup_elem(def, key.cast()) }.cast())
 }
+
+/// A read-only global value that will be initialized by the loader.
+/// Prefer using this to a plain `static` variable to avoid compiler optimizations eliding reads.
+#[repr(transparent)]
+pub struct BpfGlobal<T> {
+    value: SyncUnsafeCell<MaybeUninit<T>>,
```
tamird: Does it need to hold an uninitialized value? Can't it just hold a T with the value passed to new serving as default?

aya/src/bpf.rs
jamie-osec: Not since we always initialize now; removed

jamie-osec:
~~I think CI failure as I added an `aya` -> `aya-ebpf` dev-dependency for the doctest, and `aya-ebpf` doesn't compile on armv7~~
Removed that, made the doctest no-build

ebpf/aya-ebpf/src/lib.rs
jamie-osec: `MaybeUninit` is used here as an extra guard against some optimizations, but it is now always initialized on construction

tamird:
> ~I think CI failure as I added an `aya` -> `aya-ebpf` dev-dependency for the doctest, and `aya-ebpf` doesn't compile on armv7~ Removed that, made the doctest no-build

Consider moving the doc test to where it can build and just mentioning the relevant BPF-side type in the userspace comment.

ebpf/aya-ebpf/src/lib.rs
tamird: Interesting. What is it guarding against? Can you please leave some comments?

jamie-osec:
Tried to clarify `load` a little, and added an explanatory comment for the safety properties of the storage. `MaybeUninit` is most likely not strictly necessary, but is zero-cost and I would recommend its usage whenever dealing with bytes that haven't come from the type system.

jamie-osec:
> I'm still confused about the word "must". Is this documentation for the implementor or the user? It should be for the user.

This is for the user, as in "if `override_global` was used, the data you loaded into it must be be a valid `T`".
This might make sense to be moved to/reiterated on `override_global` itself if you prefer.

tamird:
> > I'm still confused about the word "must". Is this documentation for the implementor or the user? It should be for the user.
> 
> This is for the user, as in "if `override_global` was used, the data you loaded into it must be be a valid `T`". This might make sense to be moved to/reiterated on `override_global` itself if you prefer.

I would prefer "is guaranteed to" rather than "must".

jamie-osec:
Well it's guaranteed to be _some_ valid type, but the type used by the userspace side is not guaranteed to be the same as the `T` here - therefore the user _must_ ensure they do so this method returns the correct result.

tamird:
Would you mind also porting https://github.com/aya-rs/aya/blob/baa7ddb3814fb1728318dd30ab43da8b139d60ce/ebpf/aya-log-ebpf/src/lib.rs#L37 to this new container type?

ebpf/aya-log-ebpf/src/lib.rs
```diff
@@ -34,12 +34,12 @@ pub mod macro_support {
     ///
     /// Userspace may patch this symbol before load via `EbpfLoader::override_global`.
     #[unsafe(no_mangle)]
-    pub static AYA_LOG_LEVEL: u8 = 0xff;
+    pub static AYA_LOG_LEVEL: EbpfGlobal<u8> = EbpfGlobal::new(0xff);
```
tamird, line 37: Looks like now that we're using this here the integration tests are failing.

jamie-osec:
```
thread 'tests::log::log_level_prevents_verif_fail' (13389) panicked at test/integration-test/src/tests/log.rs:305:17:
called `Result::unwrap()` on an `Err` value: LoadError { io_error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, verifier_log: 0: R1=ctx() R10=fp0
0: (18) r1 = 0xffff90759484c508       ; R1_w=map_value(map=.data,ks=4,vs=1)
2: (71) r1 = *(u8 *)(r1 +0)           ; R1_w=scalar(smin=smin32=0,smax=umax=smax32=umax32=255,var_off=(0x0; 0xff))
3: (b7) r2 = 4                        ; R2_w=4
4: (2d) if r2 > r1 goto pc+5          ; R1_w=scalar(smin=umin=smin32=umin32=4,smax=umax=smax32=umax32=255,var_off=(0x0; 0xff)) R2_w=4
5: (b7) r1 = 0                        ; R1_w=0
6: (63) *(u32 *)(r10 -4) = r1         ; R1_w=0 R10=fp0 fp-8=0000????
7: (bf) r1 = r10                      ; R1_w=fp0 R10=fp0
8: (07) r1 += -4                      ; R1_w=fp-4
9: (05) goto pc-1
9: (05) goto pc-1
infinite loop detected at insn 9
```
Odd. I'll see if I can understand what's happening here and get it working locally

tamird:
FYI we use an infinite loop as our panic handler in BPF

jamie-osec:
Interesting, might be to do with SUC causing `LOG_LEVEL` to be in .data and not .rodata? Removing SUC and just using MU passes fine.
If I'm reading the output right it seems like it's unconditionally panicking?
I might close this if you don't have any ideas for the cause, I'm not familiar enough with eBPF to know exactly what the issue is here :confused:

tamird:
I don't think it makes sense to close this. Even if we can't use SUC and/or MU we should have a wrapper type that builds-in the volatile read. Code is better than docs.

jamie-osec:
I removed the `SyncUnsafeCell` for now and kept `MaybeUninit`. This should still be a strict hardening of globals

tamird:
Looks like you'll need to bless the API again

tamird:
Thanks!

PR #1479, opened 2026-02-17
maps: add ring buffer batch API

tamird:
This implementation should have lower overhead than `RingBuffer::next` and
by avoiding the overhead associated with `RingBufItem`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479)

### DIFF
--- a/aya/src/maps/ring_buf.rs
+++ b/aya/src/maps/ring_buf.rs
@@ -132,7 +132,9 @@ impl<T> RingBuf<T> {
     )]
     pub fn next(&mut self) -> Option<RingBufItem<'_>> {
         let Self {
-            consumer, producer, ..
+            map: _,
+            consumer,
+            producer,
         } = self;
         producer.next(consumer)
     }
@@ -165,7 +167,7 @@ impl Deref for RingBufItem<'_> {
     type Target = [u8];
 
     fn deref(&self) -> &Self::Target {
-        let Self { data, .. } = self;
+        let Self { data, consumer: _ } = self;
         data
     }
 }
@@ -351,7 +353,7 @@ impl ProducerData {
             let caught_up = *pos == *pos_cache;
 
             if !caught_up {
-                match read_item(data_pages, *mask, consumer) {
+                match consumer.read_item(data_pages, *mask) {
                     Item::Busy => {}
                     Item::Discard { len } => {
                         consumer.consume(len);
@@ -375,44 +377,45 @@ impl ProducerData {
                 *pos_cache = load_producer_pos(mmap);
             }
         }
+    }
+}
 
-        enum Item<'a> {
-            Busy,
-            Discard { len: usize },
-            Data(&'a [u8]),
-        }
+enum Item<'a> {
+    Busy,
+    Discard { len: usize },
+    Data(&'a [u8]),
+}
 
-        fn read_item<'data>(data: &'data [u8], mask: u32, pos: &ConsumerPos) -> Item<'data> {
-            let ConsumerPos { pos, .. } = pos;
-            let offset = pos & usize::try_from(mask).unwrap();
-            #[expect(
-                clippy::panic,
-                reason = "invalid ring buffer layout is a fatal internal error"
-            )]
-            let must_get_data = |offset, len| {
-                data.get(offset..offset + len).unwrap_or_else(|| {
-                    panic!("{:?} not in {:?}", offset..offset + len, 0..data.len())
-                })
-            };
-            let header_ptr: *const AtomicU32 = must_get_data(offset, size_of::<AtomicU32>())
-                .as_ptr()
-                .cast();
-            // The kernel commits the record with a fully ordered xchg [1]. Pair it with an Acquire
-            // load so data written by the producer is visible after the busy bit is cleared.
-            //
-            // [1]: https://github.com/torvalds/linux/blob/eb26cbb1/kernel/bpf/ringbuf.c#L488
-            let header = unsafe { &*header_ptr }.load(Ordering::Acquire);
-            if header & BPF_RINGBUF_BUSY_BIT != 0 {
-                Item::Busy
+impl ConsumerPos {
+    fn read_item<'data>(&self, data: &'data [u8], mask: u32) -> Item<'data> {
+        let Self { pos, metadata: _ } = self;
+        let offset = pos & usize::try_from(mask).unwrap();
+        #[expect(
+            clippy::panic,
+            reason = "invalid ring buffer layout is a fatal internal error"
+        )]
+        let must_get_data = |offset, len| {
+            data.get(offset..offset + len)
+                .unwrap_or_else(|| panic!("{:?} not in {:?}", offset..offset + len, 0..data.len()))
+        };
+        let header_ptr: *const AtomicU32 = must_get_data(offset, size_of::<AtomicU32>())
+            .as_ptr()
+            .cast();
+        // The kernel commits the record with a fully ordered xchg [1]. Pair it with an Acquire
+        // load so data written by the producer is visible after the busy bit is cleared.
+        //
+        // [1]: https://github.com/torvalds/linux/blob/eb26cbb1/kernel/bpf/ringbuf.c#L488
+        let header = unsafe { &*header_ptr }.load(Ordering::Acquire);
+        if header & BPF_RINGBUF_BUSY_BIT != 0 {
+            Item::Busy
+        } else {
+            let len = usize::try_from(header & mask).unwrap();
+            if header & BPF_RINGBUF_DISCARD_BIT != 0 {
+                Item::Discard { len }
             } else {
-                let len = usize::try_from(header & mask).unwrap();
-                if header & BPF_RINGBUF_DISCARD_BIT != 0 {
-                    Item::Discard { len }
-                } else {
-                    let data_offset = offset + usize::try_from(BPF_RINGBUF_HDR_SZ).unwrap();
-                    let data = must_get_data(data_offset, len);
-                    Item::Data(data)
-                }
+                let data_offset = offset + usize::try_from(BPF_RINGBUF_HDR_SZ).unwrap();
+                let data = must_get_data(data_offset, len);
+                Item::Data(data)
             }
         }
     }
--- a/aya/src/maps/ring_buf.rs
+++ b/aya/src/maps/ring_buf.rs
@@ -6,8 +6,9 @@
 
 use std::{
     borrow::Borrow,
+    convert::Infallible,
     fmt::{self, Debug, Formatter},
-    ops::Deref,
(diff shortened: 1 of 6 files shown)


tamird:
@codex review

aya/src/maps/ring_buf.rs
```diff
@@ -314,98 +416,155 @@ impl ProducerData {
...
+            match consumer.read_item(data_pages, self.mask) {
+                Item::Busy => {
+                    if !flush(consumer, advanced) {
+                        break;
+                    }
+                }
+                Item::Discard { len } => consume(consumer, advanced, len),
+                Item::Data(data) => {
+                    // This must be deferred in case `f` panics.
+                    scopeguard::defer! { consume(consumer, advanced, data.len()) };
```
ajwerner, line 525: If f panics do you definitely want to consume the item? I think it's a reasonable contract for the use case because panicking in a loop is bad, but it is worth documenting and I could see it being pushed into f if needed (like the caller of this function in some way tracks if it panicked)

ajwerner (commented):
Did something concrete motivate this change? When I first saw the title I was an expecting a different sort of API that might enable the consumer to do vectorized processing of items and then to selectively choose where to commit from the returned batch. None of the existing APIs let the consumer client examine multiple items with a shared lifetime so you can't really make a tight loop over a batch. Maybe that doesn't matter. 

I can see how this patch helps reduce contention between the producer and consumer on the consumer position. I'm okay with this change but would love to see a motivating benchmark or something.

tamird:
> Did something concrete motivate this change? When I first saw the title I was an expecting a different sort of API that might enable the consumer to do vectorized processing of items and then to selectively choose where to commit from the returned batch. None of the existing APIs let the consumer client examine multiple items with a shared lifetime so you can't really make a tight loop over a batch. Maybe that doesn't matter.

It was sort of like that initially, but it ended up having worse performance than the existing API. You're right that there's no API for iterating in batches. I'm also not exactly sure how you're implement such an API because if you field 2 items and the later item is dropped, what do you do? How do you express this ordering in the type system?

> I can see how this patch helps reduce contention between the producer and consumer on the consumer position. I'm okay with this change but would love to see a motivating benchmark or something.

In addition to reducing contention it also optimizes the consumer; the consumer code path is up to 75% faster. I need to figure out how to properly integrate the benchmarks into the integration test harness.

alessandrod:
I haven't reviewed yet, but batching makes sense, I've been working on XDP stuff recently and batching vs non batching makes GB/s of difference

aya/src/maps/ring_buf.rs
tamird: That's a reasonable thing to call out. It also suggests that you might want control over consumption - you can imagine a combinator like TakeWhile that needs to peek and thus may elect to halt iteration without consuming the last item.

tamird (commented):
Reviews appreciated! I haven't had time to investigate the bench harness - running benchmarks in a VM is annoying (for collecting results)

> @tamird made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479#-:-OrTx8OX4cFknXK9_DJZ:ba6vffc)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on ajwerner and vadorovsky).

aya/src/maps/ring_buf.rs
tamird: Actually the `take_while` docs explicitly mention this; the rejected element is itself consumed.

https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.take_while

Still, this is a unlike a normal iterator because it's yielding borrowed data. Do you think it would be useful to allow the caller to explicitly (or implicitly, e.g. via panic) decide to not consume data?

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on ajwerner and vadorovsky).

vadorovsky:
> I haven't had time to investigate the bench harness - running benchmarks in a VM is annoying (for collecting results)

And also not indicative IMO, we should definitely bench on bare metal Linux hosts. If you tell me how do you intend to benchmark it - or even better - add some benchmarks alongside the tests, I'm happy to check it out on a 24 core Epyc machine.

tamird:
> > I haven't had time to investigate the bench harness - running benchmarks in a VM is annoying (for collecting results)
> 
> And also not indicative IMO, we should definitely bench on bare metal Linux hosts. If you tell me how do you intend to benchmark it - or even better - add some benchmarks alongside the tests, I'm happy to check it out on a 24 core Epyc machine.

Yeah i agree with this of course but we still need to support running the benchmarks in the VM so that we can easily check that they actually work. Maybe that narrows the scope enough to be feasible. I have it in a local branch, I'll try to find time to get this over the finish line.

aya/src/maps/ring_buf.rs
ajwerner: I think it's not worth doing anything about. If some consumer is so worried about panicking they can copy the data out or something like that.

ajwerner (changes requested):
> @ajwerner reviewed 14 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479)* status: all files reviewed, 1 unresolved discussion (waiting on tamird and vadorovsky).

___
*[`aya/src/maps/ring_buf.rs` line 467 at r3](https://reviewable.io/reviews/aya-rs/aya/1479#-OrVwn4z12-aQ64NEUFf:-OrVwn5-9J1hS-whPOYS:b-eyv5p2) ([raw file](https://github.com/aya-rs/aya/blob/403a44fe6d8e9a96cff5cc05233f79fe1df41df8/aya/src/maps/ring_buf.rs#L467)):*
> ```XML
>         consumer: &'a mut ConsumerPos,
>         mut advanced: Option<&'a mut bool>,
>     ) -> Option<RingBufItem<'a>> {
> ```

I feel like I must be missing something -- this function was introduced to take this `advanced` parameter but then as far as I can tell, it's only ever called with None. It seems to me like this function was intended to be used by `try_fold` but then some AI bot got frustrated with a borrow-checker and decided to back that out and just inline the code from here into `try_fold`?

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479)* status: all files reviewed, 1 unresolved discussion (waiting on ajwerner and vadorovsky).

___
*[`aya/src/maps/ring_buf.rs` line 467 at r3](https://reviewable.io/reviews/aya-rs/aya/1479#-OrVwn4z12-aQ64NEUFf:-OrW-d5Y-4z573eLO609:btra9bj) ([raw file](https://github.com/aya-rs/aya/blob/403a44fe6d8e9a96cff5cc05233f79fe1df41df8/aya/src/maps/ring_buf.rs#L467)):*

I feel like I must be missing something -- this function was introduced to take this `advanced` parameter but then as far as I can tell, it's only ever called with None. It seems to me like this function was intended to be used by `try_fold` but then some AI bot got frustrated with a borrow-checker and decided to back that out and just inline the code from here into `try_fold`?
</blockquote></details>

It's been a while since I did this but I think it was used in one of the earlier API shapes I tried.

I think this function just goes away completely if we go all in on the batch API.

ajwerner (changes requested):
> @ajwerner made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479)* status: all files reviewed, 1 unresolved discussion (waiting on tamird and vadorovsky).

___
*[`aya/src/maps/ring_buf.rs` line 467 at r3](https://reviewable.io/reviews/aya-rs/aya/1479#-OrVwn4z12-aQ64NEUFf:-OrW1Yqf2kFZV2fts6lX:bw3xdwv) ([raw file](https://github.com/aya-rs/aya/blob/403a44fe6d8e9a96cff5cc05233f79fe1df41df8/aya/src/maps/ring_buf.rs#L467)):*

It's been a while since I did this but I think it was used in one of the earlier API shapes I tried.

I think this function just goes away completely if we go all in on the batch API.
</blockquote></details>

fine, but why keep the `advanced` param if nobody uses it?

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on ajwerner and vadorovsky).

___
*[`aya/src/maps/ring_buf.rs` line 467 at r3](https://reviewable.io/reviews/aya-rs/aya/1479#-OrVwn4z12-aQ64NEUFf:-OrW35gzAA_OtWaZ7Apq:b-kbnuny) ([raw file](https://github.com/aya-rs/aya/blob/403a44fe6d8e9a96cff5cc05233f79fe1df41df8/aya/src/maps/ring_buf.rs#L467)):*

fine, but why keep the `advanced` param if nobody uses it?
</blockquote></details>

It's detritus, I'll remove it. I remember now - an earlier shape of this PR tried to yield a batch of items and do the consumer update only once, but that turned out not to be the expensive part, it was just all the bookkeeping of the RingBufItem itself.

I wasn't suggesting that this parameter would stay, I was just trying to redirect the discussion: does this API look reasonable? do we want to keep the old API?

Anyway, this parameter is removed and I reduced the diff a bit by undoing some of the unnecessary movement.

ajwerner (approved):
:lgtm:

> @ajwerner made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1479#-:-OrW810_8WsOLsA3LOHx:bnfp4nl)* status: 0 of 14 files reviewed, all discussions resolved (waiting on vadorovsky).

___
*[`aya/src/maps/ring_buf.rs` line 467 at r3](https://reviewable.io/reviews/aya-rs/aya/1479#-OrVwn4z12-aQ64NEUFf:-OrW684x10RhErK_EW_p:b6ph4nk) ([raw file](https://github.com/aya-rs/aya/blob/403a44fe6d8e9a96cff5cc05233f79fe1df41df8/aya/src/maps/ring_buf.rs#L467)):*
> does this API look reasonable?

Yes, it's reasonable. 

> do we want to keep the old API?

I think the old API should stay at least with a deprecation for existing users. For use cases that aren't super high throughput what was there was fine.

Not to be overly inspired by bad code, but cilium/ebpf is extremely widely used in Go and [it's just always copying out and committing the consumer offset](https://github.com/cilium/ebpf/blob/fd33a781ea9ebf9d1bff748707793deccc412c05/ringbuf/ring.go#L98).

> Anyway, this parameter is removed and I reduced the diff a bit by undoing some of the unnecessary movement.

Nice, looks good.

PR #1482, opened 2026-02-20
fix: tc attachment failure with long program names   

### LINKED ISSUE #610: tc load report 'no space left'

cong-or:
The TcRequest netlink attribute buffer was 64 bytes, which isn't enough when program names approach the kernel's 256-byte limit (CLS_BPF_NAME_LEN). This caused a misleading "no space left" error on attach.

Bumped the buffer to 292 bytes to fit all attributes with a max-length name.

Fixes #610
Closes #728

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1482)

### DIFF
--- a/aya/src/sys/netlink.rs
+++ b/aya/src/sys/netlink.rs
@@ -754,6 +754,8 @@ unsafe fn request_attributes<T>(req: &mut T, msg_len: usize) -> &mut [u8] {
 mod tests {
     use std::ffi::CString;
 
+    use assert_matches::assert_matches;
+
     use super::*;
 
     #[test]
@@ -880,6 +882,14 @@ mod tests {
         assert_eq!(name.to_str().unwrap(), "foo");
     }
 
+    fn tc_request(name: &[u8]) -> io::Result<()> {
+        let mut req = unsafe { mem::zeroed::<TcRequest>() };
+        let nlmsg_len = size_of::<nlmsghdr>() + size_of::<tcmsg>();
+        req.header.nlmsg_len = nlmsg_len as u32;
+
+        write_tc_attach_attrs(&mut req, nlmsg_len, 0, name)
+    }
+
     /// Verify that [`TcRequest`] fits all the attributes [`write_tc_attach_attrs`]
     /// writes, even with the kernel's maximum TC name length (`CLS_BPF_NAME_LEN`).
     ///
@@ -887,12 +897,7 @@ mod tests {
     /// long names failed with "no space left".
     #[test]
     fn tc_request_fits_max_length_name() {
-        let max_name = [b'a'; CLS_BPF_NAME_LEN];
-        let mut req = unsafe { mem::zeroed::<TcRequest>() };
-        let nlmsg_len = size_of::<nlmsghdr>() + size_of::<tcmsg>();
-        req.header.nlmsg_len = nlmsg_len as u32;
-
-        write_tc_attach_attrs(&mut req, nlmsg_len, 0, &max_name).unwrap();
+        assert_matches!(tc_request(&[b'a'; CLS_BPF_NAME_LEN]), Ok(()));
     }
 
     /// Verify that a name exceeding `CLS_BPF_NAME_LEN` is rejected.
@@ -900,11 +905,12 @@ mod tests {
     fn tc_request_rejects_oversized_name() {
         // One byte over the kernel's maximum — the attribute buffer is sized
         // exactly for CLS_BPF_NAME_LEN, so this should fail with "no space left".
-        let oversized_name = [b'a'; CLS_BPF_NAME_LEN + 1];
-        let mut req = unsafe { mem::zeroed::<TcRequest>() };
-        let nlmsg_len = size_of::<nlmsghdr>() + size_of::<tcmsg>();
-        req.header.nlmsg_len = nlmsg_len as u32;
-
-        assert!(write_tc_attach_attrs(&mut req, nlmsg_len, 0, &oversized_name).is_err());
+        assert_matches!(
+            tc_request(&[b'a'; CLS_BPF_NAME_LEN + 1]),
+            Err(err) => {
+                assert_eq!(err.kind(), io::ErrorKind::Other);
+                assert_eq!(err.to_string(), "no space left");
+            }
+        );
     }
 }


tamird:
Hi, thanks for the PR! There's a very old attempt to fix this in https://github.com/aya-rs/aya/pull/728 - I believe the sticking point there was absence of satisfying testing. Would you like to take a stab at writing a test?

cong-or:
> Hi, thanks for the PR! There's a very old attempt to fix this in #728 - I believe the sticking point there was absence of satisfying testing. Would you like to take a stab at writing a test?

Hey @tamird , thanks! I've added a test - it writes all the same netlink attributes as netlink_qdisc_attach into a TcRequest but with a max-length name (255 bytes). With the old buffer size it fails with "no space left", passes with the fix. Hopefully this addresses the testing gap from #728.

aya/src/sys/netlink.rs
```diff
@@ -852,4 +855,42 @@ mod tests {
         let name = CStr::from_bytes_with_nul(inner.data).unwrap();
         assert_eq!(name.to_str().unwrap(), "foo");
     }
+
+    /// Verify that [`TcRequest`] fits all the attributes [`netlink_qdisc_attach`]
+    /// writes, even with the kernel's maximum TC name length (`CLS_BPF_NAME_LEN` = 256).
+    ///
+    /// Before the buffer was enlarged, names approaching the 256-byte kernel
+    /// limit caused a "no space left" error during attachment.
```
tamird: you say that this happened during attachment, but this isn't testing attachment. what am I missing?

cong-or: the comment was misleading since the test only exercises serialization, not actual attachment. Fixed.

aya/src/sys/netlink.rs
```diff
@@ -369,7 +369,10 @@ unsafe impl Pod for Request {}
 struct TcRequest {
     header: nlmsghdr,
     tc_info: tcmsg,
-    attrs: [u8; 64],
+    // The kernel allows TC names up to 256 bytes (CLS_BPF_NAME_LEN). 292 is the
```
tamird: is it possible to do some const fn math to get this value? we should define a private constant CLS_BPF_NAME_LEN with a link to https://github.com/torvalds/linux/blob/v6.19/net/sched/cls_bpf.c#L28 in a comment

aya/src/sys/netlink.rs
```diff
@@ -852,4 +855,42 @@ mod tests {
...
+    /// Verify that [`TcRequest`] fits all the attributes [`netlink_qdisc_attach`]
+    /// writes, even with the kernel's maximum TC name length (`CLS_BPF_NAME_LEN` = 256).
+    ///
+    /// Before the buffer was enlarged, serializing the netlink attributes for
+    /// names approaching the 256-byte kernel limit failed with "no space left".
+    #[test]
+    fn tc_request_fits_max_length_name() {
+        // The kernel's CLS_BPF_NAME_LEN is 256 (including null terminator),
+        // so the longest valid name is 255 bytes.
+        let max_name = CString::new(vec![b'a'; 255]).unwrap();
```
tamird: you could just make this an array of size CLS_BPF_NAME_LEN, right? no need for the final byte to be NULL i suspect

aya/src/sys/netlink.rs
```diff
@@ -852,4 +855,42 @@ mod tests {
         let name = CStr::from_bytes_with_nul(inner.data).unwrap();
         assert_eq!(name.to_str().unwrap(), "foo");
     }
+
+    /// Verify that [`TcRequest`] fits all the attributes [`netlink_qdisc_attach`]
```
tamird: Ah, I think I better understand what's going on here. This test is duplicating a bunch of code from `netlink_qdisc_attach`. Instead, could you extract the relevant bits so that they can be tested?

aya/src/sys/netlink.rs
cong-or: Done, added the constant with the kernel link and derived the buffer size from it.

aya/src/sys/netlink.rs
cong-or: changed it to `[b'a'; CLS_BPF_NAME_LEN] `

aya/src/sys/netlink.rs
cong-or: extracted the attribute serialization into a write_tc_attach_attrs helper so both netlink_qdisc_attach and the test use the same code. Much cleaner, thanks for the suggestion.

aya/src/sys/netlink.rs
```diff
@@ -154,6 +158,26 @@ pub(crate) unsafe fn netlink_qdisc_add_clsact(if_index: i32) -> Result<(), Netli
...
 
+fn write_tc_attach_attrs(
+    req: &mut TcRequest,
+    nlmsg_len: usize,
+    prog_fd: i32,
+    prog_name: &[u8],
+) -> Result<(), io::Error> {
+    let attrs_buf = unsafe { request_attributes(req, nlmsg_len) };
+
+    let kind_len = write_attr_bytes(attrs_buf, 0, TCA_KIND as u16, b"bpf\0")?;
```
tamird: while you're here, make this `c"bpf"`

aya/src/sys/netlink.rs
```diff
@@ -369,7 +377,9 @@ unsafe impl Pod for Request {}
 struct TcRequest {
     header: nlmsghdr,
     tc_info: tcmsg,
-    attrs: [u8; 64],
+    // Must fit all netlink attributes written by netlink_qdisc_attach:
+    // TCA_KIND, and nested TCA_OPTIONS containing TCA_BPF_FD, TCA_BPF_NAME, TCA_BPF_FLAGS.
+    attrs: [u8; 5 * NLA_HDR_LEN + 4 + size_of::<i32>() + CLS_BPF_NAME_LEN + size_of::<u32>()],
```
tamird: could you put this calculation in a const fn so that we can also have a static assertion on its value? that way we can see both the value and the calculation

aya/src/sys/netlink.rs
```diff
@@ -852,4 +879,19 @@ mod tests {
...
         assert_eq!(name.to_str().unwrap(), "foo");
     }
+
+    /// Verify that [`TcRequest`] fits all the attributes [`write_tc_attach_attrs`]
+    /// writes, even with the kernel's maximum TC name length (`CLS_BPF_NAME_LEN`).
+    ///
+    /// Before the buffer was enlarged, serializing the netlink attributes for
+    /// long names failed with "no space left".
+    #[test]
+    fn tc_request_fits_max_length_name() {
```
tamird, line 899: could you write another test that asserts that a name with size CLS_BPF_NAME_LEN + 1 fails?

aya/src/sys/netlink.rs
cong-or: done

aya/src/sys/netlink.rs
cong-or: done

aya/src/sys/netlink.rs
cong-or: yes np, done.

tamird:
Thanks!

PR #1483, opened 2026-02-21
netlink: misc improvements

tamird:
- **netlink: use checked accessors**
- **netlink: use constants more consistently**
- **netlink: advance through write buffers**
- **netlink: avoid some allocations**

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1483)

### DIFF
--- a/aya/src/programs/tc.rs
+++ b/aya/src/programs/tc.rs
@@ -1,10 +1,5 @@
 //! Network traffic control programs.
-use std::{
-    ffi::{CStr, CString},
-    io,
-    os::fd::AsFd as _,
-    path::Path,
-};
+use std::{ffi::CString, io, os::fd::AsFd as _, path::Path};
 
 use aya_obj::generated::{
     TC_H_CLSACT, TC_H_MIN_EGRESS, TC_H_MIN_INGRESS,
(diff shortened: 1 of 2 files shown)


aya/src/sys/netlink.rs
```diff
@@ -473,58 +492,66 @@ impl NetlinkSocket {
...
+                                        source: io::Error::from_raw_os_error(-err.error),
+                                    },
+                                    None => NetlinkErrorInternal::IoError(
+                                        io::Error::from_raw_os_error(-err.error),
+                                    ),
+                                };
+                                return Err(e);
+                            }
+                            NLMSG_DONE => break 'out,
+                            _ => return Ok(Some(message)),
```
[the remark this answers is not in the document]
tamird: @codex review

aya/src/programs/tc.rs
```diff
@@ -610,30 +605,16 @@ pub fn qdisc_detach_program(
...
-    attach_type: TcAttachType,
-    name: &CStr,
-) -> Result<(), TcError> {
     let if_index = ifindex_from_ifname(if_name)? as i32;
 
-    let filter_info = unsafe { netlink_find_filter_with_name(if_index, attach_type, name)? };
+    let sock = NetlinkSocket::open().map_err(NetlinkError::from)?;
+    let filter_info = netlink_find_filter_with_name(&sock, if_index, attach_type, &cstr)?;
+    // Check for errors before detaching any programs.
+    let filter_info: Vec<_> = filter_info.collect::<Result<_, _>>()?;
```
vadorovsky, line 613: Since the commit message says "avoid some allocations", I think we could avoid this one as well. 😄

One way would be to call `peekable()` on the iterator returned by `netlink_find_filter_with_name` and do:

```rust
    if filter_info.peek().is_none() {
        return Err(TcError::IoError(io::Error::new(
            io::ErrorKind::NotFound,
            name.to_owned(),
        )));
    }
```

tamird: That addresses the `is_empty` check below, but not the `?` at the end of this line. The point is that we have an iterator of `Result` and we want to bail if any of them are `Err` before we start processing *any* of the `Ok` ones.

vadorovsky: Good point, let's leave it then

PR #1484, opened 2026-02-21
test: replace loopback with veth pairs in integration tests

cong-or:
The integration tests used to run on loopback, but that only supports generic (SKB-mode) XDP, which doesn’t really reflect a real production setup. Now, they use a veth pair instead, which supports native XDP and behaves much more like an actual NIC.

NetNsGuard sets up the veth pair. For the traffic-heavy tests like af_xdp, PeerNsGuard moves the peer into its own namespace, assigns IPs on both ends, and installs static ARP entries to prevent random ARP traffic from interfering with the XDP redirect programs.

The cpumap_chain test stays on loopback for now, because cpumap chaining doesn’t reliably work through veth on arm64. As a result, the old 5.15 kernel version gate for generic XDP mode remains in place.

Additionally, there’s a small veth_connectivity test that sends a plain UDP packet across the pair with no BPF attached. If anything’s broken in the basic setup, it fails quickly and clearly, avoiding confusing issues down the line.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)


cong-or:
Closes #422

cong-or:
> The integration tests were running on loopback, which only supports generic (SKB-mode) XDP — not something that reflects real production setups. They now use a veth pair instead, which supports native XDP and behaves much more like a real NIC.
> 
> NetNsGuard sets up the veth pair. For the traffic-heavy tests (af_xdp, cpumap_chain), PeerNsGuard moves the peer into its own namespace, assigns IPs on both ends, and installs static ARP entries so random ARP traffic doesn’t get picked up by XDP redirect programs.
> 
> Native XDP on veth supports cpumap chaining starting from kernel 5.9, so the old 5.15 version gate (which was only there for generic mode on loopback) can be dropped.
> 
> There’s also a small veth_connectivity test that sends a plain UDP packet across the pair without any BPF attached. If the basic wiring is broken, it fails fast and clearly instead of causing confusing downstream failures.
> 
> This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)

  **Follow-up:** replaced `ip` command invocations with direct netlink syscalls. The CI VM doesn't have iproute2 installed.

tamird (changes requested):
Seems like you tried to solve this, but then the LLM ran into some problems, gave up, and reverted most of it?

> @tamird reviewed 7 files and all commit messages, and made 22 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484#-:-On3Nrpg3H-MOyTNhF-s:b-fw38al)* status: all files reviewed, 21 unresolved discussions (waiting on cong-or).

___
*[`-- commits` line 35 at r8](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Nf6C6yEvi7Y6EN2I:-On3Nf6C6yEvi7Y6EN2J:bvu0dsd):*
so the LLM just gave up in the end?
___
*[`test/integration-test/src/netlink.rs` line 4 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3N21iDn02W6s2naHe:-On3N21iDn02W6s2naHf:b-mqcj75) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L4)):*
> ```XML
> //!
> //! These live here (rather than in the `aya` crate) to avoid adding test-only
> //! public API surface to the production library.
> ```

on the other hand...there's a fair bit of duplication here



_Code quote:_
```XML
//! Minimal netlink helpers for integration-test network setup.
//!
//! These live here (rather than in the `aya` crate) to avoid adding test-only
//! public API surface to the production library.
```
___
*[`test/integration-test/src/netlink.rs` line 32 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3MtEz7J69qohXpYCb:-On3MtEz7J69qohXpYCc:b-460xdl) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L32)):*
> ```XML
> #[derive(Copy, Clone)]
> #[repr(C)]
> struct Ifinfomsg {
> ```

can we use bindgen for this?
___
*[`test/integration-test/src/utils.rs` line 148 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3K-MW612KIbeaMjeP:-On3K-MW612KIbeaMjeQ:bsufw58) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L148)):*
> ```XML
>     const PEER_IFACE: &str = "veth1";
>     pub(crate) const IFACE_ADDR: &str = "10.0.0.1";
>     pub(crate) const PEER_ADDR: &str = "10.0.0.2";
> ```

while I understand that these are strings, does it make sense to use them unless you have a `NetNsGuard` in scope? if not, perhaps these should be methods that return a &str with a lifetime matching that of the NetNsGuard/Peer (even though they'd really be static strings)



_Code quote:_
```XML
    pub(crate) const IFACE_ADDR: &str = "10.0.0.1";
    pub(crate) const PEER_ADDR: &str = "10.0.0.2";
```
___
*[`test/integration-test/src/utils.rs` line 152 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3KFsp76RAKYx07Che:-On3KFsp76RAKYx07Chf:b-boe6at) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L152)):*
> ```XML
>     pub(crate) fn name(&self) -> &str {
>         &self.name
>     }
> ```

do you need this? it's only used in this same file, where the field is accessible



_Code quote:_
```XML
    pub(crate) fn name(&self) -> &str {
        &self.name
    }
```
___
*[`test/integration-test/src/utils.rs` line 215 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LxBe3qswNQ-iYAUB:-On3LxBe3qswNQ-iYAUC:b-qsuilr) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L215)):*
> ```XML
>             "name",
>             Self::PEER_IFACE,
>         ]);
> ```

shouldn't this happen only on the creation of the peer?



_Code quote:_
```XML
        // Create a veth pair for tests that attach XDP/TC programs. Veth supports
        // native XDP (unlike loopback which only supports SKB/generic mode), so tests
        // exercise the same code path used in production.
        run_ip(&[
            "link",
            "add",
            Self::IFACE,
            "type",
            "veth",
            "peer",
            "name",
            Self::PEER_IFACE,
        ]);
```
___
*[`test/integration-test/src/utils.rs` line 219 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3L86eE3ClH67Atw_Z:-On3L86eE3ClH67Atw__:bzacvye) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L219)):*
> ```XML
>         // Bring up both ends.
>         for iface in [Self::IFACE, Self::PEER_IFACE] {
>             let name = CString::new(iface).unwrap();
> ```

consider making the constants C-strings

EDIT: this is also duplicated above
___
*[`test/integration-test/src/utils.rs` line 223 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LJU5Aiyi4ps2DYo3:-On3LJU5Aiyi4ps2DYo4:b-i13nu2) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L223)):*
> ```XML
>                 let idx = if_nametoindex(name.as_ptr());
>                 assert!(
>                     idx != 0,
> ```

i suppose this can be `assert_ne` (but keep the custom message)
___
*[`test/integration-test/src/utils.rs` line 278 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LVtJEmNANg3WoT6M:-On3LVtJEmNANg3WoT6N:b-8nnu6g) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L278)):*
> ```XML
>         .status()
>         .unwrap_or_else(|e| panic!("ip {}: {e}", args.join(" ")));
>     assert!(status.success(), "ip {} failed", args.join(" "));
> ```

you can just keep `cmd` in scope and use its debug output (here and below
___
*[`test/integration-test/src/utils.rs` line 302 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LkBJ6B0Hwhhdm2Ku:-On3LkBJ6B0Hwhhdm2Kv:b-sihhu4) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L302)):*
> ```XML
>     full_args.extend(args);
>     run_ip_output(&full_args)
> }
> ```

this is a lot of helpers and they don't make it easy to understand what's going on

we also have existing use of "ip", so either use these everywhere or nowhere please



_Code quote:_
```XML
fn run_ip_output(args: &[&str]) -> String {
    let output = Command::new("ip")
        .args(args)
        .output()
        .unwrap_or_else(|e| panic!("ip {}: {e}", args.join(" ")));
    assert!(output.status.success(), "ip {} failed", args.join(" "));
    String::from_utf8(output.stdout).unwrap()
}

/// Run `ip netns exec <ns> ip <args...>`.
fn run_ip_netns(ns: &str, args: &[&str]) {
    let mut full_args = vec!["netns", "exec", ns, "ip"];
    full_args.extend(args);
    run_ip(&full_args);
}

/// Run `ip netns exec <ns> ip <args...>` and return stdout.
fn run_ip_netns_output(ns: &str, args: &[&str]) -> String {
    let mut full_args = vec!["netns", "exec", ns, "ip"];
    full_args.extend(args);
    run_ip_output(&full_args)
}
```
___
*[`test/integration-test/src/utils.rs` line 405 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3M4yI5P2urcWBOu0r:-On3M4yI5P2urcWBOu0s:bg6m81w) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L405)):*
> ```XML
>             .unwrap_or_else(|e| panic!("open(\"{}\"): {e}", peer_ns_path.display()));
> 
>         std::thread::scope(|s| {
> ```

scope for one thread? why?
___
*[`test/integration-test/src/tests/xdp.rs` line 18 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3IecoEMVOktetRc7O:-On3IecoEMVOktetRc7P:b-f1sqvo) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L18)):*
> ```XML
> #[test_log::test]
> fn veth_connectivity() {
>     // peer must be declared after netns so it drops first (see PeerNsGuard docs).
> ```

useless comment; this is enforced by the compiler
___
*[`test/integration-test/src/tests/xdp.rs` line 22 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ihxt0wp9_FEkRc0X:-On3Ihxt0wp9_FEkRc0Y:b10mw6i) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L22)):*
> ```XML
>     let peer = PeerNsGuard::new(&netns);
> 
>     let sock = UdpSocket::bind(format!("{}:0", NetNsGuard::IFACE_ADDR)).unwrap();
> ```

I think you can pass `(NetNsGuard::IFACE_ADDR, 0)`
___
*[`test/integration-test/src/tests/xdp.rs` line 28 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ilwz7P811xpdEYXA:-On3Ilwz7P811xpdEYXB:b7imlg5) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L28)):*
> ```XML
> 
>     peer.run(|| {
>         let sock = UdpSocket::bind(format!("{}:0", NetNsGuard::PEER_ADDR)).unwrap();
> ```

ditto, but here i think there's no need to bind at all?
___
*[`test/integration-test/src/tests/xdp.rs` line 34 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3InfT0NTLZkykZJdP:-On3InfT0NTLZkykZJdQ:b4yywok) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L34)):*
> ```XML
>     let mut buf = [0u8; 16];
>     let n = sock.recv(&mut buf).unwrap();
>     assert_eq!(&buf[..n], b"veth ok");
> ```

extract constant (and use its size for the buf above)
___
*[`test/integration-test/src/tests/xdp.rs` line 43 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3J8174xdZ568f7Xp-:-On3J8174xdZ568f7Xp0:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L43)):*
> ```XML
> )]
> fn af_xdp() {
>     // peer must be declared after netns so it drops first (see PeerNsGuard docs).
> ```

ditto
___
*[`test/integration-test/src/tests/xdp.rs` line 108 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3J9o57_eQe54PyI0A:-On3J9o57_eQe54PyI0B:bywx0z9) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L108)):*
> ```XML
>     writer.commit();
> 
>     let dst = format!("{}:1777", NetNsGuard::IFACE_ADDR);
> ```

same here, can be a tuple
___
*[`test/integration-test/src/tests/xdp.rs` line 197 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3JN9Q0eOzlrB8gX-n:-On3JN9Q0eOzlrB8gX-o:b-7ropln) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L197)):*
> ```XML
> #[test_log::test]
> fn cpumap_chain() {
>     // peer must be declared after netns so it drops first (see PeerNsGuard docs).
> ```

ditto
___
*[`test/integration-test/src/tests/xdp.rs` line 238 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3JwJ2-7cddqhGlCBo:-On3JwJ2-7cddqhGlCBp:bb7od0r) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L238)):*
> ```XML
>     const PAYLOAD: &str = "hello cpumap";
> 
>     let sock = UdpSocket::bind(format!("{}:0", NetNsGuard::IFACE_ADDR)).unwrap();
> ```

same
___
*[`test/integration-test/src/tests/xdp.rs` line 243 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Jwm95U0YDLlIniRq:-On3Jwm95U0YDLlIniRr:bb7od0r) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L243)):*
> ```XML
>         .unwrap();
>     peer.run(|| {
>         let sock = UdpSocket::bind(format!("{}:0", NetNsGuard::PEER_ADDR)).unwrap();
> ```

same
___
*[`test/integration-test/src/tests/xdp.rs` line 227 at r6](https://reviewable.io/reviews/aya-rs/aya/1484#-On3NB35-CHAdmNDhMDA:-On3NB35-CHAdmNDhMDB:b-oau3l) ([raw file](https://github.com/aya-rs/aya/blob/45480e3ce05a8bc4c6a980063b8b2c75e5615bf7/test/integration-test/src/tests/xdp.rs#L227)):*
> ```XML
>     // not reliably deliver packets through veth on older kernels (confirmed
>     // failing on 5.10). Gate at 6.1 which is the oldest CI-tested kernel where
>     // this works end-to-end.
> ```

a bit unsatisfying :(



_Code quote:_
```XML
    // While veth supports native XDP attachment from 5.9, cpumap chaining does
    // not reliably deliver packets through veth on older kernels (confirmed
    // failing on 5.10). Gate at 6.1 which is the oldest CI-tested kernel where
    // this works end-to-end.
```

cong-or (commented):
The veth work is actually all there — PeerNsGuard creates the pair via raw netlink (CI VM doesn't have iproute2), moves one end into a peer namespace, and wires up IP + static ARP. The XDP and load tests are all using veth now.                                                                                                                                                     
                                         
The only exception is cpumap_chain — spent a while trying to get CI to pass with that one on veth but cpumap chaining doesn't work with veth on arm64 since it needs generic/SKB-mode XDP, which only loopback supports. So that's a kernel limitation rather than something I gave up on — loopback is the correct interface for that test.

Commit history is a bit messy from all the CI iteration, I can clean it up?

> @cong-or made 22 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484#-:-On8yzotAfJqPkOlV6Lw:b-vz59n9)* status: 0 of 27 files reviewed, 21 unresolved discussions (waiting on tamird).

___
*[`-- commits` line 35 at r8](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Nf6C6yEvi7Y6EN2I:-On9-DkO1sX15IO-u3BT:b-896fix):*

so the LLM just gave up in the end?
</blockquote></details>

Done.
___
*[`test/integration-test/src/netlink.rs` line 4 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3N21iDn02W6s2naHe:-On4VMmd21b2HjSs3DZ5:bs8btx4) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L4)):*

on the other hand...there's a fair bit of duplication here
</blockquote></details>

Yeah, it's a trade-off — didn't want to add test-only stuff to the aya crate API. Open to suggestions if you have a better idea though. 
___
*[`test/integration-test/src/netlink.rs` line 32 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3MtEz7J69qohXpYCb:-On91dP85lXBpiDuIAsW:b-ab0jwk) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L32)):*

can we use bindgen for this?
</blockquote></details>

ifinfomsg is already coming from bindgen via aya_obj::generated. I could add Ifaddrmsg and Ndmsg to the codegen too — they're just not there yet. VETH_INFO_PEER and NLMSG_ALIGNTO are one-liner constants that libc doesn't export, didn't seem worth pulling them through bindgen but can do if you'd prefer.
___
*[`test/integration-test/src/utils.rs` line 148 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3K-MW612KIbeaMjeP:-On93BY2CNQ2Hwvg8bjw:b2775ni) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L148)):*

while I understand that these are strings, does it make sense to use them unless you have a `NetNsGuard` in scope? if not, perhaps these should be methods that return a &str with a lifetime matching that of the NetNsGuard/Peer (even though they'd really be static strings)
</blockquote></details>

Good point, changed them to methods on PeerNsGuard so you need a live instance to access them.
___
*[`test/integration-test/src/utils.rs` line 152 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3KFsp76RAKYx07Che:-On93_816s_dQM_XZ-Za:bx922lj) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L152)):*

do you need this? it's only used in this same file, where the field is accessible
</blockquote></details>

I removed it
___
*[`test/integration-test/src/utils.rs` line 215 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LxBe3qswNQ-iYAUB:-On94l4Y8dyUdCQzd9YL:byw3fmf) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L215)):*

shouldn't this happen only on the creation of the peer?
</blockquote></details>

Yeah that's fixed now — moved it into PeerNsGuard::new() so it only runs when a test actually needs it
___
*[`test/integration-test/src/utils.rs` line 219 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3L86eE3ClH67Atw_Z:-On9DZ380Oyw9fJJYvTw:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L219)):*

consider making the constants C-strings

EDIT: this is also duplicated above
</blockquote></details>

Done.
___
*[`test/integration-test/src/utils.rs` line 223 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LJU5Aiyi4ps2DYo3:-On9c_FI54EVK_2NeQcn:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L223)):*

i suppose this can be `assert_ne` (but keep the custom message)
</blockquote></details>

Done.
___
*[`test/integration-test/src/utils.rs` line 278 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LVtJEmNANg3WoT6M:-On9dA6V8-s33-KA7Xcr:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L278)):*

you can just keep `cmd` in scope and use its debug output (here and below
</blockquote></details>

Done.
___
*[`test/integration-test/src/utils.rs` line 302 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3LkBJ6B0Hwhhdm2Ku:-On9dOgF8n1uM9Zbond2:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L302)):*

this is a lot of helpers and they don't make it easy to understand what's going on

we also have existing use of "ip", so either use these everywhere or nowhere please
</blockquote></details>

Done.
___
*[`test/integration-test/src/utils.rs` line 405 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3M4yI5P2urcWBOu0r:-On9ehclEeZOK93GxqgV:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/utils.rs#L405)):*

scope for one thread? why?
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 18 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3IecoEMVOktetRc7O:-On9eoH22LcH0xgFyqSB:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L18)):*

useless comment; this is enforced by the compiler
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 22 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ihxt0wp9_FEkRc0X:-On9euayAcNpQIzVAd7X:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L22)):*

I think you can pass `(NetNsGuard::IFACE_ADDR, 0)`
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 28 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ilwz7P811xpdEYXA:-On9fFxC5bguaBzNhuH1:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L28)):*

ditto, but here i think there's no need to bind at all?
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 34 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3InfT0NTLZkykZJdP:-On9fHrl0gH2f60GQa4U:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L34)):*

extract constant (and use its size for the buf above)
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 43 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3J8174xdZ568f7Xp-:-On9fRD9AsiACVx6X262:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L43)):*

ditto
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 108 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3J9o57_eQe54PyI0A:-On9fRfyAtZG4MeFq-e-:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L108)):*

same here, can be a tuple
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 197 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3JN9Q0eOzlrB8gX-n:-On9fZSL5sprbaXrnc3d:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L197)):*

ditto
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 238 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3JwJ2-7cddqhGlCBo:-On9fc-oENjDfiV8UygD:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L238)):*

same
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 243 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Jwm95U0YDLlIniRq:-On9fbmL9cKtCkA2Mw-r:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L243)):*

same
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/xdp.rs` line 227 at r6](https://reviewable.io/reviews/aya-rs/aya/1484#-On3NB35-CHAdmNDhMDA:-On9frXp8h03TMKI-5Vd:b3md3tb) ([raw file](https://github.com/aya-rs/aya/blob/45480e3ce05a8bc4c6a980063b8b2c75e5615bf7/test/integration-test/src/tests/xdp.rs#L227)):*

a bit unsatisfying :(
</blockquote></details>

Yeah I'm not thrilled about it either. Couldn't figure out why cpumap chaining wouldn't work through veth on arm64 so I just kept it on loopback. Open to ideas

tamird (commented):
Yeah cleaning up would be good. Apologies for the delay getting back to this. Just a few comments, thanks for figuring out some of the kernel bugs.

> @tamird reviewed 27 files and all commit messages, made 9 comments, and resolved 19 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484#-:-OqXFRBK1Q-a-xMPqsR_:b-oaikzp)* status: all files reviewed, 7 unresolved discussions (waiting on cong-or).

___
*[`test/integration-test/src/netlink.rs` line 4 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3N21iDn02W6s2naHe:-Oq_mYjVBX7GsJLHmQPE:b6ad417) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L4)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Yeah, it's a trade-off — didn't want to add test-only stuff to the aya crate API. Open to suggestions if you have a better idea though.
</blockquote></details>

@alessandrod pinky promises he's going to rewrite a lot of the netlink machinery Real Soon Now so I guess we can leave it here.

Note that I did a pretty big rewrite of the netlink stuff since this PR opened, so if you followed the earlier style you could consider revising. Up to you.
___
*[`test/integration-test/src/netlink.rs` line 32 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3MtEz7J69qohXpYCb:-OqXFplI-D3xohpNvN6a:be3czd2) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L32)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

ifinfomsg is already coming from bindgen via aya_obj::generated. I could add Ifaddrmsg and Ndmsg to the codegen too — they're just not there yet. VETH_INFO_PEER and NLMSG_ALIGNTO are one-liner constants that libc doesn't export, didn't seem worth pulling them through bindgen but can do if you'd prefer.
</blockquote></details>

The more we can generate the better, I think.
___
*[`test/integration-test/src/tests/xdp.rs` line 28 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ilwz7P811xpdEYXA:-OqXG-VO47DULxDi540i:bbwfy6w) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L28)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Done.
</blockquote></details>

you can use either

https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html#associatedconstant.LOCALHOST

or

https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html#associatedconstant.UNSPECIFIED

but the choice just be justified, i guess you need UNSPECIFIED so you can send on non-localhost? the best way to document would be to extract a local constant or value and assign it UNSPECIFIED; the name you give it would document its intent.
___
*[`test/integration-test/src/tests/xdp.rs` line 33 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-OqXGBHS-1auzlkbSeyz:-OqXGBHS-1auzlkbSez-:bj6yd27) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/tests/xdp.rs#L33)):*
> ```XML
>     });
> 
>     let mut buf = [0u8; PAYLOAD.len()];
> ```

this should be oversized since you're asserting a precise read
___
*[`test/integration-test/src/utils.rs` line 239 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nLgREl-KZDQIgwxw:-Oq_nLgREl-KZDQIgwxx:b-ju2tx8) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L239)):*
> ```XML
> /// Run a closure inside a network namespace identified by a file handle.
> ///
> /// Uses a scoped thread because `setns` changes the network namespace of the
> ```

why does it need to be *scoped*? why can't it just be a scopeless std::thread::spawn?
___
*[`test/integration-test/src/utils.rs` line 279 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nbYkCsmJoQNtta_2:-Oq_nbYkCsmJoQNtta_3:b-adj9e9) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L279)):*
> ```XML
> 
>     #[expect(clippy::unused_self, reason = "access requires a live PeerNsGuard")]
>     pub(crate) fn iface(&self) -> &'static str {
> ```

i think the whole point is for this to be *not* static since you want the lifetime to be bound by the guard. ditto below
___
*[`test/integration-test/src/utils.rs` line 323 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nmwO1sMh26V97Yr-:-Oq_nmwO1sMh26V97Yr0:b-ebjgz9) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L323)):*
> ```XML
>         }
> 
>         // Create peer netns: create a persist file, then use a scoped thread
> ```

same here, you don't need the scope or at least it's not clear why you do
___
*[`test/integration-test/src/utils.rs` line 467 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_o2Tw1D59jBf8cZat:-Oq_o2Tw1D59jBf8cZau:b-1g1at6) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L467)):*
> ```XML
>     /// Run a closure inside the peer network namespace.
>     ///
>     /// Uses a scoped thread because `setns` changes the network namespace of the
> ```

would be good to reduce the number of copies of this

cong-or (commented):
> @cong-or made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)* status: 22 of 29 files reviewed, 7 unresolved discussions (waiting on tamird).

___
*[`test/integration-test/src/netlink.rs` line 4 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3N21iDn02W6s2naHe:-OqbhV8U-IzscjAJjHBJ:bv10ufr) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L4)):*

@alessandrod pinky promises he's going to rewrite a lot of the netlink machinery Real Soon Now so I guess we can leave it here.

Note that I did a pretty big rewrite of the netlink stuff since this PR opened, so if you followed the earlier style you could consider revising. Up to you.
</blockquote></details>

I'll leave it here for now, happy to adapt once the rewrite lands.
___
*[`test/integration-test/src/netlink.rs` line 32 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3MtEz7J69qohXpYCb:-Oqbh_EmB9NMATc1WXyA:b-ajd1mv) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L32)):*

The more we can generate the better, I think.
</blockquote></details>

Pulled NLMSG_ALIGNTO from aya_obj::generated. For VETH_INFO_PEER I added linux/veth.h to the wrapper header and the allowlist in xtask, but I don't have cross sysroots locally so couldn't run codegen — left a TODO on the local constant for now
___
*[`test/integration-test/src/utils.rs` line 239 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nLgREl-KZDQIgwxw:-OqbiYXp187R_qXWTM9T:b-ajmsoj) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L239)):*

why does it need to be *scoped*? why can't it just be a scopeless std::thread::spawn?
</blockquote></details>

Needs to be scoped because ns_file is borrowed, and f isn't 'static — once we tie the iface_addr() lifetime to the guard, callers end up capturing non-'static refs. Could dup the fd but we'd still need F: 'static which would break those call sites. Added a note in the doc.
___
*[`test/integration-test/src/utils.rs` line 279 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nbYkCsmJoQNtta_2:-OqbhwBx9lOuprIahlx1:b5870to) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L279)):*

i think the whole point is for this to be *not* static since you want the lifetime to be bound by the guard. ditto below
</blockquote></details>

Fixed, returns &str now so the lifetime is bound by the guard.
___
*[`test/integration-test/src/utils.rs` line 323 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nmwO1sMh26V97Yr-:-Oqbi-tn9e9f_XF95rRA:br31e4f) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L323)):*

same here, you don't need the scope or at least it's not clear why you do
</blockquote></details>

You're right, switched to std::thread::spawn — just needed to clone the path.
___
*[`test/integration-test/src/utils.rs` line 467 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_o2Tw1D59jBf8cZat:-Oqbi3q13ThW4eNr5l-I:bprkekf) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L467)):*

would be good to reduce the number of copies of this
</blockquote></details>

Consolidated on run_in_netns, PeerNsGuard::run just points there now.
___
*[`test/integration-test/src/tests/xdp.rs` line 28 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ilwz7P811xpdEYXA:-OqbihcK8RAHr16QVU30:b80bxqp) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L28)):*

you can use either

https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html#associatedconstant.LOCALHOST

or

https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html#associatedconstant.UNSPECIFIED

but the choice just be justified, i guess you need UNSPECIFIED so you can send on non-localhost? the best way to document would be to extract a local constant or value and assign it UNSPECIFIED; the name you give it would document its intent.
</blockquote></details>

Switched to Ipv4Addr::UNSPECIFIED, added a comment — we need it so the kernel picks the right source (10.0.0.2) for the veth route.
___
*[`test/integration-test/src/tests/xdp.rs` line 33 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-OqXGBHS-1auzlkbSeyz:-Oqbi8ujEoqDMD7joc6W:bk5znoz) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/tests/xdp.rs#L33)):*

this should be oversized since you're asserting a precise read
</blockquote></details>

Good catch, oversized to 512

tamird (changes requested):
> @tamird reviewed 7 files and all commit messages, made 8 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)* status: all files reviewed, 6 unresolved discussions (waiting on cong-or).

___
*[`test/integration-test/src/netlink.rs` line 4 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3N21iDn02W6s2naHe:-Oqbs8n7DPLojYiB3WIG:b-r7u18z) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L4)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

I'll leave it here for now, happy to adapt once the rewrite lands.
</blockquote></details>

It landed in https://github.com/aya-rs/aya/pull/1483
___
*[`test/integration-test/src/netlink.rs` line 32 at r4](https://reviewable.io/reviews/aya-rs/aya/1484#-On3MtEz7J69qohXpYCb:-OqbsDxT9HSUfTFMeXlg:b-boffj5) ([raw file](https://github.com/aya-rs/aya/blob/959100c9ad6eebcad0bad8b6929187f4ea547a12/test/integration-test/src/netlink.rs#L32)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Pulled NLMSG_ALIGNTO from aya_obj::generated. For VETH_INFO_PEER I added linux/veth.h to the wrapper header and the allowlist in xtask, but I don't have cross sysroots locally so couldn't run codegen — left a TODO on the local constant for now
</blockquote></details>

Cool. Codegen should run automatically once this PR lands.
___
*[`test/integration-test/src/utils.rs` line 239 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nLgREl-KZDQIgwxw:-OqbsLTX2OOPf1XGiJID:b-8bz2jy) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L239)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Needs to be scoped because ns_file is borrowed, and f isn't 'static — once we tie the iface_addr() lifetime to the guard, callers end up capturing non-'static refs. Could dup the fd but we'd still need F: 'static which would break those call sites. Added a note in the doc.
</blockquote></details>

Great; I think it's fine to keep the scoped thread to avoid cloning or whatever, I just don't think we should document that here.

As mentioned below, a better phrasing of this comment would focus on the Send bound and explain that the closure must run on another thread - without going into irrelevant details like the thread being scoped.
___
*[`test/integration-test/src/utils.rs` line 323 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nmwO1sMh26V97Yr-:-OqbsSnN1Hx1YxW76Qp9:bumabm8) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L323)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

You're right, switched to std::thread::spawn — just needed to clone the path.
</blockquote></details>

I think it's fine to use a scoped thread to save an allocation, I just wanted us to be precise about the reasoning.
___
*[`test/integration-test/src/utils.rs` line 467 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_o2Tw1D59jBf8cZat:-Oqbsad2Bfywu2Oe9hpd:b-jz0iz4) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L467)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Consolidated on run_in_netns, PeerNsGuard::run just points there now.
</blockquote></details>

I think the comment should be here; `run_in_netns` is private. The comment should talk about the bounds rather than the underlying thread being scoped or not.
___
*[`test/integration-test/src/tests/xdp.rs` line 28 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ilwz7P811xpdEYXA:-Oqbt1TA12Bmn6mUoVHQ:b-vj47sy) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L28)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Switched to Ipv4Addr::UNSPECIFIED, added a comment — we need it so the kernel picks the right source (10.0.0.2) for the veth route.
</blockquote></details>

Again, the comment should cover all uses of this constant, not just one. That's why I suggested giving that constant a local alias with a descriptive name. The comment can then be anchored on that alias.
___
*[`test/integration-test/src/tests/xdp.rs` line 33 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-OqXGBHS-1auzlkbSeyz:-Oqbsy2k6L4Ic4n0dnt4:b-zaq1xr) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/tests/xdp.rs#L33)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

Good catch, oversized to 512
</blockquote></details>

I think it can just be `PAYLOAD.len() + 1` to avoid magic numbers.
___
*[`test/integration-test/src/tests/xdp.rs` line 33 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-OqbtCunDHK1hB9gIvB1:-OqbtCunDHK1hB9gIvB2:b-7rz6xb) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/tests/xdp.rs#L33)):*
> ```XML
> 
>     // Send from the peer namespace. Bind UNSPECIFIED so the kernel picks the
>     // right source address (10.0.0.2) for the veth route.
> ```

if you want the source address to be X, why not bind to X?

tamird (changes requested):
> @tamird made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)* status: all files reviewed, 8 unresolved discussions (waiting on cong-or).

___
*[`test/integration-test/src/netlink.rs` line 29 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-OqbtbOp9GOF4GmjOAAn:-OqbtbOp9GOF4GmjOAAo:bco4juc) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/netlink.rs#L29)):*
> ```XML
> const VETH_INFO_PEER: u16 = 1;
> 
> /// `struct ifaddrmsg` from `linux/if_addr.h`.
> ```

can we codegen these structures too?
___
*[`test/integration-test/src/netlink.rs` line 55 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-Oqbtdso-gbwj_vVtrS9:-Oqbtdso-gbwj_vVtrSA:b-crahoj) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/netlink.rs#L55)):*
> ```XML
> #[derive(Copy, Clone)]
> #[repr(C)]
> struct LinkRequest {
> ```

these 4 don't have comments; should they?

cong-or (commented):
> @cong-or made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)* status: 24 of 29 files reviewed, 8 unresolved discussions (waiting on tamird).

___
*[`test/integration-test/src/netlink.rs` line 29 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-OqbtbOp9GOF4GmjOAAn:-OqbzwJN-ponj33YPQk9:b-pkjww) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/netlink.rs#L29)):*

can we codegen these structures too?
</blockquote></details>

Added linux/if_addr.h and linux/neighbour.h to the wrapper and ifaddrmsg/ndmsg to the types allowlist. Same as VETH_INFO_PEER — local definitions stay until codegen is re-run.
___
*[`test/integration-test/src/netlink.rs` line 55 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-Oqbtdso-gbwj_vVtrS9:-Oqc-0pfAlNoZt2EIsHP:bwplui) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/netlink.rs#L55)):*

these 4 don't have comments; should they?
</blockquote></details>

They're internal request envelopes (header + message + attr buffer), don't think comments add much. Can add them if you'd prefer.
___
*[`test/integration-test/src/utils.rs` line 239 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nLgREl-KZDQIgwxw:-Oqc1Ou6D532dWYKz_Oq:b4euhlo) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L239)):*

Great; I think it's fine to keep the scoped thread to avoid cloning or whatever, I just don't think we should document that here.

As mentioned below, a better phrasing of this comment would focus on the Send bound and explain that the closure must run on another thread - without going into irrelevant details like the thread being scoped.
</blockquote></details>

Stripped the doc here down to a one-liner, moved the explanation to PeerNsGuard::run and focused on the bounds.
___
*[`test/integration-test/src/utils.rs` line 323 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_nmwO1sMh26V97Yr-:-Oqc1VH94jf3IGLMik6q:b-rlconm) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L323)):*

I think it's fine to use a scoped thread to save an allocation, I just wanted us to be precise about the reasoning.
</blockquote></details>

Reverted back to scoped, agreed.
___
*[`test/integration-test/src/utils.rs` line 467 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-Oq_o2Tw1D59jBf8cZat:-Oqc1_ke4X1wWfB_m_zN:b-amcs58) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/utils.rs#L467)):*

I think the comment should be here; `run_in_netns` is private. The comment should talk about the bounds rather than the underlying thread being scoped or not.
</blockquote></details>

Done — doc is here now, talks about F: Send and not 'static rather than scoped threads
___
*[`test/integration-test/src/tests/xdp.rs` line 28 at r1](https://reviewable.io/reviews/aya-rs/aya/1484#-On3Ilwz7P811xpdEYXA:-Oqc1mMc3gsVywliFYNw:bnyzv45) ([raw file](https://github.com/aya-rs/aya/blob/1ae048a0450bc261b8b0beb0e820ce8d76d566e4/test/integration-test/src/tests/xdp.rs#L28)):*

Again, the comment should cover all uses of this constant, not just one. That's why I suggested giving that constant a local alias with a descriptive name. The comment can then be anchored on that alias.
</blockquote></details>

Ended up going a different direction — added a peer_addr() method and bind to it explicitly. let peer_ip = peer.peer_addr() is the local alias used everywhere now. 
___
*[`test/integration-test/src/tests/xdp.rs` line 33 at r14](https://reviewable.io/reviews/aya-rs/aya/1484#-OqXGBHS-1auzlkbSeyz:-Oqc0Xpo5vvwU4opOtVw:bb7f2mf) ([raw file](https://github.com/aya-rs/aya/blob/8ed05f050dd86fb3e1bfc01f1225554b73d1eb2a/test/integration-test/src/tests/xdp.rs#L33)):*

I think it can just be `PAYLOAD.len() + 1` to avoid magic numbers.
</blockquote></details>

done
___
*[`test/integration-test/src/tests/xdp.rs` line 33 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-OqbtCunDHK1hB9gIvB1:-Oqc0VCr63fTOfTBA38o:bbxhaca) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/tests/xdp.rs#L33)):*

if you want the source address to be X, why not bind to X?
</blockquote></details>

Fair point, added a peer_addr() method and bind to it directly now.

tamird (changes requested):
> @tamird reviewed 3 files and all commit messages, made 4 comments, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)* status: 26 of 29 files reviewed, 6 unresolved discussions (waiting on cong-or).

___
*[`test/integration-test/src/netlink.rs` line 55 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-Oqbtdso-gbwj_vVtrS9:-Oqc8WNK4hJrcTdSMP3b:b-3g7onj) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/netlink.rs#L55)):*
<details><summary><i>Previously, cong-or wrote…</i></summary><blockquote>

They're internal request envelopes (header + message + attr buffer), don't think comments add much. Can add them if you'd prefer.
</blockquote></details>

Ah, I hadn't squinted at them properly. Now that I am, I don't quite understand the sizes of these attrs buffers. Are they significant? If so, can you document? if not, and the only requirement is that they are big enough so that you can use `bytes_of` to write them in a single libc::send call, then I wonder if we could do this with a vectorized write instead e.g. using [`writev`](https://docs.rs/libc/latest/libc/fn.writev.html).

EDIT: I ended up playing with this [here](https://github.com/aya-rs/aya/pull/1541) (and that's why it took me this long to get back to this review). I think we can do much better with macros and get layouts that are entirely known at compile time, but for this PR I'd be happy with better commentary on magic numbers.
___
*[`test/integration-test/src/utils.rs` line 272 at r17](https://reviewable.io/reviews/aya-rs/aya/1484#-OrZ2RafBtH3ULuiWCqY:-OrZ2RafBtH3ULuiWCqZ:b-ny58kp) ([raw file](https://github.com/aya-rs/aya/blob/b41d616c1ecd02e15234256a3df0ee2e1190bd54/test/integration-test/src/utils.rs#L272)):*
> ```XML
>     const IFACE_IP: Ipv4Addr = Ipv4Addr::new(10, 0, 0, 1);
>     const PEER_IP: Ipv4Addr = Ipv4Addr::new(10, 0, 0, 2);
>     const IFACE_ADDR: &str = "10.0.0.1";
> ```

there's no reason for this to exist AFAICT because all places where it is used can just as well use an `Ipv4Addr`
___
*[`test/integration-test/src/utils.rs` line 285 at r17](https://reviewable.io/reviews/aya-rs/aya/1484#-OrZ1pJt9vSwKPHWwJRK:-OrZ1pJt9vSwKPHWwJRL:b7k9sgw) ([raw file](https://github.com/aya-rs/aya/blob/b41d616c1ecd02e15234256a3df0ee2e1190bd54/test/integration-test/src/utils.rs#L285)):*
> ```XML
> 
>     #[expect(clippy::unused_self, reason = "access requires a live PeerNsGuard")]
>     pub(crate) fn peer_addr(&self) -> Ipv4Addr {
> ```

this is a bit silly because this returns a thing that is Copy
___
*[`test/integration-test/src/utils.rs` line 289 at r17](https://reviewable.io/reviews/aya-rs/aya/1484#-OrZ2pOY99tJ27bwxb2y:-OrZ2pOY99tJ27bwxb2z:bmhyxp8) ([raw file](https://github.com/aya-rs/aya/blob/b41d616c1ecd02e15234256a3df0ee2e1190bd54/test/integration-test/src/utils.rs#L289)):*
> ```XML
>     }
> 
>     pub(crate) fn new(netns: &NetNsGuard) -> Self {
> ```

what would happen on a second call? seems like nothing good! This makes me wonder if instead of this shape we should have an alternative NetNsGuard function that returns the NetNsGuard and its peer so that it is clear that the relationship is importantly 1:1.

cong-or (commented):
> @cong-or made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1484)* status: 23 of 29 files reviewed, 6 unresolved discussions (waiting on tamird).

___
*[`test/integration-test/src/netlink.rs` line 55 at r15](https://reviewable.io/reviews/aya-rs/aya/1484#-Oqbtdso-gbwj_vVtrS9:-Ori12xj82zTfvhWxu6B:b-4th5wi) ([raw file](https://github.com/aya-rs/aya/blob/1f9270dd299aa2256bdc87fc7fbd70c48e6ba0b8/test/integration-test/src/netlink.rs#L55)):*

Ah, I hadn't squinted at them properly. Now that I am, I don't quite understand the sizes of these attrs buffers. Are they significant? If so, can you document? if not, and the only requirement is that they are big enough so that you can use `bytes_of` to write them in a single libc::send call, then I wonder if we could do this with a vectorized write instead e.g. using [`writev`](https://docs.rs/libc/latest/libc/fn.writev.html).

EDIT: I ended up playing with this [here](https://github.com/aya-rs/aya/pull/1541) (and that's why it took me this long to get back to this review). I think we can do much better with macros and get layouts that are entirely known at compile time, but for this PR I'd be happy with better commentary on magic numbers.
</blockquote></details>

The sizes just need to be large enough for the attributes — I've added comments on each struct breaking down the payload math. Happy to move to a compile-time layout approach in a follow-up if you want to take that direction. 
___
*[`test/integration-test/src/utils.rs` line 272 at r17](https://reviewable.io/reviews/aya-rs/aya/1484#-OrZ2RafBtH3ULuiWCqY:-Ori4F823hxmi8ELI1YO:b-1wz6bc) ([raw file](https://github.com/aya-rs/aya/blob/b41d616c1ecd02e15234256a3df0ee2e1190bd54/test/integration-test/src/utils.rs#L272)):*

there's no reason for this to exist AFAICT because all places where it is used can just as well use an `Ipv4Addr`
</blockquote></details>

Removed, callers use Ipv4Addr directly now. 
___
*[`test/integration-test/src/utils.rs` line 285 at r17](https://reviewable.io/reviews/aya-rs/aya/1484#-OrZ1pJt9vSwKPHWwJRK:-Ori3j170qmoodm5iVXG:b-xcukvp) ([raw file](https://github.com/aya-rs/aya/blob/b41d616c1ecd02e15234256a3df0ee2e1190bd54/test/integration-test/src/utils.rs#L285)):*

this is a bit silly because this returns a thing that is Copy
</blockquote></details>

 Done, switched to pub constants.
___
*[`test/integration-test/src/utils.rs` line 289 at r17](https://reviewable.io/reviews/aya-rs/aya/1484#-OrZ2pOY99tJ27bwxb2y:-Ori3tRf4y4-BE_hh307:bga68vs) ([raw file](https://github.com/aya-rs/aya/blob/b41d616c1ecd02e15234256a3df0ee2e1190bd54/test/integration-test/src/utils.rs#L289)):*

what would happen on a second call? seems like nothing good! This makes me wonder if instead of this shape we should have an alternative NetNsGuard function that returns the NetNsGuard and its peer so that it is clear that the relationship is importantly 1:1.
</blockquote></details>

Agreed — made PeerNsGuard::new() private, only way to get one is NetNsGuard::with_peer() which returns both together.

tamird (approved):
Reviewed at the author’s direct request. The current revision addresses the earlier concerns about one-to-one namespace ownership, explicit peer addressing, packet-length assertions, documented netlink request sizes, and the existing kernel-version fallback. The cross-architecture and Linux integration checks passed. The branch still needs its merge conflict resolved.

— tamirdex

cong-or:
> Reviewed at the author’s direct request. The current revision addresses the earlier concerns about one-to-one namespace ownership, explicit peer addressing, packet-length assertions, documented netlink request sizes, and the existing kernel-version fallback. The cross-architecture and Linux integration checks passed. The branch still needs its merge conflict resolved.
> 
> — tamirdex

@tamird  Rebased on `main` and resolved the conflicts.

One thing to call out: `main` moved the netns helpers into `aya::test_helpers` and made `NetNsGuard::new()` fallible.

Per your point about keeping test-only stuff out of the `aya` crate API, I left `PeerNsGuard` and the netlink code in the integration tests. The only awkward bit is that `NetNsGuard` is a foreign type now, so `NetNsGuard::with_peer()` couldn't stay as an inherent method. I changed it to `PeerNsGuard::with_netns() -> (NetNsGuard, PeerNsGuard)` instead.

Happy to change that if you had something else in mind.

PR #1485, opened 2026-02-22
Implement BPF_MAP_TYPE_TASK_STORAGE on kernel side.

sugarraysam:
As titled. Kernel side of #203 

Will wait for #1161 to land so we can complete the integration test. Mostly copied the `sk_storage` type, and only adding support for `#[btf_map]` assuming this is the way going forward for BPF.

Regen public API files /w `❯ cargo xtask public-api --bless`

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1485)

### DIFF
--- a/test/integration-ebpf/src/task_storage.rs
+++ b/test/integration-ebpf/src/task_storage.rs
@@ -1,0 +1,59 @@
+#![no_std]
+#![no_main]
+#![expect(unused_crate_dependencies, reason = "used in other bins")]
+
+use aya_ebpf::{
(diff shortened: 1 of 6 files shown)


test/integration-ebpf/src/task_storage.rs
```diff
@@ -0,0 +1,59 @@
+#![no_std]
```
sugarraysam, line 1: Please ignore this program for now, will modify it in due time.

PR #1489, opened 2026-03-03
aya-ebpf: document nul termination

tamird:
See https://docs.ebpf.io/linux/helper-function/bpf_get_current_comm/.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1489)

### DIFF
--- a/ebpf/aya-ebpf/src/helpers.rs
+++ b/ebpf/aya-ebpf/src/helpers.rs
@@ -569,8 +569,11 @@ pub unsafe fn bpf_probe_write_user<T>(dst: *mut T, src: *const T) -> Result<(), 
     if ret == 0 { Ok(()) } else { Err(ret as i32) }
 }
 
-/// Read the `comm` field associated with the current task struct
-/// as a `[u8; 16]`.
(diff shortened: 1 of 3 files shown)


ebpf/aya-ebpf/src/helpers.rs
```diff
@@ -569,8 +569,11 @@ pub unsafe fn bpf_probe_write_user<T>(dst: *mut T, src: *const T) -> Result<(),
     if ret == 0 { Ok(()) } else { Err(ret as i32) }
 }
 
-/// Read the `comm` field associated with the current task struct
-/// as a `[u8; 16]`.
+pub const TASK_COMM_LEN: usize = 16;
```
vadorovsky, line 572: It is already defined here https://github.com/aya-rs/aya/blob/db64fcf73b620c3e3c5ca42140bf238de1dbd47b/ebpf/aya-ebpf/src/lib.rs#L62

tamird: This is moved from there.

PR #1497, opened 2026-03-17
aya-obj: expand bpf_map_def to 36 bytes for tc maps

### LINKED ISSUE #156: Load Classifier BPF failed:  RelocationError
The Classifier BPF source contains a hashmap in `maps` section. It is referenced by a program in `egress` section as documented in `tc-bpf` manual.

However, when loading via `Bpf::load_file(xx)`, I get an `RelocationError` saying:

```
RelocationError { function: "egress", error: SectionNotFound { section_index: 5, symbol_index: 4, symbol_name: Some("config_map") } }
```

This bpf object can be loaded via `tc filter` command successfully, and the map can be shown via `bpftool map`.

output of `llvm-readelf -S`:

```
Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .strtab           STRTAB          0000000000000000 000220 000073 00      0   0  1
  [ 2] .text             PROGBITS        0000000000000000 000040 000000 00  AX  0   0  4
  [ 3] classifier/egress PROGBITS        0000000000000000 000040 0000c8 00  AX  0   0  8
  [ 4] .relclassifier/egress REL         0000000000000000 000200 000010 10     10   3  8
  [ 5] maps              PROGBITS        0000000000000000 000108 000024 00  WA  0   0  4
  [ 6] license           PROGBITS        0000000000000000 00012c 000004 00  WA  0   0  1
  [ 7] .rodata.str1.16   PROGBITS        0000000000000000 000130 000010 01 AMS  0   0 16
  [ 8] .eh_frame         PROGBITS        0000000000000000 000140 000030 00   A  0   0  8
  [ 9] .rel.eh_frame     REL             0000000000000000 000210 000010 10     10   8  8
  [10] .symtab           SYMTAB          0000000000000000 000170 000090 18      1   3  8
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  R (retain), p (processor specific)
```

ramos-99:
bpf_map_def was 28 bytes (7 fields); iproute2/tc emits bpf_elf_map at
36 bytes (9 fields), so inner_id and inner_idx were truncated.

This widens bpf_map_def to 36 bytes and rejects what aya doesn't support:

- tc pinning -> `UnsupportedLegacyPinning`
- map-of-maps via inner_id/inner_idx -> `UnsupportedLegacyMapInMap`

Note this is a behavior change: tc objects declaring pinning previously
loaded and now fail.

The fixtures compile against iproute2's own bpf_elf.h via a submodule at
v7.2.0. Two things to flag:

- `bpf_elf.h` includes `<asm/types.h>`, which iproute2 doesn't ship and
  `-nostdlibinc` excludes. `vmlinux.h` already defines the typedefs, so
  `bpf/headers/asm/types.h` is an empty stub.
- Bazel fetches the release tarball, via `http_archive` since its CI job checks out without
  submodules, same split libbpf has.

Fixes: #156

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1497)

### DIFF
--- a/test/integration-test/src/tests/load.rs
+++ b/test/integration-test/src/tests/load.rs
@@ -4,7 +4,7 @@ use std::{
 
 use assert_matches::assert_matches;
 use aya::{
-    Ebpf,
+    Ebpf, EbpfError,
     maps::{Array, RingBuf},
     pin::PinError,
     programs::{
@@ -21,7 +21,7 @@ use aya::{
     },
     util::KernelVersion,
 };
-use aya_obj::programs::XdpAttachType;
+use aya_obj::{ParseError, programs::XdpAttachType};
 use rstest::rstest;
 
 const MAX_RETRIES: usize = 100;
@@ -704,4 +704,15 @@ fn pin_lifecycle_uprobe() {
 
     // Make sure the function isn't optimized out.
     uprobe_function();
+}
+
+#[test_log::test]
+fn legacy_map_in_map_fails_at_load() {
+    let err = Ebpf::load(crate::TC_LEGACY_MAP_IN_MAP).unwrap_err();
+    assert_matches!(
+        err,
+        EbpfError::ParseError(ParseError::UnsupportedLegacyMapInMap { name }) => {
+            assert_eq!(name, "tc_map_in_map");
+        }
+    );
 }
--- a/aya-obj/src/maps.rs
+++ b/aya-obj/src/maps.rs
@@ -109,7 +109,11 @@ impl TryFrom<u32> for PinningType {
     }
 }
 
-/// Map definition in legacy BPF map declaration style
+/// Map definition in legacy BPF map declaration style.
+///
+/// This mirrors `struct bpf_elf_map`, the layout used by iproute2/tc's
+/// legacy BPF loader. See
+/// <https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/include/bpf_elf.h>.
 #[repr(C)]
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub struct bpf_map_def {
@@ -127,8 +131,17 @@ pub struct bpf_map_def {
     // optional features
     /// Id
     pub id: u32,
-    /// Pinning type
+    /// Pinning type. Only [`PinningType::None`] is supported for legacy maps.
     pub pinning: PinningType,
+    /// Inner map id. In `bpf_elf_map`, `inner_id` and `inner_idx` wire up a
+    /// map-of-maps: a non-zero `inner_id` identifies another map in the same
+    /// object to use as the inner map type. Aya does not implement
+    /// map-of-maps for legacy maps, so parsing rejects a non-zero `inner_id`
+    /// or `inner_idx` with
+    /// [`ParseError::UnsupportedLegacyMapInMap`](crate::ParseError::UnsupportedLegacyMapInMap).
+    pub inner_id: u32,
+    /// Inner map index; see [`inner_id`](Self::inner_id).
+    pub inner_idx: u32,
 }
 
 /// The first five __u32 of `bpf_map_def` must be defined.
@@ -304,6 +317,8 @@ impl Map {
                 map_flags: flags,
                 id: 0,
                 pinning: PinningType::None,
+                inner_id: 0,
+                inner_idx: 0,
             },
             inner_def: None,
             section_index: 0,
--- a/aya-obj/src/obj.rs
+++ b/aya-obj/src/obj.rs
@@ -1030,6 +1030,13 @@ pub enum ParseError {
     /// No BTF parsed for object
     #[error("no BTF parsed for object")]
     NoBTF,
+
+    #[error("map `{name}` uses unsupported legacy pinning")]
+    UnsupportedLegacyPinning { name: String },
+
+    /// Map-of-maps is not supported for legacy (non-BTF) maps
+    #[error("map `{name}` uses unsupported legacy map-in-map")]
+    UnsupportedLegacyMapInMap { name: String },
 }
 
 /// Invalid bindings to the bpf type from the parsed/received value.
@@ -1273,16 +1280,28 @@ fn parse_map_def(name: &str, data: &[u8]) -> Result<bpf_map_def, ParseError> {
         });
     }
 
-    if data.len() < size_of::<bpf_map_def>() {
+    let map_def = if data.len() < size_of::<bpf_map_def>() {
         let mut map_def = bpf_map_def::default();
         unsafe {
             let map_def_ptr = from_raw_parts_mut(ptr::from_mut(&mut map_def).cast(), data.len());
             map_def_ptr.copy_from_slice(data);
         }
-        Ok(map_def)
+        map_def
     } else {
-        Ok(unsafe { ptr::read_unaligned(data.as_ptr().cast()) })
+        unsafe { ptr::read_unaligned(data.as_ptr().cast()) }
+    };
+
+    if map_def.pinning != PinningType::None {
+        return Err(ParseError::UnsupportedLegacyPinning {
+            name: name.to_owned(),
+        });
     }
+    if map_def.inner_id != 0 || map_def.inner_idx != 0 {
+        return Err(ParseError::UnsupportedLegacyMapInMap {
+            name: name.to_owned(),
+        });
+    }
+    Ok(map_def)
 }
 
 fn parse_btf_map_def(
@@ -1465,6 +1484,8 @@ pub const fn parse_map_info(info: bpf_map_info, pinned: PinningType) -> Map {
                 map_flags: info.map_flags,
                 pinning: pinned,
                 id: info.id,
+                inner_id: 0,
+                inner_idx: 0,
             },
             inner_def: None,
             section_index: 0,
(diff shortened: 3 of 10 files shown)


ramos-99:
Ping @tamird, can you check this when you have time?

tamird:
Looks to me like this needs an integration test.

@codex review

test/integration-test/src/tests/load.rs
```diff
@@ -675,3 +675,14 @@ fn pin_lifecycle_uprobe() {
...
+#[test_log::test]
+fn legacy_36_byte_maps() {
+    let mut bpf = Ebpf::load(crate::TC_LEGACY_MAP).unwrap();
+
+    let map_1: Array<_, u32> = bpf.take_map("tc_map_1").unwrap().try_into().unwrap();
+    let map_2: Array<_, u32> = bpf.take_map("tc_map_2").unwrap().try_into().unwrap();
+
+    assert_eq!(map_1.len(), 1);
+    assert_eq!(map_2.len(), 2);
+}
```
tamird, line 771: what is being tested here? reading the change description and this test name i am not understanding what i should be checking here.

ramos-99: each map has a distinct max_elem (1 and 2) to verify both are parsed correctly from the 36-byte struct. the parser is backwards compatible, so it also handles the 28-byte variant without needing separate test cases

alessandrod:
Thanks for the patch!

We should probably document that we don't support iproute2's style of map-of-maps, or maybe make it work after https://github.com/aya-rs/aya/pull/1446 lands.

With the PR as is I think we have UB because of this https://kernel.googlesource.com/pub/scm/network/iproute2/iproute2-next/%2B/refs/tags/v6.10.0/include/bpf_elf.h#29

In general iproute2 pinning seems to work differently from aya/libbpf, the paths don't match. Maybe we should reject loading if pinning != 0?

ramos-99:
Thanks for the review and the context!

Totally agree that silently ignoring pinning is a problem. Since iproute2's pinning paths differ from aya's, returning an error if pinning != 0 makes a lot of sense to fail fast.

On the map-of-maps side: since iproute2's style of map-of-maps isn't supported yet, would it also make sense to reject inner_id != 0 and inner_idx != 0 for now? That way users get a clear error instead of silent failure, and the restriction can be dropped once support is added.

Happy to implement both checks and add a documentation note if that sounds good.

ramos-99:
I've gone ahead and implemented the pinning fail-fast and added documentation notes for pinning and map-of-maps. Let me know if it aligns with what you were thinking.

ramos-99:
Friendly ping @alessandrod, could you take a look at this when you have the time? Thanks!

tamird:
@ramos-99 is your intention that we preserve the commits here, or squash on merge? if you prefer a squash, can you do that yourself please? if you want to preserve commits, please confirm the current history makes sense.

ramos-99:
Happy to squash, will do.

test/integration-test/src/tests/load.rs
tamird: i think in general this PR is missing citations. `tc_legacy_map` defines `struct bpf_elf_map` with these other fields but has no citations for where this definition comes from. This PR claims it is adding support for a style of map used by some specific toolchain but nothing in the code anchors the support to that toolchain's actual definitions.

tamird (changes requested):
> @tamird reviewed 5 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1497)* status: 5 of 7 files reviewed, 4 unresolved discussions (waiting on alessandrod and ramos-99).

___
*[`aya-obj/src/maps.rs` line 132 at r5](https://reviewable.io/reviews/aya-rs/aya/1497#-OxpdGZr3xNarYw6Lg8O:-OxpdGZr3xNarYw6Lg8P:b-as374) ([raw file](https://github.com/aya-rs/aya/blob/4de5d14a2a9549781ecf024624c89c2b3c09a803/aya-obj/src/maps.rs#L132)):*
> ```XML
>     /// Pinning type. Only [`PinningType::None`] is supported for legacy maps.
>     pub pinning: PinningType,
>     /// Inner map id. Map-of-maps is not supported for legacy maps.
> ```

Can these comments be a bit clearer i.e. that these fields are for map-of-maps and thus are always zero for legacy maps? Or should these be `Option<NonZero>`?

tamird (changes requested):
> @tamird reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1497)* status: 7 of 9 files reviewed, 5 unresolved discussions (waiting on alessandrod and ramos-99).

___
*[`test/integration-test/bpf/tc_legacy_map.bpf.c` line 7 at r6](https://reviewable.io/reviews/aya-rs/aya/1497#-Oxum_d63U1j3yh1NjbU:-Oxum_d63U1j3yh1NjbV:b-a5edm7) ([raw file](https://github.com/aya-rs/aya/blob/855d2745973763a0fb7379f6cccd6741611385fc/test/integration-test/bpf/tc_legacy_map.bpf.c#L7)):*
> ```C
> 
> // Layout of iproute2/tc's legacy map declaration struct. See
> // https://git.kernel.org/pub/scm/linux/kernel/git/jkirsher/iproute2.git/tree/include/bpf_elf.h
> ```

these tests aren't very convincing, they are just about tautological. is the intention to allow aya to load iproute2-flavored programs? if yes, can we just test that end-to-end?

ramos-99:
*[`test/integration-test/bpf/tc_legacy_map.bpf.c` line 7 at r6](https://reviewable.io/reviews/aya-rs/aya/1497#-Oxum_d63U1j3yh1NjbU:-Oxv6D49Cb3Gi7BB3Doo:bdfgnd0) ([raw file](https://github.com/aya-rs/aya/blob/855d2745973763a0fb7379f6cccd6741611385fc/test/integration-test/bpf/tc_legacy_map.bpf.c#L7)):*

these tests aren't very convincing, they are just about tautological. is the intention to allow aya to load iproute2-flavored programs? if yes, can we just test that end-to-end?
</blockquote></details>

dropped the tautological len check, replaced it with tc_legacy_map_run: loads a tc object, runs the classifier, reads the maps back. it'd still pass on main since plain tc maps already loaded, so it's just a guard that loading isn't broken. the actual fix (rejecting unsupported pinning/map-in-map) is covered by a unit test and legacy_map_in_map_fails_at_load, both of which fail on main.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1497)* status: 6 of 10 files reviewed, 5 unresolved discussions (waiting on alessandrod and ramos-99).

___
*[`test/integration-test/bpf/tc_legacy_map.bpf.c` line 7 at r6](https://reviewable.io/reviews/aya-rs/aya/1497#-Oxum_d63U1j3yh1NjbU:-OxvBgpC8VwkboWCfnI4:b-r77kh2) ([raw file](https://github.com/aya-rs/aya/blob/855d2745973763a0fb7379f6cccd6741611385fc/test/integration-test/bpf/tc_legacy_map.bpf.c#L7)):*
<details><summary><i>Previously, ramos-99 (Martim Ramos) wrote…</i></summary><blockquote>

dropped the tautological len check, replaced it with tc_legacy_map_run: loads a tc object, runs the classifier, reads the maps back. it'd still pass on main since plain tc maps already loaded, so it's just a guard that loading isn't broken. the actual fix (rejecting unsupported pinning/map-in-map) is covered by a unit test and legacy_map_in_map_fails_at_load, both of which fail on main.
</blockquote></details>

It's still tautological, right? Both the test and non-test changes are citing iproute's data structure but not actually using that toolchain which includes various macros that may affect behavior. Can we do what we do with libbpf and test against a real thing produced by iproute2/tc?

ramos-99:
*[`test/integration-test/bpf/tc_legacy_map.bpf.c` line 7 at r6](https://reviewable.io/reviews/aya-rs/aya/1497#-Oxum_d63U1j3yh1NjbU:-Oxzsoqi-3OmJ9JkNs_d:b-kjiu0q) ([raw file](https://github.com/aya-rs/aya/blob/855d2745973763a0fb7379f6cccd6741611385fc/test/integration-test/bpf/tc_legacy_map.bpf.c#L7)):*

It's still tautological, right? Both the test and non-test changes are citing iproute's data structure but not actually using that toolchain which includes various macros that may affect behavior. Can we do what we do with libbpf and test against a real thing produced by iproute2/tc?
</blockquote></details>

makes sense. i looked at building it like the libbpf fixtures, but iproute2's `bpf_api.h` needs glibc's `stdint.h` and the kernel uapi headers, while our clang setup uses `-nostdlibinc` + `vmlinux.h` and excludes those. so it needs its own compile path (system headers, ci deps, bazel), which is a lot for one fixture.

instead, what if i commit a `.o` built by real iproute2/tc, with a script pinning the exact version and command? real artifact, all their macros, test just loads it. integration tests only run little-endian so one prebuilt object is enough. wdyt?

ramos-99:
CI red seems unrelated, both failures are dependency downloads failing.

@tamird thanks for the reviews. UB and citations should be addressed now. The one thing I'd like your call on is the fixture, since I'd rather not build the wrong shape twice.

On the tautological point, from what I can tell, compiling the fixture against the real bpf_elf.h and against the hand-written struct gives byte-identical maps/classifier/license sections. I think your point still holds though, since that only shows the transcription is right today, not that it stays right, so I'm happy to use the real header regardless.

Where I got stuck is using iproute2's own examples. As far as I can tell examples/bpf/legacy/*.c include bpf_api.h, which pulls <linux/pkt_cls.h>, <linux/bpf.h>, <linux/filter.h> and <asm/byteorder.h>. Our BPF compile runs -nostdlibinc with only libbpf's headers and vmlinux.h, and @kernel_headers doesn't look reachable from none_bpfel since @platforms//os:none maps to an empty arg set.

bpf_elf.h on its own looks much cheaper. To my knowledge it only needs <asm/types.h>, and vmlinux.h already provides __u32. Options I can see:

1. iproute2 as a submodule like xtask/libbpf, closest to what you suggested, though it needs Bazel wiring since there's no registry module for it.
2. Verbatim copy of bpf_elf.h plus a stub asm/types.h, needs one -I in build.rs and bazel/bpf.bzl.
3. Copy with the asm/types.h include dropped and documented, no build changes.

Whichever you prefer and I'll do it. For what it's worth, bpf_elf.h looks unchanged from v6.10.0 through v7.2.0.

On the rest, pinning is Option<NonZeroU32> as you suggested. The old check read map_def.pinning after ptr::read_unaligned had already reinterpreted the bytes, so PIN_GLOBAL_NS was UB before the check ran. inner_id/inner_idx use the same type for consistency. Added and fixed citations/docs as well.

tamird:
Let’s use option 1: add iproute2 as a submodule and compile the fixture against its upstream headers. The Cargo and Bazel wiring is worthwhile to keep the test anchored to the actual format we’re supporting. Keep the upstream headers unmodified.

ramos-99:
Thanks for the direction. Went with option 1. Two decisions I wasn't sure about are flagged in the description; happy to change either.

I also built tc from the submodule at v7.2.0 with `--libbpf_force off`, so it uses tc's own loader rather than libbpf, and it loads the fixture and creates both maps.

Split into two commits for easier reviewing; happy to squash.

PR #1513, opened 2026-04-01
fix unit test faiulre due to path/string mismatch

KirillLykov:
`cargo test` currently fails with errors like the following one:

```
error[E0277]: can't compare `std::path::Path` with `str`
    --> aya/src/programs/uprobe.rs:1160:17
     |
1160 |                 assert_eq!(path, "/usr/lib64/libcrypto.so.3.0.9", "path: {}", path.display());
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `std::path::Path == str`
     |
     = help: the trait `PartialEq<str>` is not implemented for `std::path::Path`
     = help: the following other types implement trait `PartialEq<Rhs>`:
               `&std::path::Path` implements `PartialEq<Cow<'_, OsStr>>`
               `&std::path::Path` implements `PartialEq<Cow<'_, std::path::Path>>`
               `&std::path::Path` implements `PartialEq<OsStr>`
               `&std::path::Path` implements `PartialEq<OsString>`
               `&std::path::Path` implements `PartialEq<PathBuf>`
               `std::path::Path` implements `PartialEq<&OsStr>`
               `std::path::Path` implements `PartialEq<Cow<'_, OsStr>>`
               `std::path::Path` implements `PartialEq<Cow<'_, std::path::Path>>`
             and 4 others
     = note: required for `&std::path::Path` to implement `PartialEq<&str>`
     = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1513)

### DIFF
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -923,7 +923,7 @@ mod tests {
         assert_matches!(
             find_debug_path_in_object(&main_obj, main_path, "symbol"),
             Ok(path) => {
-                assert_eq!(&*path, "/path/to/main.debug", "path: {}", path.display());
+                assert_eq!(path, Path::new("/path/to/main.debug"), "path: {}", path.display());
             }
         );
     }
@@ -1140,7 +1140,7 @@ mod tests {
         assert_matches!(
             proc_map_libs.find_library_path_by_name(Path::new("libcrypto.so.3.0.9")),
             Ok(Some(path)) => {
-                assert_eq!(path, "/usr/lib64/libcrypto.so.3.0.9", "path: {}", path.display());
+                assert_eq!(path, Path::new("/usr/lib64/libcrypto.so.3.0.9"), "path: {}", path.display());
             }
         );
     }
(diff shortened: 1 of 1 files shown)


tamird:
They pass, you just need Rust 1.91.0 or later.

https://doc.rust-lang.org/stable/std/path/struct.Path.html#impl-PartialEq%3Cstr%3E-for-Path

KirillLykov:
> They pass, you just need Rust 1.91.0 or later.
> 
> https://doc.rust-lang.org/stable/std/path/struct.Path.html#impl-PartialEq%3Cstr%3E-for-Path

I expected that there is `rust-toolchain.toml`, is there any reason why aya doesn't have it or I'm free to add?

tamird:
I think most of us use nightly because ebpf is still tier 3.

vadorovsky:
> I expected that there is `rust-toolchain.toml`, is there any reason why aya doesn't have it or I'm free to add?

I think it would make sense to add it. The open problem you're welcome to explore and find solution for is that:

* Most users use stable toolchain for the user-space part. For this case, we would like to always stick to the newest stable (so we have the newest features, lints etc.).
* eBPF part requires nightly.
* FWIK there is no way to define different toolchains per-crate in `rust-toolchain.toml`.
  * But from what I understand, it wouldn't take any effect if you run `cargo build -p aya-ebpf` from the main directory, because cargo starts the look up from the current directory.

I guess it's fine to specify just the newest stable, since the eBPF crates are being built with aya-build anyway. But you're welcome to find some better solutions.

There are plenty of users using stable toolchains for the user-space part (including Anza :slightly_smiling_face: ), so I would be against specifying nightly globally, to not break them by accident (by using some nightly-only features).

PR #1516, opened 2026-04-02
Fix: panic in from_pin().attach() for cgroup program types

hieuit095:
was debugging a user's report about a panic when re-attaching a pinned cgroup program. the issue is that rom_pin doesnt set expected_attach_type on the program data, so calling ttach() later hits an unwrap on None. fixed by setting it in the three affected rom_pin methods (cgroup_sock, cgroup_sockopt, cgroup_sock_addr).

note: tests cant run locally on windows due to aya-obj using unix-only std::os::fd APIs. this is a pre-existing platform constraint, not related to this change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1516)

### DIFF
--- a/aya/src/programs/cgroup_sock.rs
+++ b/aya/src/programs/cgroup_sock.rs
@@ -115,7 +115,8 @@ impl CgroupSock {
         path: P,
         attach_type: CgroupSockAttachType,
     ) -> Result<Self, ProgramError> {
-        let data = ProgramData::from_pinned_path(path, VerifierLogLevel::default())?;
+        let mut data = ProgramData::from_pinned_path(path, VerifierLogLevel::default())?;
+        data.expected_attach_type = Some(attach_type.into());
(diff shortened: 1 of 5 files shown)


tamird (changes requested):
Please write an integration test.

hieuit095:
added integration test.

tamird:
Thanks. I opened https://github.com/aya-rs/aya/pull/1517 with an alternative fix that removes the `expected_attach_type` field altogether.

hieuit095:
Added an integration test (cgroup_sock_addr_from_pin_attach) that verifies CgroupSockAddr::from_pin() correctly sets expected_attach_type before ttach() is called. The test pins a cgroup_sock_addr program, reloads it via rom_pin(), and attaches it to a cgroup — without the fix this would panic at the unwrap() of expected_attach_type inside ttach().

PR #1522, opened 2026-04-04
aya-ebpf: fix bpf_printk! passing pointers instead of values

### LINKED ISSUE #1171: bpf_printk! prints garbage
According to a report on discord:

```
let test = 42i32;
bpf_printk!(b"hi there! dec: %d",test);
```

Outputs:

```
bpf_trace_printk: hi there! dec: 1228242320
```

Seems to me like this may work with integer literals, but that it's not properly dereferencing variables.

one-d-wide:
C doesn't have an array value same way Rust does, therefore [u8; 8] passed as an argument to variadic bpf_trace_printk() is actually received as &[u8; 8], resulting in garbage in tracing output.

Fixes https://github.com/aya-rs/aya/issues/1171

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1522)

### DIFF
--- a/ebpf/aya-ebpf/src/helpers.rs
+++ b/ebpf/aya-ebpf/src/helpers.rs
@@ -702,13 +702,13 @@ pub use bpf_printk;
 /// Argument ready to be passed to `printk` BPF helper.
 #[repr(transparent)]
 #[derive(Copy, Clone)]
-pub struct PrintkArg([u8; 8]);
+pub struct PrintkArg(u64);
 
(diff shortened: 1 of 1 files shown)


tamird:
Dupe of https://github.com/aya-rs/aya/pull/1451?

one-d-wide:
Oops, forgot to check other PRs. My diff looks a bit cleaner IMHO.

tamird:
Thanks, merged #1451.

Issue #1185, opened 2025-02-27
Disable Public API checks in CI


dave-tucker:
The original intent behind adding the `public-api` checks was to give us an automated way of flagging things that required an explicit ACK by @alessandrod.

However, I feel like:

1. It's too fragile - it's broken multiple times in the last few months, which causes havok
1. It makes it harder to contribute to the project
1. There is too little signal (real API changes) in the noise

My feeling is that we'd be better off removing the checks in CI for pull-requests.

There is an argument to be made that checking the public API this way is still useful for determining the version bump used when doing a release, but I'm not convinced the value it adds at the PR level is worth the hassle.

@vadorovsky @tamird thoughts?

alessandrod:
Uh no? Many times we’ve almost accidentally broken API. I don’t like that
_running_ the public-api command to bless changes requires a specific
version of nightly and to run on x86, but other than that we should
definitely have API breakage checks

tamird:
I think it's fine :/ we now automatically create PRs to regenerate so it's not that big of a deal.

We could implement automation to push API updates to people's PRs, some other projects do that to e.g. format code. We'd probably have to start squashing on merge.

vadorovsky:
I'm also for keeping it public-api

dave-tucker:
> Uh no? Many times we’ve almost accidentally broken API. I don’t like that
> _running_ the public-api command to bless changes requires a specific
> version of nightly and to run on x86, but other than that we should
> definitely have API breakage checks

I'm not against having checks, I'm just losing faith that `public-api` is the right way to do it.
Maybe I should rename the issue to `Keep public API checks but do them differently`.

The only alternative I can see would be to write `examples/*` that exercise every public-api that we have.
examples fail to compile? you broke an API
adding a new API? not without an example

We have some of these already as integration tests, but might be better suited to being example applications anyway.

Obviously the challenge there is getting full coverage, which public-api gives us for free.

tamird:
I do think we should be eyeing ways to reduce contributor and maintenance burden, so thanks for creating this discussion!

vadorovsky:
> I do think we should be eyeing ways to reduce contributor and maintenance burden, so thanks for creating this discussion!

Same. I also think there are problems with public-api which have to be solved. Not being able to run it on ARM is the most annoying one. But I would still try to fix that without dropping the project all together.

>The only alternative I can see would be to write `examples/*` that exercise every public-api that we have.
examples fail to compile? you broke an API

[...]

PR #1529, opened 2026-04-12
public-api: omit blanket impls

swananan:
### Context

While working on PR #1499 and #1417, I hit a [CI failure on the public-api check](https://github.com/aya-rs/aya/actions/runs/24302798438/job/70958997032?pr=1499) that I couldn't reproduce locally — even after matching the exact same nightly version as CI (`rustc 1.97.0-nightly (2026-04-11)`).

After some investigation, I traced the root cause to `Cargo.lock`: since it's gitignored (correctly, for a library), CI resolves dependencies from scratch and may pick up different transitive dependency versions than my local environment. In this case, a different `hashbrown` version introduced different `Equivalent` blanket impls, causing the diff.

This got me thinking about whether we can avoid this class of failure entirely. I also found #1185 where the friction of maintaining public-api snapshots was discussed.

### Approach

Blanket impls (`From<T>`, `Into`, etc. and third-party ones like `hashbrown::Equivalent`) **should not be part of Aya's API contract**, correct me if I understand it wrongly.

This PR adds `.omit_blanket_impls(true)` to the `public-api` builder to filter them out.

### Impact & Trade-off

1. Open PRs that include `xtask/public-api/*.txt` files will need to rebase and re-bless.
2. Aya currently has **no public blanket impls** of its own, but If Aya introduces a public blanket impl in the future, it would be filtered out and not caught by the public-api check.

### Added/updated tests?

- [ ] Yes
- [x] No, and this is why:  this change only include public-api

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1529)

### DIFF
--- a/xtask/public-api/aya-build.txt
+++ b/xtask/public-api/aya-build.txt
@@ -12,22 +12,6 @@ impl<'a> core::marker::Unpin for aya_build::Toolchain<'a>
 impl<'a> core::marker::UnsafeUnpin for aya_build::Toolchain<'a>
 impl<'a> core::panic::unwind_safe::RefUnwindSafe for aya_build::Toolchain<'a>
 impl<'a> core::panic::unwind_safe::UnwindSafe for aya_build::Toolchain<'a>
-impl<T, U> core::convert::Into<U> for aya_build::Toolchain<'a> where U: core::convert::From<T>
-pub fn aya_build::Toolchain<'a>::into(self) -> U
-impl<T, U> core::convert::TryFrom<U> for aya_build::Toolchain<'a> where U: core::convert::Into<T>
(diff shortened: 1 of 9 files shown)


tamird (approved):
This looks good to me.  Do you want to preserve any of the PR description detail in the commit message?

> @tamird reviewed 9 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1529#-:-Oq17JFD530gplAi7mLR:bxztaic)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on swananan).

swananan:
> This looks good to me. Do you want to preserve any of the PR description detail in the commit message?

No, I think the commit message covers it, thanks!

PR #1218, opened 2025-03-15
feat: Added bounds check for `SkBuff::load_bytes` and relevant integration test.

RyanAlameddine:
Alternative to https://github.com/aya-rs/aya/pull/1187, attempting to resolve the `invalid zero-sized read` error found on discord and in my personal projects. 

This PR contains: 

The correct inclusive bounds check.

An integration test which fails before the update and succeeds after.

PR #1539, opened 2026-04-19
feat: Added bounds check for `SkBuff::load_bytes` and relevant integration test.

benaryorg:
This is a reboot of #1218, with some minor commit juggling (i.e. moving changes from one commit to the other to make them a little more atomic).

> […], attempting to resolve the invalid zero-sized read error found on discord and in my personal projects.
> 
> This PR contains:
> 
> The correct inclusive bounds check.
> 
> An integration test which fails before the update and succeeds after.

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- [ ] I have blessed any API changes with `cargo xtask public-api --bless`.

(the last two err out on my machine due to problems in the development environment, not the changes themselves)

Unit test failures:

- `maps::tests::test_create_perf_event_array` because `/sys/devices/system/cpu/possible` is not available in the sandbox I run tests in
- `maps::hash_map::per_cpu_hash_map::tests::test_get_not_found`: probably the same thing:

```text
assertion failed: `Err(SyscallError(SyscallError { call: "bpf_map_lookup_elem", io_error: Os { code: 2, kind: NotFound, message: "No such file or directory" } }))` does not match `Err(MapError::KeyNotFound)`
```

Me trying to run the tests without setting up the full development environment.
Also: me running the tests against the wrong branch.

bonk

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1539)

### DIFF
--- a/ebpf/aya-ebpf/src/programs/sk_buff.rs
+++ b/ebpf/aya-ebpf/src/programs/sk_buff.rs
@@ -7,7 +7,7 @@ use aya_ebpf_bindings::helpers::{
 };
 use aya_ebpf_cty::c_long;
 
-use crate::{EbpfContext, bindings::__sk_buff};
+use crate::{EbpfContext, bindings::__sk_buff, check_bounds_signed};
 
 pub struct SkBuff {
     pub skb: *mut __sk_buff,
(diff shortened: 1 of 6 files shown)


benaryorg:
Apologies for that.… turns out having two shells on two different machines in the same repo is not the best idea.
The original test did end up tripping *clippy* due to panic handling (and some other stuff), so I brought it up to speed based on another test.

Since I usually develop in sandboxed environments (forgive me, but I really don't trust other people's build scripts to not write where they shouldn't) a few things on that end failed.
However, `unshare --user --net --mount --map-root-user --keep-caps` is a pretty suitable replacement for `sudo` in case of the unit tests, so those all pass now (leaving the OP untouched in that regard since I'm not 100% certain they operate as they should).
The minimum access required (beyond bubblewrap defaults) seems to be `/sys/devices/system/cpu/possible` and `/sys/kernel/btf/vmlinux`.
The integration tests on the other hand are less happy about their lack of privileges and continue to fail, so I'm relying on the CI here.

tamird:
Looks like this has merge conflicts.

benaryorg:
Unfortunate timing with 8edfe49 getting merged in-between. Should be fixed now.

PR #1542, opened 2026-04-20
aya-ebpf: add BTF map definition for stack trace

Brskt:
BTF `.maps` support for `BPF_MAP_TYPE_STACK_TRACE`. Keys are `u32`
stack IDs, values are `[u64; DEPTH]` with `DEPTH` defaulting to
`PERF_MAX_STACK_DEPTH` (127). Compile-time assertions enforce
non-zero `DEPTH`/`max_entries` and reject `BPF_F_STACK_BUILD_ID`.
A preparatory commit exposes `PERF_MAX_STACK_DEPTH` from
`aya-ebpf-bindings` so neither BTF nor legacy `StackTrace` needs
to hardcode the value.

Another preparatory commit rejects `BPF_F_STACK_BUILD_ID` in
`aya::maps::StackTraceMap`. The flag switches the kernel element
layout to `struct bpf_stack_build_id` (32 bytes), but the wrapper
unconditionally decodes values as `[u64]` and silently corrupts
reads. Adds `MapError::UnsupportedMapFlags` and
`MapError::InvalidValueStride`, and drops the now-redundant
sysctl read.

The integration test is parameterized via `test_case` over both
the BTF and legacy map definitions. Each case attaches a uprobe
that records a stack id via `get_stackid` and verifies the
resulting trace has at least one non-zero IP frame.

A final commit cleans up the lpm_trie test's `TestResult::ran`
from `u32` to `bool`.

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -144,6 +144,15 @@ pub enum MapError {
         expected: usize,
     },
 
+    /// Invalid value stride
+    #[error("invalid value size {size}, expected a non-zero multiple of {stride}")]
+    InvalidValueStride {
+        /// Size encountered
+        size: usize,
+        /// Required stride
+        stride: usize,
+    },
+
     /// Index is out of bounds
     #[error("the index is {index} but `max_entries` is {max_entries}")]
     OutOfBounds {
@@ -196,6 +205,15 @@ pub enum MapError {
         name: String,
         /// The map type
         map_type: bpf_map_type,
+    },
+
+    /// Unsupported map flags
+    #[error("unsupported map flags {flags:#x}: {reason}")]
+    UnsupportedMapFlags {
+        /// The map flags
+        flags: u32,
+        /// The reason
+        reason: &'static str,
     },
 }
 
--- a/aya/src/maps/stack_trace.rs
+++ b/aya/src/maps/stack_trace.rs
@@ -4,11 +4,10 @@
 
 use std::{
     borrow::{Borrow, BorrowMut},
-    fs, io,
     os::fd::AsFd as _,
-    path::Path,
-    str::FromStr,
 };
+
+use aya_obj::generated::BPF_F_STACK_BUILD_ID;
 
 use crate::{
     maps::{IterableMap, MapData, MapError, MapIter, MapKeys, hash_map},
@@ -78,24 +77,43 @@ pub struct StackTraceMap<T> {
     max_stack_depth: usize,
 }
 
+// A stack trace entry is a single `u64` instruction pointer, matching the
+// kernel layout for `BPF_MAP_TYPE_STACK_TRACE` values.
+type StackEntry = u64;
+
 impl<T: Borrow<MapData>> StackTraceMap<T> {
     pub(crate) fn new(map: T) -> Result<Self, MapError> {
         let data = map.borrow();
-        let expected = size_of::<u32>();
-        let size = data.obj.key_size() as usize;
-        if size != expected {
-            return Err(MapError::InvalidKeySize { size, expected });
+
+        let key_size = data.obj.key_size() as usize;
+        let expected_key = size_of::<u32>();
+        if key_size != expected_key {
+            return Err(MapError::InvalidKeySize {
+                size: key_size,
+                expected: expected_key,
+            });
         }
 
-        let max_stack_depth =
-            sysctl::<usize>("kernel/perf_event_max_stack").map_err(|io_error| SyscallError {
-                call: "sysctl",
-                io_error,
-            })?;
-        let size = data.obj.value_size() as usize;
-        if size > max_stack_depth * size_of::<u64>() {
-            return Err(MapError::InvalidValueSize { size, expected });
+        // BPF_F_STACK_BUILD_ID switches stack entries to
+        // `struct bpf_stack_build_id` (32 bytes), which `get` decodes as
+        // `[u64]` and would silently corrupt. Reject it.
+        let flags = data.obj.map_flags();
+        if flags & BPF_F_STACK_BUILD_ID != 0 {
+            return Err(MapError::UnsupportedMapFlags {
+                flags,
+                reason: "StackTraceMap does not support bpf_stack_build_id entries",
+            });
         }
+
+        let value_size = data.obj.value_size() as usize;
+        let expected_stride = size_of::<StackEntry>();
+        if value_size == 0 || !value_size.is_multiple_of(expected_stride) {
+            return Err(MapError::InvalidValueStride {
+                size: value_size,
+                stride: expected_stride,
+            });
+        }
+        let max_stack_depth = value_size / expected_stride;
 
         Ok(Self {
             inner: map,
@@ -112,7 +130,7 @@ impl<T: Borrow<MapData>> StackTraceMap<T> {
     pub fn get(&self, stack_id: &u32, flags: u64) -> Result<StackTrace, MapError> {
         let fd = self.inner.borrow().fd().as_fd();
 
-        let mut frames = vec![0; self.max_stack_depth];
+        let mut frames: Vec<StackEntry> = vec![0; self.max_stack_depth];
         bpf_map_lookup_elem_ptr(fd, Some(stack_id), frames.as_mut_ptr(), flags)
             .map_err(|io_error| SyscallError {
                 call: "bpf_map_lookup_elem",
@@ -191,11 +209,4 @@ impl StackTrace {
 pub struct StackFrame {
     /// The instruction pointer of this frame.
     pub ip: u64,
-}
-
-fn sysctl<T: FromStr>(key: &str) -> Result<T, io::Error> {
-    let val = fs::read_to_string(Path::new("/proc/sys").join(key))?;
-    val.trim()
-        .parse::<T>()
-        .map_err(|_err: T::Err| io::Error::new(io::ErrorKind::InvalidData, val))
 }
--- a/ebpf/aya-ebpf-bindings/include/bindings.h
+++ b/ebpf/aya-ebpf-bindings/include/bindings.h
@@ -7,6 +7,8 @@ typedef __u32 __bitwise __wsum;
 #include <bpf/bpf_helpers.h>
 #include <linux/bpf.h>
 #include <linux/bpf_perf_event.h>
+// needed for PERF_MAX_STACK_DEPTH
+#include <linux/perf_event.h>
 // needed for TC_ACT_*
 #include <linux/pkt_cls.h>
 #include <linux/ptrace.h>
--- a/ebpf/aya-ebpf-bindings/src/aarch64/bindings.rs
+++ b/ebpf/aya-ebpf-bindings/src/aarch64/bindings.rs
@@ -255,6 +255,7 @@ pub const BPF_F_TEST_XDP_LIVE_FRAMES: u32 = 2;
 pub const BPF_BUILD_ID_SIZE: u32 = 20;
 pub const BPF_OBJ_NAME_LEN: u32 = 16;
 pub const BPF_TAG_SIZE: u32 = 8;
+pub const PERF_MAX_STACK_DEPTH: u32 = 127;
 pub const TC_ACT_UNSPEC: i32 = -1;
 pub const TC_ACT_OK: u32 = 0;
 pub const TC_ACT_RECLASSIFY: u32 = 1;
--- a/ebpf/aya-ebpf-bindings/src/armv7/bindings.rs
+++ b/ebpf/aya-ebpf-bindings/src/armv7/bindings.rs
@@ -255,6 +255,7 @@ pub const BPF_F_TEST_XDP_LIVE_FRAMES: u32 = 2;
 pub const BPF_BUILD_ID_SIZE: u32 = 20;
 pub const BPF_OBJ_NAME_LEN: u32 = 16;
 pub const BPF_TAG_SIZE: u32 = 8;
+pub const PERF_MAX_STACK_DEPTH: u32 = 127;
 pub const TC_ACT_UNSPEC: i32 = -1;
 pub const TC_ACT_OK: u32 = 0;
 pub const TC_ACT_RECLASSIFY: u32 = 1;
--- a/ebpf/aya-ebpf-bindings/src/loongarch64/bindings.rs
+++ b/ebpf/aya-ebpf-bindings/src/loongarch64/bindings.rs
@@ -255,6 +255,7 @@ pub const BPF_F_TEST_XDP_LIVE_FRAMES: u32 = 2;
 pub const BPF_BUILD_ID_SIZE: u32 = 20;
 pub const BPF_OBJ_NAME_LEN: u32 = 16;
 pub const BPF_TAG_SIZE: u32 = 8;
+pub const PERF_MAX_STACK_DEPTH: u32 = 127;
 pub const TC_ACT_UNSPEC: i32 = -1;
 pub const TC_ACT_OK: u32 = 0;
 pub const TC_ACT_RECLASSIFY: u32 = 1;
--- a/ebpf/aya-ebpf-bindings/src/mips/bindings.rs
+++ b/ebpf/aya-ebpf-bindings/src/mips/bindings.rs
@@ -255,6 +255,7 @@ pub const BPF_F_TEST_XDP_LIVE_FRAMES: u32 = 2;
 pub const BPF_BUILD_ID_SIZE: u32 = 20;
 pub const BPF_OBJ_NAME_LEN: u32 = 16;
 pub const BPF_TAG_SIZE: u32 = 8;
+pub const PERF_MAX_STACK_DEPTH: u32 = 127;
 pub const TC_ACT_UNSPEC: i32 = -1;
 pub const TC_ACT_OK: u32 = 0;
 pub const TC_ACT_RECLASSIFY: u32 = 1;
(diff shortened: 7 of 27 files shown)


tamird (changes requested):
> @tamird reviewed 12 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: all files reviewed, 7 unresolved discussions (waiting on Brskt).

___
*[`-- commits` line 22 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfd8fdCa-6YfTQ-rdY:-Oqfd8fdCa-6YfTQ-rdZ:b9yng26):*
It might be neat to support this in the future, seems useful.
___
*[`aya/src/maps/stack_trace.rs` line 105 at r1](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfb__l1KF_wKd61Ucl:-Oqfb__l1KF_wKd61Ucm:b-yug11j) ([raw file](https://github.com/aya-rs/aya/blob/bc0c0b4014b10994d6b6fb8fb589e9e9f06cd5ac/aya/src/maps/stack_trace.rs#L105)):*
> ```XML
> 
>         let value_size = data.obj.value_size() as usize;
>         let expected_stride = size_of::<u64>();
> ```

this is coupled with something in `get`, right? it would be good extract a common constant or otherwise make the coupling more apparent.
___
*[`aya/src/maps/stack_trace.rs` line 107 at r1](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfcBep9T6q-IiXwZs_:-OqfcBep9T6q-IiXwZsa:bnmx5eq) ([raw file](https://github.com/aya-rs/aya/blob/bc0c0b4014b10994d6b6fb8fb589e9e9f06cd5ac/aya/src/maps/stack_trace.rs#L107)):*
> ```XML
>         let expected_stride = size_of::<u64>();
>         if value_size == 0 || !value_size.is_multiple_of(expected_stride) {
>             return Err(MapError::InvalidValueSize {
> ```

should we clarify this error? it will currently produce "invalid value size {size}, expected {expected}" which is pretty odd here.
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfdn1L0GFO8G03D_cG:-Oqfdn1L0GFO8G03D_cH:b3hw10z) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L7)):*
> ```XML
> // Default maximum stack-trace depth enforced by the kernel (UAPI
> // `PERF_MAX_STACK_DEPTH` in <linux/perf_event.h>).
> const PERF_MAX_STACK_DEPTH: usize = 127;
> ```

we actually have this constant in the generated bindings already, let's use it from there (both in the BTF map and the legacy map)
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 58 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfe49uCkk72zjwUHLH:-Oqfe49uCkk72zjwUHLI:b-uto0js) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L58)):*
> ```XML
>     // in each public method.
>     const _CHECK: () = {
>         assert!(DEPTH >= 1, "stack trace DEPTH must be at least 1.");
> ```

why not write `> 0`?
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 75 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfeH6r4rJGhfGmaNeI:-OqfeH6r4rJGhfGmaNeJ:btqkfdo) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L75)):*
> ```XML
>     /// `bpf_get_stackid` is only available to tracing program types
>     /// (kprobe, tracepoint, `perf_event`, `raw_tracepoint`). Calling
>     /// it from any other program type fails verification.
> ```

i wonder if a better API would put this method on those programs' context types rather than publicly on the map. WDYT?



_Code quote:_
```XML
    /// `bpf_get_stackid` is only available to tracing program types
    /// (kprobe, tracepoint, `perf_event`, `raw_tracepoint`). Calling
    /// it from any other program type fails verification.
```
___
*[`test/integration-ebpf/src/stack_trace.rs` line 22 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfdX_r-imGGjXbgLPI:-OqfdX_r-imGGjXbgLPJ:b-o3ko9y) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/test/integration-ebpf/src/stack_trace.rs#L22)):*
> ```XML
> 
> #[btf_map]
> static RESULT: BtfArray<u32, { NUM_SLOTS as usize }, 0> = BtfArray::new();
> ```

is it possible for this value type to be `bool`?

Brskt (commented):
> @Brskt made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: 0 of 12 files reviewed, 7 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/stack_trace.rs` line 105 at r1](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfb__l1KF_wKd61Ucl:-OqgOt4E8fmmtD1ftGWC:burvl5z) ([raw file](https://github.com/aya-rs/aya/blob/bc0c0b4014b10994d6b6fb8fb589e9e9f06cd5ac/aya/src/maps/stack_trace.rs#L105)):*

this is coupled with something in `get`, right? it would be good extract a common constant or otherwise make the coupling more apparent.
</blockquote></details>

Done with a private type alias.
___
*[`aya/src/maps/stack_trace.rs` line 107 at r1](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfcBep9T6q-IiXwZs_:-OqgiaQ_9RMgunJ7e6gI:borl1q4) ([raw file](https://github.com/aya-rs/aya/blob/bc0c0b4014b10994d6b6fb8fb589e9e9f06cd5ac/aya/src/maps/stack_trace.rs#L107)):*

should we clarify this error? it will currently produce "invalid value size {size}, expected {expected}" which is pretty odd here.
</blockquote></details>

Done via a new InvalidValueLayout { size, reason } variant.
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfdn1L0GFO8G03D_cG:-OqgllQkDMektSpdQXzf:bdqaaa5) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L7)):*

we actually have this constant in the generated bindings already, let's use it from there (both in the BTF map and the legacy map)
</blockquote></details>

it's only in `aya_obj::generated`, not `aya-ebpf-bindings`. follow-up PR to expose it, or here?
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 58 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfe49uCkk72zjwUHLH:-Oqgm-bw6BjUJsTG2yeG:b9izszt) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L58)):*

why not write `> 0`?
</blockquote></details>

Yup, done.
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 75 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfeH6r4rJGhfGmaNeI:-OqgoN8iA6PM2ZMABzqF:bpbyf7) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L75)):*

i wonder if a better API would put this method on those programs' context types rather than publicly on the map. WDYT?
</blockquote></details>

The `unsafe` here isn't UB, just a load-time verifier failure. Context-method would move the error to compile-time, no real safety gain.

Bigger question: `PerfEventArray::output`/`RingBuf::output_bytes` use the same `&C: EbpfContext` shape. Are you thinking a sweep across all tracing helpers, or just `StackTrace`?
___
*[`test/integration-ebpf/src/stack_trace.rs` line 22 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfdX_r-imGGjXbgLPI:-OqjYXgmDtYwi0ZU7k_K:b9bz3k) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/test/integration-ebpf/src/stack_trace.rs#L22)):*

is it possible for this value type to be `bool`?
</blockquote></details>

No, the other slot stores the `u32` stack id. Bool only fits if we split it into two maps.

tamird (changes requested):
> @tamird reviewed 12 files and all commit messages, made 4 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: all files reviewed, 4 unresolved discussions (waiting on Brskt).

___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfdn1L0GFO8G03D_cG:-Oqk-lRwAYpDt8hArwyJ:b-ba0gz4) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L7)):*
<details><summary><i>Previously, Brskt wrote…</i></summary><blockquote>

it's only in `aya_obj::generated`, not `aya-ebpf-bindings`. follow-up PR to expose it, or here?
</blockquote></details>

i think you can just do it here, as the first commit maybe?
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 75 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfeH6r4rJGhfGmaNeI:-Oqk-vj5D0I7O-Hv2ETL:b-3wnq3i) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L75)):*
<details><summary><i>Previously, Brskt wrote…</i></summary><blockquote>

The `unsafe` here isn't UB, just a load-time verifier failure. Context-method would move the error to compile-time, no real safety gain.

Bigger question: `PerfEventArray::output`/`RingBuf::output_bytes` use the same `&C: EbpfContext` shape. Are you thinking a sweep across all tracing helpers, or just `StackTrace`?
</blockquote></details>

Maybe across all tracing helpers, if we're very confident we can enumerate the support matrix accurately.

Compile-time is a huge win over verifier failure especially as agents do more and more of this work; the earlier that they can find out that an approach doesn't work the faster the feedback loop will be.

I suppose this can be a follow-up, if you're willing to do the sweep across all maps.
___
*[`test/integration-ebpf/src/stack_trace.rs` line 22 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfdX_r-imGGjXbgLPI:-Oqk-U6_EXXIK2mt1ARI:b-rgakyv) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/test/integration-ebpf/src/stack_trace.rs#L22)):*
<details><summary><i>Previously, Brskt wrote…</i></summary><blockquote>

No, the other slot stores the `u32` stack id. Bool only fits if we split it into two maps.
</blockquote></details>

ah, i see. why do we need two slots? if you want to guard against the case where stackid is somehow 0 you can put a struct in the map with both items. wdyt?
___
*[`aya/src/maps/stack_trace.rs` line 113 at r3](https://reviewable.io/reviews/aya-rs/aya/1542#-OqjzwIF7oMIhPKnZWhE:-OqjzwIG-6x-2tEAGiSE:bbqji3p) ([raw file](https://github.com/aya-rs/aya/blob/8ab4cdb63558cee699c3ce3f2c85a1789c2918dd/aya/src/maps/stack_trace.rs#L113)):*
> ```XML
>             return Err(MapError::InvalidValueLayout {
>                 size: value_size,
>                 reason: "expected a non-zero multiple of 8",
> ```

bummer to repeat in prose something encoded in code. maybe craft this message in the error type and put the stride in its data

Brskt (commented):
> @Brskt made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: all files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/stack_trace.rs` line 113 at r3](https://reviewable.io/reviews/aya-rs/aya/1542#-OqjzwIF7oMIhPKnZWhE:-Oqkg95R1XWF3NUfGAS0:b-3ffgtr) ([raw file](https://github.com/aya-rs/aya/blob/8ab4cdb63558cee699c3ce3f2c85a1789c2918dd/aya/src/maps/stack_trace.rs#L113)):*

bummer to repeat in prose something encoded in code. maybe craft this message in the error type and put the stride in its data
</blockquote></details>

Done. Renamed to `InvalidValueStride { size, stride }`
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-Oqfdn1L0GFO8G03D_cG:-Oqkfuh07kFYTCTt1Vt_:br2f38f) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L7)):*

i think you can just do it here, as the first commit maybe?
</blockquote></details>

Okay, done.
___
*[`ebpf/aya-ebpf/src/btf_maps/stack_trace.rs` line 75 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfeH6r4rJGhfGmaNeI:-OqlIBGU-8mZwWiQ507L:bw64y2d) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/ebpf/aya-ebpf/src/btf_maps/stack_trace.rs#L75)):*

Maybe across all tracing helpers, if we're very confident we can enumerate the support matrix accurately.

Compile-time is a huge win over verifier failure especially as agents do more and more of this work; the earlier that they can find out that an approach doesn't work the faster the feedback loop will be.

I suppose this can be a follow-up, if you're willing to do the sweep across all maps.
</blockquote></details>

Agreed. I'll open a follow-up after this lands: sealed tracing-context trait covering `get_stackid` and the perf/ringbuf output helpers.
___
*[`test/integration-ebpf/src/stack_trace.rs` line 22 at r2](https://reviewable.io/reviews/aya-rs/aya/1542#-OqfdX_r-imGGjXbgLPI:-Oql5Tpy41SqeICk0KXl:by6y6kg) ([raw file](https://github.com/aya-rs/aya/blob/15bb3ccf66387a836985ea22039f57fcf1f2bc73/test/integration-ebpf/src/stack_trace.rs#L22)):*

ah, i see. why do we need two slots? if you want to guard against the case where stackid is somehow 0 you can put a struct in the map with both items. wdyt?
</blockquote></details>

Yeah, good idea. Switched to a `TestResult` struct with `stack_id: u32` and `ran: u32`, same as the lpm_trie test.

tamird (changes requested):
> @tamird reviewed 25 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: all files reviewed, 1 unresolved discussion (waiting on Brskt).

___
*[`test/integration-common/src/lib.rs` line 175 at r7](https://reviewable.io/reviews/aya-rs/aya/1542#-OqlaG7t-fikwSrtYAFL:-OqlaG7t-fikwSrtYAFM:b-g9ltb4) ([raw file](https://github.com/aya-rs/aya/blob/cdb409918543936eae77f59c10fe5f3e5e4246c0/test/integration-common/src/lib.rs#L175)):*
> ```XML
>     pub struct TestResult {
>         pub stack_id: u32,
>         pub ran: u32,
> ```

surely now this can be a bool :) a comment to explain that this is only here to distinguish from the zero value would be good

maybe do the same for the lpm trie test if it also needlessly uses u32

Brskt (commented):
> @Brskt made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: 16 of 27 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`test/integration-common/src/lib.rs` line 175 at r7](https://reviewable.io/reviews/aya-rs/aya/1542#-OqlaG7t-fikwSrtYAFL:-OqlmJ3e5cmmtNLAdiq9:boro3w2) ([raw file](https://github.com/aya-rs/aya/blob/cdb409918543936eae77f59c10fe5f3e5e4246c0/test/integration-common/src/lib.rs#L175)):*

surely now this can be a bool :) a comment to explain that this is only here to distinguish from the zero value would be good

maybe do the same for the lpm trie test if it also needlessly uses u32
</blockquote></details>

Aight, done.

tamird (approved):
> @tamird reviewed 11 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1542)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Brskt).

PR #1543, opened 2026-04-21
aya: fix pid=0 handling in uprobe attach

swananan:
### Summary

While working on multi-uprobe support (#1417), @alessandrod pointed out that `Some(0)` has inconsistent semantics between the single-attach (`perf_event_open`: pid=0 → current process) and multi-attach (`bpf_link_create`: pid=0 → all processes) paths.

Investigating this, I found that `Some(0)` never actually worked in aya's existing single-attach path either: `UProbe::attach` unconditionally reads `/proc/{pid}/maps` when pid is `Some`, and  `/proc/0/maps` does not exist (pid 0 is the kernel swapper process). The attach fails before even reaching the kernel.

This PR uses the real pid only for the ProcMap lookup, and keeps the original pid=0 for the actual attach call, matching libbpf's single-attach behavior which passes pid=0 directly to  `perf_event_open_probe`  ([ref](https://github.com/torvalds/linux/blob/8541d8f725/tools/lib/bpf/libbpf.c#L12845-L12847)).

And replace the pid Option in UProbe::attach with UProbeScope so the all-processes, calling-process, and explicit-pid cases are all expressed directly in the type.

This removes pid=0 as a magic value from the public API, keeps the calling-process case explicit, and uses the current pid for ProcMap resolution introduced in the previous commit. Update docs and integration tests to use the new scope-based API.

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -1099,11 +1099,11 @@ impl Ebpf {
     ///
     /// ```no_run
     /// # let mut bpf = aya::Ebpf::load(&[])?;
-    /// use aya::programs::UProbe;
+    /// use aya::programs::{uprobe::UProbeScope, UProbe};
     ///
     /// let program: &mut UProbe = bpf.program_mut("SSL_read").unwrap().try_into()?;
     /// program.load()?;
-    /// program.attach("SSL_read", "libssl", None)?;
+    /// program.attach("SSL_read", "libssl", UProbeScope::AllProcesses)?;
     /// # Ok::<(), aya::EbpfError>(())
     /// ```
     pub fn program_mut(&mut self, name: &str) -> Option<&mut Program> {
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -109,7 +109,8 @@ impl UProbe {
     /// Attaches the uprobe to the function `fn_name` defined in the `target`.
     /// If the attach point specifies an offset, it is added to the address of
     /// the target function. If `pid` is not `None`, the program executes only
-    /// when the target function is executed by the given `pid`.
+    /// when the target function is executed by the given `pid`; `Some(0)`
+    /// means the calling process/thread.
     ///
     /// The `target` argument can be an absolute path to a binary or library, or
     /// a library name (eg: `"libc"`).
@@ -131,7 +132,10 @@ impl UProbe {
         pid: Option<u32>,
     ) -> Result<UProbeLinkId, ProgramError> {
         let UProbeAttachPoint { location, cookie } = point.into();
-        let proc_map = pid.map(ProcMap::new).transpose()?;
+        // /proc/0/maps does not exist; use the current pid only for ProcMap
+        // resolution and preserve pid=0 for the actual attach semantics.
+        let proc_map_pid = pid.map(|pid| if pid == 0 { std::process::id() } else { pid });
+        let proc_map = proc_map_pid.map(ProcMap::new).transpose()?;
         let path = resolve_attach_path(target.as_ref(), proc_map.as_ref())?;
         let (symbol, offset) = match location {
             UProbeAttachLocation::Symbol(s) => (Some(s), 0),
--- a/aya/src/programs/uprobe.rs
+++ b/aya/src/programs/uprobe.rs
@@ -6,6 +6,7 @@ use std::{
     fmt::{self, Write},
     fs,
     io::{self, BufRead as _, Cursor, Read as _},
+    num::NonZeroU32,
     os::unix::ffi::{OsStrExt as _, OsStringExt as _},
     path::{Path, PathBuf},
     sync::LazyLock,
@@ -88,6 +89,17 @@ impl<'a, L: Into<UProbeAttachLocation<'a>>> From<L> for UProbeAttachPoint<'a> {
     }
 }
 
+/// Specifies which processes a uprobe should fire for.
+#[derive(Debug, Clone, Copy)]
+pub enum UProbeScope {
+    /// Fire for any process that hits the attach point.
+    AllProcesses,
+    /// Fire only when the calling process/thread hits the attach point.
+    CallingProcess,
+    /// Fire only when the given process hits the attach point.
+    OneProcess(NonZeroU32),
+}
+
 impl UProbe {
     /// The type of the program according to the kernel.
     pub const PROGRAM_TYPE: ProgramType = ProgramType::KProbe;
@@ -108,9 +120,8 @@ impl UProbe {
     ///
     /// Attaches the uprobe to the function `fn_name` defined in the `target`.
     /// If the attach point specifies an offset, it is added to the address of
-    /// the target function. If `pid` is not `None`, the program executes only
-    /// when the target function is executed by the given `pid`; `Some(0)`
-    /// means the calling process/thread.
+    /// the target function. `scope` specifies which processes should trigger
+    /// the uprobe.
     ///
     /// The `target` argument can be an absolute path to a binary or library, or
     /// a library name (eg: `"libc"`).
@@ -129,12 +140,19 @@ impl UProbe {
         &mut self,
         point: Point,
         target: T,
-        pid: Option<u32>,
+        scope: UProbeScope,
     ) -> Result<UProbeLinkId, ProgramError> {
         let UProbeAttachPoint { location, cookie } = point.into();
-        // /proc/0/maps does not exist; use the current pid only for ProcMap
-        // resolution and preserve pid=0 for the actual attach semantics.
-        let proc_map_pid = pid.map(|pid| if pid == 0 { std::process::id() } else { pid });
+        let (proc_map_pid, perf_event_pid) = match scope {
+            UProbeScope::AllProcesses => (None, None),
+            // /proc/0/maps does not exist, so use the real pid for ProcMap
+            // resolution while keeping the kernel's pid=0 sentinel for attach.
+            UProbeScope::CallingProcess => (Some(std::process::id()), Some(0)),
+            UProbeScope::OneProcess(pid) => {
+                let pid = pid.get();
+                (Some(pid), Some(pid))
+            }
+        };
         let proc_map = proc_map_pid.map(ProcMap::new).transpose()?;
         let path = resolve_attach_path(target.as_ref(), proc_map.as_ref())?;
         let (symbol, offset) = match location {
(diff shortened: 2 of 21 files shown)


tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/programs/uprobe.rs` line 111 at r1](https://reviewable.io/reviews/aya-rs/aya/1543#-Oql3o9y43mQoiIHYEih:-Oql3o9y43mQoiIHYEii:b-f1ro1m) ([raw file](https://github.com/aya-rs/aya/blob/344a3725270f8df5f35aafdbc59fddf992c176de/aya/src/programs/uprobe.rs#L111)):*
> ```XML
>     /// Attaches the uprobe to the function `fn_name` defined in the `target`.
>     /// If the attach point specifies an offset, it is added to the address of
>     /// the target function. If `pid` is not `None`, the program executes only
> ```

Should we use a tri-variant enum instead of Option? The integer case would carry a [`NonZero`](https://doc.rust-lang.org/stable/std/num/struct.NonZero.html)

swananan (commented):
UProbeScope mirrors PerfEventScope as prior art, and I keep the two commits for clarity. PerfEventScope::OneProcess needs NonZeroU32 too, I can raise a follow-up PR for this. 

> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543#-:-Oqp9nLh-2bzRUb9YPWF:bw7qbnt)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/programs/uprobe.rs` line 111 at r1](https://reviewable.io/reviews/aya-rs/aya/1543#-Oql3o9y43mQoiIHYEih:-Oqp925X4pwlk1Fk4BbY:bb6un1j) ([raw file](https://github.com/aya-rs/aya/blob/7c9c720c0f9b0b97020d7ef8961062dbcf47b61a/aya/src/programs/uprobe.rs#L111)):*

Should we use a tri-variant enum instead of Option? The integer case would carry a [`NonZero`](https://doc.rust-lang.org/stable/std/num/struct.NonZero.html)
</blockquote></details>

Done

tamird (changes requested):
> @tamird partially reviewed 6 files and all commit messages, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543)* status: 5 of 21 files reviewed, 4 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/uprobe.rs` line 2 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpDkd46IT_rRw1ipyj:-OqpDkd46IT_rRw1ipyk:b-kvkz7n) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/aya/src/programs/uprobe.rs#L2)):*
> ```XML
> //! User space probes.
> use core::num::NonZeroU32;
> ```

you can get this from std for slightly tidier imports
___
*[`aya/src/programs/uprobe.rs` line 93 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpDqL6-rNYUgIFu8NH:-OqpDqL6-rNYUgIFu8NI:bmcztdc) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/aya/src/programs/uprobe.rs#L93)):*
> ```XML
> 
> /// Specifies which processes a uprobe should fire for.
> #[derive(Debug, Clone, Copy, PartialEq, Eq)]
> ```

we probably don't need anything here except maybe Debug
___
*[`aya/src/programs/uprobe.rs` line 121 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpE1NH3j-s9p6JYidi:-OqpE1NH3j-s9p6JYidj:b1d43zx) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/aya/src/programs/uprobe.rs#L121)):*
> ```XML
>             Self::OneProcess(pid) => Some(pid.get()),
>         }
>     }
> ```

both of these are called in the same function -- i think we can just bake the knowledge into that function until we have a real second user



_Code quote:_
```XML
    const fn perf_event_pid(self) -> Option<u32> {
        match self {
            Self::AllProcesses => None,
            // Keep the kernel's pid=0 sentinel for the calling process/thread.
            Self::CallingProcess => Some(0),
            Self::OneProcess(pid) => Some(pid.get()),
        }
    }

    fn proc_map_pid(self) -> Option<u32> {
        match self {
            Self::AllProcesses => None,
            // /proc/0/maps does not exist, so resolve the calling process's
            // ProcMap with its real pid; the actual attach still uses pid=0.
            Self::CallingProcess => Some(std::process::id()),
            Self::OneProcess(pid) => Some(pid.get()),
        }
    }
```
___
*[`test/integration-test/src/tests/array.rs` line 74 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpDXHVDViueEy9YBDa:-OqpDXHVDViueEy9YBDb:bt3vzcd) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/test/integration-test/src/tests/array.rs#L74)):*
> ```XML
>             symbol,
>             "/proc/self/exe",
>             aya::programs::uprobe::UProbeScope::AllProcesses,
> ```

maybe add an import? there are already several at the top of the file

here and all others please

swananan (commented):
> @swananan made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543)* status: 0 of 21 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/uprobe.rs` line 2 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpDkd46IT_rRw1ipyj:-OqpGtIn7GmM03dkFLiy:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/aya/src/programs/uprobe.rs#L2)):*

you can get this from std for slightly tidier imports
</blockquote></details>

Done.
___
*[`aya/src/programs/uprobe.rs` line 93 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpDqL6-rNYUgIFu8NH:-OqpGvBL48adtP3_HvMk:b-elm4sy) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/aya/src/programs/uprobe.rs#L93)):*

we probably don't need anything here except maybe Debug
</blockquote></details>

Done. Same mistake, will keep it in the mind.

Restored Clone + Copy, since load.rs now uses #[test_case] to run the same uprobe unload test across multiple scopes.
___
*[`aya/src/programs/uprobe.rs` line 121 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpE1NH3j-s9p6JYidi:-OqpJfkjBT5bTL8mvHUm:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/aya/src/programs/uprobe.rs#L121)):*

both of these are called in the same function -- i think we can just bake the knowledge into that function until we have a real second user
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/array.rs` line 74 at r2](https://reviewable.io/reviews/aya-rs/aya/1543#-OqpDXHVDViueEy9YBDa:-OqpJwsS9IXuVLDFPGJQ:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/d8136e10238afcac096fc12386ee2eb709196441/test/integration-test/src/tests/array.rs#L74)):*

maybe add an import? there are already several at the top of the file

here and all others please
</blockquote></details>

Done.

tamird (approved):
That would be great, ty.

@codex review

> @tamird reviewed 21 files and all commit messages, made 2 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543#-:-Oqpdeqh62pw_LuoWZRy:bh0786f)* status: all files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/programs/uprobe.rs` line 176 at r4](https://reviewable.io/reviews/aya-rs/aya/1543#-Oqpe5up6xdkMZJQgqdB:-Oqpe5up6xdkMZJQgqdC:bc0z651) ([raw file](https://github.com/aya-rs/aya/blob/b2059395312eec25f8c2f12c67566da5057968ab/aya/src/programs/uprobe.rs#L176)):*
> ```XML
>         let Self { data, kind } = self;
>         let path = path.as_os_str();
>         attach::<Self, _>(data, *kind, path, offset, perf_event_pid, cookie)
> ```

I wonder if the pid-to-int conversion should be down inside this method. Depends who else calls it

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543)* status: all files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/programs/uprobe.rs` line 176 at r4](https://reviewable.io/reviews/aya-rs/aya/1543#-Oqpe5up6xdkMZJQgqdB:-Oqq3GMPAjEFCwzVBL5C:bgiuuam) ([raw file](https://github.com/aya-rs/aya/blob/b2059395312eec25f8c2f12c67566da5057968ab/aya/src/programs/uprobe.rs#L176)):*

I wonder if the pid-to-int conversion should be down inside this method. Depends who else calls it
</blockquote></details>

Looked into this, only uprobe will use this method with pid, even multi uprobe doesn't need this. I would keep this conversion here.

tamird (commented):
> @tamird resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1543)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

tamird:
@swananan could you update aya-template? see [these build failures](https://github.com/aya-rs/aya-template/actions/runs/24819530455/job/72641009904). thank you!

swananan:
> @swananan could you update aya-template? see [these build failures](https://github.com/aya-rs/aya-template/actions/runs/24819530455/job/72641009904). thank you!

Sure

PR #1547, opened 2026-04-23
aya: use NonZeroU32 for perf event process scope

swananan:
### Summary
Follow-up on the recent PR #1543

### Testings

`perf_event_open_trace_point` is only reachable on kernels < 4.17. aya CI runs ≥ 5.10, so uprobe e2e tests on CI never enter this branch. Coverage is provided by a mocked-syscall unit test driving all three pid-scope cases (`None` / `Some(0)` / `Some(non-zero)`) through `override_syscall`.

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)

### DIFF
--- a/aya/src/programs/kprobe.rs
+++ b/aya/src/programs/kprobe.rs
@@ -15,6 +15,7 @@ use crate::{
         ProgramData, ProgramError, ProgramType, define_link_wrapper, impl_try_from_fdlink,
         impl_try_into_fdlink, load_program_without_attach_type,
         perf_attach::{PerfLinkIdInner, PerfLinkInner},
+        perf_event::PerfEventScope,
         probe::{Probe, ProbeKind, attach},
     },
 };
@@ -87,7 +88,9 @@ impl KProbe {
             *kind,
             fn_name.as_ref(),
             offset,
-            None, // pid
+            // For all-processes attachment, perf_event_open requires an explicit
+            // CPU. Use CPU 0 only to open the backing perf event.
+            PerfEventScope::AllProcessesOneCpu { cpu: 0 },
             None, // cookie
         )
     }
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -77,6 +77,7 @@ use std::{
     borrow::Cow,
     ffi::CString,
     io,
+    num::NonZeroU32,
     os::fd::{AsFd, BorrowedFd},
     path::{Path, PathBuf},
     sync::Arc,
@@ -267,6 +268,32 @@ impl AsFd for ProgramFd {
     fn as_fd(&self) -> BorrowedFd<'_> {
         let Self(fd) = self;
         fd.as_fd()
+    }
+}
+
+/// A non-zero process identifier.
+///
+/// This type represents a real process ID. Kernel sentinel values such as `0`
+/// are modeled separately by the APIs that accept them.
+#[derive(Debug, Clone, Copy)]
+pub struct Pid(NonZeroU32);
+
+impl Pid {
+    /// Creates a process identifier from a raw PID.
+    ///
+    /// Returns `None` if `pid` is zero.
+    pub fn new(pid: u32) -> Option<Self> {
+        NonZeroU32::new(pid).map(Self)
+    }
+
+    /// Returns the identifier of the current process.
+    pub fn current() -> Self {
+        Self(NonZeroU32::new(std::process::id()).unwrap())
+    }
+
+    /// Returns the raw PID value.
+    pub const fn get(self) -> u32 {
+        self.0.get()
     }
 }
 
--- a/aya/src/programs/perf_event.rs
+++ b/aya/src/programs/perf_event.rs
@@ -11,7 +11,7 @@ use aya_obj::generated::{
 
 use crate::{
     programs::{
-        ProgramData, ProgramError, ProgramType, impl_try_from_fdlink, impl_try_into_fdlink,
+        Pid, ProgramData, ProgramError, ProgramType, impl_try_from_fdlink, impl_try_into_fdlink,
         links::define_link_wrapper,
         load_program_without_attach_type,
         perf_attach::{PerfLinkIdInner, PerfLinkInner, perf_attach},
@@ -372,7 +372,7 @@ pub enum PerfEventScope {
     /// one process
     OneProcess {
         /// process id
-        pid: u32,
+        pid: Pid,
         /// cpu id or any cpu if None
         cpu: Option<u32>,
     },
--- a/aya/src/programs/probe.rs
+++ b/aya/src/programs/probe.rs
@@ -12,7 +12,8 @@ use std::{
 use crate::{
     programs::{
         Link, ProgramData, ProgramError, perf_attach, perf_attach::PerfLinkInner,
-        perf_attach_debugfs, trace_point::read_sys_fs_trace_point_id, utils::find_tracefs_path,
+        perf_attach_debugfs, perf_event::PerfEventScope, trace_point::read_sys_fs_trace_point_id,
+        utils::find_tracefs_path,
     },
     sys::{SyscallError, perf_event_open_probe, perf_event_open_trace_point},
     util::KernelVersion,
@@ -145,7 +146,7 @@ pub(crate) fn attach<P: Probe, T: Link + From<PerfLinkInner>>(
     // separate argument.
     fn_name: &OsStr,
     offset: u64,
-    pid: Option<u32>,
+    scope: PerfEventScope,
     cookie: Option<u64>,
 ) -> Result<T::Id, ProgramError> {
     // https://github.com/torvalds/linux/commit/e12f03d7031a977356e3d7b75a68c2185ff8d155
@@ -153,13 +154,13 @@ pub(crate) fn attach<P: Probe, T: Link + From<PerfLinkInner>>(
     let prog_fd = program_data.fd()?;
     let prog_fd = prog_fd.as_fd();
     let link = if KernelVersion::at_least(4, 17, 0) {
-        let perf_fd = create_as_probe::<P>(kind, fn_name, offset, pid)?;
+        let perf_fd = create_as_probe::<P>(kind, fn_name, offset, scope)?;
         perf_attach(prog_fd, perf_fd, cookie)
     } else {
         if cookie.is_some() {
             return Err(ProgramError::AttachCookieNotSupported);
         }
-        let (perf_fd, event) = create_as_trace_point::<P>(kind, fn_name, offset, pid)?;
+        let (perf_fd, event) = create_as_trace_point::<P>(kind, fn_name, offset, scope)?;
         perf_attach_debugfs(prog_fd, perf_fd, event)
     }?;
     program_data.links.insert(T::from(link))
@@ -176,7 +177,7 @@ fn create_as_probe<P: Probe>(
     kind: ProbeKind,
     fn_name: &OsStr,
     offset: u64,
-    pid: Option<u32>,
+    scope: PerfEventScope,
 ) -> Result<crate::MockableFd, ProgramError> {
     let perf_ty = read_sys_fs_perf_type(P::PMU)
         .map_err(|(filename, io_error)| P::file_error(filename, io_error).into())?;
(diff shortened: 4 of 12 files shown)


tamird (commented):
> @tamird reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on swananan).

___
*[`aya/src/sys/perf_event.rs` line 273 at r1](https://reviewable.io/reviews/aya-rs/aya/1547#-OquPSKN4rZToyeMTPyu:-OquPSKN4rZToyeMTPyv:b-dp1pru) ([raw file](https://github.com/aya-rs/aya/blob/0dec3c2e7ff7cfd145c35417a456f71d47533be2/aya/src/sys/perf_event.rs#L273)):*
> ```XML
>     const EVENT_ID: u64 = 123;
> 
>     // `override_syscall` takes a `fn` pointer that can't capture, so per-case
> ```

would life be easier if you could change it to take `Box<Fn<....>>`?
___
*[`test/integration-test/src/tests/perf_event_bp.rs` line 124 at r1](https://reviewable.io/reviews/aya-rs/aya/1547#-OquPgzT9etgmPEHkSXm:-OquPgzT9etgmPEHkSXn:b-6k66in) ([raw file](https://github.com/aya-rs/aya/blob/0dec3c2e7ff7cfd145c35417a456f71d47533be2/test/integration-test/src/tests/perf_event_bp.rs#L124)):*
> ```XML
> 
>     let pid = std::process::id();
>     let scope_pid = NonZeroU32::new(pid).unwrap();
> ```

can this just be called pid?

swananan (commented):
> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`aya/src/sys/perf_event.rs` line 273 at r1](https://reviewable.io/reviews/aya-rs/aya/1547#-OquPSKN4rZToyeMTPyu:-Oqv9Les34dZ-guHuy54:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/0dec3c2e7ff7cfd145c35417a456f71d47533be2/aya/src/sys/perf_event.rs#L273)):*

would life be easier if you could change it to take `Box<Fn<....>>`?
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/perf_event_bp.rs` line 124 at r1](https://reviewable.io/reviews/aya-rs/aya/1547#-OquPgzT9etgmPEHkSXm:-Oqv9Adn4E-oiX6a2mOe:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/0dec3c2e7ff7cfd145c35417a456f71d47533be2/test/integration-test/src/tests/perf_event_bp.rs#L124)):*

can this just be called pid?
</blockquote></details>

Done.

tamird (changes requested):
> @tamird reviewed 3 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/sys/perf_event.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/1547#-OqxmlfU7o64kg-w-_EB:-OqxmlfU7o64kg-w-_EC:blg97dt) ([raw file](https://github.com/aya-rs/aya/blob/88181de95f916269daeaceb18379e5b023ff7466/aya/src/sys/perf_event.rs#L170)):*
> ```XML
> pub(crate) fn perf_event_open_trace_point(
>     event_id: u64,
>     pid: Option<u32>,
> ```

how come we aren't plumbing this all the way up?

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/sys/perf_event.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/1547#-OqxmlfU7o64kg-w-_EB:-OqzGRlp9NFTe8GAd9ey:b4dzmx1) ([raw file](https://github.com/aya-rs/aya/blob/88181de95f916269daeaceb18379e5b023ff7466/aya/src/sys/perf_event.rs#L170)):*

how come we aren't plumbing this all the way up?
</blockquote></details>

Right — `UProbe::attach` is the only public API that exposes `pid`today. I assume you mean adding `pid` to `TracePoint::attach`.

Looking at libbpf as reference, `bpf_program__attach_tracepoint` doesn't expose `pid` either; perf-level pid filtering doesn't constrain BPF execution on the tracepoint path anyway, so users would filter with `bpf_get_current_pid_tgid()` inside the program. I'd leave `TracePoint::attach` as-is. WDYT?

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/sys/perf_event.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/1547#-OqxmlfU7o64kg-w-_EB:-Or7fukB0yl-lnNGYCOK:b-c6nwlf) ([raw file](https://github.com/aya-rs/aya/blob/9c32f803560364168f2c894895d06dab811bca20/aya/src/sys/perf_event.rs#L170)):*

Right — `UProbe::attach` is the only public API that exposes `pid`today. I assume you mean adding `pid` to `TracePoint::attach`.

Looking at libbpf as reference, `bpf_program__attach_tracepoint` doesn't expose `pid` either; perf-level pid filtering doesn't constrain BPF execution on the tracepoint path anyway, so users would filter with `bpf_get_current_pid_tgid()` inside the program. I'd leave `TracePoint::attach` as-is. WDYT?
</blockquote></details>

I rethought this a bit. Plumbing a scope only into `perf_event_open_trace_point` doesn’t seem to buy us much: plain tracepoints don’t need it, uprobes already translate their scope before this point, and kprobes don’t need this distinction. So I’d prefer to keep this helper scoped to the raw perf_event pid mapping.

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/sys/perf_event.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/1547#-OqxmlfU7o64kg-w-_EB:-Or94aCi7SGu1B_2EntW:bot3qsg) ([raw file](https://github.com/aya-rs/aya/blob/9c32f803560364168f2c894895d06dab811bca20/aya/src/sys/perf_event.rs#L170)):*

I rethought this a bit. Plumbing a scope only into `perf_event_open_trace_point` doesn’t seem to buy us much: plain tracepoints don’t need it, uprobes already translate their scope before this point, and kprobes don’t need this distinction. So I’d prefer to keep this helper scoped to the raw perf_event pid mapping.
</blockquote></details>

I wasn't clear enough. What I really meant was that `UProbe::attach` takes a `UProbeScope`, translates it to an `Option<u32>` and then here we translate it again to `PerfEventScope`. My suggestion is to plumb `PerfEventScope` up so that the conversion from `UProbeScope` to `PerfEventScope` can happen directly.

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547)* status: 5 of 10 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/sys/perf_event.rs` line 170 at r2](https://reviewable.io/reviews/aya-rs/aya/1547#-OqxmlfU7o64kg-w-_EB:-Or9fz9w1yj3bCTPfNdI:b1yptmk) ([raw file](https://github.com/aya-rs/aya/blob/9c32f803560364168f2c894895d06dab811bca20/aya/src/sys/perf_event.rs#L170)):*

I wasn't clear enough. What I really meant was that `UProbe::attach` takes a `UProbeScope`, translates it to an `Option<u32>` and then here we translate it again to `PerfEventScope`. My suggestion is to plumb `PerfEventScope` up so that the conversion from `UProbeScope` to `PerfEventScope` can happen directly.
</blockquote></details>

Got it, that makes sense.

tamird (approved):
LGTM. @alessandrod had thoughts on NonZero in the public API, so will let him voice those here.

> @tamird reviewed 5 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1547#-:-Or9i04D2ivjBccWlEge:b-bo1dpd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on swananan).

alessandrod:
As I mentioned on discord, I'm not a fan of NonZeroU32 in public APIs.

If we're fixing up this corner of the API, I'd do something like:

```
pub struct Pid(NonZeroU32);

impl Pid {
    pub fn new(pid: u32) -> Option<Self> {
        NonZeroU32::new(pid).map(Self)
    }

    pub fn current() -> Self {
        Self(NonZeroU32::new(std::process::id()).unwrap())
    }

    pub fn get(self) -> u32 {
        self.0.get()
    }
}
```

and then the scope APIs also look weird, instead of having CallingProcess OneProcess etc we could have

```
pub struct PerfEventScope {
    pub process: ProcessScope,
    pub cpu: Option<u32>,
}

pub enum ProcessScope {
   All,
   Pid(Pid)
}
```

and I think this covers all combinations?

tamird:
The kernel exposes "calling process", IMO we should retain the ability to express that. I don't have strong feelings on the newtype wrapper around NonZeroU32, but if we introduce it then we should also use it in `UProbeScope`.

Maybe we should make these enums more opaque and control ingress with constructors. But IMO something like `pub fn new(pid: u32) -> Option<Self>` is worse than NonZero* because it's not clear from the signature or the name when it would return None.

alessandrod:
> is worse than NonZero* because it's not clear from the signature or the name when it would return None.

By this argument no fallible APIs are good because in the name they don’t express why they’re fallible. 

NonZeroU32 is bad for a number of reasons, starting from the name: the fact that a pid is a u32 is an implementation detail.

Ack on having a Current variant, but seems like the process and the cpu id should be decoupled

tamird:
You can't decouple them because when you say "all processes" you *must* specify a CPU.

> By this argument no fallible APIs are good because in the name they don’t express why they’re fallible.

No, they are just worse than the ones that say in the name why they are fallible. But I struggle to understand your beef with `NonZeroU32`; I agree u32 is an implementation detail, but even your sketch has u32 in the API. Can't get away from this implementation detail unless you *completely* hide which we can't do because there's no way to get an opaque PID of another process.

alessandrod:
> You can't decouple them because when you say "all processes" you _must_ specify a CPU.
> 
> > By this argument no fallible APIs are good because in the name they don’t express why they’re fallible.
> 
> No, they are just worse than the ones that say in the name why they are fallible. But I struggle to understand your beef with `NonZeroU32`

not sure we need to rehash the history of PID handling, even posix does pid_t in the year of the lord 2026. It's "just" a typedef but it clearly communicates that you're dealing with a pid and not an arbitrary integer. Abstractions are cool!

> You can't decouple them because when you say "all processes" you must specify a CPU.

Right now I remember this curse

tamird:
> not sure we need to rehash the history of PID handling, even posix does pid_t in the year of the lord 2026. It's "just" a typedef but it clearly communicates that you're dealing with a pid and not an arbitrary integer. Abstractions are cool!

of course they are, but `std::process::id()` does not abstract, it returns u32

alessandrod:
> of course they are, but `std::process::id()` does not abstract, it returns u32

And so do we in master. If we're trying to abstract more, the next step is a pid abstraction, not "still u32 but not zero ok!!"

tamird:
Right but it is not an abstraction! If the user wants to target any pid that isn't the current process, there is no way for us to do that without exposing `u32` in the API.

swananan:
Thanks, I missed the some of the discord context earlier😂, now I updated the PR accordingly:

1. Introduced a `Pid` wrapper
2. Updated the `UProbeScope::OneProcess` to use `Pid` as well

alessandrod (approved):

tamird:
FWIW I still think this Pid type is the wrong direction. It's a newtype around nonzerou32 with no clear benefit.

swananan:
Would you be comfortable merging this when you get a chance? Thanks!

Issue #1540, opened 2026-04-19
Add support for Netkit attachment types


rcanderson23:
The types `BPF_NETKIT_PRIMARY` and `BPF_NETKIT_PEER` are already present in in aya-obj. I took a quick look at what it would take and the existing `SchedClassifier` APIs potentially make it a little awkward. I am willing to put in some work to get this in if there is any guidance or preference on what the API might look like.

tamird:
Perhaps you could sketch a proposal.

rcanderson23:
Regardless of option a new enum for netkit attachments will be created:

```rust
pub enum NetkitAttachType {
    Primary,
    Peer,
}

impl NetkitAttachType {
    pub(crate) const fn netkit_attach_type(self) -> bpf_attach_type {
        match self {
            Self::Primary => BPF_NETKIT_PRIMARY,
            Self::Peer => BPF_NETKIT_PEER,
        }
    }
}
```

## Option 1:

Additive methods like `attach_netkit()` and `attach_netkit_with_options()` related netkit specific functions. This the smallest change possible and doesn't break current users but has the downside that we have `attach()` and `attach_with_options` for `tc` and `tcx` and a separate function for netkit specific. This is what I did locally to ensure there was nothing I was missing and it worked.


## Option 2:

The `attach()` call currently hides the attach type based on kernel version. Something like this makes everything explicit but changes the attach apis and requires changes to all the call sites.

```rust
pub enum TcNlAttachType {
    Ingress,
    Egress,
    Custom(u32),
}

pub enum TcxAttachType {
    Ingress,
    Egress,
}

pub enum NetkitAttachType {
    Primary,
    Peer,
}
pub enum SchedClassifierAttach {
    Tc {
        attach_type: TcNlAttachType,
        options: NlOptions,
    },
    Tcx {
        attach_type: TcxAttachType,
        order: LinkOrder,
    },
    Netkit {
        attach_type: NetkitAttachType,
        order: LinkOrder,
    },
}
```

While option 2 is much more involved and a breaking change, it feels more correct as the user knows when they are using netlink vs tcx for attachment instead of hiding it from the user and the three attachments types are explicit.

tamird:
Option 2 looks reasonable to me.

rcanderson23:
[...]

PR #1553, opened 2026-04-27
aya: add support for netkit attachments

rcanderson23:
Adds support for netkit primary and peer attachments. This refactors the
attachments from one that covers tc and tcx and hiding the attachment
type from the user to making the user explicitly choose which attachment
they are using.

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1553)

### DIFF
--- a/aya/src/programs/links.rs
+++ b/aya/src/programs/links.rs
@@ -648,12 +648,17 @@ bitflags::bitflags! {
 ///
 /// ```no_run
 /// # let mut bpf = aya::Ebpf::load(&[])?;
-/// use aya::programs::{tc, SchedClassifier, TcAttachType, tc::TcAttachOptions, LinkOrder};
+/// use aya::programs::{LinkOrder, SchedClassifier, SchedClassifierAttachment, TcxAttachType};
 ///
 /// let prog: &mut SchedClassifier = bpf.program_mut("redirect_ingress").unwrap().try_into()?;
 /// prog.load()?;
-/// let options = TcAttachOptions::TcxOrder(LinkOrder::first());
-/// prog.attach_with_options("eth0", TcAttachType::Ingress, options)?;
+/// prog.attach(
+///     "eth0",
+///     SchedClassifierAttachment::Tcx {
+///         attach_type: TcxAttachType::Ingress,
+///         link_order: LinkOrder::first(),
+///     },
+/// )?;
 ///
 /// # Ok::<(), aya::EbpfError>(())
 /// ```
--- a/aya/src/programs/links.rs
+++ b/aya/src/programs/links.rs
@@ -656,12 +656,17 @@ bitflags::bitflags! {
 ///
 /// ```no_run
 /// # let mut bpf = aya::Ebpf::load(&[])?;
-/// use aya::programs::{tc, SchedClassifier, TcAttachType, tc::TcAttachOptions, LinkOrder};
+/// use aya::programs::{LinkOrder, SchedClassifier, SchedClassifierAttachment, TcxAttachType};
 ///
 /// let prog: &mut SchedClassifier = bpf.program_mut("redirect_ingress").unwrap().try_into()?;
 /// prog.load()?;
-/// let options = TcAttachOptions::TcxOrder(LinkOrder::first());
-/// prog.attach_with_options("eth0", TcAttachType::Ingress, options)?;
+/// prog.attach(
+///     "eth0",
+///     SchedClassifierAttachment::Tcx {
+///         attach_type: TcxAttachType::Ingress,
+///         link_order: LinkOrder::first(),
+///     },
+/// )?;
 ///
 /// # Ok::<(), aya::EbpfError>(())
 /// ```
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -133,7 +133,10 @@ pub use crate::programs::{
     sk_skb::{SkSkb, SkSkbKind},
     sock_ops::SockOps,
     socket_filter::{ReusePortSocketFilter, SocketFilter, SocketFilterError},
-    tc::{SchedClassifier, TcAttachType, TcError, TcHandle},
+    tc::{
+        NetkitAttachType, SchedClassifier, SchedClassifierAttachment, TcAttachType, TcError,
+        TcHandle, TcxAttachType,
+    },
     tp_btf::BtfTracePoint,
     trace_point::{TracePoint, TracePointError},
     uprobe::{UProbe, UProbeError},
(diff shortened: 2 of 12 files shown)


tamird (approved):
> @tamird+AGNT reviewed 13 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1553)* status: all files reviewed, 3 unresolved discussions (waiting on rcanderson23).

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/1553#-P1pJfz07HFszlFE1wTw:-P1pJfz07HFszlFE1wTx:by9rbav) (no related file):*
Codex-assisted review of both commits through `d66ae319`.

The explicit attachment variants remove the invalid TCX `Custom` combination, and I did not find a production Netkit attachment/query ABI defect. The requested changes concern tests that currently hide setup failures, duplicated TCX coverage without exercising the Peer hook, and stale public error documentation.

I inspected the code, kernel interfaces, existing discussions, and reported CI results; I did not run local builds or tests. The current head has merge conflicts with `main`.
___
*[`test/integration-test/src/tests/netkit.rs` line 21 at r12](https://reviewable.io/reviews/aya-rs/aya/1553#-P1pJejgE3yCOd3xedis:-P1pJejgE3yCOd3xedit:b-icsqrc) ([raw file](https://github.com/aya-rs/aya/blob/d66ae319d199948665fc68d3c85fc972b94cb4e9/test/integration-test/src/tests/netkit.rs#L21)):*
> ```XML
> 
>     let _netns = NetNsGuard::new().unwrap();
>     if let Err(err) = create_netkit_link(primary, peer) {
> ```

Please let unexpected setup errors fail this test. After the kernel-version check, this catches every `create_netkit_link` error and returns success: malformed netlink attributes, interface setup failures, and regressions in the new helper all become a green skip before any attachment is exercised. `pin_netkit_link` has the same issue.

If a supported kernel can lack Netkit configuration, distinguish that unsupported case narrowly and propagate the other errors.
___
*[`test/integration-test/src/tests/netkit.rs` line 41 at r12](https://reviewable.io/reviews/aya-rs/aya/1553#-P1pJf958-CmVQRqbz-V:-P1pJf958-CmVQRqbz-W:b1pcjtf) ([raw file](https://github.com/aya-rs/aya/blob/d66ae319d199948665fc68d3c85fc972b94cb4e9/test/integration-test/src/tests/netkit.rs#L41)):*
> ```XML
>                     primary,
>                     SchedClassifierAttachment::Netkit {
>                         attach_type: NetkitAttachType::Primary,
> ```

Please share the existing TCX ordering and pinning test bodies across attachment cases, and include both Netkit `Primary` and `Peer`. This file copies the TCX ordering test, and `pin_netkit_link` copies `pin_tcx_link`, but both new tests only exercise `Primary`.

A useful Netkit-specific assertion is hook isolation: attach through the primary interface to `Peer`, then query both hooks and verify the program is present only on `Peer`. That would catch an accidental `Peer` → `Primary` mapping while keeping the common ordering and link-lifetime coverage in one place.
___
*[`aya/src/programs/tc.rs` line 353 at r11](https://reviewable.io/reviews/aya-rs/aya/1553#-P1pJfa_Db5zsNsn7B5c:-P1pJfa_Db5zsNsn7B5d:bfgv6sf) ([raw file](https://github.com/aya-rs/aya/blob/ff6a4bfc409c5ec34a8008cf69954ef171f1fefe/aya/src/programs/tc.rs#L353)):*
> ```XML
>     /// # Errors
>     ///
>     /// When attaching fails, [`ProgramError::SyscallError`] is returned for
> ```

This still describes the removed automatic kernel-version selection. With the new API, choosing `SchedClassifierAttachment::Tc` uses netlink even on a kernel newer than 6.6. Please describe the error path by the selected attachment variant, and keep the `clsact` requirement attached to the `Tc` case.

tamird (changes requested):
> @tamird partially reviewed 6 files and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1553)* status: all files reviewed, 6 unresolved discussions (waiting on rcanderson23).

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/aya-rs/aya/1553#-OrFjQNFEvAOmbmJRg0D:-OrFjQNG43aq4H6pQLK3:byi7kp9):*
"Fixes #1540." please
___
*[`aya/src/programs/tc.rs` line 167 at r1](https://reviewable.io/reviews/aya-rs/aya/1553#-OrJFpKo59w7Zz4gJvHX:-OrJFpKo59w7Zz4gJvHY:b-olncrc) ([raw file](https://github.com/aya-rs/aya/blob/18153c74fa12c8771e4f7904ed6abcb85261ac96/aya/src/programs/tc.rs#L167)):*
> ```XML
>     /// On kernels >= 6.6.0, it will attempt to use the TCX interface and attach as
>     /// the last TCX program. On older kernels, it will fallback to using the
>     /// legacy netlink interface.
> ```

can you help me understand what's going on here? does this mean users will now need to pick the appropriate interface depending on the kernel version they are using?



_Code quote:_
```XML
    /// On kernels >= 6.6.0, it will attempt to use the TCX interface and attach as
    /// the last TCX program. On older kernels, it will fallback to using the
    /// legacy netlink interface.
```
___
*[`test/integration-test/src/tests/tcx.rs` line 13 at r4](https://reviewable.io/reviews/aya-rs/aya/1553#-OrJGyqV2RIXfhfdjzVf:-OrJGyqV2RIXfhfdjzVg:bonof4m) ([raw file](https://github.com/aya-rs/aya/blob/c300e3c326c0258eb6b52f7978dc78bf916aed99/test/integration-test/src/tests/tcx.rs#L13)):*
> ```XML
> #[test_log::test]
> fn tcx() {
>     let kernel_version = KernelVersion::current().unwrap();
> ```

See my other comment; I think this is hiding the fact that this change break older kernels. See https://github.com/aya-rs/aya/pull/1555.

rcanderson23:
*[`aya/src/programs/tc.rs` line 167 at r1](https://reviewable.io/reviews/aya-rs/aya/1553#-OrJFpKo59w7Zz4gJvHX:-P1qzuMY8aZtTjfH6q0u:b-rjmivn) ([raw file](https://github.com/aya-rs/aya/blob/18153c74fa12c8771e4f7904ed6abcb85261ac96/aya/src/programs/tc.rs#L167)):*

can you help me understand what's going on here? does this mean users will now need to pick the appropriate interface depending on the kernel version they are using?
</blockquote></details>

Yes, this is a breaking change that would force users to choose their attachment type or manually implement tcx attempt and fall back on tc. This approach was mentioned in the issue before implementation and was seen as reasonable https://github.com/aya-rs/aya/issues/1540#issuecomment-4285765649. I also commented on that there was previous desire to hide netlink vs bpf_link from the users: https://github.com/aya-rs/aya/issues/1540#issuecomment-4308446030. 

I can refactor this to approach 1 mentioned in the issue and leave `attach` untouched and implement netkit specific methods if you would rather go that direction.

PR #1554, opened 2026-04-28
aya: PerfEventArrayBuffer::next_event

Brskt:
`PerfEventArrayBuffer` exposes `try_fold`, `fold`, and `for_each` that
drain available events through a closure. The closure receives
`PerfEvent<'_>`, which is `Sample { head: &[u8], tail: &[u8] }` (slices
borrowed directly from the kernel-mapped ring buffer; `tail` is empty
unless the sample straddles the wrap boundary) or `Lost { count }`.
`data_tail` is committed once per drain instead of once per event.
Follow-up to #1550.

A first commit fixes a latent acquire-fence bug in `read_events`
between the `data_head` load and the record reads, present since 2021.

A second commit replaces `read_events` with the closure-based API
and migrates the perf event array integration test to `for_each`.
`data_head` is read with `ptr::read_volatile` + Acquire fence to pair
with the kernel's smp_wmb() + WRITE_ONCE() publish; `data_tail` is
read with a plain field access since userspace is the sole writer;
writes use `ptr::write_volatile` + SeqCst fence to pair with the
kernel's READ_ONCE(). The shared u64 path matches
`tools/include/linux/ring_buffer.h` on every supported arch, including
32-bit hosts that lack a 64-bit acquire-load. Removes `read_events`,
`Events`, `PerfBufferError::NoBuffers`, `PerfBufferError::MoreSpaceNeeded`,
and the `bytes` dependency.

A final commit migrates the perf event array integration test to
`for_each`

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)

### DIFF
--- a/aya/Cargo.toml
+++ b/aya/Cargo.toml
@@ -20,7 +20,6 @@ workspace = true
 assert_matches = { workspace = true }
 aya-obj = { path = "../aya-obj", version = "^0.2.2" }
 bitflags = { workspace = true }
-bytes = { workspace = true }
 # TODO(https://github.com/rust-lang/rust/issues/60896): Remove once
 # `std::collections::hash_set::Entry` is stabilized.
 hashbrown = { workspace = true, features = ["default-hasher", "equivalent"] }
@@ -28,6 +27,7 @@ libc = { workspace = true }
 log = { workspace = true }
 object = { workspace = true, features = ["elf", "read_core", "std", "write"] }
 once_cell = { workspace = true }
+scopeguard = { workspace = true }
 thiserror = { workspace = true }
 
 [dev-dependencies]
--- a/aya/src/maps/perf/perf_buffer.rs
+++ b/aya/src/maps/perf/perf_buffer.rs
@@ -220,6 +220,9 @@ impl PerfBuffer {
 
         let head = unsafe { (*header).data_head } as usize;
         let mut tail = unsafe { (*header).data_tail } as usize;
+        // Pair with the kernel's smp_store_release on `data_head`: subsequent
+        // record reads must not be reordered before observing the head.
+        atomic::fence(Ordering::Acquire);
         let result = loop {
             if head == tail {
                 break Ok(());
--- a/aya/src/maps/perf/perf_buffer.rs
+++ b/aya/src/maps/perf/perf_buffer.rs
@@ -1,7 +1,10 @@
 use std::{
+    convert::Infallible,
     io,
+    ops::ControlFlow,
     os::fd::{AsFd, BorrowedFd},
-    ptr, slice,
+    ptr::{self, NonNull},
+    slice,
     sync::atomic::{self, Ordering},
 };
 
@@ -9,7 +12,6 @@ use aya_obj::generated::{
     PERF_FLAG_FD_CLOEXEC, perf_event_header, perf_event_mmap_page,
     perf_event_type::{PERF_RECORD_LOST, PERF_RECORD_SAMPLE},
 };
-use bytes::BytesMut;
 use libc::{MAP_SHARED, PROT_READ, PROT_WRITE};
 use thiserror::Error;
 
@@ -55,34 +57,41 @@ pub enum PerfBufferError {
         io_error: io::Error,
     },
 
-    /// `read_events()` was called with no output buffers.
-    #[error("read_events() was called with no output buffers")]
-    NoBuffers,
-
-    /// `read_events()` was called with a buffer that is not large enough to
-    /// contain the next event in the perf buffer.
-    #[deprecated(
-        since = "0.10.8",
-        note = "read_events() now calls BytesMut::reserve() internally, so this error is never returned"
-    )]
-    #[error("the buffer needs to be of at least {size} bytes")]
-    MoreSpaceNeeded {
-        /// expected size
-        size: usize,
-    },
-
     /// An IO error occurred.
     #[error(transparent)]
     IOError(#[from] io::Error),
 }
 
-/// Return type of `read_events()`.
-#[derive(Debug, PartialEq, Eq)]
-pub struct Events {
-    /// The number of events read.
-    pub read: usize,
-    /// The number of events lost.
-    pub lost: usize,
+/// An event read from a perf event array buffer.
+///
+/// Yielded by [`PerfEventArrayBuffer::try_fold`], [`PerfEventArrayBuffer::fold`],
+/// and [`PerfEventArrayBuffer::for_each`].
+///
+/// [`PerfEventArrayBuffer::try_fold`]: super::PerfEventArrayBuffer::try_fold
+/// [`PerfEventArrayBuffer::fold`]: super::PerfEventArrayBuffer::fold
+/// [`PerfEventArrayBuffer::for_each`]: super::PerfEventArrayBuffer::for_each
+#[derive(Debug)]
+pub enum PerfEvent<'a> {
+    /// A sample emitted by `bpf_perf_event_output()`. The bytes are exposed
+    /// as up to two slices borrowed directly from the kernel-mapped ring
+    /// buffer; the second slice is empty for samples that fit contiguously,
+    /// and both are populated when a sample straddles the ring boundary. The
+    /// bytes include any kernel-side alignment padding that follows the
+    /// payload.
+    #[doc(alias = "PERF_RECORD_SAMPLE")]
+    Sample {
+        /// First chunk of the sample, or the entire sample when it does not wrap.
+        head: &'a [u8],
+        /// Second chunk; empty unless the sample straddles the ring boundary.
+        tail: &'a [u8],
+    },
+    /// A signal from the kernel that samples were dropped because the ring
+    /// buffer was full.
+    #[doc(alias = "PERF_RECORD_LOST")]
+    Lost {
+        /// Number of dropped samples.
+        count: u64,
+    },
 }
 
 #[cfg_attr(test, derive(Debug))]
@@ -135,131 +144,189 @@ impl PerfBuffer {
         Ok(perf_buf)
     }
 
-    const fn buf(&self) -> ptr::NonNull<perf_event_mmap_page> {
+    const fn buf(&self) -> NonNull<perf_event_mmap_page> {
         self.mmap.ptr().cast()
     }
 
-    pub(crate) fn readable(&self) -> bool {
-        let header = self.buf().as_ptr();
-        let head = unsafe { (*header).data_head } as usize;
-        let tail = unsafe { (*header).data_tail } as usize;
-        head != tail
+    /// Read `data_head`, the kernel's producer position. Pairs with the
+    /// kernel's `smp_wmb() + WRITE_ONCE()` publish [1].
+    ///
+    /// [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/ring_buffer.c#L113-L114
+    fn data_head(&self) -> u64 {
+        // SAFETY: `self.buf()` points to the mmap'd `perf_event_mmap_page` for
+        // the lifetime of `self`, so the field projection is in-bounds and
+        // aligned.
+        let ptr = unsafe { &raw const (*self.buf().as_ptr()).data_head };
+        // SAFETY: the kernel writes `data_head` concurrently; a plain read
+        // would be UB.
+        let value = unsafe { ptr::read_volatile(ptr) };
+        atomic::fence(Ordering::Acquire);
+        value
     }
 
-    pub(crate) fn read_events(
-        &mut self,
-        buffers: &mut [BytesMut],
-    ) -> Result<Events, PerfBufferError> {
-        if buffers.is_empty() {
-            return Err(PerfBufferError::NoBuffers);
-        }
-        let header = self.buf().as_ptr();
-        let base = unsafe { header.byte_add(self.page_size) };
+    /// Read `data_tail`, the userspace consumer position. Userspace is the
+    /// sole writer, so a plain read suffices.
+    fn data_tail(&self) -> u64 {
+        // SAFETY: `self.buf()` points to the mmap'd `perf_event_mmap_page`
+        // for the lifetime of `self`. Only userspace writes this field, so
+        // there is no concurrent kernel write to race with.
+        unsafe { (*self.buf().as_ptr()).data_tail }
+    }
 
-        let mut events = Events { read: 0, lost: 0 };
-        let mut buf_n = 0;
+    /// Mutable pointer to `data_tail`, used by [`try_fold`] to commit the
+    /// local position with a release-store on scope exit.
+    ///
+    /// [`try_fold`]: Self::try_fold
+    fn data_tail_ptr(&mut self) -> NonNull<u64> {
+        // SAFETY: `self.buf()` points to the mmap'd `perf_event_mmap_page` for
+        // the lifetime of `self`, so the field projection is in-bounds and
+        // aligned.
+        let ptr = unsafe { &raw mut (*self.buf().as_ptr()).data_tail };
+        // SAFETY: `self.buf()` is `NonNull`, so a field projection within it
+        // is non-null.
+        unsafe { NonNull::new_unchecked(ptr) }
+    }
 
-        let fill_buf = |start_off, base: *const u8, mmap_size, out_buf: &mut [u8]| {
-            let len = out_buf.len();
+    /// Pointer to the start of the data pages following the header page.
+    const fn data_pages(&self) -> *const u8 {
+        // SAFETY: the mmap spans `page_size + size` bytes; the data area
+        // starts at offset `page_size`.
+        unsafe { self.buf().as_ptr().byte_add(self.page_size) }.cast::<u8>()
+    }
 
-            let end = (start_off + len) % mmap_size;
-            let start = start_off % mmap_size;
+    pub(crate) fn readable(&self) -> bool {
+        self.data_head() != self.data_tail()
+    }
 
-            if start < end {
-                out_buf.copy_from_slice(unsafe { slice::from_raw_parts(base.add(start), len) });
-            } else {
-                let size = mmap_size - start;
+    pub(crate) fn try_fold<B, C, F>(&mut self, init: C, mut f: F) -> ControlFlow<B, C>
+    where
+        F: FnMut(C, PerfEvent<'_>) -> ControlFlow<B, C>,
+    {
+        let base = self.data_pages();
+        let mmap_size = self.size;
+        let data_head = self.data_head();
+        let initial_tail = self.data_tail();
+        let data_tail_ptr = self.data_tail_ptr();
+
+        // Defer the userspace release-store of `data_tail` to scope exit so
+        // it pairs with the kernel's `READ_ONCE()` [1] in one SeqCst barrier
+        // instead of per event, and remains panic-safe.
+        //
+        // [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/ring_buffer.c#L202
+        let mut guard = scopeguard::guard(initial_tail, |tail| {
+            if tail != initial_tail {
+                atomic::fence(Ordering::SeqCst);
+                // SAFETY: `data_tail_ptr` was derived from a `&mut PerfBuffer`
+                // outliving this scope; userspace is the sole writer of
+                // `data_tail`.
                 unsafe {
-                    out_buf[..size].copy_from_slice(slice::from_raw_parts(base.add(start), size));
-                    out_buf[size..].copy_from_slice(slice::from_raw_parts(base, len - size));
+                    ptr::write_volatile(data_tail_ptr.as_ptr(), tail);
                 }
             }
-        };
+        });
 
-        let read_event = |event_start, event_type, base, buf: &mut BytesMut| {
-            let sample_size = match event_type {
-                x if x == PERF_RECORD_SAMPLE as u32 || x == PERF_RECORD_LOST as u32 => {
-                    let mut size = [0u8; size_of::<u32>()];
-                    fill_buf(
-                        event_start + size_of::<perf_event_header>(),
-                        base,
-                        self.size,
-                        &mut size,
-                    );
-                    u32::from_ne_bytes(size)
-                }
-                _ => return Ok(None),
-            } as usize;
+        let tail = &mut *guard;
 
-            let sample_start =
-                (event_start + size_of::<perf_event_header>() + size_of::<u32>()) % self.size;
+        let mut acc = init;
+        while *tail != data_head {
+            let event_start = (*tail % mmap_size as u64) as usize;
+            debug_assert_eq!(
+                event_start % 8,
+                0,
+                "perf records are 8-byte aligned (event_start={event_start})"
+            );
+            // SAFETY: the kernel pads each record to 8 bytes [1] and the
+            // buffer is page-aligned, so `base + event_start` is aligned for
+            // `perf_event_header` and the 8-byte header fits before the wrap
+            // boundary.
+            //
+            // [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/core.c#L8451
+            let event: perf_event_header = unsafe { ptr::read(base.add(event_start).cast()) };
+            *tail = tail.wrapping_add(u64::from(event.size));
 
-            match event_type {
+            let perf_event = match event.type_ {
                 x if x == PERF_RECORD_SAMPLE as u32 => {
-                    buf.clear();
-                    buf.reserve(sample_size);
-                    unsafe { buf.set_len(sample_size) }
+                    // The `u32` size prefix follows the header at an 8-aligned
+                    // offset, so it never spans the wrap.
+                    let size_offset = (event_start + size_of::<perf_event_header>()) % mmap_size;
+                    // SAFETY: same kernel-alignment guarantee as the header read.
+                    let sample_size: u32 = unsafe { ptr::read(base.add(size_offset).cast()) };
+                    let sample_size = sample_size as usize;
 
-                    fill_buf(sample_start, base, self.size, buf);
-
-                    Ok(Some((1, 0)))
+                    // The sample payload follows the size prefix and may span
+                    // the wrap.
+                    let sample_start =
+                        (event_start + size_of::<perf_event_header>() + size_of::<u32>())
+                            % mmap_size;
+                    debug_assert!(sample_size <= mmap_size);
+                    let (head, tail) = if let Some(second) =
+                        (sample_start + sample_size).checked_sub(mmap_size)
+                    {
+                        let first = mmap_size - sample_start;
+                        // SAFETY: `[sample_start, mmap_size)` and `[0, second)`
+                        // are disjoint ranges within the mmap region; the
+                        // kernel will not overwrite them until `data_tail`
+                        // is committed on scope exit.
+                        let head = unsafe { slice::from_raw_parts(base.add(sample_start), first) };
+                        let tail = unsafe { slice::from_raw_parts(base, second) };
+                        (head, tail)
+                    } else {
+                        // SAFETY: same as above; `sample_start + sample_size
+                        // <= mmap_size`, contiguous within mmap.
+                        let head =
+                            unsafe { slice::from_raw_parts(base.add(sample_start), sample_size) };
+                        (head, &[][..])
+                    };
+                    PerfEvent::Sample { head, tail }
                 }
                 x if x == PERF_RECORD_LOST as u32 => {
-                    let mut count = [0u8; size_of::<u64>()];
-                    fill_buf(
-                        event_start + size_of::<perf_event_header>() + size_of::<u64>(),
-                        base,
-                        self.size,
-                        &mut count,
-                    );
-                    Ok(Some((0, u64::from_ne_bytes(count) as usize)))
+                    // `PERF_RECORD_LOST` layout is
+                    // `{ header, u64 id, u64 lost, sample_id }` [1]; skip past
+                    // `id` to read the `lost` count.
+                    //
+                    // [1]: https://github.com/torvalds/linux/blob/05f7e89a/include/uapi/linux/perf_event.h#L906-L914
+                    let lost_offset =
+                        (event_start + size_of::<perf_event_header>() + size_of::<u64>())
+                            % mmap_size;
+                    // SAFETY: same kernel-alignment guarantee as the header read; the
+                    // fixed 8-byte field at an 8-aligned offset cannot span the wrap.
+                    let count: u64 = unsafe { ptr::read(base.add(lost_offset).cast()) };
+                    PerfEvent::Lost { count }
                 }
-                _ => Ok(None),
-            }
-        };
-
-        let head = unsafe { (*header).data_head } as usize;
-        let mut tail = unsafe { (*header).data_tail } as usize;
-        // Pair with the kernel's smp_store_release on `data_head`: subsequent
-        // record reads must not be reordered before observing the head.
-        atomic::fence(Ordering::Acquire);
-        let result = loop {
-            if head == tail {
-                break Ok(());
-            }
-            if buf_n == buffers.len() {
-                break Ok(());
-            }
-
-            let buf = &mut buffers[buf_n];
-
-            let event_start = tail % self.size;
-            let event: perf_event_header =
-                unsafe { ptr::read_unaligned(base.byte_add(event_start).cast()) };
-            let event_size = event.size as usize;
-
-            match read_event(event_start, event.type_, base.cast(), buf) {
-                Ok(Some((read, lost))) => {
-                    if read > 0 {
-                        buf_n += 1;
-                        events.read += read;
-                    }
-                    events.lost += lost;
+                event_type => {
+                    // `PerfBuffer::open` configures `SoftwareEvent::BpfOutput`
+                    // with no side-band attr flags [1]; the kernel only emits
+                    // SAMPLE and LOST.
+                    //
+                    // [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/core.c#L5182-L5200
+                    debug_assert!(false, "unexpected perf record type: {event_type}");
+                    continue;
                 }
-                Ok(None) => { /* skip unknown event type */ }
-                Err(e) => {
-                    // we got an error and we didn't process any events, propagate the error
-                    // and give the caller a chance to increase buffers
-                    break Err(e);
-                }
+            };
+
+            match f(acc, perf_event) {
+                ControlFlow::Continue(next) => acc = next,
+                ControlFlow::Break(v) => return ControlFlow::Break(v),
             }
-            tail += event_size;
-        };
+        }
 
-        atomic::fence(Ordering::SeqCst);
-        unsafe { (*header).data_tail = tail as u64 }
+        ControlFlow::Continue(acc)
+    }
 
-        result.map(|()| events)
+    pub(crate) fn fold<C, F>(&mut self, init: C, mut f: F) -> C
+    where
+        F: FnMut(C, PerfEvent<'_>) -> C,
+    {
+        let ControlFlow::Continue(acc) = self
+            .try_fold::<Infallible, _, _>(init, |acc, event| ControlFlow::Continue(f(acc, event)));
+        acc
+    }
+
+    pub(crate) fn for_each<F>(&mut self, mut f: F)
+    where
+        F: FnMut(PerfEvent<'_>),
+    {
+        self.fold((), |(), event| f(event))
     }
 }
 
@@ -281,6 +348,7 @@ mod tests {
     use std::fmt::Debug;
 
     use assert_matches::assert_matches;
+    use test_case::test_case;
 
     use super::*;
     use crate::sys::{Syscall, TEST_MMAP_RET, override_syscall};
@@ -325,34 +393,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn test_no_out_bufs() {
-        let mut mmapped_buf = MMappedBuf {
-            data: [0; PAGE_SIZE * 2],
-        };
-
-        fake_mmap(&mut mmapped_buf);
-        let mut buf = PerfBuffer::open(1, PAGE_SIZE, 1).unwrap();
-
-        assert_matches!(buf.read_events(&mut []), Err(PerfBufferError::NoBuffers))
-    }
-
-    #[test]
-    fn test_no_events() {
-        let mut mmapped_buf = MMappedBuf {
-            data: [0; PAGE_SIZE * 2],
-        };
-
-        fake_mmap(&mut mmapped_buf);
-        let mut buf = PerfBuffer::open(1, PAGE_SIZE, 1).unwrap();
-
-        let out_buf = BytesMut::with_capacity(4);
-        assert_eq!(
-            buf.read_events(&mut [out_buf]).unwrap(),
-            Events { read: 0, lost: 0 }
-        );
-    }
-
     fn write<T: Debug>(mmapped_buf: &mut MMappedBuf, offset: usize, value: T) -> usize {
         let dst: *mut _ = mmapped_buf;
         let head = offset + size_of::<T>();
@@ -363,8 +403,70 @@ mod tests {
         head
     }
 
+    #[repr(C)]
+    #[derive(Debug)]
+    struct TestPerfRecord<T: Debug> {
+        s_hdr: Sample,
+        value: T,
+    }
+
+    fn write_sample<T: Debug>(mmapped_buf: &mut MMappedBuf, offset: usize, value: T) -> usize {
+        write(
+            mmapped_buf,
+            offset,
+            TestPerfRecord {
+                s_hdr: Sample {
+                    header: perf_event_header {
+                        type_: PERF_RECORD_SAMPLE as u32,
+                        misc: 0,
+                        size: size_of::<TestPerfRecord<T>>() as u16,
+                    },
+                    size: size_of::<T>() as u32,
+                },
+                value,
+            },
+        )
+    }
+
+    fn u32_from_buf(buf: &[u8]) -> u32 {
+        u32::from_ne_bytes(buf[..4].try_into().unwrap())
+    }
+
+    fn sample_bytes(head: &[u8], tail: &[u8]) -> Vec<u8> {
+        let mut v = Vec::with_capacity(head.len() + tail.len());
+        v.extend_from_slice(head);
+        v.extend_from_slice(tail);
+        v
+    }
+
+    #[test_case(&[] ; "empty")]
+    #[test_case(&[0xCAFEBABEu32] ; "single")]
+    #[test_case(&[0xCAFEBABEu32, 0xBADCAFEu32] ; "consecutive")]
+    fn test_for_each_samples(expected: &[u32]) {
+        let mut mmapped_buf = MMappedBuf {
+            data: [0; PAGE_SIZE * 2],
+        };
+
+        let mut offset = 0;
+        for &v in expected {
+            offset = write_sample(&mut mmapped_buf, offset, v);
+        }
+
+        fake_mmap(&mut mmapped_buf);
+        let mut buf = PerfBuffer::open(1, PAGE_SIZE, 1).unwrap();
+
+        let mut payloads = Vec::new();
+        buf.for_each(|event| match event {
+            PerfEvent::Sample { head, tail } => {
+                payloads.push(u32_from_buf(&sample_bytes(head, tail)));
+            }
+            PerfEvent::Lost { count } => panic!("unexpected lost: {count}"),
+        });
+        assert_eq!(payloads, expected);
+    }
+
     #[test]
-    fn test_read_first_lost() {
+    fn test_for_each_lost() {
         #[repr(C)]
         #[derive(Debug)]
         struct LostSamples {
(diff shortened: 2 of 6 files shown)


tamird (changes requested):
Haven't looked at the new API yet but can we aim to replace the old with the new? I also see that the new API allocates; can we make it zero copy? I wouldn't copy the ring buffer API, see #1479 .

> @tamird partially reviewed 1 file and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-OrJ8-Hj--ejSqmtcbPy:b-u62j06)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 221 at r1](https://reviewable.io/reviews/aya-rs/aya/1554#-OrJ7Xa85ZZYYPqOxY6E:-OrJ7Xa95AKttJyP0d-b:bajnxp6) ([raw file](https://github.com/aya-rs/aya/blob/2999294f6efcd1a5aeba7b10b23fe234dcdcbd88/aya/src/maps/perf/perf_buffer.rs#L221)):*
> ```XML
>         };
> 
>         let head = unsafe { (*header).data_head } as usize;
> ```

Instead of manual fences, can we treat these as atomics? We do something like this in the ring buffer implementation.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 223 at r1](https://reviewable.io/reviews/aya-rs/aya/1554#-OrJ7g5sCWPmurpAwCkK:-OrJ7g5sCWPmurpAwCkL:b-o19x21) ([raw file](https://github.com/aya-rs/aya/blob/2999294f6efcd1a5aeba7b10b23fe234dcdcbd88/aya/src/maps/perf/perf_buffer.rs#L223)):*
> ```XML
>         let head = unsafe { (*header).data_head } as usize;
>         let mut tail = unsafe { (*header).data_tail } as usize;
>         // Pair with the kernel's smp_store_release on `data_head`: subsequent
> ```

This kind of thing needs citations. See the ring buffer implementation which I believe has specific kernel code references.

Brskt (commented):
Sure, done. `read_events` removed, `next_event` is zero-copy. `PerfSample::as_slices` exposes both halves of wrap-spanning samples (no `Deref<[u8]>` since the perf ring isn't double-mapped). Drop advances `data_tail`.

> @Brskt made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-OrOh0mV1c6-hpOR3zZp:bihb9g5)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 221 at r1](https://reviewable.io/reviews/aya-rs/aya/1554#-OrJ7Xa85ZZYYPqOxY6E:-OrNrIsQ2jE7qW7VZVO9:b-dugiwj) ([raw file](https://github.com/aya-rs/aya/blob/2999294f6efcd1a5aeba7b10b23fe234dcdcbd88/aya/src/maps/perf/perf_buffer.rs#L221)):*

Instead of manual fences, can we treat these as atomics? We do something like this in the ring buffer implementation.
</blockquote></details>

Ring_buf uses `AtomicUsize` because its kernel positions are `unsigned long`. Perf `data_head`/`data_tail` are fixed `__u64`, so the equivalent `AtomicU64` doesn't compile on 32-bit MIPS. Using `ptr::read_volatile` + `atomic::fence` helpers instead.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 223 at r1](https://reviewable.io/reviews/aya-rs/aya/1554#-OrJ7g5sCWPmurpAwCkK:-OrO8URW6lNNbZKOWw_o:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2999294f6efcd1a5aeba7b10b23fe234dcdcbd88/aya/src/maps/perf/perf_buffer.rs#L223)):*

This kind of thing needs citations. See the ring buffer implementation which I believe has specific kernel code references.
</blockquote></details>

Done.

tamird (changes requested):
I think you ignored this part:

> I wouldn't copy the ring buffer API, see #1479 .

> @tamird partially reviewed 6 files, made 8 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-OrOwIPP9XacZOLrnQWP:by75hcy)* status: 3 of 6 files reviewed, 7 unresolved discussions (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 69 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOrcX40zLF5EcFNNal:-OrOrcX40zLF5EcFNNam:beww9k0) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L69)):*
> ```XML
> /// See [`PerfEventArrayBuffer::next_event`].
> ///
> /// [`PerfEventArrayBuffer::next_event`]: crate::maps::perf::PerfEventArrayBuffer::next_event
> ```

s/crate::maps::perf/super/ ?
___
*[`aya/src/maps/perf/perf_buffer.rs` line 72 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOrmYd0mVzhuoBNCM_:-OrOrmYd0mVzhuoBNCMa:bywpowc) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L72)):*
> ```XML
> #[derive(Debug)]
> pub enum PerfEvent<'a> {
>     /// A `PERF_RECORD_SAMPLE` event. The sample bytes emitted by
> ```

instead of saying `PERF_RECORD_SAMPLE` you should use the doc alias attribute, that way a search in the docs for PERF_RECORD_SAMPLE would produce the right result.

Ditto below.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 97 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOsD6V2lnDZHbosQmP:-OrOsD6V2lnDZHbosQmQ:bb7od0r) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L97)):*
> ```XML
> /// the kernel observes the buffer as full and stops writing new events.
> ///
> /// [`PerfEventArrayBuffer::next_event`]: crate::maps::perf::PerfEventArrayBuffer::next_event
> ```

same
___
*[`aya/src/maps/perf/perf_buffer.rs` line 103 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOsK46DjlcAV_wNKU0:-OrOsK46DjlcAV_wNKU1:b-3b74hv) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L103)):*
> ```XML
>     data_tail: NonNull<u64>,
>     advance: u64,
>     _marker: PhantomData<&'a mut ()>,
> ```

There should be a comment here indicating this is the lifetime of the PerfBuffer from which this data is owned and also explaining why it isn't written as literally `&'a mut PerfBuffer`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 133 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOvn5O3Gyu8gfPS4Te:-OrOvn5O3Gyu8gfPS4Tf:b-mchm1z) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L133)):*
> ```XML
>     pub const fn is_empty(&self) -> bool {
>         self.len() == 0
>     }
> ```

do we need this?



_Code quote:_
```XML
    /// Total length of the sample in bytes.
    pub const fn len(&self) -> usize {
        let Self { head, tail, .. } = self;
        head.len() + tail.len()
    }

    /// Returns `true` if the sample is empty.
    pub const fn is_empty(&self) -> bool {
        self.len() == 0
    }
```
___
*[`aya/src/maps/perf/perf_buffer.rs` line 246 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrPogV55U-myNv0nLEH:-OrPogV55U-myNv0nLEI:b3hh0ri) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L246)):*
> ```XML
>                     load_volatile_u64(&raw const (*header).data_tail) as usize,
>                 )
>             };
> ```

this code is repeated above. we are lacking some abstractions that would make this code reasonable to read. See also comments below. LLMs are not good at this, it's going to take significant human steering to get this right.



_Code quote:_
```XML
            // SAFETY: `header` is valid for the lifetime of `self`.
            let (head, tail) = unsafe {
                (
                    load_acquire_u64(&raw const (*header).data_head) as usize,
                    load_volatile_u64(&raw const (*header).data_tail) as usize,
                )
            };
```
___
*[`aya/src/maps/perf/perf_buffer.rs` line 357 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrPmpQh5q2GyowHmTiO:-OrPmpQh5q2GyowHmTiP:b-admrsq) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L357)):*
> ```XML
> ///
> /// [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/ring_buffer.c#L113-L114
> fn load_acquire_u64(ptr: *const u64) -> u64 {
> ```

at the caller we have a reference, is there a good reason to pass a pointer here?

also, we always use `load_acquire_u64` to read `data_head` and always use `load_volatile_u64` to read `data_tail`, can we make these into newtypes so we have reasonable type safety? same for store_release_u64.

lastly: why do we need read_volatile? if userspace is the sole writer, can we just plain old read the thing?

Brskt (commented):
Should be ok now ?

> @Brskt made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-OrSMxLaE_zxtYPrF3SJ:b635rae)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 69 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOrcX40zLF5EcFNNal:-OrSNH3J1DOXa-Fvi_nD:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L69)):*

s/crate::maps::perf/super/ ?
</blockquote></details>

Done.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 72 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOrmYd0mVzhuoBNCM_:-OrSNsTM1-koYQUc3Ud8:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L72)):*

instead of saying `PERF_RECORD_SAMPLE` you should use the doc alias attribute, that way a search in the docs for PERF_RECORD_SAMPLE would produce the right result.

Ditto below.
</blockquote></details>

Done.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 97 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOsD6V2lnDZHbosQmP:-OrSNt6q8IYS1vCaISSc:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L97)):*

same
</blockquote></details>

Done.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 103 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOsK46DjlcAV_wNKU0:-OrSONBW6ydktP374H4K:b3inpax) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L103)):*

There should be a comment here indicating this is the lifetime of the PerfBuffer from which this data is owned and also explaining why it isn't written as literally `&'a mut PerfBuffer`.
</blockquote></details>

The struct is gone. `PerfEvent::Sample { head: &'a [u8], tail: &'a [u8] }` carries the lifetime on the slices.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 133 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrOvn5O3Gyu8gfPS4Te:-OrSOaaK43WOnw_9bUr-:b-emtfjx) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L133)):*

do we need this?
</blockquote></details>

Gone with `PerfSample`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 246 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrPogV55U-myNv0nLEH:-OrSPoCfEeloP_ndnY-0:b-6vapo8) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L246)):*

this code is repeated above. we are lacking some abstractions that would make this code reasonable to read. See also comments below. LLMs are not good at this, it's going to take significant human steering to get this right.
</blockquote></details>

Done. Pulled out `data_head`/`data_tail`/`data_tail_ptr`/`data_pages` helpers.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 357 at r5](https://reviewable.io/reviews/aya-rs/aya/1554#-OrPmpQh5q2GyowHmTiO:-OrSU1ml8lyzCpULLE1h:bhfzkvy) ([raw file](https://github.com/aya-rs/aya/blob/55d9e03112e85ac3b051dad5f082e265cd9d3429/aya/src/maps/perf/perf_buffer.rs#L357)):*
> at the caller we have a reference, is there a good reason to pass a pointer here?

Yes for `data_head` (kernel writes it; `&u64` would be UB). No for `data_tail`; folded into a method.

> also, we always use `load_acquire_u64` to read `data_head` and always use `load_volatile_u64` to read `data_tail`, can we make these into newtypes so we have reasonable type safety? same for store\_release\_u64.

Done via typed methods (`data_head`, `data_tail`, `data_tail_ptr` on `PerfBuffer`).

> lastly: why do we need read\_volatile? if userspace is the sole writer, can we just plain old read the thing?

We don't. `data_tail` is now a plain read.

tamird (commented):
> @tamird partially reviewed 3 files, made 4 comments, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 3 of 6 files reviewed, 4 unresolved discussions (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 159 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrTvV-WDUDkxx6IhXED:-OrTvV-WDUDkxx6IhXEE:b-ix3nj8) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L159)):*
> ```XML
>         // for the lifetime of `self`. The kernel writes this field
>         // concurrently, so we must use `read_volatile`; an `&u64` would be UB.
>         let value = unsafe { ptr::read_volatile(&raw const (*self.buf().as_ptr()).data_head) };
> ```

there are two different unsafe operations here, no? can we separate so each can have its own comment?



_Code quote:_
```XML
        // SAFETY: `self.buf()` points to the mmap'd `perf_event_mmap_page`
        // for the lifetime of `self`. The kernel writes this field
        // concurrently, so we must use `read_volatile`; an `&u64` would be UB.
        let value = unsafe { ptr::read_volatile(&raw const (*self.buf().as_ptr()).data_head) };
```
___
*[`aya/src/maps/perf/perf_buffer.rs` line 176 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrTvYnTBzuquD7wWAS2:-OrTvYnTBzuquD7wWAS3:b-55gtd) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L176)):*
> ```XML
>     /// commits the local position with a release-store on drop.
>     fn data_tail_ptr(&mut self) -> NonNull<u64> {
>         // SAFETY: same as `data_head`.
> ```

see above: this doesn't make sense to me. the thing that is happening here is effectively NonNull field projection, which is different from what's happening in data_head.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 244 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrU6-QF7Nq2nKjbe_Ui:-OrU6-QGEF_WH-bbFOiG:b-kmvpu2) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L244)):*
> ```XML
>                         // `TailTracker::drop`).
>                         let head = unsafe { slice::from_raw_parts(base.add(sample_start), first) };
>                         let tail = unsafe { slice::from_raw_parts(base, sample_size - first) };
> ```

in the ring buffer implementation we use a neat trick: mmap the data page twice, which ensures you can always read an event contiguously. It's a bigger change here, but could we do it?



_Code quote:_
```XML
                        let head = unsafe { slice::from_raw_parts(base.add(sample_start), first) };
                        let tail = unsafe { slice::from_raw_parts(base, sample_size - first) };
```
___
*[`aya/src/maps/perf/perf_buffer.rs` line 362 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrU4r151690TWUT4BeD:-OrU4r151690TWUT4BeE:b-nifau) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L362)):*
> ```XML
>     let len = out_buf.len();
>     let end = (start_off + len) % mmap_size;
>     let start = start_off % mmap_size;
> ```

does this mean start_off can be past-the-end? that doesn't seem right to me, there is already mod happening in the caller

Brskt (commented):
> @Brskt made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 159 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrTvV-WDUDkxx6IhXED:-OrUkTVt3gxpuhnXDACS:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L159)):*

there are two different unsafe operations here, no? can we separate so each can have its own comment?
</blockquote></details>

Done.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 176 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrTvYnTBzuquD7wWAS2:-OrV5BP3EQ3k2gNZMQpZ:by1gdx0) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L176)):*

see above: this doesn't make sense to me. the thing that is happening here is effectively NonNull field projection, which is different from what's happening in data_head.
</blockquote></details>

Done. Two `unsafe` blocks now: one for the field projection, one for `NonNull::new_unchecked`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 244 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrU6-QF7Nq2nKjbe_Ui:-OrV2o7F12iu47nVzfzq:b6uerbu) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L244)):*

in the ring buffer implementation we use a neat trick: mmap the data page twice, which ensures you can always read an event contiguously. It's a bigger change here, but could we do it?
</blockquote></details>

Not from userspace. `perf_mmap` rejects any `vm_pgoff` other than 0 or `rb->aux_pgoff` ([core.c](https://github.com/torvalds/linux/blob/v6.13/kernel/events/core.c#L6624)), and `__perf_mmap_to_page` is bounded by `nr_pages` ([ring_buffer.c:805](https://github.com/torvalds/linux/blob/v6.13/kernel/events/ring_buffer.c#L805)). The trick would need a kernel patch to `rb_alloc` + `__perf_mmap_to_page` mirroring `bpf_ringbuf_area_alloc`; libbpf, `tools/lib/perf`, and bcc all do the wrap memcpy for the same reason.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 362 at r8](https://reviewable.io/reviews/aya-rs/aya/1554#-OrU4r151690TWUT4BeD:-OrV4SLl6Iq0FyjtP7qC:b-oewlim) ([raw file](https://github.com/aya-rs/aya/blob/007e56b84b944469d8d181227221b36efddf2ab3/aya/src/maps/perf/perf_buffer.rs#L362)):*

does this mean start_off can be past-the-end? that doesn't seem right to me, there is already mod happening in the caller
</blockquote></details>

Yes. `event_start` is modded but the call sites add `sizeof(header)` (and `sizeof(u64)` for LOST), which can push past `mmap_size`. Moved the mod to both call sites and added `debug_assert!(start < mmap_size)`.

tamird (changes requested):
> @tamird partially reviewed 6 files, made 6 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 1 of 6 files reviewed, 6 unresolved discussions (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 207 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYy5skEyQtMCsjOQcj:-OrYy5skEyQtMCsjOQck:b-nqwxvo) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L207)):*
> ```XML
>         let data_head = self.data_head();
>         let initial_tail = self.data_tail();
>         let mut tracker = TailTracker::new(self.data_tail_ptr(), initial_tail);
> ```

in the ringbuf implementation i used scopeguard::defer, which I found to be easier to reason about because it didn't require bouncing between this function and the Drop impl. do you think that would work here?
___
*[`aya/src/maps/perf/perf_buffer.rs` line 212 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYyjx7DtcCylcELR6A:-OrYyjx7DtcCylcELR6B:b-aoiioh) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L212)):*
> ```XML
>         while tracker.tail != data_head {
>             let event_start = (tracker.tail % mmap_size as u64) as usize;
>             // SAFETY: the kernel guarantees event headers are 8-byte aligned
> ```

would you mind adding a citation for this and perhaps a debug assert?

also if the kernel promises they are aligned, why are we using read_unaligned?
___
*[`aya/src/maps/perf/perf_buffer.rs` line 254 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYzDHW4j2Kl_QEwrkV:-OrYzDHW4j2Kl_QEwrkW:b-irppg7) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L254)):*
> ```XML
>                         let tail = unsafe { slice::from_raw_parts(base, sample_size - first) };
>                         (head, tail)
>                     };
> ```

this code is _very_ similar to the body of `fill_from_mmap`. Extract this so the helper does this (return two slices) and the caller can do the trivial copy into a 4-byte buffer for reading the length?



_Code quote:_
```XML
                    let (head, tail) = if sample_start + sample_size <= mmap_size {
                        // SAFETY: the if-condition guarantees the slice is
                        // entirely within the mmap data area. The kernel will
                        // not overwrite these bytes until `data_tail` is
                        // committed (via `TailTracker::drop`).
                        let s =
                            unsafe { slice::from_raw_parts(base.add(sample_start), sample_size) };
                        (s, &[][..])
                    } else {
                        let first = mmap_size - sample_start;
                        // SAFETY: `[sample_start, mmap_size)` and
                        // `[0, sample_size - first)` are disjoint ranges within
                        // the mmap data area; together they cover the wrapping
                        // sample exactly. The kernel will not overwrite either
                        // range until `data_tail` is committed (via
                        // `TailTracker::drop`).
                        let head = unsafe { slice::from_raw_parts(base.add(sample_start), first) };
                        let tail = unsafe { slice::from_raw_parts(base, sample_size - first) };
                        (head, tail)
                    };
```
___
*[`aya/src/maps/perf/perf_buffer.rs` line 260 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYyb041J0iaymAIfbn:-OrYyb041J0iaymAIfbo:bve52km) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L260)):*
> ```XML
>                     let mut count_buf = [0u8; size_of::<u64>()];
>                     fill_from_mmap(
>                         (event_start + size_of::<perf_event_header>() + size_of::<u64>())
> ```

if i'm reading this right we skip over the header plus 8 bytes and then read 8 bytes. Can you write a comment explaining why please?
___
*[`aya/src/maps/perf/perf_buffer.rs` line 372 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYxmBV-i2OD7-6j3b7:-OrYxmBV-i2OD7-6j3b8:b-rq401y) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L372)):*
> ```XML
> /// `base` must point to a valid mmap region of at least `mmap_size` bytes.
> fn fill_from_mmap(start: usize, base: *const u8, mmap_size: usize, out_buf: &mut [u8]) {
>     debug_assert!(start < mmap_size);
> ```

consider emitting the values in case of mismatch. here and below
___
*[`aya/src/maps/perf/perf_buffer.rs` line 376 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYw-AP1hPC9Byv_b5H:-OrYw-AP1hPC9Byv_b5I:b-gjiaxu) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L376)):*
> ```XML
>     debug_assert!(len <= mmap_size);
> 
>     if start + len <= mmap_size {
> ```

```
if let Some(second) = (start + len).checked_sub(mmap_size) {
    let first = mmap_size - start;
    # split
} else {
    # contiguous
}
```
if you enjoy checked arithmetic

Brskt (commented):
> @Brskt made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 207 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYy5skEyQtMCsjOQcj:-OrZJDslBBH7CBCGijY2:b-r3a9iv) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L207)):*

in the ringbuf implementation i used scopeguard::defer, which I found to be easier to reason about because it didn't require bouncing between this function and the Drop impl. do you think that would work here?
</blockquote></details>

Yes, done. Used `scopeguard::guard` instead of `defer!` since the loop needs to mutate the captured `tail`/`dirty` state.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 212 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYyjx7DtcCylcELR6A:-OrZPj5j2CI2LeVukCJW:b-2yk699) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L212)):*

would you mind adding a citation for this and perhaps a debug assert?

also if the kernel promises they are aligned, why are we using read_unaligned?
</blockquote></details>

Citation and `debug_assert_eq!(event_start % 8, 0)` added. No good reason for `read_unaligned`, switched to `read`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 254 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYzDHW4j2Kl_QEwrkV:-OrZV65k8SqtpS-OqQXg:b-668it1) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L254)):*

this code is _very_ similar to the body of `fill_from_mmap`. Extract this so the helper does this (return two slices) and the caller can do the trivial copy into a 4-byte buffer for reading the length?
</blockquote></details>

Done. Extracted into `slice_from_mmap` returning `(&[u8], &[u8])`; size and count copy into their stack buffers.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 260 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYyb041J0iaymAIfbn:-Or_4jX8031ALrrre5Cc:b-l6ra1u) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L260)):*

if i'm reading this right we skip over the header plus 8 bytes and then read 8 bytes. Can you write a comment explaining why please?
</blockquote></details>

Comment added explaining the skip past `id`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 372 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYxmBV-i2OD7-6j3b7:-Or_5SMm3g8au-oUuwYs:bxrp70u) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L372)):*

consider emitting the values in case of mismatch. here and below
</blockquote></details>

Done. Values now emitted on mismatch.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 376 at r11](https://reviewable.io/reviews/aya-rs/aya/1554#-OrYw-AP1hPC9Byv_b5H:-Or_5VzH0ep6cy-udi08:b1u7uiv) ([raw file](https://github.com/aya-rs/aya/blob/367c3a7497931124a1df6519b58dd15314d7215c/aya/src/maps/perf/perf_buffer.rs#L376)):*

```
if let Some(second) = (start + len).checked_sub(mmap_size) {
    let first = mmap_size - start;
    # split
} else {
    # contiguous
}
```
if you enjoy checked arithmetic
</blockquote></details>

Done. Switched to `checked_sub`.

tamird (commented):
This is looking really good! Thanks for iterating.

> @tamird reviewed 6 files and all commit messages, made 5 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-OrhvZuaALduy-rEXL2j:bb9qgp6)* status: all files reviewed, 4 unresolved discussions (waiting on Brskt).

___
*[`-- commits` line 54 at r15](https://reviewable.io/reviews/aya-rs/aya/1554#-OrhvTUJ8LELrw-deWAl:-OrhvTUJ8LELrw-deWAm:bkaha2z):*
this should probably be squashed into the prior commit since it would otherwise not compile. right?
___
*[`aya/src/maps/perf/perf_buffer.rs` line 226 at r14](https://reviewable.io/reviews/aya-rs/aya/1554#-Orhsklz0efwZjwb87f2:-Orhsklz0efwZjwb87f3:b1rcoqk) ([raw file](https://github.com/aya-rs/aya/blob/37e867e14305d892923457e52cc85843dd0ba9d5/aya/src/maps/perf/perf_buffer.rs#L226)):*
> ```XML
>                 }
>             }
>         });
> ```

i wonder if you could write `let (tail, dirty) = &mut *guard;` here to avoid all the guard.0, guard.1 below
___
*[`aya/src/maps/perf/perf_buffer.rs` line 243 at r14](https://reviewable.io/reviews/aya-rs/aya/1554#-OrhtoCp-0uNA_D38MjX:-OrhtoCp-0uNA_D38MjY:bbwyw2o) ([raw file](https://github.com/aya-rs/aya/blob/37e867e14305d892923457e52cc85843dd0ba9d5/aya/src/maps/perf/perf_buffer.rs#L243)):*
> ```XML
>             // [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/core.c#L8451
>             let event: perf_event_header = unsafe { ptr::read(base.add(event_start).cast()) };
>             let event_size = event.size as usize;
> ```

can you help me understand the relationship between `event_size` and all the reads we do below?

currently we call `slice_from_mmap` twice down there but it seem to me we could do it just once for the whole event (and possibly out here, unconditionally).

that might even allow slice_from_mmap to be inlined here
___
*[`aya/src/maps/perf/perf_buffer.rs` line 297 at r14](https://reviewable.io/reviews/aya-rs/aya/1554#-Orhv3DG5X9L9EWwyXOc:-Orhv3DG5X9L9EWwyXOd:b-m34etc) ([raw file](https://github.com/aya-rs/aya/blob/37e867e14305d892923457e52cc85843dd0ba9d5/aya/src/maps/perf/perf_buffer.rs#L297)):*
> ```XML
>                     }
>                 }
>                 _ => continue,
> ```

this predates you but do you know if we expect other event types here? would be good to write a comment and/or a debug assertion

Brskt (commented):
> @Brskt made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`-- commits` line 54 at r15](https://reviewable.io/reviews/aya-rs/aya/1554#-OrhvTUJ8LELrw-deWAl:-OriQb_S4QkBGhs9UVBF:b-un4k39):*

this should probably be squashed into the prior commit since it would otherwise not compile. right?
</blockquote></details>

Squashed.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 226 at r14](https://reviewable.io/reviews/aya-rs/aya/1554#-Orhsklz0efwZjwb87f2:-OriQeyvB9QOafqTSdYR:b-1eh3wi) ([raw file](https://github.com/aya-rs/aya/blob/37e867e14305d892923457e52cc85843dd0ba9d5/aya/src/maps/perf/perf_buffer.rs#L226)):*

i wonder if you could write `let (tail, dirty) = &mut *guard;` here to avoid all the guard.0, guard.1 below
</blockquote></details>

Sure, done.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 243 at r14](https://reviewable.io/reviews/aya-rs/aya/1554#-OrhtoCp-0uNA_D38MjX:-OriSC1Y7pyCzQSwTsVh:bhcoj8h) ([raw file](https://github.com/aya-rs/aya/blob/37e867e14305d892923457e52cc85843dd0ba9d5/aya/src/maps/perf/perf_buffer.rs#L243)):*

can you help me understand the relationship between `event_size` and all the reads we do below?

currently we call `slice_from_mmap` twice down there but it seem to me we could do it just once for the whole event (and possibly out here, unconditionally).

that might even allow slice_from_mmap to be inlined here
</blockquote></details>

event_size covers the whole record. Inlined the wrap into the SAMPLE arm and dropped `slice_from_mmap`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 297 at r14](https://reviewable.io/reviews/aya-rs/aya/1554#-Orhv3DG5X9L9EWwyXOc:-OrjoaHc95w03b9JoI4U:b-lig3kv) ([raw file](https://github.com/aya-rs/aya/blob/37e867e14305d892923457e52cc85843dd0ba9d5/aya/src/maps/perf/perf_buffer.rs#L297)):*

this predates you but do you know if we expect other event types here? would be good to write a comment and/or a debug assertion
</blockquote></details>

No, only SAMPLE and LOST. Added a comment and `debug_assert!`.

tamird (commented):
Pretty much there, just a few nits and a question about the lost events.

> @tamird reviewed 6 files and all commit messages, made 5 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-OrjqoOS6xXJzoQyaDZs:b-wr7ngo)* status: all files reviewed, 4 unresolved discussions (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 247 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-OrjpuePAFgcJ4LN-Ep2:-OrjpuePAFgcJ4LN-Ep3:bp1go98) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L247)):*
> ```XML
>             let event_size = event.size as usize;
>             let event_type = event.type_;
>             *tail = tail.wrapping_add(event_size as u64);
> ```

what is the type of `event.size`? we're `as usize`ing it above and then `as u64` here
___
*[`aya/src/maps/perf/perf_buffer.rs` line 250 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-Orjq-8uCe8o-DhTTwXg:-Orjq-8uCe8o-DhTTwXh:b-8s1fax) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L250)):*
> ```XML
>             *dirty = true;
> 
>             let perf_event = match event_type {
> ```

consider just event.type_ here and dropping the local variable
___
*[`aya/src/maps/perf/perf_buffer.rs` line 289 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-OrjqUM5CPgcUHKPHKQh:-OrjqUM5CPgcUHKPHKQi:b-7zqphf) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L289)):*
> ```XML
>                     // `{ header, u64 id, u64 lost, sample_id }` [1]; skip past
>                     // `id` to read the `lost` count, which lives at an 8-aligned
>                     // offset and cannot span the wrap.
> ```

is there a citation for this?



_Code quote:_
```XML
                    // `id` to read the `lost` count, which lives at an 8-aligned
                    // offset and cannot span the wrap.
```
___
*[`aya/src/maps/perf/perf_buffer.rs` line 299 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-Orjq_vR417IG_JTNGP8:-Orjq_vR417IG_JTNGP9:busbwnu) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L299)):*
> ```XML
>                     PerfEvent::Lost { count }
>                 }
>                 _ => {
> ```

you can put `event_type` here if you write `match event.type_` above

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: all files reviewed, 5 unresolved discussions (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 216 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-Orjs0ft--3dP292NICX:-Orjs0ft--3dP292NICY:b37yrgj) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L216)):*
> ```XML
>         //
>         // [1]: https://github.com/torvalds/linux/blob/05f7e89a/kernel/events/ring_buffer.c#L202
>         let mut guard = scopeguard::guard((initial_tail, false), |(tail, dirty)| {
> ```

another thought: we don't need `dirty`, do we? we can close over `initial_tail` and replace `if dirty` with `if tail != initial_tail`?

Brskt (commented):
> @Brskt made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on tamird).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 216 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-Orjs0ft--3dP292NICX:-OrmFIKi5NacQz7IyT1A:b-dqtbua) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L216)):*

another thought: we don't need `dirty`, do we? we can close over `initial_tail` and replace `if dirty` with `if tail != initial_tail`?
</blockquote></details>

Correct, we don't need it, removed.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 247 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-OrjpuePAFgcJ4LN-Ep2:-Orm-wyc3JJrLyYbix6w:bo3e3x9) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L247)):*

what is the type of `event.size`? we're `as usize`ing it above and then `as u64` here
</blockquote></details>

`u16`. Inlined as `u64::from(event.size)`.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 250 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-Orjq-8uCe8o-DhTTwXg:-Orm0S3q7R4pPLzH_qqe:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L250)):*

consider just event.type_ here and dropping the local variable
</blockquote></details>

Done.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 289 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-OrjqUM5CPgcUHKPHKQh:-OrmF91P4CVryf2Uf2gW:bqmjwcz) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L289)):*

is there a citation for this?
</blockquote></details>

Added.
___
*[`aya/src/maps/perf/perf_buffer.rs` line 299 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-Orjq_vR417IG_JTNGP8:-OrmF36E8uuuO4G18F14:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L299)):*

you can put `event_type` here if you write `match event.type_` above
</blockquote></details>

Done.

tamird (commented):
> @tamird reviewed 6 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: all files reviewed, 1 unresolved discussion (waiting on Brskt).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 289 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-OrjqUM5CPgcUHKPHKQh:-OrmYI2eAw1UiLJbCFMQ:b-7oxohn) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L289)):*
<details><summary><i>Previously, Brskt wrote…</i></summary><blockquote>

Added.
</blockquote></details>

this is the same citation as used above for the header. sorry to belabor this. can you help me understand why this cannot span the wrap but the sample case above can?

Brskt (commented):
> @Brskt made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/maps/perf/perf_buffer.rs` line 289 at r17](https://reviewable.io/reviews/aya-rs/aya/1554#-OrjqUM5CPgcUHKPHKQh:-Ormatkd82pe5EVjQSm6:b-qvp4mk) ([raw file](https://github.com/aya-rs/aya/blob/1e930017704a10e70aae92fe019190883d752641/aya/src/maps/perf/perf_buffer.rs#L289)):*

this is the same citation as used above for the header. sorry to belabor this. can you help me understand why this cannot span the wrap but the sample case above can?
</blockquote></details>

I'm sorry for the misunderstanding. Fixed 8 bytes at an 8-aligned offset cannot cross an 8-aligned boundary; the sample payload is variable-length and can. Updated citation.

tamird (approved):
@codex review

> @tamird reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1554#-:-Ormfp33CM0XjCuHfTKn:b-fv7n3s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Brskt).

tamird:
Thanks!

tamird:
@Brskt would you mind fixing up the build failure [here](https://github.com/aya-rs/book/actions/runs/25360955976/job/74360275338)?

Brskt:
> @Brskt would you mind fixing up the build failure [here](https://github.com/aya-rs/book/actions/runs/25360955976/job/74360275338)?

Sure, I will take a look.

PR #1555, opened 2026-04-28
test: cover TC across supported kernels

tamird:
Extend TC integration coverage across the existing kernel matrix, including kernels that predate TCX.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1555)


alessandrod (approved):
lgtm

kicked the test because it's flakey downloading some ubuntu stuff

PR #1556, opened 2026-04-29
aya-obj: Remove the `no_std` support

vadorovsky:
The original motivation for `no_std` support in `aya-obj` was to keep it closer to `object`. In practice, though, one of `aya-obj`'s main jobs is sanitizing bytecode and BTF, and that logic relies on dynamic data structures.

Given that, keeping `no_std` support no longer reflects how the crate is actually used and only adds maintenance overhead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1556)

### DIFF
--- a/aya/Cargo.toml
+++ b/aya/Cargo.toml
@@ -18,10 +18,12 @@ workspace = true
 
 [dependencies]
 assert_matches = { workspace = true }
-aya-obj = { path = "../aya-obj", version = "^0.2.2", features = ["std"] }
+aya-obj = { path = "../aya-obj", version = "^0.2.2" }
 bitflags = { workspace = true }
 bytes = { workspace = true }
-hashbrown = { workspace = true }
+# TODO(https://github.com/rust-lang/rust/issues/60896): Remove once
+# `std::collections::hash_set::Entry` is stabilized.
+hashbrown = { workspace = true, features = ["default-hasher", "equivalent"] }
 libc = { workspace = true }
 log = { workspace = true }
 object = { workspace = true, features = ["elf", "read_core", "std", "write"] }
--- a/aya-obj/Cargo.toml
+++ b/aya-obj/Cargo.toml
@@ -18,14 +18,10 @@ workspace = true
 
 [dependencies]
 bytes = { workspace = true }
-hashbrown = { workspace = true, features = ["default-hasher", "equivalent"] }
 log = { workspace = true }
 object = { workspace = true, features = ["elf", "read_core"] }
-thiserror = { workspace = true }
+thiserror = { workspace = true, features = ["std"] }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
 rbpf = { workspace = true }
-
-[features]
-std = ["thiserror/std"]
--- a/aya-obj/src/btf/btf.rs
+++ b/aya-obj/src/btf/btf.rs
@@ -1,14 +1,8 @@
-use alloc::{
+use std::{
     borrow::{Cow, ToOwned as _},
-    format,
-    string::String,
-    vec,
-    vec::Vec,
-};
-use core::{
     cell::OnceCell,
     ffi::{CStr, FromBytesUntilNulError},
-    mem, ptr,
+    format, mem, ptr,
 };
 
 use bytes::BufMut as _;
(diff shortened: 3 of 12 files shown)


aya/Cargo.toml
```diff
@@ -18,10 +18,10 @@ workspace = true
 
 [dependencies]
 assert_matches = { workspace = true }
-aya-obj = { path = "../aya-obj", version = "^0.2.2", features = ["std"] }
+aya-obj = { path = "../aya-obj", version = "^0.2.2" }
 bitflags = { workspace = true }
 bytes = { workspace = true }
-hashbrown = { workspace = true }
+hashbrown = { workspace = true, features = ["default-hasher", "equivalent"] }
```
vadorovsky, line 26: Note: `aya` always needed these `hashbrown` features. The reason why we didn't have to specify them here is that `aya-obj` was enabling them already. In this change, the dependency on `hashbrown` in `aya-obj` is removed, so the features have to be moved here.

`aya` still needs `hashbrown`, because [`Entry` is not yet stable in `std`](https://doc.rust-lang.org/std/collections/hash_set/enum.Entry.html).

tamird: Perhaps leave a comment?

tamird (commented):
> @tamird reviewed 12 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1556)* status: all files reviewed, 8 unresolved discussions (waiting on vadorovsky).

___
*[`test/integration-test/Cargo.toml` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1556#-OrOr6_s9aOAMCZs0TIf:-OrOr6_s9aOAMCZs0TIg:b8llfrk) ([raw file](https://github.com/aya-rs/aya/blob/715ce838ac2cf61b5febcb0e42cf3cada35bc3d2/test/integration-test/Cargo.toml#L23)):*
> ```TOML
> assert_matches = { workspace = true }
> aya-log = { path = "../../aya-log", version = "^0.2.2", default-features = false }
> aya-obj = { path = "../../aya-obj", version = "^0.2.2" }
> ```

why did this need to change?

aya/Cargo.toml
vadorovsky: I don't think it's needed. I left the comment here, on the PR, because the diff might raise a question "why is he adding thee features, if the code remains untouched?".

But adding an inline comment in Cargo.toml would be probably an overkill, unless we intend to comment on all feature additions.

tamird: This is different from other features because the justification is "XXX is unstable". That means the comment takes the shape of a TODO.

vadorovsky: OK, good call, added.

vadorovsky (commented):
> @vadorovsky made 2 comments and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1556)* status: all files reviewed, 3 unresolved discussions (waiting on tamird).

___
*[`test/integration-test/Cargo.toml` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1556#-OrOr6_s9aOAMCZs0TIf:-OrOuIOf72YGK78el__5:b-ur4i5k) ([raw file](https://github.com/aya-rs/aya/blob/715ce838ac2cf61b5febcb0e42cf3cada35bc3d2/test/integration-test/Cargo.toml#L23)):*

why did this need to change?
</blockquote></details>

After this change, `aya-obj` has no features, so `default-features = false` is redundant.

tamird (approved):
:lgtm:

> @tamird reviewed 8 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1556#-:-OrXqMF_B3YJHVIUNzIJ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on vadorovsky).

___
*[`aya/Cargo.toml` line 24 at r2](https://reviewable.io/reviews/aya-rs/aya/1556#-OrXq7wO2Ethwza6t1VS:-OrXq7wPDzav91NGlZVg:b-k8ni1t) ([raw file](https://github.com/aya-rs/aya/blob/235a98830b14f24252d12d7f00ac77372f8aeb8b/aya/Cargo.toml#L24)):*
> ```TOML
> bitflags = { workspace = true }
> bytes = { workspace = true }
> # TODO: Remove once `std::collections::hash_set::Entry` is stabilized.
> ```

please TODO(https://github.com/rust-lang/rust/issues/60896): ...

tamird (approved):
> @tamird reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1556)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on vadorovsky).

Issue #441, opened 2022-11-16
attach() for SocketFilter needs an enum


dave-tucker:
SocketFilter programs can be attached to a socket with two socket options:
- SO_ATTACH_BPF
- SO_ATTACH_REUSEPORT_EBPF

See: man socket(7)

We should add an enum to `attach()` for SocketFilter programs so the caller can pick the attach type they want.

The latter program type MUST return an integer - we should ensure that the SocketFilter program in aya-bpf returns one (i32 or u32 probably) and verify the behaviour of the return code in both modes of attachment so we can document any verifier hiccups. For example, int returns could either be ignored when attached with SO_ATTACH_BPF or the verifier will get upset.

bboyleonp666:
Hi @dave-tucker, could you provide me more insights on this one? I have searched for the `SO_ATTACH_BPF` and `SO_ATTACH_REUSEPORT_EBPF`. It seems the way aya handles attach point enum is to put them in some `binding.rs` files. I found that they are in different directories like `ebpf/aya-ebpf-bindings`, `aya-obj/src/generated`, `aya-obj/src/generated`. It get me confused where I shoud put the mappings.

mahmudsudo:
is this available to be taken ?

femiagbabiaka:
@dave-tucker 
Working this issue, (need it for an experiment) had a few comments/questions:
- Adding an enum param to `SocketFilter::attach()` is a breaking change, but a simple one. I made it and updated the doc test, which passes. I ran integration tests but noticed there's no integration test covering socket programs yet. I'm looking at adding one.
- The original suggestion for this issue was to add an enum for SocketFilter, but I noticed, for example, in the feature_probe integration test, that we check to see if a particular feature is available for a particular kernel version. SocketFilter was available in 3.19, but REUSEPORT_EBPF wasn't available until 4.5. Is breaking out this functionality something that makes sense? Also avoids the breaking change to the existing API.
- It was mentioned that the SocketFilter program needed to be updated to return a return code. That totally makes sense. I think in this case you're referring to the macro in aya-ebpf-macros? Right now it returns an i64, it looks like you made that decision, wonder if there are other load-bearing assumptions there that I'm missing?

Current work here: https://github.com/femiagbabiaka/aya/tree/femi/socket-reuseport-ebpf

femiagbabiaka:
[...]

PR #1563, opened 2026-05-07
aya: remove SocketFilter link abstraction

swananan:
While working on #441, I found that SocketFilter's link abstraction was incomplete. During review, we also found a deeper issue: the kernel models socket filters as a single per-socket filter slot, so a link-level RAII abstraction is not suitable for SocketFilter.

This PR removes the SocketFilter link abstraction and switches to explicit detach semantics, matching the kernel behavior more closely.

### Added/updated tests?

- [x] Yes

### Checklist

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)

### DIFF
--- a/aya/src/programs/mod.rs
+++ b/aya/src/programs/mod.rs
@@ -819,9 +819,10 @@ macro_rules! impl_program_unload {
             impl $struct_name {
                 /// Unloads the program from the kernel.
                 ///
-                /// Links will be detached before unloading the program.  Note
-                /// that owned links obtained using `take_link()` will not be
-                /// detached.
+                /// Tracked links will be detached before unloading the program.
+                /// Attachment mechanisms that do not create tracked links are
+                /// not affected. Note that owned links obtained using
+                /// `take_link()` will not be detached.
                 pub fn unload(&mut self) -> Result<(), ProgramError> {
                     unload_program(&mut self.data)
                 }
@@ -1263,7 +1264,6 @@ macro_rules! impl_from_pin {
 // Use impl_from_pin if the program doesn't require additional data
 impl_from_pin!(
     TracePoint,
-    SocketFilter,
     SkMsg,
     CgroupSysctl,
     LircMode2,
--- a/aya/src/programs/sk_reuseport.rs
+++ b/aya/src/programs/sk_reuseport.rs
@@ -171,7 +171,7 @@ impl SkReuseport {
     /// operation. Dropping the program object does not detach it; call
     /// [`SkReuseport::detach`] with any socket from the same group to remove it
     /// again, or close all sockets in the group.
-    pub fn attach<T: AsFd>(&mut self, socket: T) -> Result<(), ProgramError> {
+    pub fn attach<T: AsFd>(&self, socket: T) -> Result<(), ProgramError> {
         let prog_fd = self.fd()?.as_fd().as_raw_fd();
         let socket = socket.as_fd().as_raw_fd();
 
@@ -186,7 +186,7 @@ impl SkReuseport {
     /// from the entire group, regardless of which socket in that group was
     /// used to attach it. Unlike [`SkReuseport::attach`], this operation does
     /// not require the program to remain loaded in this process.
-    pub fn detach<T: AsFd>(&mut self, socket: T) -> Result<(), ProgramError> {
+    pub fn detach<T: AsFd>(socket: T) -> Result<(), ProgramError> {
         let socket = socket.as_fd().as_raw_fd();
         let dummy: libc::c_int = 0;
 
--- a/aya/src/programs/socket_filter.rs
+++ b/aya/src/programs/socket_filter.rs
@@ -1,7 +1,9 @@
 //! Socket filter programs.
 use std::{
     io,
-    os::fd::{AsFd, AsRawFd as _, RawFd},
+    os::fd::{AsFd, AsRawFd as _},
+    path::Path,
+    ptr,
 };
 
 use aya_obj::generated::{
@@ -10,17 +12,45 @@ use aya_obj::generated::{
 use libc::{SOL_SOCKET, setsockopt};
 use thiserror::Error;
 
-use crate::programs::{
-    Link, ProgramData, ProgramError, ProgramType, id_as_key, load_program_without_attach_type,
+use crate::{
+    VerifierLogLevel,
+    programs::{
+        ProgramData, ProgramError, ProgramType, links::FdLink, load_program_without_attach_type,
+    },
 };
 
-/// The type returned when attaching a [`SocketFilter`] fails.
+macro_rules! setsockopt_socket_filter {
+    ($socket:expr, $option:ident, $value:expr) => {{
+        let value = $value;
+        let ret = unsafe {
+            setsockopt(
+                $socket,
+                SOL_SOCKET,
+                $option as libc::c_int,
+                ptr::from_ref(value).cast(),
+                size_of_val(value) as libc::socklen_t,
+            )
+        };
+        if ret < 0 {
+            Err(SocketFilterError::SetsockoptError {
+                option: stringify!($option),
+                io_error: io::Error::last_os_error(),
+            })
+        } else {
+            Ok(())
+        }
+    }};
+}
+
+/// The type returned when a [`SocketFilter`] socket option operation fails.
 #[derive(Debug, Error)]
 pub enum SocketFilterError {
-    /// Setting the `SO_ATTACH_BPF` socket option failed.
-    #[error("setsockopt SO_ATTACH_BPF failed")]
-    SoAttachEbpfError {
-        /// original [`io::Error`]
+    /// Setting a socket filter socket option failed.
+    #[error("setsockopt {option} failed")]
+    SetsockoptError {
+        /// Socket option passed to `setsockopt`.
+        option: &'static str,
+        /// Underlying OS error.
         #[source]
         io_error: io::Error,
     },
@@ -31,9 +61,21 @@ pub enum SocketFilterError {
 /// [`SocketFilter`] programs are attached on sockets and can be used to inspect
 /// and filter incoming packets.
 ///
+/// Each socket has one filter slot. Attaching a new [`SocketFilter`] replaces
+/// the socket's current filter, and detaching clears that current filter
+/// regardless of which program installed it. Aya therefore does not expose a
+/// link-style attachment handle for [`SocketFilter`] or automatically track
+/// socket filter attachments for cleanup. Dropping [`SocketFilter`] or
+/// [`crate::Ebpf`] does not detach the filter; call [`SocketFilter::detach`]
+/// explicitly when you want to remove it, or close the socket.
+///
 /// # Minimum kernel version
 ///
-/// The minimum kernel version required to use this feature is 4.0.
+/// The minimum kernel version required to use this feature is 3.19.
+/// `BPF_PROG_TYPE_SOCKET_FILTER` and `SO_ATTACH_BPF` are present in Linux
+/// v3.19:
+/// <https://github.com/torvalds/linux/blob/v3.19/include/uapi/linux/bpf.h#L118-L120>
+/// <https://github.com/torvalds/linux/blob/v3.19/include/uapi/asm-generic/socket.h#L87-L88>
 ///
 /// # Examples
 ///
@@ -62,7 +104,7 @@ pub enum SocketFilterError {
 #[derive(Debug)]
 #[doc(alias = "BPF_PROG_TYPE_SOCKET_FILTER")]
 pub struct SocketFilter {
-    pub(crate) data: ProgramData<SocketFilterLink>,
+    pub(crate) data: ProgramData<FdLink>,
 }
 
 impl SocketFilter {
@@ -77,82 +119,53 @@ impl SocketFilter {
 
     /// Attaches the filter on the given socket.
     ///
-    /// The returned value can be used to detach from the socket, see [`SocketFilter::detach`].
-    pub fn attach<T: AsFd>(&mut self, socket: T) -> Result<SocketFilterLinkId, ProgramError> {
+    /// If the socket already has a filter attached, attaching again replaces
+    /// the current filter instead of returning an already-attached error. This
+    /// follows the kernel model: each socket has one filter slot and cannot run
+    /// multiple socket filters together. The kernel detach API also clears the
+    /// socket's current filter slot; it cannot detach a specific program
+    /// attachment. For that reason, Aya does not provide link-level RAII
+    /// semantics for socket filters. Dropping [`SocketFilter`] or [`crate::Ebpf`]
+    /// does not detach the filter. Call [`SocketFilter::detach`] explicitly when
+    /// you want to remove it, or close the socket.
+    pub fn attach<T: AsFd>(&self, socket: T) -> Result<(), ProgramError> {
         let prog_fd = self.fd()?;
-        let prog_fd = prog_fd.as_fd();
-        let prog_fd = prog_fd.as_raw_fd();
-        let socket = socket.as_fd();
-        let socket = socket.as_raw_fd();
+        let prog_fd = prog_fd.as_fd().as_raw_fd();
+        let socket = socket.as_fd().as_raw_fd();
 
-        let ret = unsafe {
-            setsockopt(
-                socket,
-                SOL_SOCKET,
-                SO_ATTACH_BPF as i32,
-                std::ptr::from_ref(&prog_fd).cast(),
-                size_of_val(&prog_fd) as u32,
-            )
-        };
-        if ret < 0 {
-            return Err(SocketFilterError::SoAttachEbpfError {
-                io_error: io::Error::last_os_error(),
-            }
-            .into());
-        }
-
-        self.data.links.insert(SocketFilterLink { socket, prog_fd })
-    }
-
-    /// Detaches the program.
-    ///
-    /// See [`Self::attach`].
-    pub fn detach(&mut self, link_id: SocketFilterLinkId) -> Result<(), ProgramError> {
-        self.data.links.remove(link_id)
-    }
-
-    /// Takes ownership of the link referenced by the provided `link_id`.
-    ///
-    /// The caller takes the responsibility of managing the lifetime of the link. When the returned
-    /// [`SocketFilterLink`] is dropped, the link is detached.
-    pub fn take_link(
-        &mut self,
-        link_id: SocketFilterLinkId,
-    ) -> Result<SocketFilterLink, ProgramError> {
-        self.data.links.forget(link_id)
-    }
-}
-
-/// The type returned by [`SocketFilter::attach`]. Can be passed to [`SocketFilter::detach`].
-#[derive(Debug, Hash, Eq, PartialEq)]
-pub struct SocketFilterLinkId(RawFd, RawFd);
-
-/// A [`SocketFilter`] Link.
-#[derive(Debug)]
-pub struct SocketFilterLink {
-    socket: RawFd,
-    prog_fd: RawFd,
-}
-
-impl Link for SocketFilterLink {
-    type Id = SocketFilterLinkId;
-
-    fn id(&self) -> Self::Id {
-        SocketFilterLinkId(self.socket, self.prog_fd)
-    }
-
-    fn detach(self) -> Result<(), ProgramError> {
-        unsafe {
-            setsockopt(
-                self.socket,
-                SOL_SOCKET,
-                SO_DETACH_BPF as i32,
-                std::ptr::from_ref(&self.prog_fd).cast(),
-                size_of_val(&self.prog_fd) as u32,
-            );
-        }
+        setsockopt_socket_filter!(socket, SO_ATTACH_BPF, &prog_fd)?;
         Ok(())
     }
-}
 
-id_as_key!(SocketFilterLink, SocketFilterLinkId);
+    /// Detaches the current filter from the given socket.
+    ///
+    /// Detaching clears the socket's current filter slot, regardless of which
+    /// program was used to attach that filter. Unlike [`SocketFilter::attach`],
+    /// this operation does not require the program to remain loaded in this
+    /// process. If another filter replaced this program on the same socket,
+    /// detaching will remove that replacement filter.
+    pub fn detach<T: AsFd>(socket: T) -> Result<(), ProgramError> {
+        let socket = socket.as_fd().as_raw_fd();
+
+        // `SO_DETACH_BPF` is an alias for `SO_DETACH_FILTER`; Linux's
+        // `sk_setsockopt()` requires an `int` optval, but the detach branch only
+        // calls `sk_detach_filter(sk)` and does not use a program fd:
+        // https://github.com/torvalds/linux/blob/v6.9/net/core/sock.c#L1413-L1414
+        let dummy: libc::c_int = 0;
+        setsockopt_socket_filter!(socket, SO_DETACH_BPF, &dummy)?;
+        Ok(())
+    }
+
+    /// Creates a program from a pinned entry on a bpffs.
+    ///
+    /// `SocketFilter` does not use link-style attachments, so this only
+    /// restores access to the pinned program itself.
+    ///
+    /// Dropping the returned value unloads the local program FD, but does not
+    /// detach the filter from any socket. This will also not unload the program
+    /// from the kernel while it remains pinned.
+    pub fn from_pin<P: AsRef<Path>>(path: P) -> Result<Self, ProgramError> {
+        let data = ProgramData::from_pinned_path(path, VerifierLogLevel::default())?;
+        Ok(Self { data })
+    }
+}
--- a/test/integration-common/src/lib.rs
+++ b/test/integration-common/src/lib.rs
@@ -164,8 +164,16 @@ pub mod sk_reuseport {
     pub const SELECT_HITS_INDEX: u32 = 0;
     pub const MIGRATE_HITS_INDEX: u32 = 1;
     pub const CLEAR_FALLBACK_HITS_INDEX: u32 = 2;
+    pub const PATH_HITS_MAX_ENTRIES: u32 = 3;
     pub const SELECT_SOCKET_INDEX: u32 = 0;
     pub const MIGRATE_SOCKET_INDEX: u32 = 2;
+}
+
+pub mod socket_filter {
+    pub const PASS_HITS_INDEX: u32 = 0;
+    pub const TRIM_HITS_INDEX: u32 = 1;
+    pub const PATH_HITS_MAX_ENTRIES: u32 = 2;
+    pub const TRIM_DELTA_BYTES: u32 = 4;
 }
 
 pub mod stack_trace {
--- a/test/integration-ebpf/Cargo.toml
+++ b/test/integration-ebpf/Cargo.toml
@@ -121,6 +121,10 @@ name = "sk_reuseport"
 path = "src/sk_reuseport.rs"
 
 [[bin]]
+name = "socket_filter"
+path = "src/socket_filter.rs"
+
+[[bin]]
 name = "strncmp"
 path = "src/strncmp.rs"
 
--- a/test/integration-ebpf/src/sk_reuseport.rs
+++ b/test/integration-ebpf/src/sk_reuseport.rs
@@ -10,8 +10,8 @@ use aya_ebpf::{
     programs::SkReuseportContext,
 };
 use integration_common::sk_reuseport::{
-    CLEAR_FALLBACK_HITS_INDEX, MIGRATE_HITS_INDEX, MIGRATE_SOCKET_INDEX, SELECT_HITS_INDEX,
-    SELECT_SOCKET_INDEX,
+    CLEAR_FALLBACK_HITS_INDEX, MIGRATE_HITS_INDEX, MIGRATE_SOCKET_INDEX, PATH_HITS_MAX_ENTRIES,
+    SELECT_HITS_INDEX, SELECT_SOCKET_INDEX,
 };
 #[cfg(not(test))]
 extern crate ebpf_panic;
@@ -30,7 +30,7 @@ static SOCKET_MAP_BTF: BtfReusePortSockArray<SOCKET_COUNT_USIZE, 0> = BtfReusePo
 #[map(name = "path_hits")]
 // Test-only counters used to verify that the expected BPF path actually ran,
 // instead of the kernel falling back to its default reuseport selection logic.
-static PATH_HITS: Array<u64> = Array::with_max_entries(3, 0);
+static PATH_HITS: Array<u64> = Array::with_max_entries(PATH_HITS_MAX_ENTRIES, 0);
 
 #[inline]
 fn record_hit(index: u32) {
--- a/test/integration-ebpf/src/socket_filter.rs
+++ b/test/integration-ebpf/src/socket_filter.rs
@@ -1,0 +1,40 @@
+#![no_std]
+#![no_main]
+#![expect(unused_crate_dependencies, reason = "used in other bins")]
+
+use aya_ebpf::{
+    macros::{map, socket_filter},
+    maps::Array,
+    programs::SkBuffContext,
+};
+use integration_common::socket_filter::{
+    PASS_HITS_INDEX, PATH_HITS_MAX_ENTRIES, TRIM_DELTA_BYTES, TRIM_HITS_INDEX,
+};
+#[cfg(not(test))]
+extern crate ebpf_panic;
+
+#[map(name = "path_hits")]
+static PATH_HITS: Array<u64> = Array::with_max_entries(PATH_HITS_MAX_ENTRIES, 0);
+
+#[inline]
+fn record_hit(index: u32) {
+    let Some(hit) = PATH_HITS.get_ptr_mut(index) else {
+        return;
+    };
+
+    unsafe {
+        *hit += 1;
+    }
+}
+
+#[socket_filter]
+fn pass_packets(ctx: SkBuffContext) -> i64 {
+    record_hit(PASS_HITS_INDEX);
+    i64::from(ctx.len())
+}
+
+#[socket_filter]
+fn trim_packets(ctx: SkBuffContext) -> i64 {
+    record_hit(TRIM_HITS_INDEX);
+    i64::from(ctx.len().saturating_sub(TRIM_DELTA_BYTES))
+}
--- a/test/integration-test/src/lib.rs
+++ b/test/integration-test/src/lib.rs
@@ -61,6 +61,7 @@ bpf_file!(
     RING_BUF => "ring_buf",
     SIMPLE_PROG => "simple_prog",
     SK_REUSEPORT => "sk_reuseport",
+    SOCKET_FILTER => "socket_filter",
     SK_STORAGE => "sk_storage",
     SOCK_HASH => "sock_hash",
     SOCK_MAP => "sock_map",
--- a/test/integration-test/src/tests.rs
+++ b/test/integration-test/src/tests.rs
@@ -39,6 +39,7 @@ mod sk_lookup;
 mod sk_reuseport;
 mod sk_storage;
 mod smoke;
+mod socket_filter;
 mod stack_trace;
 mod stack_trace_lsm;
 mod strncmp;
--- a/test/integration-test/src/tests/sk_reuseport.rs
+++ b/test/integration-test/src/tests/sk_reuseport.rs
@@ -187,15 +187,13 @@ async fn sk_reuseport_selects_expected_listener(socket_map: &str, select_prog: &
     assert_eq!(read_hits(&path_hits, CLEAR_FALLBACK_HITS_INDEX), 0);
     assert_connection_works(client, server).await;
 
-    let prog: &mut SkReuseport = ebpf.program_mut(select_prog).unwrap().try_into().unwrap();
-
     // `SkReuseport` attachments are group-scoped: detaching through any socket in
     // the reuseport group removes the program from the whole group, even if that
     // socket was not used for attach, so a second detach through listener A yields
     // ENOENT.
-    prog.detach(&second).unwrap();
+    SkReuseport::detach(&second).unwrap();
 
-    let err = prog.detach(&first).unwrap_err();
+    let err = SkReuseport::detach(&first).unwrap_err();
     assert_matches!(
         err,
         ProgramError::SkReuseportError(SkReuseportError::SetsockoptError {
@@ -267,7 +265,7 @@ async fn sk_reuseport_clear_index_changes_selection(socket_map: &str, clear_prog
 #[test_case("select_socket"; "legacy")]
 #[test_case("select_socket_btf"; "btf")]
 #[test_log::test(tokio::test)]
-async fn sk_reuseport_detaches_after_unload(select_prog: &str) {
+async fn sk_reuseport_stays_attached_until_explicit_detach(select_prog: &str) {
     let _netns = NetNsGuard::new();
 
     let mut ebpf = Ebpf::load(crate::SK_REUSEPORT).unwrap();
@@ -276,14 +274,14 @@ async fn sk_reuseport_detaches_after_unload(select_prog: &str) {
     let prog: &mut SkReuseport = ebpf.program_mut(select_prog).unwrap().try_into().unwrap();
     prog.load().unwrap();
     prog.attach(&first).unwrap();
-    prog.unload().unwrap();
+    drop(ebpf);
 
     // `SO_DETACH_REUSEPORT_BPF` identifies the group solely from the
-    // socket, so detaching should still succeed after the local program FD
-    // has been unloaded.
-    prog.detach(&second).unwrap();
+    // socket, so detaching should still succeed after the local program has
+    // been unloaded.
+    SkReuseport::detach(&second).unwrap();
 
-    let err = prog.detach(&first).unwrap_err();
+    let err = SkReuseport::detach(&first).unwrap_err();
     assert_matches!(
         err,
         ProgramError::SkReuseportError(SkReuseportError::SetsockoptError {
--- a/test/integration-test/src/tests/socket_filter.rs
+++ b/test/integration-test/src/tests/socket_filter.rs
@@ -1,0 +1,203 @@
+use std::time::Duration;
+
+use aya::{
+    Ebpf,
+    maps::{Array, MapData},
+    programs::{ProgramType, SocketFilter},
+    sys::is_program_supported,
+};
+use integration_common::socket_filter::{PASS_HITS_INDEX, TRIM_DELTA_BYTES, TRIM_HITS_INDEX};
+use tokio::{net::UnixDatagram, time::timeout};
+
+const IO_TIMEOUT: Duration = Duration::from_secs(10);
+const TEST_PAYLOAD: &[u8] = b"hello-world";
+// Leave room for one extra byte so an unexpectedly long datagram is not
+// truncated to the expected length.
+const RECV_BUF_LEN: usize = TEST_PAYLOAD.len() + 1;
+
+fn read_hits(hits: &Array<MapData, u64>, index: u32) -> u64 {
+    hits.get(&index, 0).unwrap()
+}
+
+async fn send_and_assert(sender: &UnixDatagram, receiver: &UnixDatagram, trim_bytes: usize) {
+    sender
+        .send(TEST_PAYLOAD)
+        .await
+        .expect("failed to send datagram");
+
+    let mut buf = [0u8; RECV_BUF_LEN];
+    let len = timeout(IO_TIMEOUT, receiver.recv(&mut buf))
+        .await
+        .expect("timed out waiting for datagram")
+        .expect("failed to receive datagram");
+    let expected_len = TEST_PAYLOAD.len().saturating_sub(trim_bytes);
+    let expected = &TEST_PAYLOAD[..expected_len];
+    assert_eq!(&buf[..len], expected);
+}
+
+#[test_log::test(tokio::test)]
+async fn socket_filter_program_can_pass_packets() {
+    if !is_program_supported(ProgramType::SocketFilter).unwrap() {
+        eprintln!("skipping test - socket_filter program not supported");
+        return;
+    }
+
+    let (sender, receiver) = UnixDatagram::pair().unwrap();
+
+    let mut ebpf = Ebpf::load(crate::SOCKET_FILTER).unwrap();
+    let path_hits: Array<_, u64> = ebpf.take_map("path_hits").unwrap().try_into().unwrap();
+    let prog: &mut SocketFilter = ebpf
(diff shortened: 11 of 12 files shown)


tamird (commented):
Nice catch. Just a few comments.

> @tamird reviewed 9 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563#-:-Os1ZAk7CYNVY0vSEefg:b-i4uceq)* status: all files reviewed, 5 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 98 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XAta5GgiVnz0dfp6:-Os1XAtb6FaICa7pkKEC:b-q84k03) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/aya/src/programs/socket_filter.rs#L98)):*
> ```XML
>         if self.data.links.contains_id(&link_id) {
>             return Err(ProgramError::AlreadyAttached);
>         }
> ```

this sounds like the better approach would be to defer creating the RAII link until the link is successfully created in the kernel. is there a reason that is infeasible?



_Code quote:_
```XML
        // The kernel allows installing the same socket filter program on the
        // same socket again, but the socket still only has one filter slot.
        // Aya represents that as one link. With the current link
        // implementation, a duplicate attach would create a rejected RAII
        // link and dropping it would detach the socket's current filter, so
        // reject duplicates before the second install reaches the socket.
        if self.data.links.contains_id(&link_id) {
            return Err(ProgramError::AlreadyAttached);
        }
```
___
*[`test/integration-ebpf/src/socket_filter.rs` line 15 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XQAq95NBzZw2Io_W:-Os1XQAq95NBzZw2Io_X:b-b1gywg) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-ebpf/src/socket_filter.rs#L15)):*
> ```XML
> 
> #[map(name = "path_hits")]
> static PATH_HITS: Array<u64> = Array::with_max_entries(2, 0);
> ```

this "2" is a magic number that should be a constant next to the indexes. agree?
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os1YR2_2292tqDG_KEY:bgbah4x) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*
> ```XML
> async fn recv_payload(socket: &UdpSocket) -> Vec<u8> {
>     let mut buf = [0u8; 64];
>     let (len, _) = timeout(IO_TIMEOUT, socket.recv_from(&mut buf))
> ```

i think you can use try_recv since you discard the sender address immediately
___
*[`test/integration-test/src/tests/socket_filter.rs` line 39 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XyXMBJn5WuOwtGJp:-Os1XyXN-FwRuw1XZdKQ:b-hxujjk) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L39)):*
> ```XML
>     let _netns = NetNsGuard::new();
>     let receiver = UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).await.unwrap();
>     let sender = UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).await.unwrap();
> ```

ths works of course, but I wonder: could we use a socketpair? https://docs.rs/nix/latest/nix/sys/socket/fn.socketpair.html



_Code quote:_
```XML
    let _netns = NetNsGuard::new();
    let receiver = UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).await.unwrap();
    let sender = UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).await.unwrap();
```
___
*[`test/integration-test/src/tests/socket_filter.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YWKj0bwpozCY9NJE:-Os1YWKj0bwpozCY9NJF:b5igywx) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L52)):*
> ```XML
> 
>     sender
>         .send_to(b"pass-packet", receiver.local_addr().unwrap())
> ```

let payload = b"pass-socket";
sender.send_to(.....);
let buf = [0; payload.len()+1];
let len = ...;
assert_eq!(buf[..len], payload);

this avoids the magic number 64 in `recv_payload`. This whole thing can be a function of course (it would have to use const generic, but should work).

swananan (commented):
> @swananan made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: all files reviewed, 5 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 98 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XAta5GgiVnz0dfp6:-Os1hUj38FLxcyrVIanI:bjl1zau) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/aya/src/programs/socket_filter.rs#L98)):*

this sounds like the better approach would be to defer creating the RAII link until the link is successfully created in the kernel. is there a reason that is infeasible?
</blockquote></details>

In my understanding, there isn't a separate step where we can first reserve the link slot and only then construct the RAII object, unless we change insert API.
___
*[`test/integration-ebpf/src/socket_filter.rs` line 15 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XQAq95NBzZw2Io_W:-Os1dOY1BRhEhwetvvvn:bgnh37b) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-ebpf/src/socket_filter.rs#L15)):*

this "2" is a magic number that should be a constant next to the indexes. agree?
</blockquote></details>

Agreed, I'll avoid magic numbers like this in future test code (:
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os1ds7N9WsQwy8NWgmO:b-wcigor) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

i think you can use try_recv since you discard the sender address immediately
</blockquote></details>

Switched to a socket pair, so this now uses `recv` instead.
___
*[`test/integration-test/src/tests/socket_filter.rs` line 39 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XyXMBJn5WuOwtGJp:-Os1e3pcAcX5unG9P7QP:b-gq7oj9) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L39)):*

ths works of course, but I wonder: could we use a socketpair? https://docs.rs/nix/latest/nix/sys/socket/fn.socketpair.html
</blockquote></details>

That makes sense, so we can removed the NetNsGuard.
___
*[`test/integration-test/src/tests/socket_filter.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YWKj0bwpozCY9NJE:-Os1eKZY7fZAsTRBfFPX:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L52)):*

let payload = b"pass-socket";
sender.send_to(.....);
let buf = [0; payload.len()+1];
let len = ...;
assert_eq!(buf[..len], payload);

this avoids the magic number 64 in `recv_payload`. This whole thing can be a function of course (it would have to use const generic, but should work).
</blockquote></details>

Done.

tamird (changes requested):
> @tamird reviewed 4 files and all commit messages, made 3 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: all files reviewed, 3 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 98 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XAta5GgiVnz0dfp6:-Os1nQsFB54nqPVpeMx7:b-9vc594) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/aya/src/programs/socket_filter.rs#L98)):*

In my understanding, there isn't a separate step where we can first reserve the link slot and only then construct the RAII object, unless we change insert API.
</blockquote></details>

Well, can we change the insert API?
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os1n4b1AdQyPa2dxZlI:b-b6cb1q) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

Switched to a socket pair, so this now uses `recv` instead.
</blockquote></details>

I think payload can be `[u8; N]` with a const generic to avoid the vector allocation and the assert_eq can be right here in this helper?
___
*[`test/integration-common/src/lib.rs` line 164 at r2](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1my4E-tCjyvTgwD18:-Os1my4E-tCjyvTgwD19:b-gmugy3) ([raw file](https://github.com/aya-rs/aya/blob/e0fb91e8da8d3a40f62788522f4e282b622d2d02/test/integration-common/src/lib.rs#L164)):*
> ```XML
> 
> pub mod sk_reuseport {
>     pub const PATH_HITS_MAX_ENTRIES: u32 = 3;
> ```

should come after indexes so the value choice is obvious

ditto below

swananan (commented):
> @swananan made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: all files reviewed, 3 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 98 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1XAta5GgiVnz0dfp6:-Os1wIRR2ABfLzj-BzSi:bojkzfk) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/aya/src/programs/socket_filter.rs#L98)):*

Well, can we change the insert API?
</blockquote></details>

I added a new `insert_with` API that takes a closure, since I didn't want to change `insert` directly and impact all existing callers.
___
*[`test/integration-common/src/lib.rs` line 164 at r2](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1my4E-tCjyvTgwD18:-Os1tfQ04XslZf5U0qm2:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/e0fb91e8da8d3a40f62788522f4e282b622d2d02/test/integration-common/src/lib.rs#L164)):*

should come after indexes so the value choice is obvious

ditto below
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os1tZhmDK399x6mhO0Y:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

I think payload can be `[u8; N]` with a const generic to avoid the vector allocation and the assert_eq can be right here in this helper?
</blockquote></details>

Done.

tamird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 6 of 10 files reviewed, 4 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/links.rs` line 105 at r3](https://reviewable.io/reviews/aya-rs/aya/1563#-Os24U1G2PEMxpUDuSLs:-Os24U1G2PEMxpUDuSLt:b-fel9rp) ([raw file](https://github.com/aya-rs/aya/blob/b4e0b27d325c52c9b3b9ae82cc79301c2b6554e2/aya/src/programs/links.rs#L105)):*
> ```XML
>         let inserted = self.links.insert(link);
>         debug_assert!(inserted);
>         Ok(link_id)
> ```

the `insert` function above already uses the entry API, which I think is what you want here. In fact, I think we should make this the API of the insert function always - to avoid exactly this bug.



_Code quote:_
```XML
        if self.links.contains(&link_id) {
            return Err(ProgramError::AlreadyAttached);
        }

        let link = make_link()?;
        use hashbrown::Equivalent as _;

        debug_assert!(link_id.equivalent(&link));
        let inserted = self.links.insert(link);
        debug_assert!(inserted);
        Ok(link_id)
```

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/links.rs` line 105 at r3](https://reviewable.io/reviews/aya-rs/aya/1563#-Os24U1G2PEMxpUDuSLs:-Os2IZpSDquZR5L5X7qs:b-k2vjb) ([raw file](https://github.com/aya-rs/aya/blob/b4e0b27d325c52c9b3b9ae82cc79301c2b6554e2/aya/src/programs/links.rs#L105)):*

the `insert` function above already uses the entry API, which I think is what you want here. In fact, I think we should make this the API of the insert function always - to avoid exactly this bug.
</blockquote></details>

 I tried using the entry API directly, but `HashSet::entry` needs the full link value up front. That means the RAII link would already have been created before duplicate detection, unless we changed the storage from `HashSet` to `HashMap`.

Kept the existing `HashSet` storage and used the borrowed `contains` lookup by link id before creating the link. This adds an extra hash table lookup, but attach is not a hot path and the lookup cost should be negligible compared with the kernel attach operation. It also avoids the extra memory overhead of storing every link id separately in a `HashMap`.

I also did not make every caller fully defer RAII link creation. Many programs only get their link id after the kernel attach succeeds, so those callers can only adapt to the new API as `insert(link_id, || Ok(link))`. 

I split this into a separate first commit for the `Links::insert` API change, followed by the socket filter fix.

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/links.rs` line 105 at r3](https://reviewable.io/reviews/aya-rs/aya/1563#-Os24U1G2PEMxpUDuSLs:-Os2Ngod97DM9-IDdBak:b840cvc) ([raw file](https://github.com/aya-rs/aya/blob/b4e0b27d325c52c9b3b9ae82cc79301c2b6554e2/aya/src/programs/links.rs#L105)):*

I tried using the entry API directly, but `HashSet::entry` needs the full link value up front. That means the RAII link would already have been created before duplicate detection, unless we changed the storage from `HashSet` to `HashMap`.

Kept the existing `HashSet` storage and used the borrowed `contains` lookup by link id before creating the link. This adds an extra hash table lookup, but attach is not a hot path and the lookup cost should be negligible compared with the kernel attach operation. It also avoids the extra memory overhead of storing every link id separately in a `HashMap`.

I also did not make every caller fully defer RAII link creation. Many programs only get their link id after the kernel attach succeeds, so those callers can only adapt to the new API as `insert(link_id, || Ok(link))`. 

I split this into a separate first commit for the `Links::insert` API change, followed by the socket filter fix.
</blockquote></details>

The method [`get_or_insert_with`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashSet.html#method.get_or_insert_with) almost has the right semantics (calls the closure only if value not present) but doesn't allow fallible operations and doesn't tell you if the call happened or not.

That function is implemented in terms of [`entry_ref`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.entry_ref) (turns out hashset is an alias for hashmap with value type `()`), which I think has the semantics we want.

Does this help?

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/links.rs` line 105 at r3](https://reviewable.io/reviews/aya-rs/aya/1563#-Os24U1G2PEMxpUDuSLs:-Os3ziCg6j8g5GFua1HJ:b-c0agsf) ([raw file](https://github.com/aya-rs/aya/blob/b4e0b27d325c52c9b3b9ae82cc79301c2b6554e2/aya/src/programs/links.rs#L105)):*

The method [`get_or_insert_with`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashSet.html#method.get_or_insert_with) almost has the right semantics (calls the closure only if value not present) but doesn't allow fallible operations and doesn't tell you if the call happened or not.

That function is implemented in terms of [`entry_ref`](https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.entry_ref) (turns out hashset is an alias for hashmap with value type `()`), which I think has the semantics we want.

Does this help?
</blockquote></details>

Thanks for the pointer, but `entry_ref` is only available for the `hashbrown::HashMap`, HashSet doesn't expose this API, would you prefer switching to the `HashMap<LinkId, Link>`?

tamird (changes requested):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/links.rs` line 105 at r3](https://reviewable.io/reviews/aya-rs/aya/1563#-Os24U1G2PEMxpUDuSLs:-Os42rtgBc0wiD9m5EsN:byc42sx) ([raw file](https://github.com/aya-rs/aya/blob/b4e0b27d325c52c9b3b9ae82cc79301c2b6554e2/aya/src/programs/links.rs#L105)):*
Yep, that's why I wrote

> turns out hashset is an alias for hashmap with value type `()`

I think you can just spell HashSet as HashMap with value type `()` with an explanatory comment. Would that work?

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/links.rs` line 105 at r3](https://reviewable.io/reviews/aya-rs/aya/1563#-Os24U1G2PEMxpUDuSLs:-Os4G3IuAkPJlz7bgEX0:bj3aiic) ([raw file](https://github.com/aya-rs/aya/blob/b4e0b27d325c52c9b3b9ae82cc79301c2b6554e2/aya/src/programs/links.rs#L105)):*

Yep, that's why I wrote

> turns out hashset is an alias for hashmap with value type `()`

I think you can just spell HashSet as HashMap with value type `()` with an explanatory comment. Would that work?
</blockquote></details>

Ah, got it, that makes sense!

tamird (changes requested):
> @tamird reviewed 31 files and all commit messages, made 4 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: all files reviewed, 4 unresolved discussions (waiting on swananan).

___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os6925B6wSSR_JLhAVm:bykisas) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

Done.
</blockquote></details>

You didn't want to keep different payloads for different assertions?
___
*[`aya/src/programs/sock_ops.rs` line 98 at r6](https://reviewable.io/reviews/aya-rs/aya/1563#-Os68IiW-faOivJ-wQ8v:-Os68IiW-faOivJ-wQ8w:bmzchzu) ([raw file](https://github.com/aya-rs/aya/blob/4863cdefe03d2a4af764e80ac6f544ba83cf62da/aya/src/programs/sock_ops.rs#L98)):*
> ```XML
>             let link = SockOpsLink::new(SockOpsLinkInner::ProgAttach(link));
>             let link_id = link.id();
>             self.data.links.insert(link_id, || Ok(link))
> ```

does it make sense to have a dedicated method for this with a name that cautions the caller about the possible danger?
___
*[`aya/src/programs/socket_filter.rs` line 83 at r7](https://reviewable.io/reviews/aya-rs/aya/1563#-Os69EFf1yZlEWv8XCt2:-Os69EFf1yZlEWv8XCt3:b-grr0hg) ([raw file](https://github.com/aya-rs/aya/blob/b70be12becea84c824576ca1159a2733b6eb98ae/aya/src/programs/socket_filter.rs#L83)):*
> ```XML
>     /// The returned value can be used to detach from the socket, see [`SocketFilter::detach`].
>     ///
>     /// The socket passed to `attach` must remain open until the link is detached. If the
> ```

or else what?
___
*[`test/integration-common/src/lib.rs` line 167 at r7](https://reviewable.io/reviews/aya-rs/aya/1563#-Os68mEIB9a72z8srx8R:-Os68mEIB9a72z8srx8S:b7d34su) ([raw file](https://github.com/aya-rs/aya/blob/b70be12becea84c824576ca1159a2733b6eb98ae/test/integration-common/src/lib.rs#L167)):*
> ```XML
>     pub const MIGRATE_HITS_INDEX: u32 = 1;
>     pub const CLEAR_FALLBACK_HITS_INDEX: u32 = 2;
>     pub const PATH_HITS_MAX_ENTRIES: u32 = CLEAR_FALLBACK_HITS_INDEX + 1;
> ```

i think it can just be 3 - you don't write CLEAR_FALLBACK = MIGRATE + 1

swananan (commented):
> @swananan made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 31 files reviewed, 4 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/sock_ops.rs` line 98 at r6](https://reviewable.io/reviews/aya-rs/aya/1563#-Os68IiW-faOivJ-wQ8v:-Os6Ozjo770FOkhDEDO1:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/4863cdefe03d2a4af764e80ac6f544ba83cf62da/aya/src/programs/sock_ops.rs#L98)):*

does it make sense to have a dedicated method for this with a name that cautions the caller about the possible danger?
</blockquote></details>

Done.
___
*[`aya/src/programs/socket_filter.rs` line 83 at r7](https://reviewable.io/reviews/aya-rs/aya/1563#-Os69EFf1yZlEWv8XCt2:-Os6JDWk1YmsmJxYzsKj:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/b70be12becea84c824576ca1159a2733b6eb98ae/aya/src/programs/socket_filter.rs#L83)):*

or else what?
</blockquote></details>

Done.
___
*[`test/integration-common/src/lib.rs` line 167 at r7](https://reviewable.io/reviews/aya-rs/aya/1563#-Os68mEIB9a72z8srx8R:-Os6Ht26-RAsEvtUBTuF:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/b70be12becea84c824576ca1159a2733b6eb98ae/test/integration-common/src/lib.rs#L167)):*

i think it can just be 3 - you don't write CLEAR_FALLBACK = MIGRATE + 1
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os6Nvc03jOCbAYlPbuO:b1awkga) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

You didn't want to keep different payloads for different assertions?
</blockquote></details>

Keeping different payloads without allocation would require two generic const parameters(N, N+1), which felt a bit awkward, so I switched to a single payload instead.

tamird (changes requested):
> @tamird reviewed 31 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: all files reviewed, 2 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 83 at r7](https://reviewable.io/reviews/aya-rs/aya/1563#-Os69EFf1yZlEWv8XCt2:-Os7dwvSEyGtcEDWCQPK:bqeih5) ([raw file](https://github.com/aya-rs/aya/blob/b70be12becea84c824576ca1159a2733b6eb98ae/aya/src/programs/socket_filter.rs#L83)):*

Done.
</blockquote></details>

Ah! Boy this is pretty scary. I wonder if we should `dup` the FD so that we can avoid this footgun, though we would need to hold the owned FD on the link so that drop does the right thing.
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os7eZeoDRWKoRfcAHK7:b-vqrwh3) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

Keeping different payloads without allocation would require two generic const parameters(N, N+1), which felt a bit awkward, so I switched to a single payload instead.
</blockquote></details>

Ah, drat - that's annoying. You could do it with a macro, but up to you.

swananan (commented):
> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 83 at r7](https://reviewable.io/reviews/aya-rs/aya/1563#-Os69EFf1yZlEWv8XCt2:-Os8PKlzBZaoHTXkLCeJ:b-cuxln9) ([raw file](https://github.com/aya-rs/aya/blob/b70be12becea84c824576ca1159a2733b6eb98ae/aya/src/programs/socket_filter.rs#L83)):*

Ah! Boy this is pretty scary. I wonder if we should `dup` the FD so that we can avoid this footgun, though we would need to hold the owned FD on the link so that drop does the right thing.
</blockquote></details>

Done. Also added a regression test for this.
___
*[`test/integration-test/src/tests/socket_filter.rs` line 23 at r1](https://reviewable.io/reviews/aya-rs/aya/1563#-Os1YR2_2292tqDG_KEX:-Os8Fm4a-lmAbdWLUZus:bkpni1s) ([raw file](https://github.com/aya-rs/aya/blob/f54a9a6537b3b4c9e46c376696f7c3c5752a5c53/test/integration-test/src/tests/socket_filter.rs#L23)):*

Ah, drat - that's annoying. You could do it with a macro, but up to you.
</blockquote></details>

I tried a macro first, but it still felt a bit more complicated than the benefit was worth. I think the main benefit of different payloads would be making each assertion easier to identify, but I prefer to keep this helper simple for now.

I'm also planning to add the reuseport socket filter next, so more socket filter test coverage is coming. I'll revisit the payload then if the assertions become harder to follow.

tamird (changes requested):
> @tamird reviewed all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 3 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 124 at r11](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8TzlK-Mp05kAShu8M:-Os8TzlK-Mp05kAShu8N:bh4rg1u) ([raw file](https://github.com/aya-rs/aya/blob/01cc84d326b0628a1bcce65909007b7bdbc86cbe/aya/src/programs/socket_filter.rs#L124)):*
> ```XML
> 
>             Ok(SocketFilterLink::new(SocketFilterLinkInner {
>                 socket,
> ```

do we need to hold both of them? why?
___
*[`aya/src/programs/socket_filter.rs` line 141 at r11](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8U5iADrb8eqUz3zhV:-Os8U5iADrb8eqUz3zhW:b-ywdwtf) ([raw file](https://github.com/aya-rs/aya/blob/01cc84d326b0628a1bcce65909007b7bdbc86cbe/aya/src/programs/socket_filter.rs#L141)):*
> ```XML
>     // Used for the link id. This is the caller's fd number, not the
>     // duplicate, so duplicate attaches to the same socket still use the same id.
>     socket_fd: RawFd,
> ```

this doesn't look right for the same reason as the original bug: FDs can be recycled



_Code quote:_
```XML
    // Used for the link id. This is the caller's fd number, not the
    // duplicate, so duplicate attaches to the same socket still use the same id.
    socket_fd: RawFd,
```

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 3 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 141 at r11](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8U5iADrb8eqUz3zhV:-Os8Uhnc8FrSoOwujSGt:b-wqaxi6) ([raw file](https://github.com/aya-rs/aya/blob/01cc84d326b0628a1bcce65909007b7bdbc86cbe/aya/src/programs/socket_filter.rs#L141)):*

this doesn't look right for the same reason as the original bug: FDs can be recycled
</blockquote></details>

Should I try the using socket identify via fstat? Since caller could dup the original fd and attach again.

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 3 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 141 at r11](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8U5iADrb8eqUz3zhV:-Os8ViqqBVS4MV7cIM0i:bdomtkh) ([raw file](https://github.com/aya-rs/aya/blob/01cc84d326b0628a1bcce65909007b7bdbc86cbe/aya/src/programs/socket_filter.rs#L141)):*

Should I try the using socket identify via fstat? Since caller could dup the original fd and attach again.
</blockquote></details>

Socket identity sounds reasonable to me.

swananan (commented):
> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 3 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 124 at r11](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8TzlK-Mp05kAShu8M:-Os8ZLLC6eWYUiRiwus5:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/01cc84d326b0628a1bcce65909007b7bdbc86cbe/aya/src/programs/socket_filter.rs#L124)):*

do we need to hold both of them? why?
</blockquote></details>

Done.
___
*[`aya/src/programs/socket_filter.rs` line 141 at r11](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8U5iADrb8eqUz3zhV:-Os8ZJdI9pXRs9rY-8wK:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/01cc84d326b0628a1bcce65909007b7bdbc86cbe/aya/src/programs/socket_filter.rs#L141)):*

Socket identity sounds reasonable to me.
</blockquote></details>

Done.

tamird (changes requested):
> @tamird made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 2 unresolved discussions (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-Os8blsc3vPM7rtFty1S:b-jln568) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*
> ```XML
>             // Duplicate the socket fd so detach/drop do not depend on the
>             // caller's fd lifetime.
>             let socket = socket.try_clone_to_owned()?;
> ```

Thinking about this some more:
- using the socket identity as the key makes sense
- duplicating the socket FD maybe does not make sense; if the user wants to close the socket, they may be in for a surprise when the filter keeps it open

since the original bug here had to do with file descriptor reuse, maybe the socket identity is a sufficient solution?

but i'm not sure fstat does what we want here. from chatgpt:

Properties in practice (Linux/Unix)
The pair (st_dev, st_ino) is:
Unique for that socket object while it exists
Can be used to distinguish FDs referring to the same underlying socket
Stable across dup() / fork() (same underlying file → same inode)
However:
Not persistent (reused after socket is closed)
Not globally meaningful outside the kernel instance
Not tied to network identity (IP/port/etc.)



_Code quote:_
```XML
            // Duplicate the socket fd so detach/drop do not depend on the
            // caller's fd lifetime.
            let socket = socket.try_clone_to_owned()?;
```

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-Os9NG9L13laxTCVWqHO:b-ktwl0v) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Thinking about this some more:
- using the socket identity as the key makes sense
- duplicating the socket FD maybe does not make sense; if the user wants to close the socket, they may be in for a surprise when the filter keeps it open

since the original bug here had to do with file descriptor reuse, maybe the socket identity is a sufficient solution?

but i'm not sure fstat does what we want here. from chatgpt:

Properties in practice (Linux/Unix)
The pair (st_dev, st_ino) is:
Unique for that socket object while it exists
Can be used to distinguish FDs referring to the same underlying socket
Stable across dup() / fork() (same underlying file → same inode)
However:
Not persistent (reused after socket is closed)
Not globally meaningful outside the kernel instance
Not tied to network identity (IP/port/etc.)
</blockquote></details>

If we don't duplicate the socket fd, there are two downsides:

1. If the caller closes the socket fd first and that fd number is reused, `detach` or link drop may operate on an unrelated socket.
2. We also lose a reliable socket identity for duplicate detection. `fstat` works well only while the socket object is kept alive, and I haven't found a good solution for socket object identity.

This also made me wonder whether `SocketFilter` should provide RAII link semantics at all, just similar to `SkReuseport`.

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-Os9_h0XDQQZigxObb3B:b60dp8h) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

If we don't duplicate the socket fd, there are two downsides:

1. If the caller closes the socket fd first and that fd number is reused, `detach` or link drop may operate on an unrelated socket.
2. We also lose a reliable socket identity for duplicate detection. `fstat` works well only while the socket object is kept alive, and I haven't found a good solution for socket object identity.

This also made me wonder whether `SocketFilter` should provide RAII link semantics at all, just similar to `SkReuseport`.
</blockquote></details>

I looked at the existing pattern in Aya and found `ProgAttachLink`, which duplicates the target fd and stores the owned fd in the link so detach/drop do not depend on the caller's fd lifetime.

So I prefer to keep the current approach, or we can remove the RAII semantics for socket filter.

tamird (approved):
> @tamird reviewed 9 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: all files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsNbcCr-wiDXL6XTpRz:buye3ra) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

I looked at the existing pattern in Aya and found `ProgAttachLink`, which duplicates the target fd and stores the owned fd in the link so detach/drop do not depend on the caller's fd lifetime.

So I prefer to keep the current approach, or we can remove the RAII semantics for socket filter.
</blockquote></details>

Interestingly `ProgAttachLink` holds both FDs as owned. Do you grok why the prog_fd is not owned here?

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsQhvmc7wyrkEb7MKdJ:b-x5u1ww) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Interestingly `ProgAttachLink` holds both FDs as owned. Do you grok why the prog_fd is not owned here?
</blockquote></details>

Good catch, I found that we don't need to keep prog_fd here.

tamird (commented):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsQjjHU8pOpyzlUPVjp:b-jsma35) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Good catch, I found that we don't need to keep prog_fd here.
</blockquote></details>

Clarifying: that's because each socket can only have one filter?

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsQkTfs7xHeZUpS4R7n:b-2v1zjx) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Clarifying: that's because each socket can only have one filter?
</blockquote></details>

Yes, so kernel can detach the filter without the prog_fd.



_Code snippet:_
```XML
        // `SO_DETACH_BPF` is an alias for `SO_DETACH_FILTER`; Linux's
        // `sk_setsockopt()` requires an `int` optval, but the detach branch only
        // calls `sk_detach_filter(sk)` and does not use a program fd:
        // https://github.com/torvalds/linux/blob/v6.9/net/core/sock.c#L1413-L1414
```

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsQmB-H9vgY31YJJib3:b-fgjeuw) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Yes, so kernel can detach the filter without the prog_fd.
</blockquote></details>

Right but also removing the program fd means we will not allow multiple filters through the attach path. Sounds like that's correct

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsRLRO93u1pPkszdeW4:b-bkigod) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Right but also removing the program fd means we will not allow multiple filters through the attach path. Sounds like that's correct
</blockquote></details>

Thinking about this more, I don't think the RAII semantics Link is suitable for the socket filter.

For example:

1. `prog_a.attach(socket_a)` succeeds.
2. `prog_b.attach(socket_a)` also succeeds, and the kernel replaces the socket's filter slot with `prog_b`.
3. If `prog_a` or its owned link is dropped afterwards, it will detach the filter currently installed on the socket, which is now `prog_b`.

The socket identity does not protect against this scenario, because links are tracked per `ProgramData`. `prog_a` and `prog_b` have independent link maps, so `prog_b.attach(socket_a)` cannot see the link recorded by `prog_a`. And once `prog_b` replaces the socket's filter slot, `prog_a`'s link still detaches by socket, not by program.

I spent some time on this PR, but I think this needs a rework😂. Maybe socket filter should behave more like `SkReuseport`, with explicit detach semantics rather than RAII. And this also matters for the reuseport support for socket filter.

tamird (commented):
> @tamird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 22 of 31 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsRfE7T6RMuyYpW5TIz:be1hxqb) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Thinking about this more, I don't think the RAII semantics Link is suitable for the socket filter.

For example:

1. `prog_a.attach(socket_a)` succeeds.
2. `prog_b.attach(socket_a)` also succeeds, and the kernel replaces the socket's filter slot with `prog_b`.
3. If `prog_a` or its owned link is dropped afterwards, it will detach the filter currently installed on the socket, which is now `prog_b`.

The socket identity does not protect against this scenario, because links are tracked per `ProgramData`. `prog_a` and `prog_b` have independent link maps, so `prog_b.attach(socket_a)` cannot see the link recorded by `prog_a`. And once `prog_b` replaces the socket's filter slot, `prog_a`'s link still detaches by socket, not by program.

I spent some time on this PR, but I think this needs a rework😂. Maybe socket filter should behave more like `SkReuseport`, with explicit detach semantics rather than RAII. And this also matters for the reuseport support for socket filter.
</blockquote></details>

Thanks for the update; this analysis makes sense.

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563)* status: 0 of 32 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`aya/src/programs/socket_filter.rs` line 110 at r13](https://reviewable.io/reviews/aya-rs/aya/1563#-Os8blsc3vPM7rtFty1R:-OsS2rzl8x8QaOfWdE9r:bbj5i6c) ([raw file](https://github.com/aya-rs/aya/blob/8090ab8cd0a9942b2f863de9cfefb6d5b7aa95fc/aya/src/programs/socket_filter.rs#L110)):*

Thanks for the update; this analysis makes sense.
</blockquote></details>

I reworked this PR and removed the link insertion API commit as well. I’ll take another pass over all program types, and if the link insertion change is still needed, I’ll open it as a separate PR.

tamird (approved):
@codex review

> @tamird reviewed 32 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563#-:-OsYMDOC4RyoXgdVxg98:b-fv7n3s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on swananan).

swananan:
@tamird, thanks again for the careful reviews, is there anything else needed from my side, or is it ready to merge when you have a chance?

tamird:
Thanks for your patience. I pushed another commit; have a look and if it looks good to you feel free to squash it. Otherwise let me know where you disagree.

swananan:
Thanks, I squashed the commit, also removed the mutable receiver from `attach` and applied the same change to `sk_reuseport`.

tamird (approved):
@codex review

> @tamird reviewed 6 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1563#-:-Ot1oycGBSiiY8rshlum:b-fv7n3s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on swananan).

Issue #36, opened 2021-08-01
Support Unit Testing of BPF Programs


dave-tucker:
It would be helpful to be able to test the `try_xdp_stats` function of my eBPF program.
This would significantly reduce the dev/test cycle time as a simple `cargo test` could run tests using the host arch.

For example:

```rust
#![cfg_attr(not(test), no_std, no_main)]

use aya_bpf::bindings::xdp_action;
use aya_bpf::cty::c_long;
use aya_bpf::macros::{map, xdp};
use aya_bpf::maps::Array;
use aya_bpf::programs::XdpContext;

use types::XdpData;

#[cfg(not(test))]
#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    unreachable!()
}

#[map]
static mut xdp_stats_map : Array<XdpData> = Array::with_max_entries(1,0);

#[xdp(name = "xdp_stats")]
pub fn xdp_stats(ctx: XdpContext) -> u32 {
    match unsafe { try_xdp_stats(ctx) } {
        Ok(ret) => ret,
        Err(_) => xdp_action::XDP_DROP,
    }
}

unsafe fn try_xdp_stats(_ctx: XdpContext) -> Result<u32, c_long> {
    let data = match xdp_stats_map.get(0) {
        Some(data) => data,
        None => return Err(0)
    };
    let xdp_data = XdpData{
        packet_count: data.packet_count + 1,
    };
    xdp_stats_map.set(0, &xdp_data, 0)?;
    Ok(xdp_action::XDP_PASS)
}

#[cfg(test)]
mod tests {
    use super::*;
    use aya_bpf::programs::XdpContext;

    #[test]
    fn test() {
        let res = unsafe { try_xdp_stats(XdpContext::empty()) };
        match res {
            Ok(action) => assert_eq!(xdp_action::XDP_PASS, action),
            Err(e) => panic!("error: {}", e),
        }
    }
}
```

There are a couple of things that would be necessary in the context of a test to make this possible:

1. The ability to construct `XdpContext` (and other contexts) using test data (e.g `XdpContext::empty()` or `XdpContext::from_data(data, metadata, data_len ...)`
2. To replace the map implementation with a `std::collections::HashMap` (or similar compatible structure)
3. To replace helper implementations with ones that don't rely on bpf-helper.h functions 

I think this would all be relatively simple, as we should be able using conditional compilation:
[...]

PR #1565, opened 2026-05-13
feat(sys): expose test-utils feature for mocking BPF syscalls

### LINKED ISSUE #999: Support Unit Testing of Userspace Programs
I have found myself writing mocks for different map types but it is a bit painful; when trying to use new map types I get similar type conflicts, etc until I realize I need to mock more things, IDE is no longer helpful for docs and autocomplete of methods not yet implemented by my mocks, etc.

Maybe there is a better way like creating my own wrapper types I can more easily mock. Ultimately I still end up having to write my own mock implementations and I think this is likely a problem faced by others so I wanted to open this issue to discuss it.

Example mock usage:
```rust
#[cfg(any(not(test), rust_analyzer))]
type PerCpuArray<V> = aya::maps::PerCpuArray<aya::maps::MapData, V>;

#[cfg(all(test, not(rust_analyzer)))]
use crate::mocks::aya::{Ebpf, PerCpuArray};
```

Example mock (quick and dirty):
```rust
#[derive(Clone)]
pub struct PerCpuArray<V: Pod> {
    inner: Arc<Mutex<Vec<Vec<V>>>>,
    _v: std::marker::PhantomData<V>,
}

impl<V: Pod> PerCpuArray<V> {
    pub(crate) fn new(len: usize, val: V) -> Self {
        let arr = vec![val.clone(); nr_cpus().unwrap()];
        PerCpuArray {
            inner: Arc::new(Mutex::new(vec![arr.clone(); len])),
            _v: core::marker::PhantomData,
        }
    }
}

impl TryFrom<Map> for PerCpuArray<u64> {
    type Error = MapError;

    fn try_from(_map: Map) -> Result<PerCpuArray<u64>, MapError> {
        Ok(PerCpuArray::new(1, 0u64))
    }
}

impl<V: Pod> PerCpuArray<V> {
    pub fn get(&self, index: &u32, _flags: u64) -> Result<PerCpuValues<V>, MapError> {
        let guard = self.inner.lock().unwrap();
        let values = guard.get(*index as usize).ok_or(MapError::OutOfBounds {
            index: *index,
            max_entries: guard.len() as u32,
        })?;
        PerCpuValues::try_from(values.clone()).map_err(|_| MapError::KeyNotFound)
    }

    pub fn set(&mut self, index: u32, values: PerCpuValues<V>, _flags: u64) -> Result<(), MapError> {
        let arr = (0..nr_cpus().unwrap())
            .map(|i| values.get(i).unwrap().clone())
            .collect::<Vec<V>>();
        let mut guard = self.inner.lock().unwrap();
        guard[index as usize] = arr;
        Ok(())
    }
}
```

Similar to https://github.com/aya-rs/aya/issues/36

----

One thing I did notice was that tests inside aya leverage `test_utils`: https://github.com/aya-rs/aya/blob/a167554b8b1b67ca329721381afae9992a3c1544/aya/src/maps/mod.rs#L1040-L1076

which uses `override_syscall`: https://github.com/aya-rs/aya/blob/a167554b8b1b67ca329721381afae9992a3c1544/aya/src/sys/fake.rs#L19

Which is used as a mock here: https://github.com/aya-rs/aya/blob/a167554b8b1b67ca329721381afae9992a3c1544/aya/src/sys/mod.rs#L91

However, this is only usable internally to the crate and when compiled to a test target. It would be really cool to expose something similar to make it easier to test for users of aya.

There are a few different ways around it but I am thinking the simplest and least intrusive could be a feature that users can enable for their tests that allows them to override syscalls. Going one step further, the creation of the maps themselves could be mocked appropriately.

OliverGavin:
## Summary

Add a `test-utils` feature that exposes `override_syscall` and the `Syscall` enum publicly via `sys::test_utils` module. This allows downstream crates to intercept BPF syscalls in their tests without needing to reimplement aya's internal types.

## Motivation

As discussed in #999, testing userspace code that interacts with BPF maps currently requires reimplementing aya's type signatures in mock crates. Aya already has an internal `override_syscall` mechanism used in its own tests — this PR exposes it behind an opt-in feature.

## Changes

- Added `test-utils` feature to `aya/Cargo.toml`
- `sys/fake.rs` now compiles under `cfg(any(test, feature = "test-utils"))`
- New `sys::test_utils` module (feature-gated) re-exports `override_syscall`, `Syscall`, `SysResult`, `bpf_attr`, `bpf_cmd`, and `mock_fd()`
- `Syscall`, `PerfEventIoctlRequest`, and `SysResult` made `pub` (previously `pub(crate)`)
- `MockableFd` internals gated on `any(test, feature = "test-utils")`
- Version bumped to 0.13.3

## Usage

```rust
use aya::sys::test_utils::{override_syscall, Syscall, bpf_cmd, mock_fd};

override_syscall(|call| match call {
    Syscall::Ebpf { cmd: bpf_cmd::BPF_MAP_CREATE, .. } => Ok(mock_fd()),
    Syscall::Ebpf { cmd: bpf_cmd::BPF_MAP_LOOKUP_ELEM, attr } => {
        // Fill attr with test data
        Ok(0)
    }
    _ => Ok(0),
});
```

Closes #999
Related: #36

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1565)


aya/src/sys/fake.rs
```diff
@@ -4,18 +4,36 @@ use super::{SysResult, Syscall};
...
 }
 
-#[cfg(test)]
 unsafe fn test_syscall(_call: Syscall<'_>) -> SysResult {
     Err((-1, io::Error::from_raw_os_error(libc::EINVAL)))
 }
 
-#[cfg(test)]
-pub(crate) fn override_syscall(call: unsafe fn(Syscall<'_>) -> SysResult) {
+/// Override the syscall implementation for testing purposes.
```
vadorovsky: super nit:


```suggestion
/// Overrides the syscall implementation for testing purposes.
```

aya/src/sys/mod.rs
```diff
@@ -25,31 +25,87 @@ pub(crate) use netlink::*;
...
     Enable,
+    /// Disable the perf event.
     Disable,
+    /// Set the BPF program for the perf event.
     SetBpf(BorrowedFd<'a>),
 }
 
-#[cfg_attr(test, expect(dead_code, reason = "test stubs cut above this"))]
-pub(crate) enum Syscall<'a> {
+/// Represents a system call that aya may perform.
```
vadorovsky:
```suggestion
/// Represents a system call that Aya may perform.
```

aya/Cargo.toml
```diff
@@ -4,7 +4,7 @@ documentation = "https://docs.rs/aya"
 keywords = ["bpf", "ebpf", "kernel", "linux"]
 name = "aya"
 readme = "README.md"
-version = "0.13.2"
+version = "0.13.3"
```
vadorovsky: 1. Let's leave the version alone for now, we can bump it when we actually want to cut a release.
2. This PR alone changes the API, and there were lots of API-breaking changes merged since last release. To respect semver, we should release 0.14.0. Personally, I think a right moment for cutting it is when all @Brskt's PRs related to maps are merged.

aya/Cargo.toml
```diff
@@ -13,6 +13,15 @@ license.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 
+[features]
+default = []
+## Expose test utilities for mocking syscalls in user tests.
+##
+## When enabled, the `sys::test_utils` module becomes available, providing
+## `override_syscall` and the `Syscall` enum so that downstream crates can
+## intercept BPF syscalls in their own `#[cfg(test)]` code.
```
vadorovsky:
```suggestion
# Expose test utilities for mocking syscalls in user tests.
#
# When enabled, the `sys::test_utils` module becomes available, providing
# `override_syscall` and the `Syscall` enum so that downstream crates can
# intercept BPF syscalls in their own `#[cfg(test)]` code.
```

vadorovsky (commented):
Thanks! Overall I agree with the direction and this looks like a good first step towards making different pieces of Aya mockable.

tamird (commented):
> @tamird reviewed 5 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1565)* status: all files reviewed, 10 unresolved discussions (waiting on Brskt and OliverGavin).

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/aya-rs/aya/1565#-OsliOhi7gETLkF-dnqv:-OsliOhi7gETLkF-dnqw:b3cvf0g):*
nit: can just be "Closes #999."
___
*[`aya/src/sys/fake.rs` line 7 at r2](https://reviewable.io/reviews/aya-rs/aya/1565#-Oslim9-1bY2wcrrt7WX:-Oslim9-1bY2wcrrt7WY:b-o33186) ([raw file](https://github.com/aya-rs/aya/blob/df1e81c1302a29c51b6d21a92ee645d637e271ff/aya/src/sys/fake.rs#L7)):*
> ```XML
> type SyscallFn = unsafe fn(Syscall<'_>) -> SysResult;
> 
> #[cfg(test)]
> ```

in other places this changed to

#[cfg(any(test, feature = "test-utils"))]

why not in this file?
___
*[`aya/src/sys/mod.rs` line 9 at r2](https://reviewable.io/reviews/aya-rs/aya/1565#-OsljcF864csPhrsrjxg:-OsljcF864csPhrsrjxh:bf7r6hp) ([raw file](https://github.com/aya-rs/aya/blob/0dd36f24e292bf58efed43e5922c2c6ef234208d/aya/src/sys/mod.rs#L9)):*
> ```XML
> 
> #[cfg(any(test, feature = "test-utils"))]
> pub(crate) mod fake;
> ```

why pub(crate) here?
___
*[`aya/src/sys/mod.rs` line 19 at r2](https://reviewable.io/reviews/aya-rs/aya/1565#-OsljTroADrWkF2Lz92U:-OsljTroADrWkF2Lz92V:bkpwxfm) ([raw file](https://github.com/aya-rs/aya/blob/0dd36f24e292bf58efed43e5922c2c6ef234208d/aya/src/sys/mod.rs#L19)):*
> ```XML
> use aya_obj::generated::{bpf_attr, bpf_cmd, bpf_stats_type, perf_event_attr};
> pub(crate) use bpf::*;
> #[cfg(any(test, feature = "test-utils"))]
> ```

i don't think you need the `pub(crate) use fake::*` under test-utils? i think it doesn't change the crate's API, so I don't see why you would
___
*[`aya/src/sys/mod.rs` line 50 at r2](https://reviewable.io/reviews/aya-rs/aya/1565#-Osljpfc0zzmRTcibMMb:-Osljpfc0zzmRTcibMMc:b-5p8g5b) ([raw file](https://github.com/aya-rs/aya/blob/0dd36f24e292bf58efed43e5922c2c6ef234208d/aya/src/sys/mod.rs#L50)):*
> ```XML
> #[cfg(feature = "test-utils")]
> #[cfg_attr(docsrs, doc(cfg(feature = "test-utils")))]
> pub mod test_utils {
> ```

could we rename `fake` to `test_utils` so that the decision to make certain testing utils externally visible lives on each item rather than having this parallel export layer?
___
*[`aya/src/sys/mod.rs` line 59 at r2](https://reviewable.io/reviews/aya-rs/aya/1565#-Oslj7B_A_VrwTPjvksT:-Oslj7B_A_VrwTPjvksU:bg5wefa) ([raw file](https://github.com/aya-rs/aya/blob/0dd36f24e292bf58efed43e5922c2c6ef234208d/aya/src/sys/mod.rs#L59)):*
> ```XML
>     /// syscalls that create FDs (e.g. `BPF_MAP_CREATE`).
>     pub const fn mock_fd() -> i64 {
>         crate::MockableFd::mock_signed_fd() as i64
> ```

why i64 rather than the underlying i32?

OliverGavin:
Thanks for the review @tamird! I've addressed all your feedback in the latest push:

> **commit message**: nit: can just be "Closes #999."

Done.

> **fake.rs line 7**: in other places this changed to `#[cfg(any(test, feature = "test-utils"))]` — why not in this file?

The entire file is now gated by the `mod` declaration in `mod.rs`, so individual `#[cfg]` annotations inside it were redundant. But this is now moot because...

> **mod.rs line 9**: why `pub(crate)` here?
> **mod.rs line 50**: could we rename `fake` to `test_utils` so that the decision to make certain testing utils externally visible lives on each item rather than having this parallel export layer?

Great suggestion — done. I've renamed `fake.rs` to `test_utils.rs` and made it the public module directly. The `pub` visibility now lives on each item (`override_syscall`, `mock_fd`) rather than going through a re-export layer. Internal items (`TEST_SYSCALL`, `TEST_MMAP_RET`) remain `pub(crate)`.

> **mod.rs line 19**: I don't think you need the `pub(crate) use fake::*` under test-utils?

Agreed. I've split it: `TEST_MMAP_RET` and `TEST_SYSCALL` are imported under `cfg(any(test, feature = "test-utils"))` (needed by `syscall()` and `mmap()`), while `override_syscall` is only imported under `cfg(test)` (for internal test convenience).

> **mod.rs line 59**: why `i64` rather than the underlying `i32`?

Good point — changed `mock_fd()` to return `i32` since that's what an FD actually is. Users can do `Ok(i64::from(mock_fd()))` in their handlers which makes the intent clearer.

tamird (approved):
@codex review

> @tamird reviewed 4 files and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1565#-:-Ot89J4H2PSEPBg_aIE8:b-fv7n3s)* status: all files reviewed, 4 unresolved discussions (waiting on Brskt and OliverGavin).

PR #1573, opened 2026-05-19
integration-test: dedupe test_case tests

swananan:
I just realized that some e2e tests were running twice. This PR fixes that issue.

```
test tests::btf_map_of_maps::btf_map_of_maps_get_value::array_of_maps ... ok
test tests::btf_map_of_maps::btf_map_of_maps_get_value::array_of_maps ... ok
test tests::btf_map_of_maps::btf_map_of_maps_get_value::hash_of_maps ... ok
test tests::btf_map_of_maps::btf_map_of_maps_get_value::hash_of_maps ... ok
test tests::btf_maps::libbpf_can_load_btf_maps ... ok
test tests::btf_relocations::relocation_tests::crate_enum_signed_32_checked_variants_reloc_bpf_none_none_0x7aaaaaaai32_as_u64_expects ... ok
test tests::btf_relocations::relocation_tests::crate_enum_signed_32_checked_variants_reloc_bpf_none_none_0x7aaaaaaai32_as_u64_expects ... ok
test tests::btf_relocations::relocation_tests::crate_enum_signed_32_checked_variants_reloc_bpf_some_crate_enum_signed_32_checked_variants_reloc_btf_none_0x7bbbbbbbi32_as_u64_expects ... ok
test tests::btf_relocations::relocation_tests::crate_enum_signed_32_checked_variants_reloc_bpf_some_crate_enum_signed_32_checked_variants_reloc_btf_none_0x7bbbbbbbi32_as_u64_expects ... ok
test tests::btf_relocations::relocation_tests::crate_enum_signed_32_reloc_bpf_none_none_0x7aaaaaaai32_as_u64_expects ... ok
test tests::btf_relocations::relocation_tests::crate_enum_signed_32_reloc_bpf_none_none_0x7aaaaaaai32_as_u64_expects ... ok

```

### Added/updated tests?

_We strongly encourage you to add a test for your changes._

- [x] Yes

### Checklist

- [x] Rust code has been formatted with `cargo +nightly fmt`.
- [x] All clippy lints have been fixed.
      You can find failing lints with `cargo xtask clippy`.
- [x] Unit tests are passing locally with `cargo test`.
- [x] The [Integration tests] are passing locally.
- [x] I have blessed any API changes with `cargo xtask public-api --bless`.

[Integration tests]: https://github.com/aya-rs/aya/blob/main/test/README.md

### (Optional) What GIF best describes this PR or how it makes you feel?

- - -
[...]

PR #1579, opened 2026-05-23
test: replace test-case with rstest

swananan:
test-case relies on cooperating attribute macros when wrapping generated parameterized tests. Stacking it with test_log can register each generated case twice, so passing tests may run more than once and mask order-sensitive or flaky behavior.

Move the parameterized tests to rstest, but do not rely on rstest's implicit test-attribute detection. rstest treats attributes whose path ends in test as test attributes; that has its own failure mode if a non-test decorator matches, potentially producing cases that do not run as intended.

Use explicit #[test_attr(...)] wrappers for every rstest case that needs test_log or tokio, so generated tests get the intended harness attribute instead of depending on implicit detection.

Refs: #1573

### Added/updated tests?

- [x] No, and this is why: this only changes the test macros.

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1579)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ public-api = { version = "0.51.0", default-features = false }
 quote = { version = "1", default-features = false }
 rand = { version = "0.10", default-features = false }
 rbpf = { version = "0.4.0", default-features = false }
+rstest = { version = "0.26.1", default-features = false }
 rustc_version = { version = "0.4.1", default-features = false }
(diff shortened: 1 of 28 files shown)


vadorovsky (approved):
> @vadorovsky reviewed 28 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1579)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on swananan).

PR #1587, opened 2026-05-28
aya, aya-ebpf: add InodeStorage and CgrpStorage

Brskt:
Adds `InodeStorage` and `CgrpStorage` to `aya::maps` and
`aya_ebpf::btf_maps`, so `BPF_MAP_TYPE_INODE_STORAGE` and
`BPF_MAP_TYPE_CGRP_STORAGE` are no longer routed to `Map::Unsupported`.
These are the modern local storage map types -- one value per inode or
cgroup -- and, unlike the legacy `CgroupStorage` family, they ship only
in `btf_maps`: the kernel rejects map creation unless BTF key and value
type ids are set and `BPF_F_NO_PREALLOC` is on.

The eBPF maps expose `get_ptr_mut`, `get_or_insert_ptr_mut` and
`delete`, wrapping `bpf_inode_storage_get`/`_delete` and
`bpf_cgrp_storage_get`/`_delete`. Each takes the kernel object pointer
the program already holds; aya exposes no kfuncs or BTF field accessors
to derive one from elsewhere, so the program sources it from a context
whose verifier type matches.

Userspace `get`/`insert`/`remove` operate on the kernel-managed
entries, keyed by a file descriptor referring to the inode or cgroup,
and the `Map::InodeStorage` and `Map::CgrpStorage` variants are wired
into map loading.

Integration tests cover both. The inode probe is an LSM
`inode_permission` hook that filters on the test's tgid and stores a
sentinel into the inode whose permission was checked; userspace opens
a temporary file and reads the sentinel back by its file descriptor.
The cgrp probe is a `cgroup_mkdir` BTF tracepoint that stores a
sentinel into the new cgroup; userspace creates a child cgroup and
reads it back by the cgroup directory descriptor. They skip on kernels
missing the respective map type, and, for cgrp, when tracing program
attach is unavailable.

Closes #202.

### Added/updated tests?

- [x] Yes

### Checklist

- [x] Rust code has been formatted with \`cargo +nightly fmt\`.
- [x] All clippy lints have been fixed.
      You can find failing lints with \`cargo xtask clippy\`.
- [x] Unit tests are passing locally with \`cargo test\`.
- [x] The integration tests are passing locally.
- [x] I have blessed any API changes with \`cargo xtask public-api --bless\`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1587)

PR #1589, opened 2026-05-29
aya-ebpf, integration-ebpf: make map APIs safe

Brskt:
The modern local-storage maps and the non-BTF sockmap expose wrapper
methods that operate on a kernel object or context pointer the BPF
program already holds. The verifier validates those pointers at load
time, so the methods can be safe, on the same basis the existing
context accessors such as `SkBuffContext::len` are safe. Writing
end-user BPF should not require `unsafe` for them. Follow-up to #1587.

`SkStorage`, `InodeStorage` and `CgrpStorage` make `get_ptr_mut`,
`get_or_insert_ptr_mut` and `delete` safe `pub fn`. `CgrpStorage::delete`
passes the `cgroup` pointer to the helper un-cast, so it carries an
`#[expect]` for `clippy::not_unsafe_ptr_arg_deref`. The dereference of
the returned `*mut T` stays the caller's `unsafe`. Callers in
`integration-ebpf` drop the now redundant `unsafe`.

The non-BTF `SockMap::redirect_msg` and `redirect_skb` become safe,
matching the already-safe BTF `SockMap`. `SockMap::update` keeps
`unsafe` because it takes a raw `bpf_sock_ops` pointer; it gains a real
safety doc in place of the previous lint suppression.

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1589)

### DIFF
--- a/ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs
+++ b/ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs
@@ -26,28 +26,16 @@ impl<T> CgrpStorage<T> {
     }
 
     /// Gets a mutable reference to the value associated with `cgroup`.
-    ///
-    /// # Safety
-    ///
-    /// This function may dereference the pointer `cgroup`.
     #[inline(always)]
-    pub unsafe fn get_ptr_mut(&self, cgroup: *mut cgroup) -> *mut T {
+    pub fn get_ptr_mut(&self, cgroup: *mut cgroup) -> *mut T {
         self.get_ptr(cgroup, ptr::null_mut(), 0)
     }
 
     /// Gets a mutable reference to the value associated with `cgroup`.
     ///
     /// If no value is associated with `cgroup`, `value` will be inserted.
-    ///
-    /// # Safety
-    ///
-    /// This function may dereference the pointer `cgroup`.
     #[inline(always)]
-    pub unsafe fn get_or_insert_ptr_mut(
-        &self,
-        cgroup: *mut cgroup,
-        value: Option<&mut T>,
-    ) -> *mut T {
+    pub fn get_or_insert_ptr_mut(&self, cgroup: *mut cgroup, value: Option<&mut T>) -> *mut T {
         self.get_ptr(
             cgroup,
             value.map_or(ptr::null_mut(), ptr::from_mut),
@@ -56,12 +44,13 @@ impl<T> CgrpStorage<T> {
     }
 
     /// Deletes the value associated with `cgroup`.
-    ///
-    /// # Safety
-    ///
-    /// This function may dereference the pointer `cgroup`.
+    #[expect(
+        clippy::not_unsafe_ptr_arg_deref,
+        reason = "the verifier validates the cgroup pointer as ARG_PTR_TO_BTF_ID at load time; \
+                  this wrapper forwards it to the helper without dereferencing it"
+    )]
     #[inline(always)]
-    pub unsafe fn delete(&self, cgroup: *mut cgroup) -> Result<(), i32> {
+    pub fn delete(&self, cgroup: *mut cgroup) -> Result<(), i32> {
         let ret = unsafe { bpf_cgrp_storage_delete(self.as_ptr(), cgroup) };
         if ret == 0 { Ok(()) } else { Err(ret as i32) }
     }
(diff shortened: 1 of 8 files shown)


tamird (changes requested):
> @tamird made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1589)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on Brskt).

___
*[`ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs` line 30 at r1](https://reviewable.io/reviews/aya-rs/aya/1589#-OtpIu7s704GiELPguZF:-OtpIu7s704GiELPguZG:b-r6jlbr) ([raw file](https://github.com/aya-rs/aya/blob/863005e6cc08fe19b53446d451b2b07663c9b315/ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs#L30)):*
> ```XML
>     /// Gets a mutable reference to the value associated with `cgroup`.
>     #[inline(always)]
>     pub fn get_ptr_mut(&self, cgroup: *mut cgroup) -> *mut T {
> ```

can we pass an optional mut ref? if not, why not?
___
*[`ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/1589#-OtpIyuHBlsAI7B9HOzY:-OtpIyuHBlsAI7B9HOzZ:b-if4bk3) ([raw file](https://github.com/aya-rs/aya/blob/863005e6cc08fe19b53446d451b2b07663c9b315/ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs#L38)):*
> ```XML
>     /// If no value is associated with `cgroup`, `value` will be inserted.
>     #[inline(always)]
>     pub fn get_or_insert_ptr_mut(&self, cgroup: *mut cgroup, value: Option<&mut T>) -> *mut T {
> ```

same here, and also can we avoid returning a raw pointer?

Brskt (commented):
> @Brskt made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1589)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on tamird).

___
*[`ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs` line 30 at r1](https://reviewable.io/reviews/aya-rs/aya/1589#-OtpIu7s704GiELPguZF:-OtpOyeoAl9sFEnygsgJ:b-uczit) ([raw file](https://github.com/aya-rs/aya/blob/863005e6cc08fe19b53446d451b2b07663c9b315/ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs#L30)):*

can we pass an optional mut ref? if not, why not?
</blockquote></details>

We can, but it just moves the `unsafe` to the caller, since `ctx.arg()` hands back a raw `*mut cgroup`:

-get_or_insert_ptr_mut(cgrp, None)
+get_or_insert_ptr_mut(unsafe { cgrp.as_mut() }, None)

And `&mut` would be a lie anyway: the helper only uses the cgroup as a lookup key, so there's nothing exclusive to borrow.
___
*[`ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/1589#-OtpIyuHBlsAI7B9HOzY:-OtpPVlb1K6Mli9WkX4O:b-4hkkb2) ([raw file](https://github.com/aya-rs/aya/blob/863005e6cc08fe19b53446d451b2b07663c9b315/ebpf/aya-ebpf/src/btf_maps/cgrp_storage.rs#L38)):*

same here, and also can we avoid returning a raw pointer?
</blockquote></details>

Same as above for the `cgroup` arg.

tamird (approved):
Reviewed at Brskt’s direct re-review request, including the previous safety discussion. Passing verifier-validated kernel object pointers to helpers can be safe; dereferencing the returned storage pointer remains the caller’s unsafe operation, and synthesizing a mutable reference would incorrectly promise exclusive ownership. The socket redirects now match the existing safe BTF sockmap, while the raw-pointer socket update correctly remains unsafe. The public API snapshot, integration consumers, and full kernel/platform CI agree.

— tamirdex

tamird (approved):
> @tamird reviewed 8 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1589)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Brskt).

PR #1571, opened 2026-05-18
aya: add classid to SchedClassifier netlink attach

Brskt:
Adds `classid` support to `SchedClassifier` netlink attach so a TC
filter can target a specific qdisc class. In direct-action mode the
kernel splits the resulting class id between the userspace-supplied
`TCA_BPF_CLASSID` (upper 16 bits) and the program-written
`__sk_buff::tc_classid` (lower 16 bits). Without setting the upper
half from userspace, programs writing `tc_classid` had no effect on
qdisc routing.

A new `TcHandle` newtype replaces the raw `u32` filter handle on
`NlOptions`/`NlLink` and is reused for `Option<TcHandle>` classid
since both encode `(major << 16) | minor`. `TcHandle::AUTO_ASSIGN`
documents the kernel's auto-allocate sentinel.

`SchedClassifierLink::attach_to_link` preserves the classid across
atomic program replacement: `cls_bpf_change` allocates a fresh
`cls_bpf_prog` on every replace and only binds `prog->res.classid`
if the request carries `TCA_BPF_CLASSID`, so the value is stored
on `NlLink` and reemitted on replace.

The integration tests in `tc_netlink.rs` cover classid preservation,
`AUTO_ASSIGN` triggering kernel allocation, and explicit handle
round-trip. They skip on kernels older than 6.10 where the
test-distro modprobe stub cannot autoload `cls_bpf`.

Fixes #886.

### Added/updated tests?

- [x] Yes

### Checklist

- [x] Rust code has been formatted with \`cargo +nightly fmt\`.
- [x] All clippy lints have been fixed.
      You can find failing lints with \`cargo xtask clippy\`.
- [x] Unit tests are passing locally with \`cargo test\`.
- [x] The integration tests are passing locally.
- [x] I have blessed any API changes with \`cargo xtask public-api --bless\`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1571)

Issue #886, opened 2024-02-19
AYA tc classifier does not support using tc_classid


danielconnor:
A tc classifier eBPF program can set the `tc_classid` field to specify the class that should be used to handle the packet. A direct-action eBPF program only supports setting the minor 16 bits of the `classid`, meaning the `classid/flowid` attribute needs to be set on the eBPF configuration for it to work. AYA currently doesn't support setting the `classid/flowid` attribute(via `TCA_BPF_CLASSID`) on a classifier program.

#### Explaination

Always `direct-action`
https://github.com/aya-rs/aya/blob/main/aya/src/sys/netlink.rs#L161-L162

`direct-action` means exts_integrated
https://github.com/torvalds/linux/blob/master/net/sched/cls_bpf.c#L480

If `exts_integrated`, the `classid` is `prog->res.classid` ([a 32bit integer](https://github.com/torvalds/linux/blob/master/include/net/sch_generic.h#L332)) `OR`ed with `qdisc_skb_cb(skb)->tc_classid`, [which is a 16bit integer](https://github.com/torvalds/linux/blob/master/include/net/sch_generic.h#L439) - meaning the bpf program can only set the minor 16 bits.
https://github.com/torvalds/linux/blob/master/net/sched/cls_bpf.c#L110-L119

Since the program can only set the minor 16 bits and the major 16 bits need to be set, setting the `tc_classid` field from bpf has no effect unless the program `classid/flowid` is set.

Note: calling the property `classid/flowid` because that's what `tc` accepts and outputs as `flowid` (parsing `classid/flowid`
https://github.com/iproute2/iproute2/blob/d9b886d745ada3b8481e041ceca579c6f3acbea3/tc/f_bpf.c#L123-L132C26)

I've verified it works with the following change:
```
diff --git a/aya/src/sys/netlink.rs b/aya/src/sys/netlink.rs
index f10c283..46a2e73 100644
--- a/aya/src/sys/netlink.rs
+++ b/aya/src/sys/netlink.rs
@@ -19,6 +19,7 @@ use crate::{
     generated::{
         ifinfomsg, tcmsg, IFLA_XDP_EXPECTED_FD, IFLA_XDP_FD, IFLA_XDP_FLAGS, NLMSG_ALIGNTO,
         TCA_BPF_FD, TCA_BPF_FLAGS, TCA_BPF_FLAG_ACT_DIRECT, TCA_BPF_NAME, TCA_KIND, TCA_OPTIONS,
[...]

PR #1595, opened 2026-06-08
aya-ebpf: TcContext, SkBuff: add set_tc_classid

ramos-99:
Follow-up to #1571, which added `classid` to `SchedClassifier` netlink
attach, setting `TCA_BPF_CLASSID` (the upper 16 bits of the class a
direct-action program routes to). The lower 16 bits are written by the
program into `__sk_buff::tc_classid`, but there was no safe way to write
that field from aya-ebpf, so a classifier had to go through a raw pointer
to the skb. This adds `set_tc_classid` on `TcContext`.

The setter is only exposed on `TcContext`. The verifier rejects writes
to `__sk_buff::tc_classid` from the program types behind `SkBuffContext`
(`cgroup_skb`, `socket_filter`, `sk_skb`), so the setter would compile
but fail to load there.

The integration test runs a classifier through `BPF_PROG_TEST_RUN` to
write the field and read it back with a volatile load, asserting
against the value observed. It can't exercise actual qdisc routing,
which needs an HTB class hierarchy; that part was verified manually by
attaching a classifier on `lo` under an HTB qdisc and confirming
packets are routed to the class the program writes.

Refs: #886, #1571

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1595)

### DIFF
--- a/ebpf/aya-ebpf/src/programs/tc.rs
+++ b/ebpf/aya-ebpf/src/programs/tc.rs
@@ -52,6 +52,13 @@ impl TcContext {
         self.skb.set_mark(mark);
     }
 
+    /// Sets `__sk_buff::tc_classid`, steering the packet to a traffic-control
+    /// (e.g. HTB) class. Write-only, and only valid for TC classifier programs.
+    #[inline]
+    pub fn set_tc_classid(&self, classid: u32) {
+        self.skb.set_tc_classid(classid);
+    }
+
     #[inline]
     pub fn cb(&self) -> &[u32] {
         self.skb.cb()
--- a/ebpf/aya-ebpf/src/programs/tc.rs
+++ b/ebpf/aya-ebpf/src/programs/tc.rs
@@ -52,6 +52,31 @@ impl TcContext {
         self.skb.set_mark(mark);
     }
 
+    /// Sets `__sk_buff::tc_classid`.
+    ///
+    /// The kernel [stores this field as 16 bits][store], so writes are
+    /// truncated to the minor id.
+    ///
+    /// Class routing and pre-population require a netlink `cls_bpf` filter.
+    /// Attach with `aya::programs::tc::SchedClassifier::attach_with_options`
+    /// and `TcAttachOptions::Netlink`, setting `NlOptions::classid` to the
+    /// intended class.
+    ///
+    /// In direct-action mode this provides the minor 16 bits of the
+    /// resulting class id; [the major 16 bits come from the `classid`
+    /// configured on the attached filter][major].
+    ///
+    /// Before the program runs, [`cls_bpf`][cls_bpf] pre-populates this
+    /// field with that same `classid`.
+    ///
+    /// [store]: https://github.com/torvalds/linux/blob/e5f0a698b/net/core/filter.c#L9768-L9781
+    /// [major]: https://github.com/torvalds/linux/blob/e5f0a698b/net/sched/cls_bpf.c#L110-L118
+    /// [cls_bpf]: https://github.com/torvalds/linux/blob/e5f0a698b/net/sched/cls_bpf.c#L90-L105
+    #[inline]
+    pub fn set_tc_classid(&self, classid: u16) {
+        self.skb.set_tc_classid(classid);
+    }
+
     #[inline]
     pub fn cb(&self) -> &[u32] {
         self.skb.cb()
--- a/test/integration-ebpf/src/tc_classid.rs
+++ b/test/integration-ebpf/src/tc_classid.rs
@@ -1,0 +1,14 @@
+#![no_std]
+#![no_main]
+#![expect(unused_crate_dependencies, reason = "used in other bins")]
+
+use aya_ebpf::{bindings::TC_ACT_OK, macros::classifier, programs::TcContext};
+use integration_common::tc_classid::EXPECTED_CLASSID;
+#[cfg(not(test))]
+extern crate ebpf_panic;
+
+#[classifier]
+fn set_classid(ctx: TcContext) -> i32 {
+    ctx.set_tc_classid(EXPECTED_CLASSID);
+    TC_ACT_OK
+}
--- a/test/integration-ebpf/src/tc_classid.rs
+++ b/test/integration-ebpf/src/tc_classid.rs
@@ -1,0 +1,19 @@
+#![no_std]
+#![no_main]
+
+use aya_ebpf::{EbpfContext as _, bindings::__sk_buff, macros::classifier, programs::TcContext};
+use integration_common::tc_classid::EXPECTED_CLASSID;
+#[cfg(not(test))]
+extern crate ebpf_panic;
+
+#[classifier]
+fn set_classid(ctx: TcContext) -> i32 {
+    ctx.set_tc_classid(EXPECTED_CLASSID);
+
+    // Read the field back and return it so the test can assert the store
+    // happened. Volatile so the compiler emits the BPF load instead of
+    // forwarding the store above.
+    let skb = ctx.as_ptr().cast::<__sk_buff>();
+    let read_back = unsafe { core::ptr::read_volatile(&raw const (*skb).tc_classid) };
+    read_back as i32
+}
--- a/test/integration-test/src/tests/tc_classid.rs
+++ b/test/integration-test/src/tests/tc_classid.rs
@@ -1,0 +1,40 @@
+use aya::{
+    Ebpf, TestRunOptions, TestRunResult,
+    programs::{SchedClassifier, TestRun as _},
+};
+
+// `sizeof(pkt_v4)` = Size(Ethernet) + Size(IPv4) + Size(TCP) = 14 + 20 + 20
+const PKT_V4_SIZE: usize = 14 + 20 + 20;
+
+// Scope: proves only that the verifier loads a sched_cls program writing
+// `tc_classid`. The written value can't be observed here (BPF_PROG_TEST_RUN
+// leaves ctx_out empty for sched_cls, and the qdisc consumes the field), and
+// checking real routing needs HTB-qdisc infra the suite lacks; that path was
+// verified manually.
+#[test_log::test]
+fn tc_classid_set() {
+    let kernel_version = aya::util::KernelVersion::current().unwrap();
+    // BPF_PROG_TEST_RUN was introduced in v4.12 (1cf1cae963c2, "bpf: introduce
+    // BPF_PROG_TEST_RUN command") with support for sched_cls (used here) and
+    // sched_act program types. On kernels before v4.12 the syscall command does
+    // not exist and the bpf(2) call returns EINVAL.
+    if kernel_version < aya::util::KernelVersion::new(4, 12, 0) {
+        return;
+    }
+
+    let mut bpf = Ebpf::load(crate::TC_CLASSID).unwrap();
+    let prog: &mut SchedClassifier = bpf.program_mut("set_classid").unwrap().try_into().unwrap();
+    prog.load().unwrap();
+
+    let data_in = vec![0u8; PKT_V4_SIZE];
+    let mut data_out = vec![0u8; PKT_V4_SIZE];
+
+    let opts = TestRunOptions {
+        data_in: Some(&data_in),
+        data_out: Some(&mut data_out),
+        ..TestRunOptions::default()
+    };
+
+    let TestRunResult { return_value, .. } = prog.test_run(opts).unwrap();
+    assert_eq!(return_value, 0, "Expected TC_ACT_OK(0)");
+}
--- a/test/integration-test/src/tests/tc_classid.rs
+++ b/test/integration-test/src/tests/tc_classid.rs
@@ -1,0 +1,46 @@
+use aya::{
+    Ebpf, TestRunOptions, TestRunResult,
+    programs::{SchedClassifier, TestRun as _},
+};
+use integration_common::tc_classid::EXPECTED_CLASSID;
+
+// `sizeof(pkt_v4)` = Size(Ethernet) + Size(IPv4) + Size(TCP) = 14 + 20 + 20
+const PKT_V4_SIZE: usize = 14 + 20 + 20;
+
+// The program writes `tc_classid` and returns what a subsequent read observes,
+// so this asserts the store reached the field.
+#[test_log::test]
+fn tc_classid_set() {
+    let kernel_version = aya::util::KernelVersion::current().unwrap();
+    // BPF_PROG_TEST_RUN was introduced in v4.12 (1cf1cae963c2, "bpf: introduce
+    // BPF_PROG_TEST_RUN command") with support for sched_cls (used here) and
+    // sched_act program types. On kernels before v4.12 the syscall command does
+    // not exist and the bpf(2) call returns EINVAL.
+    if kernel_version < aya::util::KernelVersion::new(4, 12, 0) {
+        return;
+    }
+
+    let mut bpf = Ebpf::load(crate::TC_CLASSID).unwrap();
+    let prog: &mut SchedClassifier = bpf.program_mut("set_classid").unwrap().try_into().unwrap();
+    prog.load().unwrap();
+
+    // The program never reads the packet, but the kernel refuses to build the
+    // skb from fewer than ETH_HLEN bytes.
+    // https://github.com/torvalds/linux/blob/e5f0a698b/net/bpf/test_run.c#L662-L669
+    let data_in = vec![0u8; PKT_V4_SIZE];
+
+    let opts = TestRunOptions {
+        data_in: Some(&data_in),
+        ..TestRunOptions::default()
+    };
+
+    let TestRunResult {
+        return_value,
+        duration,
+        data_size_out: _,
+        ctx_size_out: _,
+    } = prog.test_run(opts).unwrap();
+
+    assert_eq!(return_value, u32::from(EXPECTED_CLASSID));
+    assert!(!duration.is_zero());
+}
--- a/xtask/public-api/aya-ebpf.txt
+++ b/xtask/public-api/aya-ebpf.txt
@@ -2227,6 +2227,7 @@ pub fn aya_ebpf::programs::tc::TcContext::load_bytes(&self, usize, &mut [u8]) ->
 pub const fn aya_ebpf::programs::tc::TcContext::new(core::ptr::non_null::NonNull<aya_ebpf_bindings::x86_64::bindings::__sk_buff>) -> Self
 pub fn aya_ebpf::programs::tc::TcContext::pull_data(&self, u32) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::set_mark(&self, u32)
+pub fn aya_ebpf::programs::tc::TcContext::set_tc_classid(&self, u32)
 pub fn aya_ebpf::programs::tc::TcContext::skb_under_cgroup<M: aya_ebpf::programs::tc::CgroupArrayMap>(&self, &M, u32) -> core::result::Result<bool, aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::store<T>(&self, usize, &T, u64) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::tc::TcContext
@@ -2726,6 +2727,7 @@ pub fn aya_ebpf::programs::tc::TcContext::load_bytes(&self, usize, &mut [u8]) ->
 pub const fn aya_ebpf::programs::tc::TcContext::new(core::ptr::non_null::NonNull<aya_ebpf_bindings::x86_64::bindings::__sk_buff>) -> Self
 pub fn aya_ebpf::programs::tc::TcContext::pull_data(&self, u32) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::set_mark(&self, u32)
+pub fn aya_ebpf::programs::tc::TcContext::set_tc_classid(&self, u32)
 pub fn aya_ebpf::programs::tc::TcContext::skb_under_cgroup<M: aya_ebpf::programs::tc::CgroupArrayMap>(&self, &M, u32) -> core::result::Result<bool, aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::store<T>(&self, usize, &T, u64) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::tc::TcContext
--- a/xtask/public-api/aya-ebpf.txt
+++ b/xtask/public-api/aya-ebpf.txt
@@ -2227,6 +2227,7 @@ pub fn aya_ebpf::programs::tc::TcContext::load_bytes(&self, usize, &mut [u8]) ->
 pub const fn aya_ebpf::programs::tc::TcContext::new(core::ptr::non_null::NonNull<aya_ebpf_bindings::x86_64::bindings::__sk_buff>) -> Self
 pub fn aya_ebpf::programs::tc::TcContext::pull_data(&self, u32) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::set_mark(&self, u32)
+pub fn aya_ebpf::programs::tc::TcContext::set_tc_classid(&self, u16)
 pub fn aya_ebpf::programs::tc::TcContext::skb_under_cgroup<M: aya_ebpf::programs::tc::CgroupArrayMap>(&self, &M, u32) -> core::result::Result<bool, aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::store<T>(&self, usize, &T, u64) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::tc::TcContext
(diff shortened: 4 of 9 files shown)


ramos-99:
Ping @tamird, can you take a look at this when you have the time?

tamird (changes requested):
Reviewed at the author’s explicit request.

`SkBuff::set_tc_classid` is public and `SkBuffContext::skb` is public, so cgroup, socket-filter, and socket-SKB programs can call a setter their verifier rejects. That contradicts the stated TC-only API. Make the `SkBuff` helper crate-private, expose the public setter only on `TcContext`, and update the public-API snapshot accordingly.

`tc_classid_set` would still pass if the setter stopped writing the class ID: it asserts only `TC_ACT_OK` and the already-documented zero output-context size. Either exercise actual qdisc routing or make the test’s verifier-load-only scope explicit and drop the unrelated zero-size assertion.

— tamirdex

ramos-99:
Both addressed in the amended commit (also rebased on main):
`SkBuff::set_tc_classid` is now `pub(crate)` with the public setter on
`TcContext` only, and the public-API snapshot is regenerated. dropped the
`ctx_size_out` assertion and made the load-only scope explicit. the value
can't be observed via `BPF_PROG_TEST_RUN` for sched_cls, so the test only
proves the verifier accepts the write; routing was verified manually as previously mentioned.

I left out real qdisc-routing coverage since the suite has no HTB
infrastructure today, but if you'd prefer it, I'm happy to build it out.

ramos-99:
@tamird friendly ping. mind taking another look when you get a chance?

tamird (changes requested):
> @tamird reviewed 9 files and all commit messages, and made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1595)* status: all files reviewed, 7 unresolved discussions (waiting on ramos-99).

___
*[`ebpf/aya-ebpf/src/programs/sk_buff.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXBah-aaeaGK9HhAK:-P-kXBah-aaeaGK9HhAL:b-p0jlsa) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/ebpf/aya-ebpf/src/programs/sk_buff.rs#L52)):*
> ```XML
> 
>     // Not public: the verifier only accepts writes to `tc_classid` from
>     // sched_cls/sched_act, so this is exposed publicly on `TcContext` only.
> ```

cool; can you add a citation?
___
*[`ebpf/aya-ebpf/src/programs/tc.rs` line 56 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXJNx7lWtdhdoiYFS:-P-kXJNx7lWtdhdoiYFT:b-9rpg66) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/ebpf/aya-ebpf/src/programs/tc.rs#L56)):*
> ```XML
> 
>     /// Sets `__sk_buff::tc_classid`, steering the packet to a traffic-control
>     /// (e.g. HTB) class. Write-only, and only valid for TC classifier programs.
> ```

the second sentence is superfluous, no?
___
*[`test/integration-common/src/lib.rs` line 285 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXQTg3OmvjvdNqzXT:-P-kXQTg3OmvjvdNqzXU:b2vfxh1) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-common/src/lib.rs#L285)):*
> ```XML
> 
> pub mod tc_classid {
>     pub const EXPECTED_CLASSID: u32 = 0x0001_0001;
> ```

why is this in common? it's not read anywhere. it may as well be a literal 0xdead_beef on the ebpf side
___
*[`test/integration-test/src/tests/tc_classid.rs` line 10 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXa9c88TB2DTfZuLE:-P-kXa9d8YqCu3hjCkjq:bu5s0j3) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L10)):*
> ```XML
> 
> // Scope: proves only that the verifier loads a sched_cls program writing
> // `tc_classid`. The written value can't be observed here (BPF_PROG_TEST_RUN
> ```

some citations here would be awesome for e.g. "BPF_PROG_TEST_RUN leaves ..." 
___
*[`test/integration-test/src/tests/tc_classid.rs` line 13 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXfi79dWpBevoi-nm:-P-kXfi79dWpBevoi-nn:b-vjdovd) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L13)):*
> ```XML
> // leaves ctx_out empty for sched_cls, and the qdisc consumes the field), and
> // checking real routing needs HTB-qdisc infra the suite lacks; that path was
> // verified manually.
> ```

"was verified manually" is fine on the commit message, but not here plz
___
*[`test/integration-test/src/tests/tc_classid.rs` line 17 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXkc20GgjbZ1aB0-a:-P-kXkc20GgjbZ1aB0-b:bb6mt9b) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L17)):*
> ```XML
> fn tc_classid_set() {
>     let kernel_version = aya::util::KernelVersion::current().unwrap();
>     // BPF_PROG_TEST_RUN was introduced in v4.12 (1cf1cae963c2, "bpf: introduce
> ```

:+1:
___
*[`test/integration-test/src/tests/tc_classid.rs` line 33 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXpYk2sC-asQok3V3:-P-kXpYk2sC-asQok3V4:b-3h52no) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L33)):*
> ```XML
> 
>     let opts = TestRunOptions {
>         data_in: Some(&data_in),
> ```

is the data buffer actually used? if not you can probably kill PKT_V4_SIZE and if yes then please leave a comment.
___
*[`test/integration-test/src/tests/tc_classid.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXvjmDTKCq6lng5Ad:-P-kXvjmDTKCq6lng5Ae:bpdkvu8) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L38)):*
> ```XML
>     };
> 
>     let TestRunResult { return_value, .. } = prog.test_run(opts).unwrap();
> ```

prefer to explicitly discard the other fields rather than `..`

ebpf/aya-ebpf/src/programs/tc.rs
```diff
@@ -52,6 +52,13 @@ impl TcContext {
         self.skb.set_mark(mark);
     }
 
+    /// Sets `__sk_buff::tc_classid`, steering the packet to a traffic-control
+    /// (e.g. HTB) class. Write-only, and only valid for TC classifier programs.
```
tamird: `tc_classid` is readable by TC programs. Linux's [TC access validator](https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/net/core/filter.c#L9023-L9067) permits reads, and its [context-rewrite selftest](https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/tools/testing/selftests/bpf/prog_tests/ctx_rewrite.c#L103-L107) explicitly covers both a read and a write of this field. Could we remove the write-only claim here and the corresponding verifier claim from the commit message? Keeping this API setter-only is a separate choice.

test/integration-test/src/tests/tc_classid.rs
```diff
@@ -0,0 +1,40 @@
+use aya::{
+    Ebpf, TestRunOptions, TestRunResult,
+    programs::{SchedClassifier, TestRun as _},
+};
+
+// `sizeof(pkt_v4)` = Size(Ethernet) + Size(IPv4) + Size(TCP) = 14 + 20 + 20
+const PKT_V4_SIZE: usize = 14 + 20 + 20;
+
+// Scope: proves only that the verifier loads a sched_cls program writing
+// `tc_classid`. The written value can't be observed here (BPF_PROG_TEST_RUN
```
tamird: `BPF_PROG_TEST_RUN` [supports output contexts](https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/net/bpf/test_run.c#L814-L868) for `sched_cls`; [the SKB conversion](https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/net/bpf/test_run.c#L960-L975) just doesn't export `tc_classid`. This isn't proof that the written value cannot be observed.

Since TC programs can read the field, could the test program report a runtime read back through its return value or a map and assert the expected minor ID? A test-side volatile read would avoid the compiler forwarding the preceding store instead of emitting the BPF load. That would catch a missing or incorrect store without requiring HTB infrastructure; verifying actual qdisc routing is a separate test.

ebpf/aya-ebpf/src/programs/tc.rs
```diff
@@ -52,6 +52,13 @@ impl TcContext {
...
+    #[inline]
+    pub fn set_tc_classid(&self, classid: u32) {
```
tamird: Could both this setter and the crate-private `SkBuff` helper take `u16`, widening with `u32::from(classid)` only for the context-field store? Although the UAPI field is `u32`, the kernel [rewrites this access to a 16-bit store](https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/net/core/filter.c#L9756-L9769). The current `0x0001_0001` test value therefore stores only `1`.

For direct-action classification, [the major bits come from the attached filter's configured classid](https://github.com/torvalds/linux/blob/038d61fd642278bab63ee8ef722c50d10ab01e8f/net/sched/cls_bpf.c#L110-L118), not this setter. Taking `u16` would make the minor-ID contract explicit and prevent callers from passing a full class handle whose major bits are silently discarded. The docs should also mention that attachment requirement.

tamird (commented):
The visibility change addresses the earlier API concern. There are two kernel-contract corrections below, including their implications for the test.

— Codex, on behalf of @tamird

ramos-99:
*[`ebpf/aya-ebpf/src/programs/sk_buff.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXBah-aaeaGK9HhAK:-P0-T17G1iM3rVcOCyjv:b-ov008k) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/ebpf/aya-ebpf/src/programs/sk_buff.rs#L52)):*

cool; can you add a citation?
</blockquote></details>

Added: link to `tc_cls_act_is_valid_access`, which is where the write is whitelisted.

ramos-99:
*[`ebpf/aya-ebpf/src/programs/tc.rs` line 56 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXJNx7lWtdhdoiYFS:-P0-TE5g6BQk8FQ_lZhW:b-ft949b) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/ebpf/aya-ebpf/src/programs/tc.rs#L56)):*

the second sentence is superfluous, no?
</blockquote></details>

Rewrote the doc entirely below: the write-only claim is gone and it now states the minor/major split instead.

ramos-99:
*[`test/integration-common/src/lib.rs` line 285 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXQTg3OmvjvdNqzXT:-P0-TjOe5i0EEc9UEGdx:b-b9x91x) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-common/src/lib.rs#L285)):*

why is this in common? it's not read anywhere. it may as well be a literal 0xdead_beef on the ebpf side
</blockquote></details>

Now read from both sides: the program returns the read-back value, the test asserts against this constant.

ramos-99:
*[`test/integration-test/src/tests/tc_classid.rs` line 10 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXa9c88TB2DTfZuLE:-P0-TqZS69j__rKv1vgk:b-9wasti) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L10)):*

some citations here would be awesome for e.g. "BPF_PROG_TEST_RUN leaves ..."
</blockquote></details>

Rewritten; the ctx_out claim is gone.

ramos-99:
*[`test/integration-test/src/tests/tc_classid.rs` line 13 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXfi79dWpBevoi-nm:-P0-UOewAGQ1uTsrauct:b-m3wme2) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L13)):*

"was verified manually" is fine on the commit message, but not here plz
</blockquote></details>

you are right, thank you. Moved to the commit message.

ramos-99:
*[`test/integration-test/src/tests/tc_classid.rs` line 33 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXpYk2sC-asQok3V3:-P0-UZ222o3ADZxkyhsL:bwcukt0) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L33)):*

is the data buffer actually used? if not you can probably kill PKT_V4_SIZE and if yes then please leave a comment.
</blockquote></details>

Not by the program. The kernel rejects skb construction below ETH_HLEN. Comment + citation added.

ramos-99:
*[`test/integration-test/src/tests/tc_classid.rs` line 38 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXvjmDTKCq6lng5Ad:-P0-UjOs2JSfumHZ5sTR:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/test/integration-test/src/tests/tc_classid.rs#L38)):*

prefer to explicitly discard the other fields rather than `..`
</blockquote></details>

Done.

ebpf/aya-ebpf/src/programs/tc.rs
ramos-99: Confirmed: reads fall through to `bpf_skb_is_valid_access`, kernel emits `BPF_LDX_MEM` for the field. Claim removed from doc and commit message. Left the API setter-only, only use case would be filter chains reading a prior program's write, seems marginal.

ebpf/aya-ebpf/src/programs/tc.rs
ramos-99: Setter and the SkBuff helper both take u16 now, u32::from(classid) only at the write site. Test constant changed to 0xbeef so it fits u16. Doc states the minor/major split and the attachment requirement. Snapshot re-blessed. TcHandle in the userspace API already splits major/minor as u16, so this matches it.

test/integration-test/src/tests/tc_classid.rs
ramos-99: Test now writes the field, reads it back with a volatile load, returns it, and asserts against the minor id. Good call on the store direction, reverting the setter call now fails the test (0 vs expected), writing a different value fails on mismatch too.

ramos-99 (commented):
u16 signature, crate-private SkBuff helper, and the test's read-back are the substantive changes; the rest is doc and citation cleanup addressed per-thread below.

> @ramos-99 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1595#-:-P0-WCDi7D2b8-iv5KQJ:bfxtcnx)* status: 3 of 9 files reviewed, 10 unresolved discussions (waiting on tamird).

ebpf/aya-ebpf/src/programs/tc.rs
tamird: ok. can you document that either in the commit message or a code comment?

ebpf/aya-ebpf/src/programs/tc.rs
tamird: use the citations please?

tamird (changes requested):
> @tamird reviewed 6 files and all commit messages, made 4 comments, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1595)* status: all files reviewed, 4 unresolved discussions (waiting on ramos-99).

___
*[`ebpf/aya-ebpf/src/programs/sk_buff.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXBah-aaeaGK9HhAK:-P0-vUrO45grF6l9wRu0:b-hv8dvp) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/ebpf/aya-ebpf/src/programs/sk_buff.rs#L52)):*
<details><summary><i>Previously, ramos-99 (Martim Ramos) wrote…</i></summary><blockquote>

Added: link to `tc_cls_act_is_valid_access`, which is where the write is whitelisted.
</blockquote></details>

could you use a permalink rather than a mutable tag? 9-digit SHA please

also i don't get how this citation proves this is only accepted from sched_cls and sched_act?
___
*[`test/integration-test/src/tests/tc_classid.rs` line 29 at r2](https://reviewable.io/reviews/aya-rs/aya/1595#-P0-we9Q8wYu97qCVKRv:-P0-we9Q8wYu97qCVKRw:b3litwr) ([raw file](https://github.com/aya-rs/aya/blob/2409c242bb5be8ca8dc79385f516d0836adfc43b/test/integration-test/src/tests/tc_classid.rs#L29)):*
> ```XML
>     // The program never reads the packet, but the kernel refuses to build the
>     // skb from fewer than ETH_HLEN bytes.
>     // https://github.com/torvalds/linux/blob/v6.17/net/bpf/test_run.c#L662-L669
> ```

:+1: permalink please (9-char SHA)

tamird:
@codex review

tamird (commented):
The `u16` setter, crate-private `SkBuff` helper, and volatile read-back test address the earlier implementation findings. Two prose points remain beyond the existing citation requests:

- The PR description still says `tc_classid` is write-only, that `sched_cls` leaves `ctx_size_out` at zero, and that the test only verifies that the write loads. Those are the old claims; the updated test now observes and checks the written value.
- The explanation that a getter would only be useful for reading a previous program's write is too narrow. Classic `cls_bpf` initializes `tc_classid` from the filter's configured class ID before each program runs, so a getter could also inspect that initial minor ID. This does not require adding a getter here, but that explanation should not be recorded as fact. [Kernel source](https://github.com/torvalds/linux/blob/e5f0a698b/net/sched/cls_bpf.c#L90-L105).

The existing visibility-citation request is also justified: the TC allowlist proves TC permits the access, not that the other program types reject it. Their validators contain explicit rejections: [socket filters](https://github.com/torvalds/linux/blob/e5f0a698b/net/core/filter.c#L8727-L8741), [cgroup SKBs](https://github.com/torvalds/linux/blob/e5f0a698b/net/core/filter.c#L8756-L8765), and [SK_SKB](https://github.com/torvalds/linux/blob/e5f0a698b/net/core/filter.c#L9369-L9380).

— Codex, on behalf of @tamird

ramos-99:
*[`ebpf/aya-ebpf/src/programs/sk_buff.rs` line 52 at r1](https://reviewable.io/reviews/aya-rs/aya/1595#-P-kXBah-aaeaGK9HhAK:-P02NxEIBlyphXS6PqIm:b-rvce4i) ([raw file](https://github.com/aya-rs/aya/blob/60ef0208f7b14af3cd3a8357e7f1f9d0ec105fd8/ebpf/aya-ebpf/src/programs/sk_buff.rs#L52)):*

could you use a permalink rather than a mutable tag? 9-digit SHA please

also i don't get how this citation proves this is only accepted from sched_cls and sched_act?
</blockquote></details>

Converted to a permalink. Also added the three rejection citations (socket_filter, cgroup_skb, sk_skb) at e5f0a698b. Confirmed via aya-ebpf-macros that these are the only three program types that construct SkBuffContext, so the four citations are exhaustive.

ebpf/aya-ebpf/src/programs/tc.rs
ramos-99: Both: the setter's doc comment states that cls_bpf pre-populates the field from the filter's classid, with a citation; 

the commit message covers the getter, which isn't added, and what one would return. happy to open an issue regarding the getter if needed.

ebpf/aya-ebpf/src/programs/tc.rs
ramos-99: Added: [store] and [major] link the exact ranges from your comment (e5f0a698b).

ramos-99:
*[`test/integration-test/src/tests/tc_classid.rs` line 29 at r2](https://reviewable.io/reviews/aya-rs/aya/1595#-P0-we9Q8wYu97qCVKRv:-P02P9nPEr0S_PUHJyU2:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/2409c242bb5be8ca8dc79385f516d0836adfc43b/test/integration-test/src/tests/tc_classid.rs#L29)):*

:+1: permalink please (9-char SHA)
</blockquote></details>

Done.

test/integration-ebpf/src/tc_classid.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[classifier]
+fn set_classid(ctx: TcContext) -> i32 {
+    ctx.set_tc_classid(EXPECTED_CLASSID);
+
+    // Read the field back and return it so the test can assert the store
+    // happened. Volatile so the compiler emits the BPF load instead of
+    // forwarding the store above.
+    let skb = ctx.as_ptr().cast::<__sk_buff>();
+    let read_back = unsafe { core::ptr::read_volatile(&raw const (*skb).tc_classid) };
```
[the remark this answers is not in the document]
ramos-99: All run-integration-test jobs pass with this read in place. Reads are permitted for sched_cls; the write-only claim was corrected earlier in this review.

tamird (commented):
The `TcContext::set_tc_classid` docs describe `cls_bpf` pre-populating the field and combining its value with the filter's major class ID without limiting that behavior to the netlink `cls_bpf` attachment path. On Linux 6.6+, Aya's ordinary [`SchedClassifier::attach`](https://github.com/aya-rs/aya/blob/8c6397dbf7d6d9b0a79499cdcfebeb611a4663d9/aya/src/programs/tc.rs#L226-L252) selects TCX. [TCX runs the program directly](https://github.com/torvalds/linux/blob/e5f0a698b/net/core/dev.c#L4330-L4450); the major/minor class-ID combination happens in [`cls_bpf_classify`](https://github.com/torvalds/linux/blob/e5f0a698b/net/sched/cls_bpf.c#L90-L120), which TCX does not call. [HTB uses packet priority or a filter result](https://github.com/torvalds/linux/blob/e5f0a698b/net/sched/sch_htb.c#L219-L253), so calling this setter in a program attached through the default TCX path does not select the described class.

Could the setter docs say explicitly that the class-routing and pre-population behavior requires a netlink `cls_bpf` filter, and point callers to `attach_with_options` with `TcAttachOptions::Netlink(NlOptions { classid: Some(...), .. })` on the intended classful filter? The existing test establishes the field write, but it does not exercise class routing.

— Codex review on behalf of @tamird, with an independent kernel-path check by a delegated reviewer.

xtask/public-api/aya-ebpf.txt
```diff
@@ -2227,6 +2227,7 @@ pub fn aya_ebpf::programs::tc::TcContext::load_bytes(&self, usize, &mut [u8]) ->
 pub const fn aya_ebpf::programs::tc::TcContext::new(core::ptr::non_null::NonNull<aya_ebpf_bindings::x86_64::bindings::__sk_buff>) -> Self
 pub fn aya_ebpf::programs::tc::TcContext::pull_data(&self, u32) -> core::result::Result<(), aya_ebpf_cty::od::c_long>
 pub fn aya_ebpf::programs::tc::TcContext::set_mark(&self, u32)
+pub fn aya_ebpf::programs::tc::TcContext::set_tc_classid(&self, u16)
```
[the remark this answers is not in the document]
tamird: This annoying finding should be gone with recent changes on main, I hope.

tamird (commented):
> @tamird reviewed 8 files and all commit messages, made 1 comment, resolved 6 discussions, and dismissed @chatgpt-codex-connector[bot] from a discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1595)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ramos-99).

tamird:
@codex review

tamird (approved):
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1595)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ramos-99).

PR #1599, opened 2026-06-14
aya, aya-ebpf: add UserRingBuf

Brskt:
Adds `UserRingBuf` to `aya::maps`, `aya_ebpf::maps`, and
`aya_ebpf::btf_maps`, so `BPF_MAP_TYPE_USER_RINGBUF` is no longer routed to
`Map::Unsupported`. It is the user-space-to-kernel counterpart of `RingBuf`:
user space produces samples and an eBPF program drains them.

Userspace `reserve` returns an entry that derefs to `[u8]`; writing the
payload and calling `submit` publishes it, while dropping it discards. The
map implements `AsFd`/`AsRawFd` so it can be polled for writability, and its
producer mapping is sized from `max_entries`.

The eBPF maps expose `drain`, wrapping `bpf_user_ringbuf_drain`, which runs a
callback per sample and stops early on `ControlFlow::Break`, and
`UserRingBufEntry::read`, which copies a value out of a sample. The legacy
and BTF forms share a single drain implementation.

A separate commit fixes reopening a pinned ring buffer: the loader now
refreshes `max_entries` from the kernel rather than the zero declared in the
ELF, without which the producer mapping would be empty. This corrects the
same latent case for `RingBuf`.

Integration tests parametrize over the legacy and BTF variants and cover
submit and discard, drop-discard, callback break, a full ring, wraparound,
short samples, and pinning across a reopen, plus asynchronous writability
through `AsyncFd` and `epoll`. They skip on kernels without user ring buffer
support.

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1599)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -840,6 +840,7 @@ fn parse_map(
         bpf_map_type::BPF_MAP_TYPE_PERF_EVENT_ARRAY => Map::PerfEventArray(map),
         bpf_map_type::BPF_MAP_TYPE_REUSEPORT_SOCKARRAY => Map::ReusePortSockArray(map),
         bpf_map_type::BPF_MAP_TYPE_RINGBUF => Map::RingBuf(map),
+        bpf_map_type::BPF_MAP_TYPE_USER_RINGBUF => Map::UserRingBuf(map),
         bpf_map_type::BPF_MAP_TYPE_SOCKHASH => Map::SockHash(map),
         bpf_map_type::BPF_MAP_TYPE_SOCKMAP => Map::SockMap(map),
         bpf_map_type::BPF_MAP_TYPE_BLOOM_FILTER => Map::BloomFilter(map),
@@ -883,9 +884,11 @@ fn max_entries_override(
     let max_entries = || user_override.unwrap_or_else(&current_value);
     Ok(match map_type {
         bpf_map_type::BPF_MAP_TYPE_PERF_EVENT_ARRAY if max_entries() == 0 => Some(num_cpus()?),
-        bpf_map_type::BPF_MAP_TYPE_RINGBUF => Some(adjust_to_page_size(max_entries(), page_size()))
-            .filter(|adjusted| *adjusted != max_entries())
-            .or(user_override),
+        bpf_map_type::BPF_MAP_TYPE_RINGBUF | bpf_map_type::BPF_MAP_TYPE_USER_RINGBUF => {
+            Some(adjust_to_page_size(max_entries(), page_size()))
+                .filter(|adjusted| *adjusted != max_entries())
+                .or(user_override)
+        }
         _ => user_override,
     })
 }
@@ -898,7 +901,7 @@ fn value_size_override(map_type: bpf_map_type) -> Option<u32> {
         bpf_map_type::BPF_MAP_TYPE_DEVMAP | bpf_map_type::BPF_MAP_TYPE_DEVMAP_HASH => {
             Some(if FEATURES.devmap_prog_id() { 8 } else { 4 })
         }
-        bpf_map_type::BPF_MAP_TYPE_RINGBUF => Some(0),
+        bpf_map_type::BPF_MAP_TYPE_RINGBUF | bpf_map_type::BPF_MAP_TYPE_USER_RINGBUF => Some(0),
         _ => None,
     }
 }
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -861,7 +861,7 @@ impl MapData {
 
     pub(crate) fn create_pinned_by_name<P: AsRef<Path>>(
         path: P,
-        obj: aya_obj::Map,
+        mut obj: aya_obj::Map,
         name: &str,
         btf_fd: Option<BorrowedFd<'_>>,
         inner_map_obj: Option<aya_obj::Map>,
@@ -883,6 +883,15 @@ impl MapData {
             }
         };
         if let Ok(fd) = bpf_get_object(&path_string) {
+            // The pinned map already exists, so its definition lives in the kernel rather than in
+            // `obj`. Ring buffers size their mmap from `max_entries`, but the ELF value is often
+            // zero (resolved at load time), so refresh it from the kernel for those map types.
+            let map_type = obj.map_type();
+            if map_type == bpf_map_type::BPF_MAP_TYPE_RINGBUF as u32
+                || map_type == bpf_map_type::BPF_MAP_TYPE_USER_RINGBUF as u32
+            {
+                obj.set_max_entries(MapInfo::new_from_fd(fd.as_fd())?.max_entries());
+            }
             Ok(Self {
                 obj,
                 fd: MapFd::from_fd(fd),
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -87,6 +87,7 @@ pub mod sk_storage;
 pub mod sock;
 pub mod stack;
 pub mod stack_trace;
+pub mod user_ring_buf;
 pub mod xdp;
 
 pub use array::{Array, CgroupArray, PerCpuArray, ProgramArray};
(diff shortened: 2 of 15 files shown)


tamird (changes requested):
> @tamird+AGNT reviewed 15 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1599)* status: all files reviewed, 5 unresolved discussions (waiting on Brskt).

___
*[a discussion](https://reviewable.io/reviews/aya-rs/aya/1599#-P1pJjdz4Ozcv-WGFrii:-P1pJjdz4Ozcv-WGFrij:bknokgx) (no related file):*
Codex-assisted review of all three commits through `2395e0c1`.

The pinned-map size refresh addresses the described reopen problem. The main issues are concurrent reservation through independently constructed safe producer handles, accepting a sample size that permanently stops draining, and manufacturing the callback instead of invoking the supplied value through the kernel's context argument. I also left two focused test comments: fail promptly on missing records, and make the wrap case actually cross the mapping boundary. The other integration cases largely cover distinct behavior.

The failed amd64 integration job reports `tests::strncmp::bpf_strncmp` in an unchanged test, rather than a UserRingBuf failure ([job log](https://github.com/aya-rs/aya/actions/runs/30303973468/job/90103543341)). I inspected source, kernel contracts, discussions, and existing CI; I did not run local builds or runtime reproducers. The current head also conflicts with `main`.
___
*[`test/integration-test/src/tests/user_ring_buf.rs` line 177 at r3](https://reviewable.io/reviews/aya-rs/aya/1599#-P1pJiqH0Ytc-QIgNgc5:-P1pJiqH0Ytc-QIgNgc6:b6ypnut) ([raw file](https://github.com/aya-rs/aya/blob/2395e0c152743b56e0a2ef86578d408fc0d6b713/test/integration-test/src/tests/user_ring_buf.rs#L177)):*
> ```XML
> fn collect(result: &mut RingBuf<MapData>, n: usize) -> Vec<u64> {
>     let mut seen = Vec::with_capacity(n);
>     while seen.len() < n {
> ```

Please consume the available records until `next()` returns `None`, then compare the resulting vector with the expected values. These callers invoke the uprobe trigger synchronously before `collect`, so draining and echo publication have already completed. If a regression loses a record, nothing will publish it later, and this loop spins forever instead of reporting the missing record. A bounded synchronous read also makes the extra `result.next() == None` assertions unnecessary.
___
*[`test/integration-test/src/tests/user_ring_buf.rs` line 352 at r3](https://reviewable.io/reviews/aya-rs/aya/1599#-P1pJjDg2MWVoerUjFLd:-P1pJjDg2MWVoerUjFLe:b-zc052h) ([raw file](https://github.com/aya-rs/aya/blob/2395e0c152743b56e0a2ef86578d408fc0d6b713/test/integration-test/src/tests/user_ring_buf.rs#L352)):*
> ```XML
>     };
> 
>     // Reserve and drain in rounds smaller than the capacity, so each round leaves room. Enough
> ```

This wraps the cursor, but does not exercise a payload crossing the end of the first mapping. Every sample has an 8-byte payload plus an 8-byte header, so the 16-byte stride divides the 64 KiB ring exactly; the last payload ends exactly at the boundary.

Please arrange for a payload larger than 8 bytes to start near the end and span both mappings, then verify all of its bytes after draining. That exercises the reason for the double mapping and distinguishes this test from the existing capacity/reuse coverage. Merely shifting an 8-byte payload to the boundary is insufficient because `reserve` masks the payload's starting offset back into the first mapping.
___
*[`aya/src/maps/user_ring_buf.rs` line 234 at r2](https://reviewable.io/reviews/aya-rs/aya/1599#-P1pJhvjEUDgHKYvw_jq:-P1pJhvjEUDgHKYvw_jr:b-wz6k78) ([raw file](https://github.com/aya-rs/aya/blob/fcfe54ebf6b3eee9ee0444d1862ff3ea69386c96/aya/src/maps/user_ring_buf.rs#L234)):*
> ```XML
>         // The top two header bits encode the busy and discard flags, so a size that sets them would
>         // be misread by the kernel; reject it like libbpf does.
>         if size as u32 & (BPF_RINGBUF_BUSY_BIT | BPF_RINGBUF_DISCARD_BIT) != 0 {
> ```

Please reject payloads that exceed the kernel's dynptr size limit before publishing a reservation. A 32 MiB ring currently accepts `reserve(1 << 24)`, but the kernel limits dynptr payloads to `(1 << 24) - 1` ([definition and check](https://github.com/torvalds/linux/blob/v6.18/kernel/bpf/helpers.c#L1663-L1705)).

This wedges subsequent draining: [`__bpf_user_ringbuf_peek` validates the size before handling the discard or busy flags](https://github.com/torvalds/linux/blob/v6.18/kernel/bpf/ringbuf.c#L671-L718), returns `-E2BIG`, and leaves the consumer position unchanged. Submitting, explicitly discarding, or dropping this reservation all leave the invalid record at the front of the ring.

Please add an integration case at this boundary that also demonstrates a subsequent valid record can still be drained. Checking only the header flag bits, as libbpf currently does, misses this kernel restriction.
___
*[`aya/src/maps/user_ring_buf.rs` line 251 at r2](https://reviewable.io/reviews/aya-rs/aya/1599#-P1pJhX381XV3Qndd9na:-P1pJhX381XV3Qndd9nb:bnpa5q8) ([raw file](https://github.com/aya-rs/aya/blob/fcfe54ebf6b3eee9ee0444d1862ff3ea69386c96/aya/src/maps/user_ring_buf.rs#L251)):*
> ```XML
>         // position cannot leave this one reserving a stale slot.
>         let producer_pos = unsafe { mmap.ptr().cast::<AtomicUsize>().as_ref() };
>         let pos = producer_pos.load(Ordering::Acquire);
> ```

Reloading the cursor handles sequential producers, but separate safe handles can still reserve the same bytes concurrently. `TryFrom<&Map>` permits two `UserRingBuf` wrappers over one map, and the wrappers are `Send`. Each thread can hold its own `&mut UserRingBuf`, read the same `pos` here, and return a mutable payload backed by the same shared storage. Different reservation sizes can also make the later store move the producer cursor backwards.

The producer exclusivity requirement needs to be enforced across handles to the same map, or made an explicit unsafe contract at the API boundary. A lock belonging only to each wrapper would not solve this. [libbpf explicitly requires callers to synchronize multiple producers](https://github.com/libbpf/libbpf/blob/v1.6.3/src/libbpf.h#L1473-L1492); the safe Rust API currently has no corresponding constraint.

A cursor CAS alone would also need more design: the busy header must be initialized before publishing the advanced cursor to the kernel.
___
*[`ebpf/aya-ebpf/src/maps/user_ring_buf.rs` line 139 at r2](https://reviewable.io/reviews/aya-rs/aya/1599#-P1pJiMJ1hsF70vQtDj2:-P1pJiMJ1hsF70vQtDj3:bcjf4te) ([raw file](https://github.com/aya-rs/aya/blob/fcfe54ebf6b3eee9ee0444d1862ff3ea69386c96/ebpf/aya-ebpf/src/maps/user_ring_buf.rs#L139)):*
> ```XML
>         // SAFETY: `F` is zero-sized (asserted in `drain`), so a value can be produced
>         // without reading any memory.
>         let callback = unsafe { mem::zeroed::<F>() };
> ```

Please pass the supplied callback through `callback_ctx` and invoke it from the generic trampoline. The kernel requires a static callback function, but does not prohibit state: [Linux 6.1 accepts a stack context and forwards it to that callback](https://github.com/torvalds/linux/blob/v6.1/kernel/bpf/ringbuf.c#L666-L726). Using that context removes the artificial noncapturing restriction and the need to manufacture `F` with `zeroed`.

`size_of::<F>() == 0` also does not justify duplicating the value. A move closure can own a zero-sized type with a `Drop` implementation and still implement `Fn`. This trampoline would drop a fabricated copy after every sample, then `drain` would drop the original callback as well. Calling the actual supplied value preserves its ownership semantics and would let a captured accumulator be supported with `FnMut`.

PR #1588, opened 2026-05-29
ci: add missing LTS VM kernels

swananan:
Ideally we would run against every active upstream LTS kernel, but I could only find usable Debian Snapshot cloud kernel packages for the 5.15 and 6.6 lines. These are not the latest upstream LTS patch releases, but they still improve coverage across older supported kernel lines.

I did not add the new kernels to the arm64 lane for now, to avoid increasing PR CI time while GitHub Actions has been unstable recently.

### Added/updated tests?

_We strongly encourage you to add a test for your changes._

- [x] No, and this is why: just some CI change.

### Checklist

- [x] Rust code has been formatted with `cargo +nightly fmt`.
- [x] All clippy lints have been fixed.
      You can find failing lints with `cargo xtask clippy`.
- [x] Unit tests are passing locally with `cargo test`.
- [x] The [Integration tests] are passing locally.
- [x] I have blessed any API changes with `cargo xtask public-api --bless`.

[Integration tests]: https://github.com/aya-rs/aya/blob/main/test/README.md

### (Optional) What GIF best describes this PR or how it makes you feel?

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1588)

PR #1604, opened 2026-06-18
xtask: Rewrite download_kernel_images in Rust

vadorovsky:
Port the shell script to a Rust subcommand that downloads kernel images and debug packages from the Debian mirror. The new command mirrors the the original behavior: fetch Packages.gz/Packages, filter debs by regex, sort by version, and extract .dbg packages.

This eliminates a shell dependency in the development tooling and brings the implementation in line with the existing Rust xtask subcommands.

Assisted-by: hf.co/unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_M

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1604)


tamird:
https://github.com/aya-rs/aya/pull/1588 wants to use Ubuntu because more versions

vadorovsky:
Sorry for missing #1588, it's def in better shape already

PR #1614, opened 2026-06-28
aya-ebpf: keep verifier-safe intrinsics

tamird:
Nightly 1.98 now rejects Aya's void definitions of the standard library
runtime symbols.

Returning the destination pointer satisfies that lint but makes linked
eBPF subprograms return caller stack pointers, which the verifier
rejects. Keep the verifier-safe definitions and expect the lint only on
compilers that provide it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1614)


vadorovsky (approved):
:shipit:

PR #1616, opened 2026-06-29
aya: add Ebpf::take_program

### LINKED ISSUE #809: aya: add take_program API
The maps have an API whereby a user of the library can take a map from the `Bpf`. There is no equivalent API for programs.

alexadamm:
Adds `Ebpf::take_program`, the program-side counterpart of the existing `Ebpf::take_map`.

It moves a program out of the `Ebpf` instance and hands ownership to the caller — useful when a `&`/`&mut` borrow is not enough, e.g. moving a program into an async task. Implementation and docs mirror `take_map` (`self.programs.remove(name)`).

Also updates the `aya` public-API snapshot (`xtask/public-api/aya.txt`).

Closes #809

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1616)

### DIFF
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -1169,6 +1169,29 @@ impl Ebpf {
         self.programs.get_mut(name)
     }
 
+    /// Takes ownership of the program with the given name.
+    ///
+    /// Use this when borrowing with [`program`](crate::Ebpf::program) or
+    /// [`program_mut`](crate::Ebpf::program_mut) is not possible (e.g. when using the
+    /// program from an async task). The caller then owns the returned program and is
+    /// responsible for managing its lifetime; like any [`Program`](crate::programs::Program),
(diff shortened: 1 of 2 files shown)


tamird:
Seems very red

aya/src/bpf.rs
```diff
@@ -1169,6 +1169,29 @@ impl Ebpf {
         self.programs.get_mut(name)
     }
 
+    /// Takes ownership of the program with the given name.
+    ///
+    /// Use this when borrowing with [`program`](crate::Ebpf::program) or
+    /// [`program_mut`](crate::Ebpf::program_mut) is not possible (eg when using the
```
vadorovsky:
```suggestion
    /// [`program_mut`](crate::Ebpf::program_mut) is not possible (e.g. when using the
```

vadorovsky (changes requested):

PR #1621, opened 2026-07-03
aya-ebpf: add ProbeContext::syscall_arg

### LINKED ISSUE #1146: Access first argument of syscall
The libbpf folks did the following here: https://github.com/libbpf/libbpf/commit/9f6e3a7a59894af6ff9221083800556e68ffcd8a

I do not have a arm64 machine currently to test whether accessing the first syscall using aya works, but I stumbled over this while looking into their implementation and though this might be relevant here, as `aya` is just using `arg[0]` for pt_regs.

Also for syscalls on other platforms there are some differences, for example on `x86_64` the 4th (starting from 1) argument is passed via `r10` instaed of `rcx` like normal. 

Maybe a note that `PtRegs` cannot be used for syscall arguments, tests or another struct for Syscall arguments might be useful. Let me know what you prefer.

alexadamm:
Closes #1146.

## Background

On architectures with `CONFIG_ARCH_HAS_SYSCALL_WRAPPER` (arm64, x86-64,
s390, etc.), the kernel wraps each syscall in a function that takes a single
`const struct pt_regs *` argument. To obtain the actual syscall arguments
from a kprobe attached to such a wrapper, users must dereference that
pointer and read registers using the **syscall** calling convention, which
differs from the regular call convention on a few architectures:

| arch    | syscall arg 1    | regular arg 1    | syscall arg 4 | regular arg 4 |
|---------|--------------------------|--------------------------------------|---------------|--------------------------------|
| aarch64 | `orig_x0` (fixed-offset read) | `regs[0]` (overwritten with retval) | `regs[3]`     | `regs[3]` |
| x86-64  | `rdi`                    | `rdi`                                | `r10`         | `rcx` (clobbered by `syscall`) |

This mirrors what libbpf's `BPF_KSYSCALL` / `PT_REGS_PARMn_SYSCALL` macros
do in `bpf_tracing.h`. Only the native syscall ABI is supported; compat
wrappers (e.g. `__ia32_sys_*` on x86-64) pass arguments in a different
register set, matching libbpf's `BPF_KSYSCALL` limitation.

## Changes

* `PtRegsLayout::syscall_arg_reg_offset` — new trait method returning the
  field offset of the `n`th syscall argument; the default implementation
  returns `None` on architectures without syscall argument support.
* `aarch64` override — syscall arg 0 lives in `orig_x0`, which follows the
  ABI-stable `user_pt_regs` struct in the kernel's internal `pt_regs`; it
  is read with `bpf_probe_read_kernel` at a fixed offset.
* `x86_64` override — syscall arg 3 is read from `r10` instead of `rcx`.
* `ProbeContext::syscall_arg` — convenience method that reads the pt_regs
  pointer from `ctx.arg(0)` (the wrapper's only argument) and dispatches to
  the architecture-specific syscall register layout.
* Integration test: `tests::syscall_args::syscall_arg_splice` — a kprobe on
  `__arm64_sys_splice` / `__x64_sys_splice` that records all six syscall
  arguments (exercising the x86-64 `r10` difference) while userspace
  invokes `splice(2)` with sentinel values. The probe filters by
  `TARGET_TGID` so only the test process's calls are recorded. Skips on
  architectures that don't implement syscall wrappers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)

### DIFF
--- a/ebpf/aya-ebpf/src/args.rs
+++ b/ebpf/aya-ebpf/src/args.rs
@@ -1,3 +1,6 @@
+#[cfg(any(bpf_target_arch = "aarch64", bpf_target_arch = "x86_64"))]
+use core::mem::offset_of;
+
 use crate::bindings::{bpf_raw_tracepoint_args, pt_regs};
 
 mod sealed {
@@ -57,6 +60,24 @@ trait PtRegsLayout {
 
     fn arg_reg(&self, index: usize) -> Option<&Self::Reg>;
     fn rc_reg(&self) -> &Self::Reg;
+
+    /// Returns the offset of the field holding the `index`th syscall argument
+    /// within the runtime [`pt_regs`] layout, or `None` if the index is out of
+    /// range. The offset may reference a field that is part of the kernel's
+    /// internal `struct pt_regs` but not of the ABI-stable `user_pt_regs`
+    /// exposed to userspace (e.g. `orig_x0` on `AArch64`).
+    ///
+    /// Callers are expected to read the value using a helper such as
+    /// [`bpf_probe_read_kernel`](crate::helpers::bpf_probe_read_kernel):
+    /// kprobes on syscall wrappers receive a `const struct pt_regs *` argument
+    /// that the verifier tracks as a scalar, so reinterpreting it as
+    /// `&pt_regs` and accessing fields directly is rejected.
+    fn syscall_arg_reg_offset(_index: usize) -> Option<usize>
+    where
+        Self: Sized,
+    {
+        None
+    }
 }
 
 #[cfg(bpf_target_arch = "aarch64")]
@@ -77,6 +98,25 @@ impl PtRegsLayout for pt_regs {
         // Return codes use libbpf's __PT_RC_REG (regs[0]/x0).
         // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L248-L251
         &self.regs[0]
+    }
+
+    fn syscall_arg_reg_offset(index: usize) -> Option<usize> {
+        // `AArch64` syscall arguments differ from regular call arguments only in
+        // the first argument: it lives in `orig_x0`, which is part of the
+        // kernel's internal `struct pt_regs` but NOT of the ABI-stable
+        // `user_pt_regs` that `pt_regs` aliases here.
+        // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/uapi/asm/ptrace.h#L88-L93
+        // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/asm/ptrace.h#L59-L66
+        // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L238-L246
+        let offset = match index {
+            // `orig_x0` immediately follows the `user_pt_regs` portion of the
+            // kernel's `struct pt_regs`.
+            0 => size_of::<Self>(),
+            // syscall args 1..5 live in `regs[1..5]`.
+            n @ 1..=5 => offset_of!(Self, regs) + n * size_of::<Self::Reg>(),
+            _ => return None,
+        };
+        Some(offset)
     }
 }
 
@@ -237,6 +277,25 @@ impl PtRegsLayout for pt_regs {
         // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L148-L152
         &self.rax
     }
+
+    fn syscall_arg_reg_offset(index: usize) -> Option<usize> {
+        // x86-64 syscall arguments differ from regular call arguments only in
+        // the 4th argument (index 3): it is passed in `r10` instead of `rcx`
+        // because `rcx` is clobbered by the `syscall` instruction (it holds the
+        // return address).
+        // https://github.com/torvalds/linux/blob/v6.17/arch/x86/include/asm/ptrace.h#L103-L155
+        // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L93-L102
+        let offset = match index {
+            0 => offset_of!(Self, rdi),
+            1 => offset_of!(Self, rsi),
+            2 => offset_of!(Self, rdx),
+            3 => offset_of!(Self, r10),
+            4 => offset_of!(Self, r8),
+            5 => offset_of!(Self, r9),
+            _ => return None,
+        };
+        Some(offset)
+    }
 }
 
 /// Coerces a `T` from the `n`th argument of a `pt_regs` context where `n` starts
@@ -251,6 +310,38 @@ pub(crate) fn arg<T: Argument>(ctx: &pt_regs, n: usize) -> Option<T> {
         reason = "architecture-specific"
     )]
     Some(T::from_register((*reg) as u64))
+}
+
+/// Coerces a `T` from the `n`th syscall argument of a `pt_regs` context where
+/// `n` starts at 0 and increases by 1 for each successive argument.
+///
+/// Unlike [`arg`], this uses the syscall calling convention rather than the
+/// regular call convention. On some architectures this differs from [`arg`]:
+///
+/// - `AArch64`: the first syscall argument lives in `orig_x0`, not `regs[0]`
+///   (which is overwritten with the return value).
+/// - `x86-64`: the 4th syscall argument (index 3) is passed in `r10` rather
+///   than `rcx` (which is clobbered by the `syscall` instruction).
+///
+/// Reads are performed with [`bpf_probe_read_kernel`] because `pt_regs`
+/// pointers obtained from a kprobe on a syscall wrapper are tracked by the
+/// verifier as scalars and cannot be dereferenced directly.
+///
+/// # Safety
+///
+/// `ctx` must point to a valid `pt_regs` value in kernel memory.
+pub(crate) unsafe fn syscall_arg<T: Argument>(ctx: *const pt_regs, n: usize) -> Option<T> {
+    let offset = pt_regs::syscall_arg_reg_offset(n)?;
+    let ptr = unsafe { ctx.byte_add(offset) }.cast::<u64>();
+    let reg = unsafe { crate::helpers::bpf_probe_read_kernel(ptr).ok() }?;
+    #[expect(clippy::allow_attributes, reason = "architecture-specific")]
+    #[allow(
+        clippy::cast_sign_loss,
+        clippy::unnecessary_cast,
+        trivial_numeric_casts,
+        reason = "architecture-specific"
+    )]
+    Some(T::from_register(reg as u64))
 }
 
 /// Coerces a `T` from the return value of a `pt_regs` context.
--- a/ebpf/aya-ebpf/src/programs/probe.rs
+++ b/ebpf/aya-ebpf/src/programs/probe.rs
@@ -1,6 +1,10 @@
 use core::ffi::c_void;
 
-use crate::{Argument, EbpfContext, args::arg, bindings::pt_regs};
+use crate::{
+    Argument, EbpfContext,
+    args::{arg, syscall_arg},
+    bindings::pt_regs,
+};
 
 pub struct ProbeContext {
     pub regs: *mut pt_regs,
@@ -36,6 +40,50 @@ impl ProbeContext {
     /// ```
     pub fn arg<T: Argument>(&self, n: usize) -> Option<T> {
         arg(unsafe { &*self.regs }, n)
+    }
+
+    /// Returns the `n`th syscall argument passed to the probed function,
+    /// starting from 0.
+    ///
+    /// This is intended for use with kprobes attached to syscall wrapper
+    /// functions (e.g., `__arm64_sys_*` on `AArch64` or `__x64_sys_*` on
+    /// `x86-64`), which are present on kernels built with
+    /// `CONFIG_ARCH_HAS_SYSCALL_WRAPPER`. Such wrappers take a single
+    /// `const struct pt_regs *` argument; retrieve that pointer with
+    /// [`Self::arg`] passing `0` as the index, then this method dereferences it
+    /// and extracts the syscall arguments using the syscall calling convention
+    /// rather than the regular call convention.
+    ///
+    /// On some architectures the syscall register layout differs from the
+    /// regular call convention; see the per-architecture details on the
+    /// `syscall_arg_reg_offset` impl in `aya_ebpf::args`.
+    ///
+    /// Currently this is implemented only for `AArch64` and `x86-64`; on other
+    /// architectures this method will return `None`.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # use aya_ebpf::programs::ProbeContext;
+    /// unsafe fn try_kprobe_sys_kill(ctx: ProbeContext) -> Result<u32, u32> {
+    ///     // `__arm64_sys_kill` / `__x64_sys_kill` take a single
+    ///     // `const struct pt_regs *` argument.
+    ///     let pid: i32 = ctx.syscall_arg(0).ok_or(1u32)?;
+    ///     let sig: i32 = ctx.syscall_arg(1).ok_or(1u32)?;
+    ///     Ok(0)
+    /// }
+    /// ```
+    pub fn syscall_arg<T: Argument>(&self, n: usize) -> Option<T> {
+        // With CONFIG_ARCH_HAS_SYSCALL_WRAPPER, the probed function's only
+        // argument is `const struct pt_regs *regs`. On x86-64 and AArch64 the
+        // kernel selects this config automatically, but on other architectures
+        // (e.g. powerpc, s390) it is user-configurable; libbpf auto-detects the
+        // absence of a wrapper at runtime (see
+        // https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/lib/bpf/bpf_tracing.h#L906).
+        // TODO: auto-detect non-wrapper kernels when support expands beyond
+        // x86-64 and AArch64.
+        let regs: *const pt_regs = self.arg(0)?;
+        unsafe { syscall_arg(regs, n) }
     }
 }
 
--- a/xtask/public-api/aya-ebpf.txt
+++ b/xtask/public-api/aya-ebpf.txt
@@ -1906,6 +1906,7 @@ pub aya_ebpf::programs::probe::ProbeContext::regs: *mut aya_ebpf_bindings::x86_6
 impl aya_ebpf::programs::probe::ProbeContext
 pub fn aya_ebpf::programs::probe::ProbeContext::arg<T: aya_ebpf::Argument>(&self, usize) -> core::option::Option<T>
 pub const fn aya_ebpf::programs::probe::ProbeContext::new(*mut core::ffi::c_void) -> Self
+pub fn aya_ebpf::programs::probe::ProbeContext::syscall_arg<T: aya_ebpf::Argument>(&self, usize) -> core::option::Option<T>
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::probe::ProbeContext
 pub fn aya_ebpf::programs::probe::ProbeContext::as_ptr(&self) -> *mut core::ffi::c_void
 pub fn aya_ebpf::programs::probe::ProbeContext::command(&self) -> core::result::Result<[u8; 16], i32>
@@ -2449,6 +2450,7 @@ pub aya_ebpf::programs::ProbeContext::regs: *mut aya_ebpf_bindings::x86_64::bind
 impl aya_ebpf::programs::probe::ProbeContext
 pub fn aya_ebpf::programs::probe::ProbeContext::arg<T: aya_ebpf::Argument>(&self, usize) -> core::option::Option<T>
 pub const fn aya_ebpf::programs::probe::ProbeContext::new(*mut core::ffi::c_void) -> Self
+pub fn aya_ebpf::programs::probe::ProbeContext::syscall_arg<T: aya_ebpf::Argument>(&self, usize) -> core::option::Option<T>
 impl aya_ebpf::EbpfContext for aya_ebpf::programs::probe::ProbeContext
 pub fn aya_ebpf::programs::probe::ProbeContext::as_ptr(&self) -> *mut core::ffi::c_void
 pub fn aya_ebpf::programs::probe::ProbeContext::command(&self) -> core::result::Result<[u8; 16], i32>
--- a/test/integration-common/src/lib.rs
+++ b/test/integration-common/src/lib.rs
@@ -286,3 +286,36 @@ pub mod test_run {
     pub const IF_INDEX: u32 = 1;
     pub const XDP_MODIFY_LEN: usize = 16;
 }
+
+pub mod syscall_args {
+    /// Index in the `RESULTS` map holding the single captured result.
+    pub const RESULT_INDEX: u32 = 0;
+
+    /// Marker written to [`TestResult::ran`] to distinguish a captured result
+    /// from an uninitialised slot.
+    pub const RAN: u32 = 1;
+
+    #[repr(C)]
+    #[derive(Clone, Copy, Default)]
+    pub struct TestResult {
+        /// Set to [`RAN`] once the kprobe has fired and overwritten the slot.
+        pub ran: u32,
+        /// `fd_in` (`splice` arg 0) as seen by the kprobe.
+        pub fd_in: u64,
+        /// `off_in` (`splice` arg 1) as seen by the kprobe.
+        pub off_in: u64,
+        /// `fd_out` (`splice` arg 2) as seen by the kprobe.
+        pub fd_out: u64,
+        /// `off_out` (`splice` arg 3) as seen by the kprobe.
+        /// On `x86-64` this is passed in `r10` (not `rcx`) for syscalls, so a
+        /// wrong value here indicates the regular calling convention was used.
+        pub off_out: u64,
+        /// `len` (`splice` arg 4) as seen by the kprobe.
+        pub len: u64,
+        /// `flags` (`splice` arg 5) as seen by the kprobe.
+        pub flags: u64,
+    }
+
+    #[cfg(feature = "user")]
+    unsafe impl aya::Pod for TestResult {}
+}
--- a/test/integration-ebpf/Cargo.toml
+++ b/test/integration-ebpf/Cargo.toml
@@ -207,3 +207,7 @@ path = "src/inode_storage.rs"
 [[bin]]
 name = "cgrp_storage"
 path = "src/cgrp_storage.rs"
+
+[[bin]]
+name = "syscall_args"
+path = "src/syscall_args.rs"
--- a/test/integration-ebpf/src/syscall_args.rs
+++ b/test/integration-ebpf/src/syscall_args.rs
@@ -1,0 +1,52 @@
+#![no_std]
+#![no_main]
+#![expect(unused_crate_dependencies, reason = "used in other bins")]
+
+use aya_ebpf::{
+    EbpfContext as _, Global,
+    macros::{kprobe, map},
+    maps::Array,
+    programs::ProbeContext,
+};
+use integration_common::syscall_args::{RAN, RESULT_INDEX, TestResult};
+#[cfg(not(test))]
+extern crate ebpf_panic;
+
+#[unsafe(no_mangle)]
+static TARGET_TGID: Global<u32> = Global::new(0);
+
+#[map]
+static RESULTS: Array<TestResult> = Array::with_max_entries(1, 0);
+
+// Kprobe on `__arm64_sys_splice` / `__x64_sys_splice` that captures all six
+// syscall arguments using `ProbeContext::syscall_arg`. The program filters by
+// `TARGET_TGID` (set from userspace) so only the test process's own `splice`
+// calls are recorded. See `integration_common::syscall_args` for the shared
+// protocol.
+#[kprobe]
+fn syscall_args_splice(ctx: ProbeContext) -> u32 {
+    if ctx.tgid() != TARGET_TGID.load() {
+        return 0;
+    }
+
+    let fd_in: u64 = ctx.syscall_arg(0).unwrap_or(0);
+    let off_in: u64 = ctx.syscall_arg(1).unwrap_or(0);
+    let fd_out: u64 = ctx.syscall_arg(2).unwrap_or(0);
+    let off_out: u64 = ctx.syscall_arg(3).unwrap_or(0);
+    let len: u64 = ctx.syscall_arg(4).unwrap_or(0);
+    let flags: u64 = ctx.syscall_arg(5).unwrap_or(0);
+    if let Some(result) = RESULTS.get_ptr_mut(RESULT_INDEX) {
+        unsafe {
+            *result = TestResult {
+                ran: RAN,
+                fd_in,
+                off_in,
+                fd_out,
+                off_out,
+                len,
+                flags,
+            };
+        }
+    }
+    0
+}
--- a/test/integration-test/src/lib.rs
+++ b/test/integration-test/src/lib.rs
@@ -85,6 +85,7 @@ bpf_file!(
     XSK_MAP => "xsk_map",
     INODE_STORAGE => "inode_storage",
     CGRP_STORAGE => "cgrp_storage",
+    SYSCALL_ARGS => "syscall_args",
 );
 
 #[cfg(test)]
--- a/test/integration-test/src/tests.rs
+++ b/test/integration-test/src/tests.rs
@@ -64,6 +64,7 @@ mod socket_filter;
 mod stack_trace;
 mod stack_trace_lsm;
 mod strncmp;
+mod syscall_args;
 mod tc_netlink;
 mod tcx;
 mod uprobe_cookie;
--- a/test/integration-test/src/tests/syscall_args.rs
+++ b/test/integration-test/src/tests/syscall_args.rs
@@ -1,0 +1,99 @@
+use aya::{
+    EbpfLoader,
+    maps::{Array, MapData},
+    programs::KProbe,
+};
+use integration_common::syscall_args::{RAN, RESULT_INDEX, TestResult};
+
+// Sentinel values passed as the six arguments of the `splice(2)` syscall.
+// The kprobe distinguishes each register slot round-trip through the syscall
+// wrapper. Mirrors the kernel selftest pattern in
+// <https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/testing/selftests/bpf/progs/bpf_syscall_macro.c#L89-L103>.
+//
+// Positive `i32` values are used for `fd_in`/`fd_out` to avoid sign-extension
+// differences between architectures. The offset pointers are invalid non-null
+// sentinels; the kernel will fail with `EFAULT` but the kprobe fires first.
+const SPLICE_FD_IN: u64 = 0x0bad_f00d;
+const SPLICE_FD_OUT: u64 = 0x0f00_d0ad;
+const SPLICE_OFF_IN: u64 = 0x1234_5678_9abc_def0;
+const SPLICE_OFF_OUT: u64 = 0x0fed_cba9_8765_4321;
+const SPLICE_LEN: u64 = 0xcafe_babe_dad0_bad0;
+const SPLICE_FLAGS: u64 = 0x0edb_ab1e;
+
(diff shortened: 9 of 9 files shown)


ebpf/aya-ebpf/src/args.rs
```diff
@@ -1,3 +1,12 @@
+#[cfg(any(bpf_target_arch = "aarch64", bpf_target_arch = "x86_64"))]
+use core::mem::offset_of;
+
```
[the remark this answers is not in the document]
alexadamm: `size_of` is part of the Rust 2024 prelude (`pub use super::v1::*`), and this crate uses `edition = "2024"` (see workspace `Cargo.toml`). That's why I dropped the explicit import after an earlier `redundant_imports` clippy complaint. also the CI confirms `build-test-aya-ebpf` passes on the aarch64 target.

swananan (commented):
The CI is failing on the public API check. You may need to run `cargo xtask public-api --bless` to fix it.

> @swananan made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621#-:-OwfxemwEfjEpBWmsLcG:b-ue0h6q)* status: 0 of 8 files reviewed, 7 unresolved discussions (waiting on alexadamm).

___
*[`-- commits` line 13 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owhvkvt4dRNerVdphhE:-Owhvkvt4dRNerVdphhF:bhkn24y):*
Should we keep these implementation details in the commit message? Maybe it’s enough to mention the supported platforms.
___
*[`-- commits` line 18 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OweyPxZDB0UfbP0JsRs:-OweyPxZDB0UfbP0JsRt:b-w66gea):*
Could we use `Fixes #1146.` here for the consistency?
___
*[`ebpf/aya-ebpf/src/args.rs` line 10 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OwfoULn5f_G_YkIzH6x:-OwfoULn5f_G_YkIzH6y:bc4ve2b) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/args.rs#L10)):*
> ```XML
> ))]
> use crate::bindings::__u64;
> use crate::bindings::{bpf_raw_tracepoint_args, pt_regs};
> ```

Could we put these imports on the same line?



_Code quote:_
```XML
use crate::bindings::__u64;
use crate::bindings::{bpf_raw_tracepoint_args, pt_regs};
```
___
*[`ebpf/aya-ebpf/src/args.rs` line 330 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owg1Ad7AWAi2GzpzMAA:-Owg1Ad7AWAi2GzpzMAB:b-6ny3js) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/args.rs#L330)):*
> ```XML
> ///   (which is overwritten with the return value).
> /// - `x86-64`: the 4th syscall argument (index 3) is passed in `r10` rather
> ///   than `rcx` (which is clobbered by the `syscall` instruction).
> ```

Should we keep these details in the syscall_arg docs? They seem to be already covered in syscall_arg_reg_offset.



_Code quote:_
```XML
/// - `AArch64`: the first syscall argument lives in `orig_x0`, not `regs[0]`
///   (which is overwritten with the return value).
/// - `x86-64`: the 4th syscall argument (index 3) is passed in `r10` rather
///   than `rcx` (which is clobbered by the `syscall` instruction).
```
___
*[`ebpf/aya-ebpf/src/programs/probe.rs` line 79 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owg2grfD-Jz-OBSVDk-:-Owg2grfD-Jz-OBSVDk0:b75wr4q) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/programs/probe.rs#L79)):*
> ```XML
>     /// }
>     /// ```
>     pub fn syscall_arg<T: Argument>(&self, n: usize) -> Option<T> {
> ```

I was thinking about the API design here. Although libbpf uses separate APIs for normal function arguments and syscall arguments, aya could potentially use a single `arg` API and choose the calling convention based on the program context. For example, `kprobe` sections would use the normal function calling convention, while future `ksyscall` sections would use the syscall calling convention.

But Aya does not support `ksyscall` sections yet, so this is probably more of a long-term direction than something to solve in this PR.
___
*[`ebpf/aya-ebpf/src/programs/probe.rs` line 81 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owg2U5EDWolxMv1GHQg:-Owg2U5EDWolxMv1GHQh:bca8hgc) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/programs/probe.rs#L81)):*
> ```XML
>     pub fn syscall_arg<T: Argument>(&self, n: usize) -> Option<T> {
>         // With CONFIG_ARCH_HAS_SYSCALL_WRAPPER, the probed function's only
>         // argument is `const struct pt_regs *regs`.
> ```

Should we add a TODO or more details here?

This currently assumes syscall wrappers. That is fine for x86-64 and AArch64, where `CONFIG_ARCH_HAS_SYSCALL_WRAPPER` is selected by the architecture rather than user-configurable, but libbpf auto-detects this and we may need the same if support expands beyond these platforms.

https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/lib/bpf/bpf_tracing.h#L906
___
*[`test/integration-ebpf/src/syscall_args.rs` line 27 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OwhxIwhC1vmPk4yYmrh:-OwhxIwhC1vmPk4yYmri:b3zfrmd) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/test/integration-ebpf/src/syscall_args.rs#L27)):*
> ```XML
> // verify the values match those that were used to invoke it.
> #[kprobe]
> fn syscall_args_kill(ctx: ProbeContext) -> u32 {
> ```

Would it be worth using the `splice` syscall here? Since it has 6 parameters and kernel selftests also use this.

https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/testing/selftests/bpf/progs/bpf_syscall_macro.c#L89-L103

alexadamm (commented):
All CIs have been successful, the regenerated `xtask/public-api/aya-ebpf.txt` changes via `cargo xtask public-api --bless` are included in this commit

> @alexadamm made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621#-:-OwlBss451hzkXRoQcN_:bz55ov5)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on swananan).

___
*[`-- commits` line 13 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owhvkvt4dRNerVdphhE:-Owl6v-i63BovWXVt6MI:b-896fix):*

Should we keep these implementation details in the commit message? Maybe it’s enough to mention the supported platforms.
</blockquote></details>

Done.
___
*[`-- commits` line 18 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OweyPxZDB0UfbP0JsRs:-Owl6tel2mld04nu7ESJ:b-896fix):*

Could we use `Fixes #1146.` here for the consistency?
</blockquote></details>

Done.
___
*[`ebpf/aya-ebpf/src/args.rs` line 10 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OwfoULn5f_G_YkIzH6x:-Owl8ygyDu2V_tFYuHo3:b-yct4i7) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/args.rs#L10)):*

Could we put these imports on the same line?
</blockquote></details>

I've tried merging `__u64` into the unconditional `use crate::bindings::{__u64, bpf_raw_tracepoint_args, pt_regs};`. It produces `warning: unused import: __u64` on x86-64 where `__u64` isn't used in type Reg. Individual items inside a `use {}` group cannot carry `#[cfg]` attributes in stable Rust, so the `cfg` gated import must remain separate... so I left as is.
___
*[`ebpf/aya-ebpf/src/args.rs` line 330 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owg1Ad7AWAi2GzpzMAA:-Owl9f48EC-SHDmSIL7U:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/args.rs#L330)):*

Should we keep these details in the syscall_arg docs? They seem to be already covered in syscall_arg_reg_offset.
</blockquote></details>

Done.
___
*[`ebpf/aya-ebpf/src/programs/probe.rs` line 79 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owg2grfD-Jz-OBSVDk-:-Owl9uYDEfvHv9kZyC1K:b-83od3n) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/programs/probe.rs#L79)):*

I was thinking about the API design here. Although libbpf uses separate APIs for normal function arguments and syscall arguments, aya could potentially use a single `arg` API and choose the calling convention based on the program context. For example, `kprobe` sections would use the normal function calling convention, while future `ksyscall` sections would use the syscall calling convention.

But Aya does not support `ksyscall` sections yet, so this is probably more of a long-term direction than something to solve in this PR.
</blockquote></details>

Noted
___
*[`ebpf/aya-ebpf/src/programs/probe.rs` line 81 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-Owg2U5EDWolxMv1GHQg:-OwlAaiY282S0y7ksR2h:b9d2oml) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/programs/probe.rs#L81)):*

Should we add a TODO or more details here?

This currently assumes syscall wrappers. That is fine for x86-64 and AArch64, where `CONFIG_ARCH_HAS_SYSCALL_WRAPPER` is selected by the architecture rather than user-configurable, but libbpf auto-detects this and we may need the same if support expands beyond these platforms.

https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/lib/bpf/bpf_tracing.h#L906
</blockquote></details>

Done, added an inline TODO comment with that refererence
___
*[`test/integration-ebpf/src/syscall_args.rs` line 27 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OwhxIwhC1vmPk4yYmrh:-OwlB2Y14nfG6uZYDohH:b3ehpjr) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/test/integration-ebpf/src/syscall_args.rs#L27)):*

Would it be worth using the `splice` syscall here? Since it has 6 parameters and kernel selftests also use this.

https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/testing/selftests/bpf/progs/bpf_syscall_macro.c#L89-L103
</blockquote></details>

Done, switched to `splice`

swananan (commented):
> @swananan reviewed all commit messages, made 3 comments, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on alexadamm).

___
*[`ebpf/aya-ebpf/src/args.rs` line 10 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OwfoULn5f_G_YkIzH6x:-Owm7G1aDnSiZlEVosJA:b-7syp1h) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/args.rs#L10)):*
<details><summary><i>Previously, alexadamm (Alexander Adam) wrote…</i></summary><blockquote>

I've tried merging `__u64` into the unconditional `use crate::bindings::{__u64, bpf_raw_tracepoint_args, pt_regs};`. It produces `warning: unused import: __u64` on x86-64 where `__u64` isn't used in type Reg. Individual items inside a `use {}` group cannot carry `#[cfg]` attributes in stable Rust, so the `cfg` gated import must remain separate... so I left as is.
</blockquote></details>

Ah, now I see why the original implementation used `type Reg = crate::bindings::__u64` . Maybe we can go back to that and use `Self::Reg` for the offset calculation.
___
*[`test/integration-common/src/lib.rs` line 312 at r7](https://reviewable.io/reviews/aya-rs/aya/1621#-OwmumQK-eemC4FMc00s:-OwmumQK-eemC4FMc00t:b-rho6ec) ([raw file](https://github.com/aya-rs/aya/blob/e52209c609943adab3fbacbc9691f18d802d3483/test/integration-common/src/lib.rs#L312)):*
> ```XML
>     pub const SPLICE_OFF_OUT: u64 = 0x0fed_cba9_8765_4321;
>     pub const SPLICE_LEN: u64 = 0xcafe_babe_dad0_bad0;
>     pub const SPLICE_FLAGS: u64 = 0x0edb_ab1e;
> ```

It looks like these variables are only used by the userspace test, so we don't need to put them in `integration-common`.



_Code quote:_
```XML
    pub const SPLICE_FD_IN: u64 = 0x0bad_f00d;
    pub const SPLICE_FD_OUT: u64 = 0x0f00_d0ad;
    pub const SPLICE_OFF_IN: u64 = 0x1234_5678_9abc_def0;
    pub const SPLICE_OFF_OUT: u64 = 0x0fed_cba9_8765_4321;
    pub const SPLICE_LEN: u64 = 0xcafe_babe_dad0_bad0;
    pub const SPLICE_FLAGS: u64 = 0x0edb_ab1e;
```
___
*[`test/integration-common/src/lib.rs` line 315 at r7](https://reviewable.io/reviews/aya-rs/aya/1621#-Own0CpI-J78O7TLli0R:-Own0CpI-J78O7TLli0S:b-io56d4) ([raw file](https://github.com/aya-rs/aya/blob/e52209c609943adab3fbacbc9691f18d802d3483/test/integration-common/src/lib.rs#L315)):*
> ```XML
> 
>     #[repr(C)]
>     #[derive(Clone, Copy, Default, Debug, Eq, PartialEq)]
> ```

Do we need `Debug`, `Eq`, and `PartialEq` here?

alexadamm (commented):
> @alexadamm made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on swananan).

___
*[`ebpf/aya-ebpf/src/args.rs` line 10 at r4](https://reviewable.io/reviews/aya-rs/aya/1621#-OwfoULn5f_G_YkIzH6x:-OwpN_ofEf3J6htDGpHI:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/b06522e4ca364c683d637e233bceedaa231e0a68/ebpf/aya-ebpf/src/args.rs#L10)):*

Ah, now I see why the original implementation used `type Reg = crate::bindings::__u64` . Maybe we can go back to that and use `Self::Reg` for the offset calculation.
</blockquote></details>

Done.
___
*[`test/integration-common/src/lib.rs` line 312 at r7](https://reviewable.io/reviews/aya-rs/aya/1621#-OwmumQK-eemC4FMc00s:-OwpMheWEoyTmz0WB8SO:b-vuneeo) ([raw file](https://github.com/aya-rs/aya/blob/e52209c609943adab3fbacbc9691f18d802d3483/test/integration-common/src/lib.rs#L312)):*

It looks like these variables are only used by the userspace test, so we don't need to put them in `integration-common`.
</blockquote></details>

Moved all 6 `SPLICE_*` constants from integration-common to the userspace test file (`test/integration-test/src/tests/syscall_args.rs`). Only `RESULT_INDEX`, `RAN`, and `TestResult` remain in `integration-common` (shared between eBPF and userspace)
___
*[`test/integration-common/src/lib.rs` line 315 at r7](https://reviewable.io/reviews/aya-rs/aya/1621#-Own0CpI-J78O7TLli0R:-OwpMqIIBpTecQMEVRMf:b7qh5vr) ([raw file](https://github.com/aya-rs/aya/blob/e52209c609943adab3fbacbc9691f18d802d3483/test/integration-common/src/lib.rs#L315)):*

Do we need `Debug`, `Eq`, and `PartialEq` here?
</blockquote></details>

Removed `Debug`, `Eq`, and `PartialEq` since the eBPF side doesn't need formatting/equality, and userspace test compares individual `u64` instead of the whole struct

swananan (commented):
LGTM from my side, but I am not a maintainer, leaving the final call to @tamird and @vadorovsky

> @swananan reviewed 1 file and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621#-:-OwpyhJq45fYNWd8wDC7:b72uan6)* status: 1 of 9 files reviewed, 1 unresolved discussion.

swananan:
Hi @alexadamm, could you please address the conflicts? Thanks!

@tamird @vadorovsky Could you take a look when you get a change?

tamird (changes requested):
> @tamird reviewed 9 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)* status: all files reviewed, 6 unresolved discussions (waiting on alexadamm and swananan).

___
*[`test/integration-test/src/tests/syscall_args.rs` line 92 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P0-xvtl-j7NVCE5nNkj:-P0-xvtl-j7NVCE5nNkk:b-gm9omy) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/test/integration-test/src/tests/syscall_args.rs#L92)):*
> ```XML
> 
>     let result = results.get(&RESULT_INDEX, 0).unwrap();
>     assert_eq!(result.ran, RAN, "kprobe did not fire");
> ```

could you destructure this please?
___
*[`ebpf/aya-ebpf/src/args.rs` line 77 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-OxLGoo0AnmNawplruqi:-OxLGoo0AnmNawplruqj:b5u0onw) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L77)):*
> ```XML
>     fn syscall_arg_reg_offset(_index: usize) -> Option<usize>
>     where
>         Self: Sized,
> ```

why this bound? just document here (or work it into the comment above)
___
*[`ebpf/aya-ebpf/src/args.rs` line 79 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-OxLGm7o4mdRCb-xZ9bd:-OxLGm7o4mdRCb-xZ9be:b-r1kw02) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L79)):*
> ```XML
>         Self: Sized,
>     {
>         None
> ```

does this need to be provided? why? if not, can you un-underscore `index` above?
___
*[`ebpf/aya-ebpf/src/args.rs` line 110 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P-yyDwp55Lhvkw3ApxE:-P-yyDwp55Lhvkw3ApxF:bx80czb) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L110)):*
> ```XML
>         // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/uapi/asm/ptrace.h#L88-L93
>         // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/asm/ptrace.h#L59-L66
>         // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L238-L246
> ```

can you please use permalinks (9-char SHA) instead of mutable tags?



_Code quote:_
```XML
        // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/uapi/asm/ptrace.h#L88-L93
        // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/asm/ptrace.h#L59-L66
        // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L238-L246
```
___
*[`ebpf/aya-ebpf/src/args.rs` line 113 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P-yypgx2S8JRvnNlmuJ:-P-yypgx2S8JRvnNlmuK:b-zbcfh1) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L113)):*
> ```XML
>         let offset = match index {
>             // `orig_x0` immediately follows the `user_pt_regs` portion of the
>             // kernel's `struct pt_regs`.
> ```

could you break up the comment above so that this here has its own citation?



_Code quote:_
```XML
            // `orig_x0` immediately follows the `user_pt_regs` portion of the
            // kernel's `struct pt_regs`.
```

tamird:
@codex review

ebpf/aya-ebpf/src/programs/probe.rs
```diff
@@ -37,6 +41,50 @@ impl ProbeContext {
...
+    /// `const struct pt_regs *` argument; retrieve that pointer with
+    /// [`Self::arg`] passing `0` as the index, then this method dereferences it
+    /// and extracts the syscall arguments using the syscall calling convention
+    /// rather than the regular call convention.
+    ///
+    /// On some architectures the syscall register layout differs from the
+    /// regular call convention; see the per-architecture details on the
+    /// `syscall_arg_reg_offset` impl in `aya_ebpf::args`.
+    ///
+    /// Currently this is implemented only for `AArch64` and `x86-64`; on other
```
tamird, line 69: Could the public contract explicitly limit this to the native syscall ABI? On an x86-64 kernel, `__ia32_sys_*` wrappers also take a `pt_regs *`, but their arguments are in `EBX, ECX, EDX, ESI, EDI, EBP`, not the native registers selected here. A caller probing one of those wrappers therefore gets unrelated register values rather than `None`, despite satisfying the stated architecture/wrapper conditions. [Kernel ABI mappings](https://github.com/torvalds/linux/blob/e5f0a698b/arch/x86/include/asm/syscall_wrapper.h#L14-L24).

libbpf explicitly calls out compat syscalls as a limitation of `BPF_KSYSCALL`; the same scope restriction belongs in these public docs. [libbpf documentation](https://github.com/torvalds/linux/blob/e5f0a698b/tools/lib/bpf/bpf_tracing.h#L885-L895).

ebpf/aya-ebpf/src/args.rs
```diff
@@ -78,6 +99,25 @@ impl PtRegsLayout for pt_regs {
...
+        // `user_pt_regs` that `pt_regs` aliases here.
+        // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/uapi/asm/ptrace.h#L88-L93
+        // https://github.com/torvalds/linux/blob/v6.17/arch/arm64/include/asm/ptrace.h#L59-L66
+        // https://github.com/torvalds/linux/blob/v6.17/tools/lib/bpf/bpf_tracing.h#L238-L246
+        let offset = match index {
+            // `orig_x0` immediately follows the `user_pt_regs` portion of the
+            // kernel's `struct pt_regs`.
+            0 => size_of::<Self>(),
+            // syscall args 1..5 live in `regs[1..5]`.
+            n @ 1..=5 => offset_of!(Self, regs) + n * size_of::<Self::Reg>(),
```
tamird, line 126: This matches libbpf's convention, but it inherits an AArch64 kernel limitation worth documenting: on kernels without the `orig_x0` synchronization fix, a ptrace syscall-entry stop can change `regs[0]` without updating `orig_x0`. The older native wrapper invokes the syscall with the changed `regs[0]`, while this accessor reports the original value. The ordinary `splice` test does not exercise that case.

The discrepancy and its fix are described in [e057b9477](https://github.com/torvalds/linux/commit/e057b94772328221405b067c3a85fe479b915dc8). The later [5a87e8c7f](https://github.com/torvalds/linux/commit/5a87e8c7f3702eba8a1ed17431a807e01da3ae51) makes `orig_x0` authoritative for native syscall invocation. This is an inherited limitation on unfixed kernels, not a claim that the offset or the general register mapping is wrong.

ebpf/aya-ebpf/src/programs/probe.rs
```diff
@@ -37,6 +41,50 @@ impl ProbeContext {
...
+    ///     let pid: i32 = ctx.syscall_arg(0).ok_or(1u32)?;
+    ///     let sig: i32 = ctx.syscall_arg(1).ok_or(1u32)?;
+    ///     Ok(0)
+    /// }
+    /// ```
+    pub fn syscall_arg<T: Argument>(&self, n: usize) -> Option<T> {
+        // With CONFIG_ARCH_HAS_SYSCALL_WRAPPER, the probed function's only
+        // argument is `const struct pt_regs *regs`. On x86-64 and AArch64 the
+        // kernel selects this config automatically, but on other architectures
+        // (e.g. powerpc, s390) it is user-configurable; libbpf auto-detects the
```
tamird: s390 is not an example of a user-configurable wrapper choice: its architecture configuration unconditionally selects `ARCH_HAS_SYSCALL_WRAPPER`. The TODO about handling non-wrapper kernels can stand without this inaccurate example. [s390 Kconfig](https://github.com/torvalds/linux/blob/e5f0a698b/arch/s390/Kconfig#L102).

tamird (commented):
The native x86-64/AArch64 register mappings and helper-based reads look consistent with the kernel, and the six distinct `splice` arguments exercise the x86-64 `r10` difference. The inline comments cover two compatibility/documentation limits and one inaccurate configuration claim; they do not ask for additional architecture support.

The PR description also needs refreshing: this version uses `syscall_arg_reg_offset` with a default of `None`, performs fixed-offset helper reads rather than CO-RE reads, and tests all six `splice` arguments rather than two `kill` arguments.

— Codex, on behalf of @tamird

tamird:
> The PR description also needs refreshing: this version uses `syscall_arg_reg_offset` with a default of `None`, performs fixed-offset helper reads rather than CO-RE reads, and tests all six `splice` arguments rather than two `kill` arguments.

Human here: FWIW I don't care about the PR description, it can just say "see commit message" as far as I'm concerned.

xtask/public-api/aya-ebpf.txt
```diff
@@ -1906,6 +1906,7 @@ pub aya_ebpf::programs::probe::ProbeContext::regs: *mut aya_ebpf_bindings::x86_6
 impl aya_ebpf::programs::probe::ProbeContext
 pub fn aya_ebpf::programs::probe::ProbeContext::arg<T: aya_ebpf::Argument>(&self, usize) -> core::option::Option<T>
 pub const fn aya_ebpf::programs::probe::ProbeContext::new(*mut core::ffi::c_void) -> Self
+pub fn aya_ebpf::programs::probe::ProbeContext::syscall_arg<T: aya_ebpf::Argument>(&self, usize) -> core::option::Option<T>
```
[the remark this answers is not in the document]
alexadamm: AGENTS.md was updated in 57326006 to require fixture updates in the same PR.

ebpf/aya-ebpf/src/args.rs
```diff
@@ -57,6 +60,25 @@ trait PtRegsLayout {
...
+    /// [`bpf_probe_read_kernel`](crate::helpers::bpf_probe_read_kernel):
+    /// kprobes on syscall wrappers receive a `const struct pt_regs *` argument
+    /// that the verifier tracks as a scalar, so reinterpreting it as
+    /// `&pt_regs` and accessing fields directly is rejected.
+    ///
+    /// The default implementation returns `None`; architectures that support
+    /// syscall argument extraction override this method.
+    fn syscall_arg_reg_offset(_index: usize) -> Option<usize> {
+        None
+    }
```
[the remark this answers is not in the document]
alexadamm: PR description updated.

ebpf/aya-ebpf/src/args.rs
alexadamm: Done.

ebpf/aya-ebpf/src/programs/probe.rs
alexadamm: Done.

ebpf/aya-ebpf/src/programs/probe.rs
alexadamm: Done.

alexadamm (commented):
> @alexadamm made 8 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)* status: 2 of 12 files reviewed, 9 unresolved discussions (waiting on swananan and tamird).

___
*[`ebpf/aya-ebpf/src/args.rs` line 77 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-OxLGoo0AnmNawplruqi:-P0DlkiD-SRX9KniCAGs:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L77)):*

why this bound? just document here (or work it into the comment above)
</blockquote></details>

Done.
___
*[`ebpf/aya-ebpf/src/args.rs` line 79 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-OxLGm7o4mdRCb-xZ9bd:-P0DlLO59sBpQ1B4oPrF:b-3tidwn) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L79)):*

does this need to be provided? why? if not, can you un-underscore `index` above?
</blockquote></details>

The default needs to exist so architectures without an override still compile; it returns `None` unconditionally, so `index` is unused there and stays `_index`.
___
*[`ebpf/aya-ebpf/src/args.rs` line 110 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P-yyDwp55Lhvkw3ApxE:-P0Dlm2UB_ZCin7V-wRG:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L110)):*

can you please use permalinks (9-char SHA) instead of mutable tags?
</blockquote></details>

Done.
___
*[`ebpf/aya-ebpf/src/args.rs` line 113 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P-yypgx2S8JRvnNlmuJ:-P0Dlmd-52E_Yc5Be9k4:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L113)):*

could you break up the comment above so that this here has its own citation?
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/syscall_args.rs` line 92 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P0-xvtl-j7NVCE5nNkj:-P0DmB6uBRe6lVPHEqIi:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/test/integration-test/src/tests/syscall_args.rs#L92)):*

could you destructure this please?
</blockquote></details>

Done.

swananan (commented):
> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)* status: 2 of 12 files reviewed, 10 unresolved discussions (waiting on alexadamm and tamird).

___
*[`ebpf/aya-ebpf/src/args.rs` line 110 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P-yyDwp55Lhvkw3ApxE:-P0NPvr64kvFZAa96JH5:b-rq0adi) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L110)):*
<details><summary><i>Previously, alexadamm (Alexander Adam) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

This is not done yet. A lot URLs return 404, and could you use a 8/9 chars SHA URL instead?
For example, https://github.com/torvalds/linux/commit/5a87e8c7
___
*[`test/integration-test/src/tests/syscall_args.rs` line 11 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P0NQanh5tCUkV5GomgB:-P0NQanh5tCUkV5GomgC:b-wbfzjg) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/test/integration-test/src/tests/syscall_args.rs#L11)):*
> ```XML
> // The kprobe distinguishes each register slot round-trip through the syscall
> // wrapper. Mirrors the kernel selftest pattern in
> // <https://github.com/torvalds/linux/blob/e5f0a698b34ed76002dc5cff3804a61c80233a7a/tools/testing/selftests/bpf/progs/bpf_syscall_macro.c#L89-L103>.
> ```

Use this URL: https://github.com/torvalds/linux/blob/e5f0a698b/tools/testing/selftests/bpf/progs/bpf_syscall_macro.c#L89-L103

alexadamm (commented):
> @alexadamm made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1621)* status: 2 of 12 files reviewed, 10 unresolved discussions (waiting on swananan and tamird).

___
*[`ebpf/aya-ebpf/src/args.rs` line 110 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P-yyDwp55Lhvkw3ApxE:-P0Xc7uj9M56iK6V2j4S:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/ebpf/aya-ebpf/src/args.rs#L110)):*

This is not done yet. A lot URLs return 404, and could you use a 8/9 chars SHA URL instead?
For example, https://github.com/torvalds/linux/commit/5a87e8c7
</blockquote></details>

Done.
___
*[`test/integration-test/src/tests/syscall_args.rs` line 11 at r8](https://reviewable.io/reviews/aya-rs/aya/1621#-P0NQanh5tCUkV5GomgB:-P0XcBJT165ByAyPngws:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/15db40337fd0c08ddcdf0a58eaf13b89b35087b3/test/integration-test/src/tests/syscall_args.rs#L11)):*

Use this URL: https://github.com/torvalds/linux/blob/e5f0a698b/tools/testing/selftests/bpf/progs/bpf_syscall_macro.c#L89-L103
</blockquote></details>

Done.

swananan:
Hi @alexadamm , could you address the conflicts? Thanks!

swananan:
@tamird all comments are addressed, PTAL, thanks.

HiChat-fog:
riscv64 selects `ARCH_HAS_SYSCALL_WRAPPER` too ([arch/riscv/Kconfig#L52](https://github.com/torvalds/linux/blob/e5f0a698b/arch/riscv/Kconfig#L52)), so with `syscall_arg_reg_offset` defaulting to `None`, `ProbeContext::syscall_arg` silently returns `None` there — an architecture aya builds for in CI (`build-test-aya-ebpf` includes `riscv64`).

libbpf reads the first riscv syscall argument from `orig_a0` ([bpf_tracing.h#L355](https://github.com/torvalds/linux/blob/e5f0a698b/tools/lib/bpf/bpf_tracing.h#L355), [#L369](https://github.com/torvalds/linux/blob/e5f0a698b/tools/lib/bpf/bpf_tracing.h#L369)), the same shape as `orig_x0` on AArch64. aya's riscv64 `pt_regs` aliases `user_regs_struct`, which has no `orig_a0`, so it would need the same fixed-offset read the AArch64 path uses.
Just flagging it in case you want the docs to note the covered architectures

swananan:
@HiChat-fog Yeah, this PR only intends to support amd64 and arm, the main reason is that aya don't have CI coverage for the other arch.

swananan:
Thinking about this some more, we currently return None to indicate that the arch is not supported, but it has two drawbacks:

1. It's the runtime check rather than the compile time check.
2. The invalid index also returns None, which could be misleading.

I'm thinking whether we should use cfg-gate to enforce the unsupported arch failing during compile time. @alexadamm WDYT?

PR #1623, opened 2026-07-04
aya-log: add {:s} format hint for null-terminated byte arrays

gmreads:
Add a new `{:s}` display hint to support logging fixed-size null-terminated byte arrays as human-readable strings. The primary use case is logging process names captured via `bpf_get_current_comm()`, which returns `[u8; 16]`.

Previously there was no way to log a process name from a BPF program using aya-log-ebpf — `[u8; 16]` with `{}` produced no output (Err(()) in the userspace formatter), and `{:i}` incorrectly formatted it as an IPv6 address.

With this change:
  let comm = bpf_get_current_comm().unwrap_or([0u8; 16]);
  info!(ctx, "pid={} comm={:s}", pid, comm);

produces userspace output like: pid=1234 comm=curl

Changes:
- aya-log-common: add Str variant to DisplayHint enum and StrFormatter trait
- aya-log-parser: recognise "s" as a valid format hint
- aya-log: implement Format for [u8; 16] with DisplayHint::Str, decoding bytes up to the first null as lossy UTF-8

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1623)


swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1623)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on gmreads).

___
*[`aya-log-common/src/lib.rs` line 182 at r1](https://reviewable.io/reviews/aya-rs/aya/1623#-OwsDZQ40Q2Oe7TUdkW1:-OwsDZQ40Q2Oe7TUdkW2:bdpmpzc) ([raw file](https://github.com/aya-rs/aya/blob/832d2c0b694e88e187cd827e5f85a8f8f2551ec3/aya-log-common/src/lib.rs#L182)):*
> ```XML
> /// Currently only `[u8; 16]` (the return type of `bpf_get_current_comm()`) implements this.
> pub trait StrFormatter {}
> impl StrFormatter for [u8; 16] {}
> ```

Considering the eBPF helpers for reading user/kernel strings, such as `bpf_probe_read_user_str_bytes()` and `bpf_probe_read_kernel_str_bytes()`, it might be worth making `{:s}` a general byte-string formatter rather than limiting it to `[u8; 16]`.

For example, `StrFormatter` could be implemented for byte slices and fixed-size byte arrays:

  ```rust
  impl StrFormatter for &[u8] {}
  impl<const N: usize> StrFormatter for [u8; N] {}
  impl<const N: usize> StrFormatter for &[u8; N] {}
```

**Review by vadorovsky — CHANGES_REQUESTED** [review]
> @vadorovsky reviewed 8 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1623)* status: all files reviewed, 2 unresolved discussions (waiting on gmreads and swananan).

___
*[`aya-log-common/src/lib.rs` line 182 at r1](https://reviewable.io/reviews/aya-rs/aya/1623#-OwsDZQ40Q2Oe7TUdkW1:-OxWs3d2303I_lQju6cQ:b-8ii570) ([raw file](https://github.com/aya-rs/aya/blob/832d2c0b694e88e187cd827e5f85a8f8f2551ec3/aya-log-common/src/lib.rs#L182)):*

Considering the eBPF helpers for reading user/kernel strings, such as `bpf_probe_read_user_str_bytes()` and `bpf_probe_read_kernel_str_bytes()`, it might be worth making `{:s}` a general byte-string formatter rather than limiting it to `[u8; 16]`.

For example, `StrFormatter` could be implemented for byte slices and fixed-size byte arrays:

  ```rust
  impl StrFormatter for &[u8] {}
  impl<const N: usize> StrFormatter for [u8; N] {}
  impl<const N: usize> StrFormatter for &[u8; N] {}
```

</blockquote></details>

+1, let's use the const generic.
___
*[`aya-log/src/lib.rs` line 410 at r1](https://reviewable.io/reviews/aya-rs/aya/1623#-OxWraO09EAFGh7RdLbY:-OxWraO09EAFGh7RdLbZ:bqvohqz) ([raw file](https://github.com/aya-rs/aya/blob/832d2c0b694e88e187cd827e5f85a8f8f2551ec3/aya-log/src/lib.rs#L410)):*
> ```XML
>             Some(DisplayHint::Str) => {
>                 let nul = self.iter().position(|&b| b == 0).unwrap_or(self.len());
>                 Ok(String::from_utf8_lossy(&self[..nul]).into_owned())
> ```

If we care about strings being valid UTF-8, we shouldn't use `from_utf8_lossy` - we return a `Result` here, so we could handle the UTF-8 related errors.

That said, we could consider adding some kind of "bytes" or "osstr" display hint, which would not pose UTF-8 limits, but would print the bytes with a lossy formatter, similar to the `OsStr::display()`.

I'll think about it more and come up with some code example.

gmreads:
Hey folks, 

Thanks for the review, I have tried to address the comments. Please take a look. 

> That said, we could consider adding some kind of "bytes" or "osstr" display hint, which would not pose UTF-8 limits, but would print the bytes with a lossy formatter, similar to the `OsStr::display()`.
> 
> I'll think about it more and come up with some code example.

@vadorovsky Do you mean instead of {:s} we have something like {:osstr} ? And do  something like this? 
```
let nul = bytes.iter().position(|&b| b == 0).unwrap_or(bytes.len());
    let trimmed = &bytes[..nul];
    let os_str: &OsStr = OsStr::from_bytes(trimmed);
    Ok(os_str.display().to_string())
```

Also I don't really care about UTF-8, so lossy was fine for me. My use case was limited to printing `/proc/<pid>/comm` which is raw null terminated C style string. 

Please let me know what is the right approach to get this merged.

vadorovsky:
> @vadorovsky Do you mean instead of {:s} we have something like {:osstr} ? And do something like this?
> 
> ```
> let nul = bytes.iter().position(|&b| b == 0).unwrap_or(bytes.len());
>     let trimmed = &bytes[..nul];
>     let os_str: &OsStr = OsStr::from_bytes(trimmed);
>     Ok(os_str.display().to_string())
> ```

Yes, something like that. Bonus points if you can think of something shorter than `{:osstr}`, but I don't have any better idea for now.

> Also I don't really care about UTF-8, so lossy was fine for me. My use case was limited to printing `/proc/<pid>/comm` which is raw null terminated C style string.

Aya is a general purpose library. General purpose libraries should be defensive against errors, and give users the choice how to handle them, instead of silently swallowing them. Having both a fallible (error for invalid UTF-8) and an infallible (lossy) path is fine, but having only an infallible one is not, as it limits the choice for people who do care about UTF-8 (the fact you don't is irrelevant).

Sorry for being pedantic and harsh here, but the "it works for me" mindset really doesn't belong to open source projects. I'm glad you adjusted the code and agreed on providing two paths, but I also want you to understand why it's important, so we can avoid having this conversation for your future contributions. :slightly_smiling_face: 

> Please let me know what is the right approach to get this merged.

To be clear, I don't consider the `{:osstr}` to be a blocker for this PR, we can do it separately. For now, doing just `{:s}` that returns an error for UTF-8 is fine. I will play with the code later today and do more thorough review.

gmreads:
> Sorry for being pedantic and harsh here, but the "it works for me" mindset really doesn't belong to open source projects. I'm glad you adjusted the code and agreed on providing two paths, but I also want you to understand why it's important, so we can avoid having this conversation for your future contributions. 🙂

No apologies necessary, I completely agree. Aya is a library, and its public APIs should be general-purpose. My intention was only to provide the motivation behind this PR.

Slightly digressing, my longer-term goal is to explore whether Aya can help solve or improve this use case, I already have this working with old BCC style architecture and C based eBPF program.

 A system with:
Network **IN** bandwidth: `1 GiB/s`
**Capturing 100%** of HTTP/1.1 traffic (request headers, request body, and full response)

What would the CPU and memory overhead look like?

From what I've researched so far, it seems Aya could support a near zero-copy pipeline:

```
Application buffer
    └──(single copy)──▶ eBPF program
                           └── reserve/commit ──▶ shared ring buffer
                                                      └──▶ userspace
```

The userspace component would then stitch together the socket receive/send buffers to reconstruct HTTP request/response pairs.

I'd be interested to hear your thoughts on whether this seems like a reasonable direction, or if there are limitations I'm overlooking. 😄 

PS: I am doing this only on weekends, so responses might be delayed.

swananan (commented):
> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1623)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on gmreads and vadorovsky).

___
*[`-- commits` line 19 at r4](https://reviewable.io/reviews/aya-rs/aya/1623#-OySk-pv7GUtiwSlMW4r:-OySk-pv7GUtiwSlMW4s:b-wvt8vz):*
Please squash the PR’s commits into a single logical commit and rebase the branch onto main instead of merging main into it.
___
*[`aya-log-parser/src/lib.rs` line 178 at r1](https://reviewable.io/reviews/aya-rs/aya/1623#-OySl4hK0E4esbW7mRMn:-OySl4hK0E4esbW7mRMo:b-f6ftk2) ([raw file](https://github.com/aya-rs/aya/blob/832d2c0b694e88e187cd827e5f85a8f8f2551ec3/aya-log-parser/src/lib.rs#L178)):*
> ```XML
>     )]
>     #[test]
>     fn test_parse_str_hint() {
> ```

Shall we incorporate this case into the test_parse test above?

Issue #1628, opened 2026-07-11
Latest upgrades broke pinned Maps


sebadob:
I am not sure which part exactly broke, but when I bump versions to the latest ones released ~2 weeks ago, I cannot load XDP programs with pinned maps anymore.

Before, I was on `aya-0.13` and `aya-ebpf-0.1.1`. I am using `Array::pinned` and `LruHashMap::pinned` inside my XDP program. I checked if a specific one causes this issue by changing the `Array` into a HashMap and out-comment the LRU maps, but all with the same result:

```bash
LoadError { io_error: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }, verifier_log: fd 25 is not pointing to valid bpf_map
verification time 19 usec
stack depth 0
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
 }
```

I did not change anything apart from bumping versions, and when I downgrade, it works again. I cleaned up the FDs manually already to start fresh, but with the same issue. The code compiles just fine. Only in userspace I had to change from `XdpFlags` to `XdpMode`, that's it. I guess internally some syscalls were changed? I even tried to start my app as `root` to get rid of possible access issues, but with the same error. It happens during `program.load()` like so:

```rust
let program = bpf.program_mut("my_xdp").unwrap();
let program: &mut Xdp = program.try_into().unwrap();

program
    .load()
    .expect("Cannot load XDP program");
```

Inside my XDP program, I am very simply doing things like:

```rust
#[map]
static CONFIG: Array<Config> = Array::pinned(1, 0);
```

I already tried to bump only partly, but the error stays the same. When I bump only `aya`, it get it, and when I stay on the old `aya` version and only bump `aya-ebpf` inside the XDP program, it's the same. It kind of seems like they don't match, or it is in a commonly used crate?

The only solution I have so far is to stay on the old versions. I could not find a way to fix it. Is there anything I missed?

tamird:
Thanks for the report. The failure still needs a reproducer; the proposed fix in #1631 does not appear to be on this path.

MapData::from_pin() is not used by EbpfLoader when it relocates an object before program.load(), and the legacy map definitions emitted by Array::pinned and LruHashMap::pinned are unchanged between aya-ebpf 0.1.1 and 0.2.1. The ordinary load, drop, reopen pinned map, then load program path also succeeds without that patch.

[...]

PR #1631, opened 2026-07-11
fix(maps): propagate correct PinningType in map initialization from FD

CaioBrz1:
When initializing maps via `MapData::from_pin`, the internal routine `from_fd_inner` hardcoded `PinningType::None` when rebuilding the underlying `Map` structure through `parse_map_info`.

This erasure of the pinning context led to critical failures during program relocation (e.g., inside `program.load()`). Because the engine treated the map as an unpinned, newly managed userspace entity rather than a persistent object inherited from bpffs, the internal file descriptor state became desynchronized. This ultimately resulted in premature drops or malformed relocation inputs, causing the kernel verifier to reject the load with:
`fd X is not pointing to valid bpf_map`

Fix this by refactoring `from_fd_inner` to accept an explicit `PinningType`. This allows `from_pin` to correctly propagate `PinningType::ByName`, while preserving the default `PinningType::None` behavior for generic ID and raw FD lookups (`from_id` / `from_fd`).

### Added/updated tests?

- [x] Yes
      have not been included_

Added an integration test (`pin_type_preserved_after_from_pin`) that verifies the end-to-end pinning context lifecycle. It ensures that a map serialized to `bpffs` preserves its structural identity and state upon subsequent retrieval via `MapData::from_pin` instead of degenerating into fallback or unsupported typings.

### Checklist

           *Verified locally using `cargo xtask integration-test`. All relevant core subsystems passed successfully.* 

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1631)

### DIFF
--- a/aya/src/maps/mod.rs
+++ b/aya/src/maps/mod.rs
@@ -944,19 +944,19 @@ impl MapData {
             io_error,
         })?;
 
-        Self::from_fd_inner(fd)
+        Self::from_fd_inner(fd, PinningType::ByName)
     }
 
     /// Loads a map from a map id.
     pub fn from_id(id: u32) -> Result<Self, MapError> {
         let fd = bpf_map_get_fd_by_id(id)?;
-        Self::from_fd_inner(fd)
+        Self::from_fd_inner(fd, PinningType::None)
     }
 
-    fn from_fd_inner(fd: crate::MockableFd) -> Result<Self, MapError> {
+    fn from_fd_inner(fd: crate::MockableFd, pinning: PinningType) -> Result<Self, MapError> {
         let MapInfo(info) = MapInfo::new_from_fd(fd.as_fd())?;
         Ok(Self {
-            obj: parse_map_info(info, PinningType::None),
+            obj: parse_map_info(info, pinning),
             fd: MapFd::from_fd(fd),
         })
     }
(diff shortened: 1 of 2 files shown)


tamird (changes requested):
@codex review

> @tamird reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1631#-:-OxKvX2j6al1LQVm-9lL:b-fv7n3s)* status: all files reviewed, 1 unresolved discussion (waiting on CaioBrz1).

___
*[`test/integration-test/src/tests/map_pin.rs` line 85 at r1](https://reviewable.io/reviews/aya-rs/aya/1631#-OxKvSjl7cFfC6Yy-6RU:-OxKvSjl7cFfC6Yy-6RV:b-sqmqfz) ([raw file](https://github.com/aya-rs/aya/blob/d08fae8aec2e9d3bc7812163214448bb8d9daeb3/test/integration-test/src/tests/map_pin.rs#L85)):*
> ```XML
> 
>     assert!(
>         matches!(reopened_map, Map::HashMap(_)),
> ```

use assert_matches please and drop the custom message



_Code quote:_
```XML
    assert!(
        matches!(reopened_map, Map::HashMap(_)),
```

tamird:
The proposed change does not address #1628.

MapData::from_pin() is not on the reported EbpfLoader to program.load() path, and the added test neither calls program.load() nor observes PinningType. The PinningType::None behavior also predates aya 0.14, while Array::pinned and LruHashMap::pinned emit the same legacy map definition across the reported aya-ebpf versions.

The ordinary pinned-map reopen and program-load path passes without this change. I asked for the missing userspace sequence on #1628; until that reproduces the failure, this PR should not claim Fixes: #1628.

CaioBrz1:
I got what happened, my bad on that mistake. I'll apply the changes suggested by the bot.

PR #1639, opened 2026-07-19
fix(aya): avoid loading kernel BTF twice

swananan:
With some time to kill while waiting for the World Cup final, I browsed through the BTF implementation in Aya and found this issue.

I also thought about removing the BTF load from the new(), but I'm concerned that this could be a behavioral breaking change.

### Added/updated tests?

- [x] No, and this is why: has been covered by the existed tests.

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639)

### DIFF
--- a/aya-obj/src/extern_types.rs
+++ b/aya-obj/src/extern_types.rs
@@ -6,6 +6,13 @@ use crate::{
 };
 
 impl Object {
+    /// Returns true if this object contains typed kernel symbols.
+    pub fn has_typed_ksyms(&self) -> bool {
+        self.btf
+            .as_ref()
+            .is_some_and(|btf| btf.externs.has_typed_ksyms())
+    }
+
     /// Resolves extern kernel symbols through kernel `BTF` and `kallsyms`.
     pub fn resolve_externs(&mut self, kernel_btf: Option<&Btf>) -> Result<(), KsymsError> {
         if self.btf.is_none() {
@@ -24,12 +31,10 @@ impl Object {
     }
 
     /// Returns whether the object contains strong typed ksyms that require kernel `BTF`.
-    pub(crate) fn has_strong_typed_ksyms(&self) -> bool {
-        self.btf.as_ref().is_some_and(|btf| {
-            btf.externs.iter().any(|(_, ext)| {
-                ext.extern_type == ExternType::Ksym && ext.type_id.is_some() && !ext.is_weak
-            })
-        })
+    pub fn has_strong_typed_ksyms(&self) -> bool {
+        self.btf
+            .as_ref()
+            .is_some_and(|btf| btf.externs.has_strong_typed_ksyms())
     }
 
     /// Resolves typed externs through kernel `BTF`.
@@ -421,6 +426,12 @@ pub(crate) struct ExternCollection {
 
     /// Index ID of `.ksyms` datasec entry in BTF types.
     pub(crate) datasec_id: Option<u32>,
+
+    /// Whether the collection contains any typed ksyms.
+    has_typed_ksyms: bool,
+
+    /// Whether the collection contains any strong typed ksyms.
+    has_strong_typed_ksyms: bool,
 }
 
 impl ExternCollection {
@@ -429,7 +440,21 @@ impl ExternCollection {
     }
 
     pub(crate) fn insert(&mut self, name: String, desc: ExternDesc) {
+        if desc.extern_type == ExternType::Ksym && desc.type_id.is_some() {
+            self.has_typed_ksyms = true;
+            if !desc.is_weak {
+                self.has_strong_typed_ksyms = true;
+            }
+        }
         self.externs.insert(name, desc);
+    }
+
+    pub(crate) const fn has_typed_ksyms(&self) -> bool {
+        self.has_typed_ksyms
+    }
+
+    pub(crate) const fn has_strong_typed_ksyms(&self) -> bool {
+        self.has_strong_typed_ksyms
     }
 
     pub(crate) fn get_mut(&mut self, name: &str) -> Option<&mut ExternDesc> {
@@ -547,6 +572,9 @@ mod tests {
 
         let mut object = object_with_externs(externs);
 
+        assert!(object.has_typed_ksyms());
+        assert!(object.has_strong_typed_ksyms());
+
         let err = object.resolve_externs(None).unwrap_err();
         assert!(matches!(err, KsymsError::TypedKsymRequiresKernelBtf));
     }
@@ -561,11 +589,25 @@ mod tests {
 
         let mut object = object_with_externs(externs);
 
+        assert!(object.has_typed_ksyms());
+        assert!(!object.has_strong_typed_ksyms());
+
         object.resolve_externs(None).unwrap();
 
         let ext = &object.btf.as_ref().unwrap().externs.externs["typed"];
         assert!(!ext.is_resolved);
         assert_eq!(ext.ksym_addr, Some(0));
+    }
+
+    #[test]
+    fn has_typed_ksyms_ignores_typeless_ksyms() {
+        let mut externs = ExternCollection::new();
+        let typeless = ExternDesc::new("typeless".into(), ExternType::Ksym, 1, false);
+        externs.insert(typeless.name.clone(), typeless);
+
+        let object = object_with_externs(externs);
+
+        assert!(!object.has_typed_ksyms());
     }
 
     #[test]
--- a/aya/src/bpf.rs
+++ b/aya/src/bpf.rs
@@ -102,12 +102,9 @@ pub fn features() -> &'static Features {
 /// # Examples
 ///
 /// ```no_run
-/// use aya::{EbpfLoader, Btf};
-/// use std::fs;
+/// use aya::EbpfLoader;
 ///
 /// let bpf = EbpfLoader::new()
-///     // load the BTF data from /sys/kernel/btf/vmlinux
-///     .btf(Btf::from_sys_fs().ok().as_ref())
 ///     // load pinned maps from /sys/fs/bpf/my-program
 ///     .default_map_pin_directory("/sys/fs/bpf/my-program")
 ///     // finally load the code
@@ -160,7 +157,7 @@ impl<'a> EbpfLoader<'a> {
     /// Creates a new loader instance.
     pub fn new() -> Self {
         Self {
-            btf: Btf::from_sys_fs().ok().map(Cow::Owned),
+            btf: None,
             default_map_pin_directory: None,
             globals: HashMap::new(),
             max_entries: HashMap::new(),
@@ -173,9 +170,10 @@ impl<'a> EbpfLoader<'a> {
 
     /// Sets the target [BTF](Btf) info.
     ///
-    /// The loader defaults to loading `BTF` info using [`Btf::from_sys_fs`].
-    /// Use this method if you want to load `BTF` from a custom location or
-    /// pass `None` to disable `BTF` relocations entirely.
+    /// By default, the loader reads target `BTF` using [`Btf::from_sys_fs`] when
+    /// the object contains CO-RE relocations or typed kernel symbols. Use this
+    /// method to load `BTF` from a custom location.
+    ///
     /// # Example
     ///
     /// ```no_run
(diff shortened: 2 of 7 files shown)


vadorovsky:
Thanks for spotting that!

But what's still wrong is, as you noticed, that `EbpfLoader::new()` unconditionally loads system BTF, but then it can be overwritten by `EbpfLoader::btf()`.

> I also thought about removing the BTF load from the new(), but I'm concerned that this could be a behavioral breaking change.

I think what we should do is:

* Replace `btf: Btf::from_sys_fs().ok().map(Cow::Owned)` with `btf: None` in `EbpfLoader::new()`.
* During `EbpfLoader::new()`, if `btf` is `None`, then call `Btf::from_sys_fs().ok().map(Cow::Owned)`.

This way we would never end up loading the system BTF if a custom BTF is provided.

Speaking of behavioral changes, I can think of only one. As of today, someone can write the following:

```rust
EbpfLoader::new().btf(None).load()
```

That would result in not loading any BTF. But I think we shouldn't allow such thing. Modern BPF requires BTF, there is literally no legitimate use case for skipping it. `None` should mean "load system BTF".

On that note, I would also change `EbpfLoader::btf` to **not** take an `Option`, but simply a `Btf`, to make the API more straightforward.

Do you want to give it a try?

swananan:
CI now completes in 17m 38s, roughly 20% faster. This also suggests that repeated BTF parsing accounted for a significant portion of the ARM CI runtime in #1588 .

swananan:
@vadorovsky Sure, will do it tomorrow, since the World Cup final is about to start (:

swananan:
Although we removed the option to explicitly disable loading system BTF `btf(None)`, I went a step further and followed libbpf’s approach: inspects the object and loads system BTF only when needed.

swananan:
Some CI jobs are still failing, will investigate further tomorrow.

swananan:
Now CI takes 12m 14s, which is fast enough to expose two flaky tests 😂

vadorovsky:
> Although we removed the option to explicitly disable loading system BTF `btf(None)`, I went a step further and followed libbpf’s approach: inspects the object and loads system BTF only when needed.

Even better, nice! I will try to do a proper review tomorrow.

vadorovsky (changes requested):
> @vadorovsky partially reviewed 6 files and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`aya/src/bpf.rs` line 497 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9-9ESERIqLa9PSBHA:-Oy9-9ESERIqLa9PSBHB:b-fzlg7q) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya/src/bpf.rs#L497)):*
> ```XML
> 
>         if btf.is_none() && (obj.has_btf_relocations() || obj.has_typed_ksyms()) {
>             *btf = Btf::from_sys_fs().ok().map(Cow::Owned);
> ```

Can we actually handle the error from `Btf::from_sys_fs()`? It usually means that the running kernel doesn't expose BTF, and that the user either needs to change the distro or rebuild the kernel with `CONFIG_DEBUG_INFO_BTF=y`.

```rust
*btf = Btf::from_sys_fs()?.map(Cow::Owned);
```

Ignoring this error makes it very difficult to figure out what's wrong.

I know we did the `.ok()` thing in even before this change, but we shouldn't do that.
___
*[`aya-obj/src/extern_types.rs` line 443 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy90yoMEtenVjtlTp7N:-Oy90yoMEtenVjtlTp7O:b-6tzmt) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya-obj/src/extern_types.rs#L443)):*
> ```XML
> 
>     pub(crate) fn insert(&mut self, name: String, desc: ExternDesc) {
>         if desc.extern_type == ExternType::Ksym && desc.type_id.is_some() {
> ```

Nice! Great that you got rid of iteration in this path.

That said, would you mind extracting this change to `aya-obj/src/extern_types.rs` to a separate preceding commit? I think it deserves a separate commit with description, since on its own it improves the performance of BTF loading.

If it's too much work, I can do it while merging.

swananan (commented):
> @swananan made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on vadorovsky).

___
*[`aya/src/bpf.rs` line 497 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9-9ESERIqLa9PSBHA:-Oy9WMIB753HWguoPrF6:b-ps0cas) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya/src/bpf.rs#L497)):*

Can we actually handle the error from `Btf::from_sys_fs()`? It usually means that the running kernel doesn't expose BTF, and that the user either needs to change the distro or rebuild the kernel with `CONFIG_DEBUG_INFO_BTF=y`.

```rust
*btf = Btf::from_sys_fs()?.map(Cow::Owned);
```

Ignoring this error makes it very difficult to figure out what's wrong.

I know we did the `.ok()` thing in even before this change, but we shouldn't do that.
</blockquote></details>

Thanks for pointing this out. I should have caught it sooner. I sometimes default to assuming that existing code is correct, which is not a good mindset.

I initially implemented the suggested approach, but codex caught an edge case during review: objects containing only weak typed ksyms. After reconsidering it, I agree that failing to load kernel BTF should not be fatal in that case, since unresolved weak ksyms can still be handled by weak extern relocation.

When kernel BTF is mandatory, for CO-RE relocations or strong typed ksyms, the loader now propagates the original BtfError, which provides a more specific diagnostic.
___
*[`aya-obj/src/extern_types.rs` line 443 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy90yoMEtenVjtlTp7N:-Oy9c-dwCJJ8nmLeMQmC:bf6fv4w) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya-obj/src/extern_types.rs#L443)):*

Nice! Great that you got rid of iteration in this path.

That said, would you mind extracting this change to `aya-obj/src/extern_types.rs` to a separate preceding commit? I think it deserves a separate commit with description, since on its own it improves the performance of BTF loading.

If it's too much work, I can do it while merging.
</blockquote></details>

Sure, happy to do that, well-structured commits matter.

tamird (commented):
Very cool.

> @tamird reviewed 7 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639#-:-Oy9e8jK2mvRFsoz8KnY:b-jdo7sw)* status: all files reviewed, 5 unresolved discussions (waiting on swananan, tamird, and vadorovsky).

___
*[`aya/src/bpf.rs` line 173 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9TRqU4tK8mYAOxo2f:-Oy9TRqU4tK8mYAOxo2g:bcwt8zc) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya/src/bpf.rs#L173)):*
> ```XML
>     /// Sets the target [BTF](Btf) info.
>     ///
>     /// By default, the loader reads target `BTF` using [`Btf::from_sys_fs`] when
> ```

interesting, i wonder if we should have a way to provide the BTF source while still preserving the loader default lazy behavior. that's a nicer interface because i can always pass the location of my custom BTF, i don't need to think about whether i'm gonna need it or not.
___
*[`aya/src/sys/bpf.rs` line 495 at r8](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9driQ1RZ3HqHgdMSp:-Oy9driQ1RZ3HqHgdMSq:bchv8yp) ([raw file](https://github.com/aya-rs/aya/blob/275261a8958802695e7b7a6d2937aaf3c2d6154c/aya/src/sys/bpf.rs#L495)):*
> ```XML
>     // Before Linux 5.11, cgroup storage allocation can turn a memlock
>     // failure into ENOMEM during attach:
>     // https://github.com/torvalds/linux/blob/b15dc4170c6317731c00fa8fe7d4d99a0a04c83c/kernel/bpf/cgroup.c#L475-L477
> ```

8 chars please and consider separating "paragraphs" with an empty comment line

// paragraph 1...
// more text
//
// paragraph 2...

oh, but this comment should probably be on the 5.11 check way below?
___
*[`test/integration-test/src/tests/info.rs` line 251 at r7](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9dkjN5Au7YNtetHfe:-Oy9dkjN5Au7YNtetHff:b-ww2tcu) ([raw file](https://github.com/aya-rs/aya/blob/b3f8dd29f08ef390a887567a4b53d2fc7ffa5daf/test/integration-test/src/tests/info.rs#L251)):*
> ```XML
>             // removed before BPF_MAP_GET_FD_BY_ID. The kernel selftests and bpftool both skip the
>             // resulting ENOENT:
>             // https://github.com/torvalds/linux/blob/b15dc4170c6317731c00fa8fe7d4d99a0a04c83c/kernel/bpf/syscall.c#L3184-L3207
> ```

8 characters of SHA should be sufficient

swananan (commented):
> @swananan made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639)* status: 5 of 7 files reviewed, 5 unresolved discussions (waiting on tamird and vadorovsky).

___
*[`aya/src/bpf.rs` line 173 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9TRqU4tK8mYAOxo2f:-Oy9kJosExwvF4GNQDAM:b-5kgi8r) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya/src/bpf.rs#L173)):*

interesting, i wonder if we should have a way to provide the BTF source while still preserving the loader default lazy behavior. that's a nicer interface because i can always pass the location of my custom BTF, i don't need to think about whether i'm gonna need it or not.
</blockquote></details>

Got it. We definitely need a new API that allows a custom BTF source to be loaded lazily. I was thinking of something along these lines:
```
pub fn btf_path<P: AsRef<Path>>(&mut self, path: P) -> &mut Self

EbpfLoader::new()
      .btf_path("/custom_btf_file")
      .load_file("file.o")?;
```
I’d prefer to open a new PR to do this, so this PR can stay focused and improve the CI time cost. Would that work for you?
___
*[`aya/src/sys/bpf.rs` line 495 at r8](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9driQ1RZ3HqHgdMSp:-Oy9pd723Wg2IdoIBBAE:bomxe7d) ([raw file](https://github.com/aya-rs/aya/blob/275261a8958802695e7b7a6d2937aaf3c2d6154c/aya/src/sys/bpf.rs#L495)):*

8 chars please and consider separating "paragraphs" with an empty comment line

// paragraph 1...
// more text
//
// paragraph 2...

oh, but this comment should probably be on the 5.11 check way below?
</blockquote></details>

Done, good to know! I also split the comment and move it to the 5.11 checkk way.
___
*[`test/integration-test/src/tests/info.rs` line 251 at r7](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9dkjN5Au7YNtetHfe:-Oy9pkB25xgcf616Vst2:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/b3f8dd29f08ef390a887567a4b53d2fc7ffa5daf/test/integration-test/src/tests/info.rs#L251)):*

8 characters of SHA should be sufficient
</blockquote></details>

Done.

tamird (approved):
> @tamird reviewed 2 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639)* status: all files reviewed, 2 unresolved discussions (waiting on tamird and vadorovsky).

___
*[`aya/src/bpf.rs` line 173 at r2](https://reviewable.io/reviews/aya-rs/aya/1639#-Oy9TRqU4tK8mYAOxo2f:-Oy9uGy34tugQaNug_YP:b-suwxj5) ([raw file](https://github.com/aya-rs/aya/blob/a547007e91b53fac10b26542122a89da9b37218e/aya/src/bpf.rs#L173)):*

Got it. We definitely need a new API that allows a custom BTF source to be loaded lazily. I was thinking of something along these lines:
```
pub fn btf_path<P: AsRef<Path>>(&mut self, path: P) -> &mut Self

EbpfLoader::new()
      .btf_path("/custom_btf_file")
      .load_file("file.o")?;
```
I’d prefer to open a new PR to do this, so this PR can stay focused and improve the CI time cost. Would that work for you?
</blockquote></details>

Sure, that's fine. We might want to pass a `Read` rather than a path, or a function that returns a `Result<Read>`. Fine to do it later.

vadorovsky (approved):
:lgtm:

> @vadorovsky partially reviewed 7 files, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1639#-:-OyAVm2B5DcVCkr5slw5:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on tamird).

PR #1655, opened 2026-08-03
ci: increase Bazel heap for macOS runners

tamird:
macOS GitHub runners provide 7 GB of memory, leaving Bazel's default JVM heap below 2 GB. Increase the CI heap to 4 GB while retaining the existing 400-job remote execution limit.

Keep remote-action throttling disabled because its semaphore also covers remote input uploads and unnecessarily limits remote throughput [1].

[1]: https://github.com/bazelbuild/bazel/issues/28734

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1655)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -31,7 +31,8 @@ common:remote --remote_executor=grpcs://remote.buildbuddy.io
 common:remote --remote_cache_compression
 common:remote --remote_download_minimal
 
-# Keep remote backends fed
+# Keep remote backends fed; throttling also waits for remote uploads.
+# https://github.com/bazelbuild/bazel/issues/28734
 common:remote --noexperimental_throttle_remote_action_building
 common:remote --jobs=400
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ on:
     - cron: 00 4 * * *
 
 env:
+  # This key is intentionally public so fork pull requests can use BuildBuddy.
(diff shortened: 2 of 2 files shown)


.bazelrc
```diff
@@ -32,7 +32,6 @@ common:remote --remote_cache_compression
 common:remote --remote_download_minimal
 
 # Keep remote backends fed
-common:remote --noexperimental_throttle_remote_action_building
```
dzbarsky, line 35: i think this will limit to much fewer than 400 in practice. maybe better to kep this and limit --jobs to 200 instead?

tamird: Wouldn't it be better to increase heap than to try guessing the right number of jobs?

dzbarsky: Downloading more RAM always helps but that's not what this change does AFAIK

tamird: done

.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ on:
     - cron: 00 4 * * *
 
 env:
+  # This key is intentionally public so fork pull requests can use BuildBuddy.
```
dzbarsky, line 12: seems unrelated but sure

dzbarsky (approved):

.github/workflows/ci.yml
tamird: yeah just matching styles

PR #1663, opened 2026-08-07
aya-obj: fix flexible array bounds check in local access specs

### LINKED ISSUE #1662: CO-RE: flexible-array bounds check compares the element index against the parent's member count
Indexing a flexible array member with anything but one specific index fails to relocate:

```
error relocating section: error relocating raw_tp/sys_enter:
invalid access string 0:0:1 for type ``: array index out of bounds
```

Repro: access element > 0 of a flexible array member, e.g. `args[1]` on

```c
struct bpf_raw_tracepoint_args {
    __u64 args[0];
};

SEC("raw_tp/sys_enter")
int prog(struct bpf_raw_tracepoint_args *ctx)
{
    return (int)ctx->args[1];   /* args[0] relocates fine */
}
```

Cause: the flexible-array exemption in `AccessSpec::new` compares the *array element index* against the parent struct's member count, instead of the array member's own position in the parent:

https://github.com/aya-rs/aya/blob/802db7ddb83718d67e9ab0a627e0e18799ea63e1/aya-obj/src/btf/relocation.rs#L828

For `bpf_raw_tracepoint_args`, `members.len() - 1` is 0, so the test reduces to "the element index must be 0". `args[0]` is exempted by coincidence; every other index falls into the normal bounds check, is compared against a declared length of 0, and is rejected.

The same confusion is over-permissive in the other direction: for a zero-length array that is *not* the last member, the element index that happens to equal `members.len() - 1` is wrongly treated as a flexible array access and accepted.

The equivalent check in `match_candidate` (for the target spec) is correct and uses the parent accessor:

https://github.com/aya-rs/aya/blob/802db7ddb83718d67e9ab0a627e0e18799ea63e1/aya-obj/src/btf/relocation.rs#L551

libbpf's `is_flex_arr()` also keys off the enclosing member's index (`acc->idx == btf_vlen(t) - 1`).

## Versions

All published versions are affected: aya-obj 0.1.0 through 0.4.0, and, before the aya-obj split, aya 0.10.0 and 0.11.0. The line is unchanged since 08d5fa60 ("Initial BTF support", 2021-02-14); the correct target-side check was added the same day in 4b65da66.

## Fix

One word:

```diff
-                                    BtfType::Struct(s) => index == s.members.len() - 1,
+                                    BtfType::Struct(s) => parent.index == s.members.len() - 1,
```

I have this with two unit tests (one per direction, both failing before the change) on https://github.com/matteo-porta/aya/tree/fix-flex-array-reloc-index and am happy to open a PR.

matteo-porta:
When building the local access spec, the flexible-array exemption compares the
array element index against the parent struct's member count, instead of the
array member's own position in the struct.

For `struct bpf_raw_tracepoint_args`, whose single member is `__u64 args[0]`,
`members.len() - 1` is 0, so only `args[0]` was exempted from the bounds check
and any other index failed to relocate with "array index out of bounds". In the
other direction, a zero-length array that is *not* the last member was wrongly
exempted for the one index that made the two counters coincide.

Compare the parent accessor's index, like `match_candidate()` already does for
the target spec, and like libbpf's `is_flex_arr()`:

https://github.com/libbpf/libbpf/blob/master/src/relo_core.c#L70-L82

Fixes: #1662

### Added/updated tests?

- [x] Yes

Two unit tests in `aya-obj/src/btf/relocation.rs`, one per direction. Both fail
on the parent commit: the first panics on `args[1]` with "array index out of
bounds", the second sees `args[1]` wrongly accepted with `bit_offset: 64`.

### Checklist

- [x] All clippy lints have been fixed. You can find failing lints with
      `cargo xtask clippy`.

On the integration tests: they do not build for me, nor on `aya-rs/aya` main,
where `integration-ebpf` fails to link with `ERROR llvm: Invalid record` from
`bpf-linker` (`main` has been red on those jobs since 2026-08-06). The remaining
jobs are green on this branch: `lint` (which covers nightly fmt, clippy, miri
and public-api), both `bazel` jobs, and `build-test-aya` on all seven targets.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1663)

### DIFF
--- a/aya-obj/src/btf/relocation.rs
+++ b/aya-obj/src/btf/relocation.rs
@@ -825,7 +825,7 @@ impl<'a> AccessSpec<'a> {
                                 let parent = accessors.last().unwrap();
                                 let parent_ty = btf.type_by_id(parent.type_id)?;
                                 match parent_ty {
-                                    BtfType::Struct(s) => index == s.members.len() - 1,
+                                    BtfType::Struct(s) => parent.index == s.members.len() - 1,
                                     _ => false,
                                 }
                             };
@@ -1294,5 +1294,80 @@ impl ComputedRelocation {
             size: 0,
             type_id: None,
         })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use assert_matches::assert_matches;
+
+    use super::*;
+    use crate::btf::Int;
+
+    fn flex_array_btf(trailing_members: usize) -> (Btf, u32) {
+        let mut btf = Btf::new();
+        let name_offset = btf.add_string("long unsigned int");
+        let u64_id = btf.add_type(BtfType::Int(Int::new(name_offset, 8, IntEncoding::None, 0)));
+        let array_id = btf.add_type(BtfType::Array(Array::new(0, u64_id, u64_id, 0)));
+
+        let name_offset = btf.add_string("args");
+        let mut members = vec![BtfMember {
+            name_offset,
+            btf_type: array_id,
+            offset: 0,
+        }];
+        for i in 0..trailing_members {
+            let name_offset = btf.add_string(&format!("trailing{i}"));
+            members.push(BtfMember {
+                name_offset,
+                btf_type: u64_id,
+                offset: 0,
+            });
+        }
+
+        let name_offset = btf.add_string("bpf_raw_tracepoint_args");
+        let struct_id = btf.add_type(BtfType::Struct(Struct::new(name_offset, members, 0)));
+        (btf, struct_id)
+    }
+
+    fn field_relocation(type_id: u32) -> Relocation {
+        Relocation {
+            kind: RelocationKind::FieldByteOffset,
+            ins_offset: 0,
+            type_id,
+            access_str_offset: 0,
+            number: 0,
+        }
+    }
+
+    #[test]
+    fn test_access_spec_flex_array_index() {
+        let (btf, struct_id) = flex_array_btf(0);
+        let relocation = field_relocation(struct_id);
+
+        // `args` is the last member of its struct and has zero declared elements, so it is a
+        // flexible array and any element index is in bounds.
+        for (spec, expected_bit_offset) in [("0:0:0", 0), ("0:0:1", 64), ("0:0:7", 448)] {
+            let access_spec = AccessSpec::new(&btf, struct_id, spec, relocation).unwrap();
+            assert_eq!(access_spec.bit_offset, expected_bit_offset);
+        }
+    }
+
+    #[test]
+    fn test_access_spec_zero_length_array_not_last_member() {
+        let (btf, struct_id) = flex_array_btf(1);
+        let relocation = field_relocation(struct_id);
+
+        // `args` is no longer the last member, so it is a zero-length array rather than a
+        // flexible one and every element index is out of bounds.
+        for spec in ["0:0:0", "0:0:1"] {
+            assert_matches!(
+                AccessSpec::new(&btf, struct_id, spec, relocation),
+                Err(RelocationError::InvalidAccessIndex {
+                    error: "array index out of bounds",
+                    ..
+                })
+            );
+        }
     }
 }


vadorovsky:
Thanks, the fix looks good to me!

CI is red, because Rust nightly switched to LLVM 23. We are preparing a fix (https://github.com/aya-rs/bpf-linker/pull/396), after it's merged, your PR should become green after a rebase. I will ping you once the fix lands.

swananan (commented):
> @swananan made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1663)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on matteo-porta).

___
*[`aya-obj/src/btf/relocation.rs` line 828 at r1](https://reviewable.io/reviews/aya-rs/aya/1663#-OznkzV-8yRBKGaZqMsa:-OznkzV-8yRBKGaZqMsb:bjo40jz) ([raw file](https://github.com/aya-rs/aya/blob/9fd820717ecee7bc79f473d31adc79af7f8fd4b9/aya-obj/src/btf/relocation.rs#L828)):*
> ```XML
>                                 let parent_ty = btf.type_by_id(parent.type_id)?;
>                                 match parent_ty {
>                                     BtfType::Struct(s) => parent.index == s.members.len() - 1,
> ```

I noticed there is a similar check on line 551. Perhaps we could extract the check into a helper.

matteo-porta:
@swananan i made the helper as suggested, i have also rebased and the pipeline looks now good

swananan (commented):
Thanks!

> @swananan made 4 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1663#-:-P-yaVLnCIXuvnyte8Ft:b-ucfknh)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on matteo-porta).

___
*[`-- commits` line 25 at r2](https://reviewable.io/reviews/aya-rs/aya/1663#-P-yXEN_5PcjBWJhagz-:-P-yXEN_5PcjBWJhagz0:bknda2n):*
Perhaps we should squash these into one commit, and not mention the implementation detail (extract the helper).
___
*[`aya-obj/src/btf/relocation.rs` line 402 at r2](https://reviewable.io/reviews/aya-rs/aya/1663#-P-yanCXAR9whixuGOZI:-P-yanCXAR9whixuGOZJ:b-b314ts) ([raw file](https://github.com/aya-rs/aya/blob/977c6b15f8fae51b92918e96a41369849a4244a4/aya-obj/src/btf/relocation.rs#L402)):*
> ```XML
> 
> // An array is potentially variable length if it has 0 elements and is the last field of the parent
> // struct, like libbpf's is_flex_arr().
> ```

This comment seems week and expose the implementation details, we should put the detailed comment  inside the function.
___
*[`aya-obj/src/btf/relocation.rs` line 557 at r2](https://reviewable.io/reviews/aya-rs/aya/1663#-P-yaKZ4AykyRU5PH6c0:-P-yaKZ4AykyRU5PH6c1:b78p38s) ([raw file](https://github.com/aya-rs/aya/blob/977c6b15f8fae51b92918e96a41369849a4244a4/aya-obj/src/btf/relocation.rs#L557)):*
> ```XML
>                         };
> 
>                         let var_len = is_flex_array(
> ```

It looks like we don't need `var_len` variable here.

matteo-porta:
@swananan all three addressed in a force-push. The PR is now a single
commit, rebased again onto current main.

swananan (commented):
Cool, now we just need to wait for @vadorovsky to review this.

> @swananan made 1 comment and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1663#-:-P02OTYQ6Yuk2ncwgcCi:bcu78ya)* status: 0 of 1 files reviewed, all discussions resolved.

tamird:
Thanks! Picked up in #1735.

PR #1671, opened 2026-08-13
aya-obj: clean up after nightly bump

tamird:
- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1671)


vadorovsky (commented):
Agree with clanker's comment, but otherwise looks good.

PR #1674, opened 2026-08-15
aya-obj: reject mismatched function info

### LINKED ISSUE #1649: `--btf`: relocating a `.text` callee that has no func_info produces a program the verifier rejects
## Summary

When aya relocates a call to a function in `.text`, it appends that function's
instructions to the program and appends its BTF func_info records. If the callee
has **no** func_info records — nothing is appended, but the kernel still counts
the relocated function as a subprogram. The program is loaded with
`func_info_cnt` one short of `subprog_cnt`, and `check_btf_func()` rejects it:

```
number of funcs in func_info doesn't match number of subprogs
```

The callees that hit this in practice are the memory intrinsics (`memset`,
`memcpy`, `memmove`, `memcmp`, `bcmp`) that come from `compiler_builtins` and
land in `.text`. bpf-linker derives BTF func_info from DWARF, so whether they
carry func_info depends on whether `compiler_builtins` was built with debug
info — which is a property of *how* the user enables debug info, not of their
own code:

| how debug info is enabled | intrinsics carry func_info | result |
|---|---|---|
| `RUSTFLAGS=-Cdebuginfo=2` (reaches build-std) | yes | loads |
| `[profile.release.package.<ebpf-crate>] debug = 2` | **no** | **rejected** |

Both are ordinary ways to ask for debug info, and only one of them works.
`aya-build` uses the first (`-Cdebuginfo=2` in `CARGO_ENCODED_RUSTFLAGS`,
alongside `-Clink-arg=--btf`), so the recommended path avoids this — but by
accident rather than by design: nothing documents `-Cdebuginfo=2` as
load-bearing for correctness, and a hand-rolled build that scopes debug info to
the eBPF package produces an object aya links into an invalid program.

## Environment

| | |
|---|---|
| kernel | `7.0.0-28-generic` (Ubuntu 26.04, x86_64) |
| rustc | `1.93.0-nightly (b84478a1c 2025-11-30)` |
| bpf-linker | `0.10.4` (built `--no-default-features --features llvm-21,rust-llvm-21`) |
| aya | `0.13.1` |
| aya-obj | `0.2.1` |
| aya-ebpf | `0.1.1` |
| target | `bpfel-unknown-none`, `-Z build-std=core`, release |

`get_func_and_line_info` and `link_func_and_line_info` were also read in aya-obj
`0.3.0` — see "Not fixed by upgrading".

## Minimal reproducer

Two kprobes in one object. They differ only in whether they call an intrinsic.

`Cargo.toml`:

```toml
[package]
name = "btf-func-info-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
aya-ebpf = { version = "0.1.1", default-features = false }

[[bin]]
name = "repro"
path = "src/main.rs"

# Scoped to this package, so build-std's compiler_builtins gets no debug info
# and the intrinsics it supplies get no BTF func_info. This one line is the
# difference between loading and not loading.
[profile.release.package.btf-func-info-repro]
debug = 2
codegen-units = 1
```

`rust-toolchain.toml`:

```toml
[toolchain]
channel = "nightly-2025-12-01"
components = ["rust-src"]
```

`src/main.rs`:

```rust
#![no_std]
#![no_main]

use aya_ebpf::{
    macros::{kprobe, map},
    maps::PerfEventArray,
    programs::ProbeContext,
};

#[repr(C)]
#[derive(Clone, Copy)]
pub struct Event {
    pub pid: u32,
    pub data: [u8; 64],
}

#[map]
static EVENTS: PerfEventArray<Event> = PerfEventArray::new(0);

/// Rejected under `--btf`. The struct literal zero-initialises `data`, and
/// because the whole struct is submitted to the perf array every byte is live,
/// so LLVM emits a call to the `memset` in `.text`. That callee has no func_info.
#[kprobe]
pub fn calls_memset(ctx: ProbeContext) -> u32 {
    let event = Event {
        pid: (unsafe { aya_ebpf::helpers::bpf_get_current_pid_tgid() } >> 32) as u32,
        data: [0u8; 64],
    };
    EVENTS.output(&ctx, &event, 0);
    0
}

/// Loads under `--btf`: it calls nothing in `.text`.
#[kprobe]
pub fn calls_nothing(_ctx: ProbeContext) -> u32 {
    (unsafe { aya_ebpf::helpers::bpf_get_current_pid_tgid() } >> 32) as u32
}

#[cfg(not(test))]
#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    loop {}
}
```

Build:

```console
$ RUSTFLAGS="-C link-arg=--btf" cargo +nightly-2025-12-01 build --release \
    --target=bpfel-unknown-none -Z build-std=core
```

Loader (aya `0.13.1`), loading each program on its own so one rejection cannot
affect the other:

```rust
use aya::{programs::KProbe, Ebpf};

fn main() {
    let path = std::env::args().nth(1).expect("usage: repro-loader <object>");
    let bytes = std::fs::read(&path).expect("read object");

    let names: Vec<String> = {
        let bpf = Ebpf::load(&bytes).expect("parse object");
        bpf.programs().map(|(n, _)| n.to_string()).collect()
    };

    for name in names {
        let mut bpf = Ebpf::load(&bytes).expect("parse object");
        let prog: &mut KProbe = bpf
            .program_mut(&name)
            .expect("program present")
            .try_into()
            .expect("kprobe");
        match prog.load() {
            Ok(()) => println!("accepted: {name}"),
            Err(e) => println!("rejected: {name}: {e}"),
        }
    }
}
```

Result:

```
rejected: calls_memset: the BPF_PROG_LOAD syscall failed. Verifier output: number of funcs in func_info doesn't match number of subprogs
verification time 9 usec
stack depth 0+0

accepted: calls_nothing
```

`processed 0 insns` — the program is rejected before verification begins.

## The A/B that isolates it

Change one line in `Cargo.toml` — move `debug = 2` from the package-scoped
profile to `[profile.release]`, so build-std's `compiler_builtins` is compiled
with debug info too. Nothing else changes: same source, same linker, same aya,
same kernel.

| `debug = 2` scope | `.text` func_info records | `calls_memset` |
|---|---:|---|
| `[profile.release.package.btf-func-info-repro]` | 0 | **rejected** |
| `[profile.release]` | 5 | accepted |

The five records in the second row are `memcpy`, `memmove`, `memset`, `bcmp`,
`memcmp`; their `.BTF.ext` `insn_off` values match those symbols' `.text`
addresses exactly. In the first row `.BTF.ext` has no `.text` entry at all,
while the `kprobe` section carries func_info for both programs in both builds.

So the trigger is not the call, and not the intrinsic. It is a relocated callee
with no func_info.

## Root cause

`aya-obj-0.2.1/src/relocation.rs:457-464`:

```rust
let func_info = &fun.func_info.func_info;
let func_info = func_info.iter().cloned().map(|mut info| {
    // `start` is the new instruction offset of `fun` within `program`
    info.insn_off = start as u32;
    info
});
program.func_info.func_info.extend(func_info);
program.func_info.num_info = program.func_info.func_info.len() as u32;
```

When `fun.func_info.func_info` is empty this extends by nothing, so `num_info`
stays put while the callee's instructions are appended and the kernel counts a
second subprogram.

The model predicts every case observed:

| program calls | func_info | subprog_cnt | result |
|---|---:|---:|---|
| nothing in `.text` | 1 | 1 | loads |
| a `.text` fn that has func_info | N+1 | N+1 | loads |
| a `.text` fn that has none | 1 + **0** | 2 | **rejected** |

The middle row is worth stating because it shows the relocation path is
otherwise correct: a program whose only `.text` callee is an ordinary Rust
function — a `bpf_loop()` callback, say — relocates and loads fine.

## Scale, on a larger object

The same reproducer conditions on an unrelated object with 50 programs
(kprobe/uprobe/tracepoint), built the same way with debug info scoped to the
eBPF package:

| build | accepted | rejected |
|---|---:|---:|
| without `--btf` | 50 | 0 |
| with `--btf` | 9 | 41 |
| with `--btf` and `RUSTFLAGS=-Cdebuginfo=2` | **50** | **0** |

All 41 rejections carried the identical message. The rejected set and the set of
programs relocating against `memset` were the same 41 names — checked program by
program, predicted from the relocation table before the load was run, and
matched name for name. In that object `.text` held one Rust function and the five
intrinsics; with debug info scoped to the package it had **1** func_info record
(the Rust function), and with `-Cdebuginfo=2` it had **6**.

162 relocations across those 41 programs were all to `memset`; `memcpy`,
`memmove`, `memcmp` and `bcmp` were synthesised into `.text` and exported but
never called. So in practice the failure has a single trigger, which should make
a fix easy to test.

**A user cannot avoid this by restructuring their own code.** aya's own helper is
enough to emit the call (`aya-ebpf-0.1.1`):

```rust
pub fn bpf_get_current_comm() -> Result<[u8; 16], c_long> {
    let mut comm: [u8; 16usize] = [0; 16];
    let ret = unsafe { gen::bpf_get_current_comm(&mut comm as *mut _ as *mut c_void, 16u32) };
```

Every program in that object that builds an event reaches it, which is the same
41 — none of them by writing a `memset`. Restructuring the event structs cannot
remove a call the helper makes.

## Workaround

Pass `-Cdebuginfo=2` in `RUSTFLAGS` (not via a package-scoped Cargo profile) so
`compiler_builtins` is compiled with debug info and its intrinsics get func_info.
Verified: 41 rejections became 0, with no other change.

The cost is object size — 693,816 → 928,392 bytes for the 50-program object, +34%
— but almost all of the growth is DWARF sections the kernel never reads. `.BTF`
itself grew 4% (24,141 → 25,124 bytes) and `.BTF.ext` 2%.

## Ruled out

Recorded so nobody repeats them.

**Not per-section func_info slicing.** The first hypothesis was that
`get_func_and_line_info`'s `retain(|f| f.insn_off == instruction_offset)`
mis-slices when one ELF section holds many programs — the larger object packs 14
into `kprobe` and 25 into `uprobe`. Three failing programs were given their own
sections via `#[kprobe(function = ...)]` and `#[uprobe(path = ..., function = ...)]`,
which aya honours (`ProgramSection::from_str` splits on `/` and keeps only the
prefix). All three failed identically and the accepted count did not move. The
two-program reproducer above settles it: both its programs live in the same
`kprobe` section and only one is rejected.

**Not fixed by upgrading.** `get_func_and_line_info` is identical between
aya-obj `0.2.1` and `0.3.0` apart from a lifetime annotation, and
`link_func_and_line_info` still extends `func_info` by however many records the
callee has, which for these callees is none. (`0.3.0` does fix the adjacent
`line_info.num_info = func_info.len()` bug at `relocation.rs:475`; the func_info
mismatch is untouched.)

**`--disable-memory-builtins` is not a workaround.** It stops bpf-linker
*exporting* the intrinsics — `.text` went from 6 functions to 4, losing `memcmp`
and `bcmp` — but those two were the ones nothing called. The relocation count is
unchanged at 162, every one still to `memset`.

## Possible fixes

Listed by where they'd live, not by preference.

1. **aya**: when relocating a callee with no func_info, synthesise a placeholder
   record so `num_info` tracks `subprog_cnt`. Smallest change, and keeps working
   with already-published bpf-linker versions.

   One constraint for whoever implements it: the record cannot point at type id
   0. `check_btf_func()` resolves each record's `type_id` and rejects it unless
   `btf_type_is_func()` holds, so the placeholder needs a real `BTF_KIND_FUNC`
   entry — which means synthesising a BTF type as well as a func_info record, or
   reusing one already in the object.
2. **bpf-linker**: emit BTF func_info for every function it puts in `.text`,
   including ones it has no DWARF for. Arguably the more correct place — the
   object is otherwise self-inconsistent under `--btf`.
3. **aya**: detect the situation at load time and return a diagnostic naming the
   func_info-less callee. Even without a fix this would turn a confusing kernel
   message into an actionable one — the verifier says nothing about *which*
   subprogram is unaccounted for, and the true cause (a Cargo profile scope) is
   several steps removed from the symptom.
4. **docs**: whatever else changes, `-Cdebuginfo=2` is currently load-bearing for
   correctness on the `--btf` path and is not documented as such. A user who
   enables debug info the other way gets an object that cannot load.

Happy to test a patch against both the reproducer and the 50-program object,
which exercises the fix and the non-regression case at once.

shivanigupta-dev:
Reject call relocation when BTF function info is present on only one of
the program and callee. A mismatch leaves the function metadata out of
sync with the linked subprograms, which the kernel later rejects with an
opaque verifier error.

The new error is raised while linking and names the mismatched callee. It
covers both mismatch directions while preserving objects where function
info is empty on both sides.

This intentionally stops `Ebpf::load` on the first mismatch, including on
kernels where Aya would not submit BTF metadata. Mixed function info
indicates an inconsistent object build regardless of kernel support.

Fixes: #1649

### Added/updated tests?

- [x] Yes

The tests cover both mismatch directions and confirm that functions still
link when function info is either present on both sides or absent on both
sides.

### Checklist

  The kernel-dependent `test_read_btf_from_sys_fs` test was skipped because
  the test machine does not expose `/sys/kernel/btf/vmlinux`; the other 111
  unit tests and the doc test passed.
  Integration tests were not run locally. The issue author has offered to
  test the branch against both original reproducers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1674)


samaepark:
Ran this branch (ee65dcc7) against both reproducers from the issue, on kernel `7.0.0-29-generic` (x86_64), rustc nightly-2025-12-01, bpf-linker 0.10.4.

Minimal two-kprobe reproducer (package-scoped `debug = 2`, `-C link-arg=--btf`):

aya 0.13.1, loading each program:

```
FAIL  calls_memset: the BPF_PROG_LOAD syscall failed. Verifier output: number of funcs in func_info doesn't match number of subprogs
PASS  calls_nothing
```

This branch, same object:

```
error relocating function
  caused by: error relocating `calls_memset`
  caused by: BTF function info for `memset` does not match the program
```

`Ebpf::load` stops before any program reaches the verifier, and the error names the callee. The other two builds behave as intended on the branch: the same source without `--btf` (func_info empty on both sides) loads 2/2, and with `--btf` plus `RUSTFLAGS="-C debuginfo=2"` (func_info present on both sides) loads 2/2.

The larger object has grown since I filed, so its counts no longer line up with the ones in the issue, but the shape is unchanged. Built the broken way (`--btf`, debug info scoped to the eBPF package), aya 0.13.1 rejects the majority of its programs, every rejection carrying the identical verifier message. This branch turns that into a single whole-object error naming `memset`; the chain matches the one above except for which program the middle line names. Without `--btf`, every program is accepted, on 0.13.1 and on this branch alike.

I also drove `Object::parse` + `relocate_calls` directly (with the same `text_sections` set `Ebpf::load` computes) over all five objects: the guard fires on exactly the two mismatched ones and nothing else.

One observation, not a blocker: which program the whole-object error names varies run to run, since `Object::programs` is a `HashMap` and `relocate_calls` reaches a different program first each time. Four consecutive runs on the same object named four different programs; the callee named was `memset` every time. Iterating in a deterministic order would make the error reproducible byte for byte, which matters mostly to people diffing CI logs.

I had no natural object for the reverse direction (program's func_info empty, callee's populated); the branch's unit test covers that case.

From the reporter's side this does exactly what the thread converged on.

tamird:
@codex review

PR #1687, opened 2026-08-21
aya: add attach_to_link for CgroupSockAddr

jibi:
Add `CgroupSockAddr::attach_to_link`, which atomically swaps the program behind an already attached cgroup link, as `Xdp` and `SchedClassifier` already do.

Together with `TryFrom<FdLink> for CgroupSockAddrLink`, added here too, a link recovered from a bpffs pin can be pointed at a newly loaded program without the cgroup ever going unattached:

    let prog: &mut CgroupSockAddr = bpf.program_mut("myprog").unwrap().try_into()?;
    prog.load()?;

    let pinned = PinnedLink::from_pin("/sys/fs/bpf/myprog_link")?;
    let link = FdLink::from(pinned).try_into()?;

    prog.attach_to_link(link)?;

Neither existed before: `CgroupSockAddr` could return a link to be pinned but would not take one back, so the only option was to unpin the old link and attach a new one, leaving a window where no cgroup program is attached.

Only `BPF_LINK_CREATE` attachments are supported: those made with `BPF_PROG_ATTACH` (kernels < 5.7) return `LinkError::InvalidLink` because, unlike XDP and TC, a cgroup configured with `BPF_F_ALLOW_MULTI` (exposed in aya as `CgroupAttachMode::AllowMultiple`) can hold several programs of one attach type, and a reattach would append the new program instead of replacing the existing one. Supporting that would mean passing `BPF_F_REPLACE` and the fd of the program to replace, which would need an extra param on `bpf_prog_attach` and all its callers. Not worth it for an outdated syscall command superseded by `BPF_LINK_CREATE`.

**Few remarks for the review**
- as stated already in the commit message, this does not support `BPF_PROG_ATTACH` attachments: deliberate choice, but happy to add support for that as well (I have the patch ready, it's ~100 lines, just wasn't sure it was worth the extra churn for an outdated bpf command)
- this touches only `CgroupSockAddr`. Implementing the same for `CgroupSkb`, `CgroupSock` and `SockOps` should be mostly a copypaste chore, while `CgroupDevice`, `CgroupSockopt` and `CgroupSysctl` would need `impl_try_into_fdlink!` first (I left them out as I wanted to get one program right first, could be a follow up in case)

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1687)

### DIFF
--- a/aya/src/programs/cgroup_sock_addr.rs
+++ b/aya/src/programs/cgroup_sock_addr.rs
@@ -2,16 +2,17 @@
 
 use std::{hash::Hash, os::fd::AsFd, path::Path};
 
-use aya_obj::generated::bpf_prog_type::BPF_PROG_TYPE_CGROUP_SOCK_ADDR;
+use aya_obj::generated::{bpf_link_type, bpf_prog_type::BPF_PROG_TYPE_CGROUP_SOCK_ADDR};
 pub use aya_obj::programs::CgroupSockAddrAttachType;
 
 use crate::{
     VerifierLogLevel,
     programs::{
-        CgroupAttachMode, FdLink, Link, ProgAttachLink, ProgramData, ProgramError, ProgramType,
-        define_link_wrapper, id_as_key, impl_try_into_fdlink, load_program_with_attach_type,
+        CgroupAttachMode, FdLink, Link, LinkError, ProgAttachLink, ProgramData, ProgramError,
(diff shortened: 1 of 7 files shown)


vadorovsky (approved):
> @vadorovsky reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1687)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jibi).

PR #1690, opened 2026-08-21
fix(aya-ebpf-macros): change return type of `stream_parser` to `i32`

### LINKED ISSUE #1411: Stream Parser return type and Stream Verdict requirement
I have a quick question about programs marked as `#[stream_parser]`. I was trying to make a test program to just log some things, but let packets through without actually performing any extra action on them. In Aya, programs of this type require a return type of `u32` but this seems to not correspond to the eBPF documentation (https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_SK_SKB/) that says:

>The return value is interpreted as follows:
>
>- 0 - indicates length of successfully parsed message
>- 0 - indicates more data must be received to parse the message
>- -ESTRPIPE - current message should not be processed by the kernel, return control of the socket to userspace which can proceed to read the messages itself
>- other < 0 - Error in parsing, give control back to userspace assuming that synchronization is lost and the stream is unrecoverable (application expected to close TCP socket)

In Aya it seems I cannot return `-ESTRPIPE` or any other negative value. I have tried to convert `-ESTRPIPE` to a u32 (twos complement) but this does not work either. In fact the only thing that seems to yield the behavior I'm looking for is returning `sk_action::SK_PASS` (aka 1), but this doesn't make sense as it should mean I have successfully parsed the message and it has length 1 (which is not the case).
`sk_action::SK_PASS` and `sk_action::SK_DROP` should only be relevant in other types of programs such as `stream_verdict` where it is the expected return.

It also seems that I must have a `stream_parser` program and a `stream_verdict` program together in order for them to be called. Having only a `stream_parser` program attached to my SockHash seems to result in my program never being called. I haven't been able to find if this is an eBPF requirement, or perhaps if it is linked to Aya. I'm curious if it is the intended behavior

zignis:
The `#[stream_parser]` macro currently uses `u32` as the return type. However, the eBPF verifier officially supports negative return values for this program type, which require an `i32` return type: https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_SK_SKB

Modified the code generation for `#[stream_parser]` to explicitly use `i32` as the return type.  Updated all related doc examples. Let me know if I need (and how) to add tests since I could not find any for these hooks.

Fixes: #1411 

### Added/updated tests?

- [x] Yes
      have not been included_

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1690)

### DIFF
--- a/aya-ebpf-macros/src/lib.rs
+++ b/aya-ebpf-macros/src/lib.rs
@@ -438,15 +438,15 @@ pub fn btf_tracepoint(attrs: TokenStream, item: TokenStream) -> TokenStream {
 ///
 ///
 ///#[stream_parser]
-///fn stream_parser(ctx: SkBuffContext) -> u32 {
+///fn stream_parser(ctx: SkBuffContext) -> i32 {
 ///    match try_stream_parser(ctx) {
 ///        Ok(ret) => ret,
 ///        Err(ret) => ret,
 ///    }
 ///}
 ///
-///fn try_stream_parser(ctx: SkBuffContext) -> Result<u32, u32> {
-///    Ok(ctx.len())
+///fn try_stream_parser(ctx: SkBuffContext) -> Result<i32, i32> {
+///    Ok(ctx.len() as i32)
 ///}
 /// ```
 #[proc_macro_attribute]
--- a/aya-ebpf-macros/src/sk_skb.rs
+++ b/aya-ebpf-macros/src/sk_skb.rs
@@ -48,10 +48,15 @@ impl SkSkb {
         } = item;
         let section_name: Cow<'_, _> = format!("sk_skb/{kind}").into();
         let fn_name = &sig.ident;
+        let return_type = match kind {
+            SkSkbKind::StreamVerdict => quote! { u32 },
+            SkSkbKind::StreamParser => quote! { i32 },
+        };
+
         quote! {
             #[unsafe(no_mangle)]
             #[unsafe(link_section = #section_name)]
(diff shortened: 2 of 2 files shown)


tamird (changes requested):
> Let me know if I need (and how) to add tests since I could not find any for these hooks.

certainly would be good to add some integration tests. I guess we don't have any for stream parsers

> @tamird reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1690#-:-P-du_AfAB4JIz_h0xMj:b-q2e1da)* status: all files reviewed, 1 unresolved discussion (waiting on zignis).

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/aya-rs/aya/1690#-P-duOzgE_tSxkrIlk3A:-P-duOzgE_tSxkrIlk3B:bbgh0e3):*
please squash this down to one commit and give it a proper message



_Code quote:_
```
New commits in r1 on 8/21/2026 at 2:59 PM:
- 62da2fb: aya-ebpf-macros: change return type of `stream_parser` to `i32`

New commits in r2 on 8/22/2026 at 2:08 AM:
- c1370ca: format code using rustfmt
```

zignis:
I'll pick this up soon and add the tests.

zignis:
I'd prefer someone already familiar with the codebase to write the stream parser tests, or point me in the right direction (what exact logic to test)

PR #1696, opened 2026-08-25
aya: add a multi-uprobe support probe

swananan:
I would like to add and expose this probe as a public API. USDT is one consumer, but it's not the only one: callers that dynamically generate eBPF bytecode also need to determine multi-uprobe support before loading. For example, my personal experimental project: https://github.com/swananan/ghostscope 😂

I hope the second commit doesn't feel premature, given that USDT support hasn't landed in the Aya yet.

### Added/updated tests?

- [x] Yes

### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1696)


tamird (commented):
I found two test issues in this change:

1. The process-scoped PID probe is not exercised against the unfixed behavior. In `test/integration-test/src/tests/feature_probe.rs`, either result passes on Linux 6.6–6.9. The [6.6 kernel used by this PR's CI](https://github.com/aya-rs/aya/actions/runs/32837818018/job/97770622477) is 6.6.153, which includes the fix; [6.6.23 still filters just the selected thread](https://github.com/gregkh/linux/blob/v6.6.23/kernel/trace/bpf_trace.c#L3101). If `ProcessScopedPidFilter` incorrectly returned true whenever link creation worked, the current matrix would pass while callers on an unfixed kernel could lose events from sibling threads. Please cover the invalid-PID `EBADF` result with a focused syscall test or test on an unfixed 6.6 kernel, ideally with a sibling-thread attachment check.

2. The positive kernel-version assertions in that test assume that uprobes are configured whenever the kernel is new enough. With `CONFIG_UPROBES=n`, [the Linux attachment stub returns `EOPNOTSUPP`](https://github.com/torvalds/linux/blob/v6.6/kernel/trace/bpf_trace.c#L3307-L3312), which this probe correctly reports as unsupported. The new test therefore fails on such kernels when run natively. Please condition those positive expectations on `CONFIG_UPROBES` as well as the version.

Reviewed with Codex, including independent checks of kernel behavior and test coverage.

swananan:
For the first comment, I added a UT to cover this scenario.

And for the second one, I added the kernel config check in the integration test.

tamird (changes requested):
The new syscall test covers the unfixed `EBADF` PID response, and the `CONFIG_UPROBES` check addresses both points from my earlier review. I found one remaining test issue on `8f55204d`:

**[P2] Account for 32-bit kernels in the positive version assertions.** [`probe_uprobe_multi`](https://github.com/aya-rs/aya/blob/8f55204dbe50ad8dccd9b43c2d0e7097fa71ac6d/test/integration-test/src/tests/feature_probe.rs#L39-L74) now asserts link support on every kernel >= 6.6 with `CONFIG_UPROBES=y`, and process-scoped PID support on every such kernel >= 6.10. But [Linux 6.18 still rejects multi-uprobe links on 32-bit kernels with `EOPNOTSUPP`](https://github.com/torvalds/linux/blob/v6.18/kernel/trace/bpf_trace.c#L3168-L3170); the same restriction exists in 6.6. On one of those kernels the probe correctly returns `false`, yet both positive assertions fail. Aya builds for armv7, while the current VM integration matrix runs only 64-bit kernels, so green CI does not exercise this case. Please gate the positive assertions on *kernel* architecture support as well as version and config. User-space pointer width alone is not sufficient when a 32-bit process runs on a 64-bit kernel.

— Codex re-review for @tamird, with independent delegated checks

tamird (commented):
@codex review

> @tamird reviewed 7 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1696#-:-P1_l1D96tnKxGFF3q4s:b-fv7n3s)* status: all files reviewed, 1 unresolved discussion (waiting on swananan).

___
*[`test/integration-test/src/tests/feature_probe.rs` line 40 at r5](https://reviewable.io/reviews/aya-rs/aya/1696#-P1_ktMDBGkHsT-rfwRU:-P1_ktMDBGkHsT-rfwRV:b-zf28tb) ([raw file](https://github.com/aya-rs/aya/blob/a72d801a4eb4012a0d20a277b3ba970ee8a95310/test/integration-test/src/tests/feature_probe.rs#L40)):*
> ```XML
> 
>     // Multi-uprobe requires a 64-bit kernel, even when CONFIG_UPROBES=y.
>     // https://github.com/torvalds/linux/blob/7d0a66e4/kernel/trace/bpf_trace.c#L3168-L3170
> ```

need 9 digits, this link is broken

swananan (commented):
> @swananan made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1696)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on tamird).

___
*[`test/integration-test/src/tests/feature_probe.rs` line 40 at r5](https://reviewable.io/reviews/aya-rs/aya/1696#-P1_ktMDBGkHsT-rfwRU:-P1_vuZQEKalnCJjt9ZQ:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/a72d801a4eb4012a0d20a277b3ba970ee8a95310/test/integration-test/src/tests/feature_probe.rs#L40)):*

need 9 digits, this link is broken
</blockquote></details>

Done.

tamird (approved):
> @tamird reviewed 4 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1696)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on swananan).

PR #1708, opened 2026-08-31
ci: add latest stable VM kernel coverage

swananan:
It would be nice to test against `bpf-next` kernel in the CI, but it looks like a big change, so I only added the latest stable kernel.

### Added/updated tests?

- [ ] No, and this is why: CI change..

### Checklist

- [ ] Rust code has been formatted with `cargo +nightly fmt`.
- [ ] All clippy lints have been fixed.
- [ ] Unit tests are passing locally with `cargo test`.
- [ ] I have blessed any API changes with `cargo xtask public-api --bless`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1708)


swananan:
https://kernel.ubuntu.com/mainline is not stable, will try it tomorrow..

tamird (changes requested):
General comment: I *think* we should be trying to move away from mutable points and prefer explicit bumps. That's what I've recently done with stable/beta/nightly versions.

> @tamird reviewed 2 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1708#-:-P0Tr-zZEXNLycyZ_Zh4:b-bfrbt8)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on swananan).

___
*[`xtask/src/ubuntu_mainline.rs` line 125 at r1](https://reviewable.io/reviews/aya-rs/aya/1708#-P0SZ_hU6ncDwCenwDt2:-P0SZ_hU6ncDwCenwDt3:b-uowd8h) ([raw file](https://github.com/aya-rs/aya/blob/b74c3888cfafa8f04681e16118a801b0819d6ad9/xtask/src/ubuntu_mainline.rs#L125)):*
> ```XML
>     // For a series such as 6.6, select the newest v6.6.y build that has packages
>     // for the requested architecture. `latest-stable` instead considers release
>     // directories across all series whose names match vMAJOR.MINOR[.PATCH].
> ```

i don't understand this sentence.
___
*[`xtask/src/ubuntu_mainline.rs` line 128 at r1](https://reviewable.io/reviews/aya-rs/aya/1708#-P0SZcHf3J8MNxQY_scz:-P0SZcHf3J8MNxQY_sd-:bwfgzql) ([raw file](https://github.com/aya-rs/aya/blob/b74c3888cfafa8f04681e16118a801b0819d6ad9/xtask/src/ubuntu_mainline.rs#L128)):*
> ```XML
>     // Ubuntu may publish newer release directories while architecture
>     // artifacts are still missing. Check only a bounded number of recent
>     // builds so an unavailable line fails instead of degrading indefinitely.
> ```

why is it a line here but series above?
___
*[`xtask/src/ubuntu_mainline.rs` line 149 at r1](https://reviewable.io/reviews/aya-rs/aya/1708#-P0TqZKV3aCFLUsBAibp:-P0TqZKV3aCFLUsBAibq:bs4yoya) ([raw file](https://github.com/aya-rs/aya/blob/b74c3888cfafa8f04681e16118a801b0819d6ad9/xtask/src/ubuntu_mainline.rs#L149)):*
> ```XML
>             if numbers.next().is_some() {
>                 return None;
>             }
> ```

slightly unfortunate that this provides no visibility into failed parse



_Code quote:_
```XML
            let mut numbers = name.strip_prefix('v')?.split('.');
            let major = numbers.next()?.parse().ok()?;
            let minor = numbers.next()?.parse().ok()?;
            let patch = numbers.next().map_or(Some(0), |patch| patch.parse().ok())?;
            if numbers.next().is_some() {
                return None;
            }
```

swananan (commented):
Agreed, but it looks like a big change, so I prefer to open a new PR for it and keep this one focused, if that works for you.

In the long term, the ubuntu mainline download is not stable recently and doesn't promise any SLA, so we should drop it in the CIs.

My plan is to add a daily workflow that checks kernel version update from https://www.kernel.org/releases.json, if LTS or latest-stable version changes, the workflow will open an bump PR.  All the kernel versions are maintained in the bazel config. In the bump PR, we will build and test the kernels with bazel from source code, and upload prebuilt artifacts for CI using.

There is one thing to confirm: I noticed that bazel upload the build outputs to the BuildBuddy, but I am not sure about its limits and cost. I also checked the cilium ebpf, they upload the artifacts to the GHCR, which I would prefer.

In the future, if GHCR provides sufficient capacity, we could support kernels for more architectures beyond Arm64 and AMD64.

Does this sound like the right direction?

> @swananan made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/aya-rs/aya/1708#-:-P0Xq6p-58Vb4_UVx8hg:b-6zx8ex)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on tamird).

___
*[`xtask/src/ubuntu_mainline.rs` line 125 at r1](https://reviewable.io/reviews/aya-rs/aya/1708#-P0SZ_hU6ncDwCenwDt2:-P0XipG_7mQvJRdJplOi:bewjlpc) ([raw file](https://github.com/aya-rs/aya/blob/b74c3888cfafa8f04681e16118a801b0819d6ad9/xtask/src/ubuntu_mainline.rs#L125)):*

i don't understand this sentence.
</blockquote></details>

I split the comments and moved each part to the right place.
___
*[`xtask/src/ubuntu_mainline.rs` line 128 at r1](https://reviewable.io/reviews/aya-rs/aya/1708#-P0SZcHf3J8MNxQY_scz:-P0Xioz1D6ENUO0fvA0T:b-896fix) ([raw file](https://github.com/aya-rs/aya/blob/b74c3888cfafa8f04681e16118a801b0819d6ad9/xtask/src/ubuntu_mainline.rs#L128)):*

why is it a line here but series above?
</blockquote></details>

Done.
___
*[`xtask/src/ubuntu_mainline.rs` line 149 at r1](https://reviewable.io/reviews/aya-rs/aya/1708#-P0TqZKV3aCFLUsBAibp:-P0XnhCX-_kq01rJ035o:b-4ybtsu) ([raw file](https://github.com/aya-rs/aya/blob/b74c3888cfafa8f04681e16118a801b0819d6ad9/xtask/src/ubuntu_mainline.rs#L149)):*

slightly unfortunate that this provides no visibility into failed parse
</blockquote></details>

The ubuntu mainline listing doesn't follow a stable name scheme, it could be a lot of error messages, I added a comment to explain why.

tamird (changes requested):
Two issues remain on `cfc2cdfd`:

- **[P2] Keep the CI kernel version explicit.** The [new `latest-stable` matrix entries](https://github.com/aya-rs/aya/blob/cfc2cdfd132fe4eb00a352952808ef32ebf41831/.github/workflows/ci.yml#L353-L365) resolve from Ubuntu's live listing at run time. Coverage can change without a reviewed version bump, which was the concern in [the earlier review](https://github.com/aya-rs/aya/pull/1708#pullrequestreview-5083252060). It also cannot guarantee coverage of the current upstream stable release: today kernel.org lists 7.2.4, while Ubuntu's newest usable stable build is 7.2.3. Please pin the version being tested and bump it explicitly.

- **[P2] Account for moving kernel packages in the VM cache.** The new selector uses the same [fixed cache key and exact-hit save condition](https://github.com/aya-rs/aya/blob/cfc2cdfd132fe4eb00a352952808ef32ebf41831/.github/workflows/ci.yml#L437-L474) after its resolved kernel changes. `xtask` downloads the new packages, but an existing cache key prevents saving them; every subsequent run restores the old packages and downloads the new ones again. That adds repeated large downloads to an already intermittent Ubuntu Mainline dependency. Please coordinate the cache refresh in #1713 or otherwise incorporate the resolved package changes into the saved key before enabling this selector.

The parser and package selection themselves look consistent with the available Ubuntu Mainline directory and package names. The current-head build and VM checks passed; the above concerns are about coverage policy and later runs after a release moves.

— Codex review for @tamird, with independent delegated checks

swananan:
For the first question, I just think it would be nice to have a latest version in the CI matrix, so the floating kernel version is OK for me, we will introduce the immutable kernel version, but not in this PR.

For the second one, this PR will be merged after the #1713.

swananan:
I have closed this PR, because @tamird is working on the immutable kernel versions.

PR #1725, opened 2026-09-13
aya-obj: correctly calculate multidimensional BTF array sizes

### LINKED ISSUE #1677: Miscalculates multidimensional BTF array sizes
## Summary

`Btf::type_size()` overwrites the accumulated element count for each nested
array instead of multiplying dimensions.

For an unmodified generated map value of type `char[2][3]`, Aya
calculates `value_size=3` instead of `6`. The kernel rejects `BPF_MAP_CREATE`
with `EINVAL`, causing Aya to reject an otherwise valid ELF.

## Reproducer

```c
typedef char matrix_t[2][3];

struct {
    __uint(type, BPF_MAP_TYPE_ARRAY);
    __uint(max_entries, 1);
    __type(key, __u32);
    __type(value, matrix_t);
} matrix SEC(".maps");
```

The program accesses `(*value)[1][2]`, which Clang emits as a byte write at
map offset 5.

## Results

Using the same ELF:

- Aya submits `value_size=3`; `BPF_MAP_CREATE` returns `EINVAL`.
- cilium/ebpf submits `value_size=6`; the verifier accepts it and `TEST_RUN`
  returns `42`.
- libbpf submits `value_size=6`; the verifier accepts it and `TEST_RUN`
  returns `42`.

## Root Cause

[`Btf::type_size()`](https://github.com/aya-rs/aya/blob/318ac8d529fb80d538d0e4aee578a8bfaabe0e6e/aya-obj/src/btf/btf.rs#L466-L488)
currently uses:

```rust
n_elems = array.len;
```

This replaces the previous dimension instead of accumulating it..

## Specification Reference

The Linux BTF specification defines `btf_array.type` as the element type and
`btf_array.nelems` as the number of elements. It explicitly permits arrays to
be chained through their element type to represent multidimensional arrays.
Its `int[5][6]` example is encoded as an array of 5 whose element type is an
array of 6:

- [Linux 6.8 BTF specification: `BTF_KIND_ARRAY`](https://docs.kernel.org/6.8/bpf/btf.html#btf-kind-array)
- [Linux kernel BTF resolver](https://github.com/torvalds/linux/blob/v6.8/kernel/bpf/btf.c#L1909-L1916)
- [libbpf `btf__resolve_size()`](https://github.com/libbpf/libbpf/blob/f7081a6baf3f54949aacb8c2fc11bb30783b83e9/src/btf.c#L931-L968)
- [cilium/ebpf `btf.Sizeof()`](https://github.com/cilium/ebpf/blob/e3a865be0e4be4a0ec78979629c2f35f3ec37c18/btf/types.go#L687-L733)

## Suggested Fix

Accumulate every array dimension with checked multiplication rather than a
plain assignment:

```rust
n_elems = n_elems
    .checked_mul(array.len)
    .ok_or(BtfError::TypeSizeOverflow {
        type_id: root_type_id,
    })?;
```

The final `size * n_elems` operation should also be checked. 

## Environment

- Aya: `318ac8d529fb80d538d0e4aee578a8bfaabe0e6e`
- Clang: `18.1.3`
- Linux: `6.8.0-117-generic`, aarch64

Aditya-9-6:
### Summary
Resolves #1677.

`Btf::type_size()` overwrote the accumulated element count for each nested array dimension (`n_elems = array.len;`) instead of multiplying them. For a type like `char[2][3]`, Aya calculated `value_size = 3` instead of `6`, causing the kernel to reject `BPF_MAP_CREATE` with `EINVAL`.

### Solution
- Accumulate nested array dimensions using `n_elems.checked_mul(array.len as usize)` in `Btf::type_size()`.
- Calculate the total type size using `(size as usize).checked_mul(n_elems)`.
- Returns `BtfError::UnexpectedBtfType { type_id }` on arithmetic overflow.
- Aligns with the Linux kernel BTF specification (`BTF_KIND_ARRAY`), libbpf (`btf__resolve_size`), and cilium/ebpf (`btf.Sizeof()`).

### Verification & Testing
- Added unit test `test_multidimensional_array_type_size` in `aya-obj/src/btf/btf.rs` verifying:
  - 1D array (`char[3]` -> 3 bytes)
  - 2D array (`char[2][3]` -> 6 bytes)
  - 3D array (`char[4][2][3]` -> 24 bytes)
  - Array of typedef / Typedef to array (`typedef char row_t[3]; row_t matrix[2]` -> 6 bytes)
- Ran `cargo +nightly fmt --check` (passes cleanly).
- Ran `cargo check --target x86_64-unknown-linux-gnu --tests -p aya-obj` (passes cleanly).
- Ran `cargo check --target x86_64-unknown-linux-gnu -p aya` (passes cleanly).

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1725)


tamird:
Thank you! Fixed in #1679.

PR #1741, opened 2026-09-18
build: update Rust 1.98.1, nightly-2026-09-18

crabby-the-crab:
Update the Rust stable and nightly pins when needed and regenerate derived files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aya-rs/aya/1741)


tamird:
@vadorovsky no need to manually force push this; the bots will do it as new commits land on master

